PR #19, opened 2015-09-23
Add es6 compilation + tests (fixes #18)

### LINKED ISSUE #18: Setup babel for es6 -> es5 transpilation

muffinresearch:
Note the code and the tests are just some dummy content to prove this is working. It can all be removed as soon as we have some real code to test.

This setup allows for a bundle made from the src/main.js entry point + one for tests.

Test runs build the bundle and test that. The sourcemaps should mean we get meaningful pointers to the code. I've followed the way the karma setup works as much as possible. 

Running a single test can be a bit of a pain due to the bundling but generally this should run very quickly.

If that proves to be a problem then we can look at this in due course. Also it should be possible to have this setup to rebuild/retest on changes if we want. That can also come later.

Other notes: Tests get `sinon` and `chai.assert` for free you don't need to import them. If we have any useful helpers we can consider doing similar for those too. Just saves on repetition.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,6 +23,7 @@
     "it": false,
     "module": false,
     "require": false,
+    "sinon": false,
   },
   "rules": {
     "block-scoped-var": 0,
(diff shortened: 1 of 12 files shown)


tofumatt:
Works for me and looks good, r+

PR #20, opened 2015-09-24
Wrap zip lib with simple promise-based API (fixes #16)

### LINKED ISSUE #16: Add code and lib wrapper to unpack an .xpi for purposes of validation
* API for extracting metadata 
* Api for extracting individual files and handing them off to the specific parser/validator.

muffinresearch:


src/xpi.js
```diff
@@ -0,0 +1,80 @@
...
+      // on this instance return that.
+      if (this.metadata && this.zipfile) {
+        return resolve(this.metadata);
+      }
+
+      this.metadata = {};
+      this.zipLib.open(this.filename, {autoClose: false},
+                       (err, zipfile) => {
+        if (err) {
+          reject(err);
```
tofumatt: Should we `return reject` here?

src/xpi.js
```diff
@@ -0,0 +1,80 @@
+import yauzl from 'yauzl';
+
+
+export default class Xpi {
+
+  constructor(filename, zipLib=yauzl) {
+    this.filename = filename;
+    this.zipLib = zipLib;
+    this.metadata = null;
```
tofumatt: Why initialise to `null`? We can test for truthiness with `undefined`.

Is it to establish the class props in the code? That's cool, but just curious. I like `this.foo = undefined;` or even `this.foo;` if it's just about expressing intent/prop names via code. `null` is different from `undefined`, but I'm being semantic.

src/xpi.js
```diff
@@ -0,0 +1,80 @@
...
+          }
+        }
+      });
+    });
+  }
+
+  getFileAsStream(path) {
+    return new Promise((resolve, reject) => {
+      if (!this.zipfile) {
+        throw new Error('this.zipfile is null. Make sure to ' +
```
tofumatt: A nitpick, but:

```js
        throw new Error(
          'this.zipfile is null. Make sure to run getMetaData first');
```

Is easier to read/edit to my eyes whilst still < 80 chars across.

src/xpi.js
```diff
@@ -0,0 +1,80 @@
...
+                        'run getMetaData first');
+      }
+
+      if (Object.keys(this.metadata).indexOf(path) === -1) {
+        throw new Error('path does not exist in metadata', path);
+      }
+
+      var entry = this.metadata[path];
```
tofumatt: `entry` is only used once on the very next line. Is this here to cut down on line length on the next line?

tests/test.xpi.js
```diff
@@ -0,0 +1,238 @@
+import Xpi from 'xpi';
+import { Readable } from 'stream';
+
+
+const defaultData = {
+  compressionMethod: 8,
```
tofumatt: This reads like a magic constant, what does this mean? ^_^

tests/test.xpi.js
```diff
@@ -0,0 +1,238 @@
...
+});
+
+const installRdfEntry = Object.assign({}, defaultData, {
+  compressedSize: 416,
+  uncompressedSize: 851,
+  fileName: 'install.rdf',
+});
+
+const chromeContentDir = {
+ compressionMethod: 0,
```
tofumatt: I'm _guessing_ this means "no compression"? But again: not sure, feels like a magic constant.

tests/test.xpi.js
```diff
@@ -0,0 +1,238 @@
...
+  it('should reject on errors in zipfile.openReadStream()', () => {
+    var myXpi = new Xpi('foo/bar', this.fakeZipLib);
+
+    this.openStub.yieldsAsync(null, this.fakeZipFile);
+    this.openReadStreamStub.yields(new Error('openReadStream test'));
+    this.entryStub.yields(chromeManifestEntry);
+
+    return myXpi.getMetaData()
+      .then(() => {
+        assert.fail(null, null, 'Unexpected success');
```
tofumatt: I chuckled reading that message :smile:

src/xpi.js
```diff
@@ -0,0 +1,80 @@
+import yauzl from 'yauzl';
+
+
+export default class Xpi {
+
+  constructor(filename, zipLib=yauzl) {
```
tofumatt: :+1: to supplying the library as a default argument, I like that touch.

tofumatt:
Looks good, just wondering about that `return reject()` and the initialisation of props. After that should be good to merge.

tests/test.xpi.js
muffinresearch: This is based on a sample of the data we get back from yauzl. So we're not defining that. But 8 means deflate, 0 means no compression. I left this is because I think we'll probably want to explode if it's not 8 or 0 as they are the only supported values.

src/xpi.js
muffinresearch: Yeah probably uncessary.

src/xpi.js
muffinresearch: Yes, we should. Good catch!

src/xpi.js
muffinresearch: Yeah this is all about intent. I chose to use null because it's not 'undefined' if that makes sense?

src/xpi.js
muffinresearch: :+1:

src/xpi.js
tofumatt: Heh, actually I guess that makes sense too :-)

Cool, let's stick with null then. Totally makes sense in my brain semantically now :-)

- tofumatt

tests/test.xpi.js
tofumatt: Defining those numbers as constants in our Xpi class and using them (`Xpi.COMPRESS_METHOD_DEFLATE` or whatnot) would work then. A comment is also fine if you find that too verbose, but I like to name magic constants like this.

- tofumatt

muffinresearch:
@tofumatt ok, I think this is ready for another look.

tests/test.xpi.js
```diff
@@ -0,0 +1,263 @@
...
+
+    this.openReadStreamStub.yields(null, rstream);
+
+    myXpi.getFileAsStream('install.rdf')
+      .then((readStream) => {
+        var chunks = '';
+        readStream
+          .on('readable', function () {
+            var chunk;
+            while (null !== (chunk = readStream.read())) {
```
tofumatt: Putting the `null` first feels a bit Pythonic, not used to that in JS.

If this is preferred style I'll just make a note of it and be sure to include it in #21.

tofumatt:
r+, let me know about that `while (null !== [...])` ordering bit, but good to merge. :+1:

tests/test.xpi.js
muffinresearch: That came from the docs example: https://nodejs.org/api/stream.html#stream_readable_read_size

tofumatt: Fair enough, seems good 😃

- tofumatt (Sent from mobile)

Issue #21, opened 2015-09-24
Add JSCS to tests


tofumatt:
I've already made a few silly style nitpicks/raised some style questions in #20.

Shall we add some [JS Style checks](https://github.com/jscs-dev/node-jscs) to the tests so we don't quibble over how we do long strings, line lengths, etc.?

It's worked well for localForage. Keeps styles in-line on the more fluid things eslint won't check.

muffinresearch:
Yeah sound good - especially if helps with code-review.

PR #26, opened 2015-09-29
Add basic zip handling from the CLI (fixes #27)

### LINKED ISSUE #27: Connect CLI to zip parser

muffinresearch:
* Hooks up zip handling to the cli. 
* Checks file exists
* Parses the zip.
* Adds colors + turns them off when --boring is passed.

Next up is starting to validate the layout + work out some of the structure for errors.

1__bash

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,10 +1,10 @@
 import argv from 'yargs';
 import { version } from 'json!../package';
 
-export var cli = argv
+
+export default argv
   .usage('Usage: ./$0 [options] addon-package \n\n' +
     'Add-ons Validator (JS Edition) v' + version)
-  // Addon type
   .option('type', {
     alias: 't',
     describe: 'The type that you expect your add-on to be detected as.',
@@ -16,7 +16,6 @@ export var cli = argv
       'search', 'multi',
     ],
   })
-  // Output type
   .option('output', {
     alias: 'o',
     describe: 'The type of output to generate',
@@ -24,7 +23,11 @@ export var cli = argv
     default: 'text',
     choices: ['json', 'text'],
   })
-  // self-hosted add-on.
+  .option('stack', {
+    describe: 'Show stacktraces when errors are thrown',
+    type: 'boolean',
+    default: false,
+  })
   .option('selfhosted', {
     describe: 'Indicates that the addon will not be hosted on ' +
       'addons.mozilla.org. This allows the <em:updateURL> element ' +
@@ -32,20 +35,17 @@ export var cli = argv
     type: 'boolean',
     default: false,
   })
-  // Determined mode.
   .option('determined', {
     describe: 'This flag will continue running tests in successive ' +
       'tests even if a lower tier fails',
     type: 'boolean',
     default: false,
   })
-  // Boring mode
   .option('boring', {
     describe: 'Disables colorful shell output',
     type: 'boolean',
     default: false,
   })
-  // target-maxversion
   .option('target-maxversion', {
     describe: "JSON string to override the package's " +
       'targetapp_maxVersion for validation. The JSON object ' +
@@ -55,7 +55,6 @@ export var cli = argv
       '{"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}": "5.*"} ',
     type: 'string',
   })
-  // target-minversion
   .option('target-minversion', {
     describe: "JSON string to override the package's " +
       'targetapp_minVersion for validation. The JSON object ' +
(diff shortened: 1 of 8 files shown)


src/cli.js
```diff
@@ -24,7 +23,11 @@ export var cli = argv
     default: 'text',
     choices: ['json', 'text'],
   })
-  // self-hosted add-on.
+  .option('stack', {
```
tofumatt: Nice :+1:

src/main.js
```diff
@@ -0,0 +1,6 @@
+import cli from 'cli';
+import Validator from 'validator';
+
+export function createInstance() {
+  return new Validator(cli.argv);
```
tofumatt: I like that API.

src/validator.js
```diff
@@ -0,0 +1,64 @@
...
+  constructor(config) {
+    this.config = config;
+    this.packagePath = config._[0];
+    this.xpi;
+    this.chalk = new chalk.constructor(
+      {enabled: !this.config.boring});
+  }
+
+  handleError(err) {
+    if (this.config.stack === true) {
```
tofumatt: This is a nitpick, but testing for truthiness here would seem to suffice, right?

Would you prefer always using the explicit `=== true` going forward? If that's how you did things on payments/you prefer it that's cool, just making notes for #21.

src/validator.js
```diff
@@ -0,0 +1,64 @@
...
+      console.error(err.stack);
+    } else {
+      console.error(this.chalk.red(err.message || err));
+    }
+  }
+
+  checkFileExists(filepath, _lstat=lstat) {
+    var invalidMsg = new Error(
+      'Path given is not a file or does not exist: "' + filepath + '"');
```
tofumatt: We should use template strings here instead, taking care of ES6 niceness. I think we did this elsewhere and I forgot to mention it, but:

```js
var invalidMessage = new Error(
  `Path given is not a file or does not exist: "${filepath}"`);
```
should work. Also, `invalidMessage` is a nicer variable name, we have space :smile:

tests/test.validator.js
```diff
@@ -0,0 +1,65 @@
...
+        assert.fail(null, null, 'Unexpected success');
+      })
+      .catch(err => {
+        assert.instanceOf(err, Error);
+        assert.include(err.message, 'Path given is not a file');
+      });
+  });
+
+  it('should detect other errors during lstat', () => {
+    var av = new Validator({_: ['foo']});
```
tofumatt: `av` tripped me up for a second, then I realised it meant `addonsValidator`. Is `validatorInstance` too verbose? I know `validator` is confusing/vague, but one/two letter variable names trip me up.

src/validator.js
tofumatt: And actually, reading the output I find this sentence hard to parse. This might read better:

```js
var invalidMessage = new Error(
  `Path "${filepath}" is not a file or does not exist.`);
```

This way you can see the path right away and the sentence is easier to mentally parse. I find that worth the trade-off of not seeing the path clearly at the end of the error (it's in quotes anyway, so it stands out).

tofumatt:
r+wc

Looks great overall; see my few comments but they're mostly minor things.

src/validator.js
muffinresearch: I'm not sure if that's a thing we particularly adhered to in payments, but I definitely tend to default using explicit comparisons over truthy/falsey ones.

src/validator.js
muffinresearch: Good call.

tests/test.validator.js
muffinresearch: Yeah in tests I might do something like this for brevity. Happy to be a bit more verbose though.

src/validator.js
tofumatt: Fair enough. I’ll just make a note in the JSCS so we keep things consistent; I tend to opt for truthy but a failed test for style will be a great reminder for me ;-)

- tofumatt

On 30 September 2015 at 09:20:17, Stuart Colville (notifications@github.com) wrote:

In src/validator.js:

> +
> +export var lstat = promisify(fs.lstat);
> +
> +
> +export default class Validator {
> +
> +  constructor(config) {
> +    this.config = config;
> +    this.packagePath = config._[0];
> +    this.xpi;
> +    this.chalk = new chalk.constructor(
> +      {enabled: !this.config.boring});
> +  }
> +
> +  handleError(err) {
> +    if (this.config.stack === true) {
I'm not sure if that's a thing we particularly adhered to in payments, but I definitely tend to default using explicit comparisons over truthy/falsey ones.

Issue #28, opened 2015-09-30
Add validation error/warning objects


muffinresearch:
From discussions we decide for the validation errors and warnings we need an object for each (they'll mostly be the same.

PR #29, opened 2015-09-30
Add message object fixes (#28)

muffinresearch:
Had to move the constants module to const as this clashed with a built-in module.

This adds a basic object for the validation output. We'll need to expand on this in due course but it will be used by the collector module to collect errors, warnings and notices as we go.


tofumatt:
Looks solid. Normally I'd say make the types a constant rather than a string (I have a pathological obsession with eliminating magic constants), but I think using strings there is a nicer API in this case. Plus we're likely to be using Message directly less and using an actual `Error` or `Warning` object directly anyway.

r+

PR #34, opened 2015-09-30
Add JSCS Tests

### LINKED ISSUE #21: Add JSCS to tests

tofumatt:
Closes #21.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,6 +11,6 @@ module.exports = function(grunt) {
 
   grunt.initConfig(configs);
 
-  grunt.registerTask('test', ['webpack:test', 'mochaTest', 'eslint']);
+  grunt.registerTask('test', ['webpack:test', 'mochaTest', 'eslint', 'jscs']);
 
 };
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Mozilla addons validator",
   "main": "index.js",
   "scripts": {
+    "start": "webpack",
     "test": "node -e \"require('grunt').cli()\" null test"
   },
   "repository": {
@@ -24,6 +25,7 @@
     "chai": "3.3.0",
     "grunt": "0.4.5",
     "grunt-eslint": "17.1.0",
+    "grunt-jscs": "2.1.0",
     "grunt-mocha-test": "0.12.7",
     "grunt-webpack": "1.0.11",
     "json-loader": "0.5.3",
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,3 +1,6 @@
+// Disable warnings about template strings in this file, as we do lots of
+// string concatenation so options/help in the CLI app display nicely.
+// jscs:disable requireTemplateStrings
 import argv from 'yargs';
 import { version } from 'json!../package';
 
--- a/src/xpi.js
+++ b/src/xpi.js
@@ -2,8 +2,8 @@ import yauzl from 'yauzl';
 
 /*
  * Simple Promise wrapper for the Yauzl unzipping lib to unpack add-on .xpis.
- * Note: We're using the autoclose feature of yauzl as a result every operation will
- * open the zip, do something and then close it implicitly.
+ * Note: We're using the autoclose feature of yauzl as a result every operation
+ * will open the zip, do something and then close it implicitly.
  * This makes the API easy to use and the consumer doesn't need to remember to
  * close the zipfile.
  *
@@ -65,7 +65,7 @@ export default class Xpi {
             }
           }
         })
-        .catch(err => reject(err));
+        .catch((err) => reject(err));
     });
   }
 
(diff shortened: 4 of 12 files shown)


src/xpi.js
```diff
@@ -85,7 +85,7 @@ export default class Xpi {
             resolve(readStream);
           });
         })
-        .catch(err => reject(err));
+        .catch((err) => reject(err));
```
tofumatt: I find using the braces unconditionally in arrow functions makes the code easier to scan and consistent. If anyone is really against this though, I can back it out.

src/cli.js
```diff
@@ -1,3 +1,6 @@
+// Disable warnings about template strings in this file, as we do lots of
+// string concatenation so options/help in the CLI app display nicely.
+// jscs:disable requireTemplateStrings
```
tofumatt: I have this here as the ESLint complains about multi-line strings and I'm not sure how others feel about them. I think, especially in this file, they're cleaner. Contrast:

```js
   .option('target-maxversion', {
     describe: "JSON string to override the package's " +
       'targetapp_maxVersion for validation. The JSON object ' +
       'should be a dict of versions keyed by application ' +
       "GUID. For example, setting a package's max Firefox " +
       'version to 5.*: ' +
       '{"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}": "5.*"} ',
     type: 'string',
   })
```

with:

```js
.option('target-maxversion', {
  describe: "JSON string to override the package's \
             targetapp_maxVersion for validation. The JSON object \
             should be a dict of versions keyed by application \
             GUID. For example, setting a package's max Firefox \
             version to 5.*: \
             {\"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}\": \"5.*\"}"
  type: 'string',
})

I find the second cleaner and easier to write. I can add in allowing multiline if you like it (everything still built and ran using them, obviously).
```

muffinresearch: Yeah sounds good - I prefer the latter too. I wanted to use template strings for this but for multi-line strings whitespace handling would result in something resembling PHP's heredoc. We could of course write a tag to handle that in time.

Gruntfile.js
```diff
@@ -11,6 +11,6 @@ module.exports = function(grunt) {
 
   grunt.initConfig(configs);
 
-  grunt.registerTask('test', ['webpack:test', 'mochaTest', 'eslint']);
+  grunt.registerTask('test', ['jscs', 'webpack:test', 'mochaTest', 'eslint']);
```
muffinresearch: It's probably preferable to have jscs at the end alongside eslint because when developing tests are more important output, than it bailing on little stylistic issues.

package.json
```diff
@@ -24,6 +25,7 @@
     "chai": "3.3.0",
     "grunt": "0.4.5",
     "grunt-eslint": "17.1.0",
+    "grunt-jscs": "^2.1.0",
```
muffinresearch: Something we did in payments was to turn off the '^' so that we get predictable versions.

This can be done globally:

```
npm config set save-prefix=''
```

Or we should use .npmrc (I've added that in #36)

src/xpi.js
muffinresearch: :+1: - on payments we ended up making them optional - but in usage I've often thought I'd prefer them to be consistent.

tasks/jscs.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  js: {
+    options: {
+      config: '.jscsrc',
+      esnext: true,
+      // fix: false,
```
muffinresearch: Can this be removed?

tasks/mochaTest.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
       function() {
         /*eslint-disable */
         assert = require('chai').assert;
-        sinon = require('sinon') //
```
muffinresearch: lol  - not sure what was going on there!

muffinresearch:
r+wc - would be good to fix the package version prefix. Otherwise all minor quibbles.

Thanks for fixing all my nits too!

tasks/jscs.js
tofumatt: Oh, yes, sorry. It's a neat feature that automatically fixes easy-to-fix nits. But I don't think it's needed and it's a bit error-prone (I tested it); I'll take it out.

package.json
tofumatt: Good call. I installed this from the CLI and I've been bit by `^` before. And we have greenkeeper anyway, so should be good.

Gruntfile.js
tofumatt: Oh, good call. :+1:

src/cli.js
tofumatt: Arg, I just noticed what you mean about the whitespace. I'll leave this one out for now, it's not too bad as-is.

PR #35, opened 2015-10-01
Add collector (fixes #30)

### LINKED ISSUE #30: Add message collector
This classes job is to collect up warnings, errors and notice objects and provide an API for doing so.

muffinresearch:
I've deliberately kept this quite simple. I don't want to bring too much baggage across from the old validator until we know we need it.

### DIFF
--- a/src/collector.js
+++ b/src/collector.js
@@ -1,0 +1,43 @@
+import { default as Message } from 'message';
+
+// "I have a display case ready and waiting for our newest acquistions!"
+// --Taneleer Tivan
+
+
+export default class Collector {
+
+  constructor() {
+    this.errors = [];
+    this.notices = [];
+    this.warnings = [];
+  }
+
+  get length() {
+    return this.errors.length +
+      this.notices.length +
+      this.warnings.length;
+  }
+
+  _addMessage(type, opts, _Message=Message) {
+    // Message will throw for incorrect types.
+    // we have a test to ensure that is the case.
+    var message = new _Message(type, opts);
+    var list = this[`${type}s`];
+    if (typeof list === 'undefined') {
+      throw new Error(`Message type "${type}" not currently collected`);
+    }
+    list.push(message);
+  }
+
+  addError(opts) {
+    this._addMessage('error', opts);
+  }
+
+  addNotice(opts) {
+    this._addMessage('notice', opts);
+  }
+
+  addWarning(opts) {
+    this._addMessage('warning', opts);
+  }
+}
--- a/tests/test.collector.js
+++ b/tests/test.collector.js
@@ -1,0 +1,68 @@
+import { default as Collector } from 'collector';
+
+
+describe('Collector', function() {
+
+  it('should be thrown an error if Message is created without a type', () => {
+    assert.throws(() => {
+      var collection = new Collector();
+      collection._addMessage();
+    }, Error, /Message type "undefined" is not/);
+  });
+
+  it('should be thrown an error if Message is created with bad type', () => {
+    assert.throws(() => {
+      var collection = new Collector();
+      collection._addMessage('whatevs');
+    }, Error, /Message type "whatevs" is not/);
+  });
+
+  it('should be throw an error type is not collected', () => {
+    assert.throws(() => {
+      var FakeMessage = sinon.stub();
+      var collection = new Collector();
+      collection._addMessage('whatevar', {}, FakeMessage);
+    }, Error, /Message type "whatevar" not currently collected/);
+  });
+
+  it('length should start at 0', () => {
+    var collection = new Collector();
+    assert.equal(collection.length, 0);
+  });
+
+  it('length should reflect number of messages', () => {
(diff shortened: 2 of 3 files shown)


src/collector.js
```diff
@@ -0,0 +1,53 @@
+import { default as Message } from 'message';
+
+// "I have a display case ready and waiting for our newest acquistions!"
```
tofumatt: Haha, I get this reference!

src/collector.js
```diff
@@ -0,0 +1,53 @@
...
+      case 'notice':
+        return this.notices;
+      case 'warning':
+        return this.warnings;
+      default:
+        throw new Error(`Invalid message type "${type}"`);
+    }
+  }
+
+  messageCount() {
```
tofumatt: Can we use fancy getters here instead? This is essentially the `length` property of a Collection (rather than the `length` property of a specific subset of messages eg `collection.notices.length`). It would be nice if you could do:

```js
var collection = new Collector();
assert.equal(collection.length, 0);
```

Seems more natural; one less API to memorise.

tests/test.collector.js
```diff
@@ -0,0 +1,67 @@
+import { default as Collector } from 'collector';
+
+
+describe('Collector', function() {
+
+  it('should be thrown an error if Message is created without a type', () => {
+    assert.throws(() => {
+      var Collection = new Collector();
```
tofumatt: We should reserve variables starting with capital letters for `Classes` or `CONSTANTS`. I think it should be `var collection = new Collector();` here.

Makes it easier to keep track of what variable is of what type.

src/collector.js
```diff
@@ -0,0 +1,53 @@
...
+
+  _addMessage(type, opts) {
+    // Message will throw for incorrect types.
+    // we have a test to ensure that is the case.
+    var message = new Message(type, opts);
+    var list = this.getMessageList(type);
+    list.push(message);
+  }
+
+  addError(opts) {
```
tofumatt: Having to supply an empty object as an argument seems silly. We should make the `opts={}` default to an empty object if that's all it needs, or make a note to make it validate the presence of certain properties in the future.

Lower in the tests there's `addError({})`, which seems awkward.

src/collector.js
```diff
@@ -0,0 +1,53 @@
...
+
+export default class Collector {
+
+  constructor() {
+    this.errors = [];
+    this.notices = [];
+    this.warnings = [];
+  }
+
+  getMessageList(type) {
```
tofumatt: In the tests you're accessing the `errors`, `notices`, and `warnings` properties all directly: https://github.com/mozilla/addons-validator/pull/35/files#diff-8f475bdab4471a61c3e772723fae0051R47

This feels like a cleaner/shorter/more obvious API, adding the `getMessageList` doesn't seem necessary.

tofumatt: Oh, I guess it allows you to use the type constants (`collection.getMessageList('error')` instead of `collection.errors`).

At the very least I think `getMessages` would be nicer, but I find a plural (`getMessages`) method name enough to imply I get back an array. If that's not explicit enough though I understand.

muffinresearch: It's needed as a look up to dry up the _addMessage call.

tofumatt: Ah, fair enough. If that's the only reason we could do:

```js
// Old:
this.getMessageList(type);
// New:
this[`${type}s`]
```
and save adding this method.

muffinresearch: I suppose I could just add 's' to the end of the type.

src/collector.js
muffinresearch: In reality it's going to be required because it won't be any use without some props. I guess we need to decide exactly what those are and deal with that into the Message object.

src/collector.js
muffinresearch: Nice idea.

src/collector.js
tofumatt: That's what I assumed. Okay, we can decide what they are later though. I'll just make an issue to make sure we don't forget.

tests/test.collector.js
muffinresearch: Yep, I'll fix that.

tofumatt:
Cool, I'm finished the review. Let me know when I should take another look :+1:

muffinresearch:
@tofumatt ok this is ready for another look.

src/collector.js
```diff
@@ -0,0 +1,43 @@
...
+      this.warnings.length;
+  }
+
+  _addMessage(type, opts, _Message=Message) {
+    // Message will throw for incorrect types.
+    // we have a test to ensure that is the case.
+    var message = new _Message(type, opts);
+    var list = this[`${type}s`];
+    if (typeof list === 'undefined') {
+      throw new Error(`Message type "${type}" not currently collected`);
```
muffinresearch: This covers us if we ever add/remove an error type from Message and forget to update this.

tofumatt:
Awesome! r+ :+1:

PR #39, opened 2015-10-01
Add template tag for concatenating multiline strings

muffinresearch:
Allows for this kind of thing:

js_bin_-_collaborative_javascript_debugging

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,0 +1,29 @@
+/*
+ * Template tag for removing whitespace and new lines
+ * in order to be able to use multiline template strings
+ * as a single string.
+ *
+ * Usage: singleLineString`foo bar baz
+ *                    whatever`;
+ *
+ * Will output: 'foo bar baz whatever'
+ *
+ */
+
+export function singleLineString(strings, ...vars) {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -0,0 +1,29 @@
...
+ * as a single string.
+ *
+ * Usage: multiConcat`foo bar baz
+ *                    whatever`;
+ *
+ * Will output: 'foo bar baz whatever'
+ *
+ */
+
+export function multiConcat(strings, ...vars) {
```
tofumatt: This is all lovely, but `multiConcat` feels too vague a name. `stringConcat` feels better but the real aim is to make a single-line string, so `singleLineString` feels the best name for this function, even if it's a bit more verbose. If we import it directly:

```js

var output = singleLineString`foo
               bar
         baz`;
```
looks pretty good to me.

tofumatt:
zomg that's awesome :+1: 

r+wc

src/utils.js
muffinresearch: Sure, that works for me.

PR #41, opened 2015-10-02
Use single line string util

muffinresearch:


tofumatt:
Yay! Ironically it's a chore to go through these ones, but next time we write anything like this it should be way easier. :smile: 

r+

muffinresearch:
Yep,  I'm starting to write some error messages and it seemed wrong to use concatenation or continuation.

PR #42, opened 2015-10-02
Restore newlines in CLI output

tofumatt:
Just realised that when we added the `singleLineString` template tag in #41, we removed intentional newlines before JSON strings (eg: https://github.com/mozilla/addons-validator/blob/master/src/cli.js#L63).

This restores them in a way I hope isn't offensive. I still think it's nicer (and more explicit/obvious) than just allowing string concat on that line.

r?

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -52,7 +52,7 @@ export default argv
     describe: singleLineString`JSON string to override the package's
       targetapp_maxVersion for validation. The JSON object should be
       a dict of versions keyed by application GUID. For example,
-      setting a package's max Firefox version to 5.*:
+      setting a package's max Firefox version to 5.*:<newline>
       {"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}": "5.*"}`,
     type: 'string',
   })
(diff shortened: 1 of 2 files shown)


muffinresearch:
Does the newline being removed matter that much? yargs own formatting seems to break up strings anyway. How different is the actual output before and after?

tofumatt:
It lets you see the JSON string on one line… but I guess the formatting is a bit weird anyway.

screenshot 2015-10-02 14 53 23

vs:

screenshot 2015-10-02 14 54 21

I added this and meant to ask if it was worth it... and if it was I'd add a test. Now I'm thinking it's silly. Feel free to close if you agree :wink:

muffinresearch:
Yeah we should probably rip the examples out. The cli help doesn't seem like a great place for that. I'll close this for now but if we want it back we can always revisit.

tofumatt:
Sounds good. Thanks. 👍

- tofumatt

On 2 October 2015 at 14:58:31, Stuart Colville (notifications@github.com) wrote:

Yeah we should probably rip the examples out. The cli help doesn't seem like a great place for that. I'll close this for now but if we want it back we can always revisit.

PR #43, opened 2015-10-02
Handle duplicate files in xpi (fixes #38, fixes #37)

### LINKED ISSUE #37: Detect duplicate files in a zip
It's possible to create a zip with 2 files being the exact same name. We need to detect that.

### LINKED ISSUE #38: Test to make sure Message has required opts
It's possible to create a Message with an empty `opts` object, which we should just make sure to test against in the future when we decide on/add the required opts.

https://github.com/mozilla/addons-validator/pull/35/files#r40915475

muffinresearch:
Ended up covering a quite a lot of ground here - as yaks reared their ugly heads I shaved them :).

* collector messageType lists are now produced through iterating over constants
* messageTypes are now constants
* required message types are handled (see #38)
* Xpi throws a custom exception DuplicateZipEntryError when an entry already seen comes up.
* Validator now has a output property and a toJSON method + a print() method.
* Fixed a race condition in the Xpi class. We now listen to `close` rather than `end` because the latter resulted in resolution of the promise before the last entry was processed (and rejected).

### DIFF
--- a/src/collector.js
+++ b/src/collector.js
@@ -1,4 +1,5 @@
 import { default as Message } from 'message';
+import * as constants from 'const';
 
 // "I have a display case ready and waiting for our newest acquisitions!"
 // --Taneleer Tivan
@@ -7,15 +8,17 @@ import { default as Message } from 'message';
 export default class Collector {
 
   constructor() {
-    this.errors = [];
-    this.notices = [];
-    this.warnings = [];
+    for (let type of constants.MESSAGE_TYPES) {
+      this[`${type}s`] = [];
+    }
   }
 
   get length() {
-    return this.errors.length +
-      this.notices.length +
-      this.warnings.length;
+    var len = 0;
+    for (let type of constants.MESSAGE_TYPES) {
+      len += this[`${type}s`].length;
+    }
+    return len;
   }
 
   _addMessage(type, opts, _Message=Message) {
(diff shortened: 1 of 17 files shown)


src/collector.js
```diff
@@ -7,15 +8,17 @@ import { default as Message } from 'message';
 export default class Collector {
 
   constructor() {
-    this.errors = [];
-    this.notices = [];
-    this.warnings = [];
+    for (let type of constants.MESSAGE_TYPES) {
```
tofumatt: Nice.

src/messages/layout.js
```diff
@@ -0,0 +1,10 @@
+import { gettext, singleLineString } from 'utils';
+
+
+export const DUPLICATE_XPI_ENTRY = {
+  code: 'DUPLICATE_XPI_ENTRY',
+  message: gettext('Package contains duplicate entries'),
```
tofumatt: It's an easy find/replace job later, but we can probably use `_` for this since we aren't likely to be using Underscore.js :wink:

tofumatt:
r+

This is splendid!

src/messages/layout.js
muffinresearch: Yeah, sounds good - we can easily do `import { gettext as _ } from 'utils'; `

tofumatt: Oh, right, of course :-)

PR #44, opened 2015-10-04
Check for mozIndexedDB; close #15

### LINKED ISSUE #15: Check for use of mozindexdb
Write a first rule for: mozindexdb_removed, mozIndexedDB has been removed

tofumatt:
This adds support for custom ESLint rules and tests for the existence of `mozIndexedDB`.

It adds a [`getFileAsString()` method](https://github.com/mozilla/addons-validator/pull/44/files#diff-9c4d55b37e780449ba143b3c579371eaR106) to the `Xpi` class so we can load entire files into memory to pass to linters like ESLint, which needs the entire file and can't deal with streams. As discussed earlier on IRC, we might not end up passing streams around much when maybe checks need the full context of a file.


tofumatt:
I think this is ready for r? now.

tests/test.javascript.js
```diff
@@ -0,0 +1,91 @@
...
+import { singleLineString } from 'utils';
+
+
+describe('JS Code Checker', function() {
+
+  it('should not allow mozIndexedDB', (done) => {
+    var code = 'var myDatabase = indexeddb || mozIndexedDB;';
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    jsScanner.scan()
```
muffinresearch: If you return a promise in a test you don't need to use `done()` it's actually quite useful as this will expose issues where the code misses returning a promise too.

tests/test.validator.js
```diff
@@ -85,6 +85,18 @@ describe('Validator', function() {
     assert.equal(parsedJSON.summary.warnings, 0);
   });
 
+  it('should send JSScanner messages to the collector', (done) => {
```
muffinresearch: I wonder about this, if it might be brittle when we get to the point of having more tests. E.g. will the count increase for example. I wonder if we should have a way to specify a specific test or list of tests or whether we should instead ignore this and work on the basis that the more granular level of tests will suffice. I'm leaning towards the latter approach.

src/xpi.js
```diff
@@ -102,4 +102,43 @@ export default class Xpi {
...
+            // TODO: Check for JS files better than this (follow require path)
+            // of add-on?
+            if (endsWith(filename, '.js')) {
+              jsFiles.push(filename);
+            }
+          }
+
+          resolve(jsFiles);
+        })
+        // NOTE: Why not just use `.catch(reject)`?
```
muffinresearch: Yeah we should just move to doing that - I can't see an obvious reaon not to.

tests/test.javascript.js
```diff
@@ -0,0 +1,91 @@
...
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 1);
+        assert.equal(validationMessages[0].id, 'mozIndexedDB_prop');
+        assert.equal(validationMessages[0].severity, SEVERITY_WARNING);
+        done();
+      });
+  });
```
muffinresearch: Could use a blank line between tests.

tests/test.javascript.js
```diff
@@ -0,0 +1,91 @@
...
+  // Not sure how to test for this one yet; it's pretty messy.
+  it.skip('should warn when mozIndexedDB is assembled into literal', (done) => {
```
muffinresearch: This is pretty important, if eslint can't provide a good way to test for this this validation is easily subverted.

src/messages/javascript.js
```diff
@@ -0,0 +1,15 @@
...
+
+// These constants aren't fullcaps so we can use filenames/rule names/
+// message names consistently across the app.
+export const mozIndexedDB = {
+  code: 'mozIndexedDB',
+  message: gettext('mozIndexedDB has been removed; use indexedDB instead.'),
+  description: gettext('mozIndexedDB has been removed; use indexedDB instead.'),
+};
+export const mozIndexedDB_prop = {
+  code: 'mozIndexedDB_prop',
```
muffinresearch: This refers to a specific JS api so I see why it's this way but I was envisaging keeping all the codes + constants uppercase.

src/xpi.js
```diff
@@ -102,4 +102,43 @@ export default class Xpi {
...
+  }
+
+  getJSFiles() {
+    return new Promise((resolve, reject) => {
+      return this.getMetaData()
+        .then((metadata) => {
+          let jsFiles = [];
+
+          for (let filename in metadata) {
+            // TODO: Check for JS files better than this (follow require path)
```
muffinresearch: Good question - I wonder how the current validator deals with the possibility of imports including requires/imports of files that don't end in .js?

src/validator.js
```diff
@@ -75,6 +88,44 @@ export default class Validator {
...
+    return new Promise((resolve, reject) => {
+      this.checkFileExists(this.packagePath)
+        .then(() => {
+          this.xpi = new _Xpi(this.packagePath);
+          return this.xpi.getMetaData();
+        }).then(() => {
+          return this.xpi.getJSFiles();
+        })
+        .then((jsFiles) => {
+          jsFiles.forEach((filename) => {
```
muffinresearch: I think there needs to be some more tests to ensure that if anything under here blows up we're seeing the errors as expected.

src/utils.js
```diff
@@ -1,4 +1,12 @@
 /*
+ * Implementation of String.endsWith, which errors in node.
+ *
+ */
+export function endsWith(string, suffix) {
+  return string.indexOf(suffix, string.length - suffix.length) !== -1;
```
muffinresearch: I don't think there's a test for this.

tests/test.javascript.js
```diff
@@ -0,0 +1,91 @@
...
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 1);
+        assert.equal(validationMessages[0].id, 'mozIndexedDB_prop');
+        assert.equal(validationMessages[0].severity, SEVERITY_WARNING);
+        done();
+      });
+  });
+
+  // Depends on: https://github.com/mozilla/addons-validator/issues/7
+  it.skip('ignores /*eslint-disable*/ comments', (done) => {
```
muffinresearch: Nice preparation for the future :+1:

muffinresearch:
Looks pretty good overall - just a few things that need addressing.

Would be good to fix up the returning promises in the tests that feature promise-based apis rather than using `done()` unless there's specific reasons why that doesn't work. 

It seems it's important to understand any limitations of eslint can a test be easily bypassed through obfuscation via concatenated propnames. Maybe FreddyB has seen good ways to handle that? Also that require / import question you raised is of a similar nature.

There's a couple of missing tests, e.g. just a basic one for the util and also I think we need some to make sure errors are being handled should eslint blow-up or the js-validator blow-up in the iteration over the js files in the xpi tests.

The rest are minor nits.

muffinresearch:
Give me shout if you want to chat over any of these bits. Some of it we might want to move to future bugs - and that's fine to do. We just need to be careful we don't end up with a huge list of things that mean that eslint for validation is easily bypassed.

tests/test.javascript.js
tofumatt: I did not know; that’s awesome, thanks!

src/const.js
```diff
@@ -1,6 +1,10 @@
 export const DEFLATE_COMPRESSION = 8;
 export const NO_COMPRESSION = 0;
 
+export const SEVERITY_ERROR = 2;
```
muffinresearch: Could this not use VALIDATION_* instead. Seems it might get confusing to have uses of similar but different constants. I'm easily confused :)

tests/test.javascript.js
muffinresearch: Yeah it's handy, I only found out about this recently too.

src/const.js
tofumatt: Yeah, that's better. I used severity as it's what ESLint uses internally, but I didn't like it either. `VALIDATION_` it is! :smile:

src/messages/javascript.js
tofumatt: We can do that. I actually think me doing it this way is a holdover from something I hacked together and never removed.

I'll aim to make these more CONSTANT_LIKE.

src/utils.js
tofumatt: Damn, good catch. I actually was thinking "I think I left out a test for something" but couldn't remember what.

src/xpi.js
tofumatt: Yeah, it's likely something for a separate issue and could be its own can of worms.

A brute-force way would be to get ESLint to try and parse every file, and anything that throws a parsing error is assumed to not be JS. There are edge cases for that implementation as well, but it would prevent people sneaking code past the validator by naming every file `.javascript` or something.

src/xpi.js
tofumatt: :+1:

tests/test.javascript.js
tofumatt: Oops! I think I grouped them because they're all testing a particular rule, but you're right, it's cramped.

tests/test.javascript.js
tofumatt: Agreed. I want to find an _easy_ way to do it... we can technically follow the variable path for any variable manually and see what it is assembled into, but I'm looking to see if there's an easier way to get the contents of a variable at any point.

I wonder how the current validator handles it. Even if they don't do anything I'd like for us to cover it. I'll have a look.

tests/test.validator.js
tofumatt: Fair enough. The test XPI is one line of JS but who knows what other tests it will trip once we have more.

Just checking that the collector increments at all is what I'm looking for, so how about I just check for `> 0`? That aspect of the code isn't tested anywhere else and I figure why not leave it in?

src/validator.js
muffinresearch: Shouldn't this use Promise.all so you can resolve when all promises have resolved?

src/validator.js
```diff
@@ -75,6 +88,44 @@ export default class Validator {
...
+              .then((code) => {
+                let jsScanner = new JavaScriptScanner(code, filename);
+                return jsScanner.scan();
+              })
+              .then((validatorMessages) => {
+                validatorMessages.forEach((message) => {
+                  this.addToCollector(message);
+                });
+
+                resolve(this.collector);
```
muffinresearch: Doesn't this mean you'll resolve on the first file - what about the following iterations.

tofumatt: I’ve been testing against things with one file, yeah. You’re right. This needs to be cleaned up, whoops.

- tofumatt

On 5 October 2015 at 16:34:30, Stuart Colville (notifications@github.com) wrote:

In src/validator.js:

> +        }).then(() => {
> +          return this.xpi.getJSFiles();
> +        })
> +        .then((jsFiles) => {
> +          jsFiles.forEach((filename) => {
> +            this.xpi.getFileAsString(filename)
> +              .then((code) => {
> +                let jsScanner = new JavaScriptScanner(code, filename);
> +                return jsScanner.scan();
> +              })
> +              .then((validatorMessages) => {
> +                validatorMessages.forEach((message) => {
> +                  this.addToCollector(message);
> +                });
> +
> +                resolve(this.collector);
Doesn't this mean you'll resolve on the first file - what about the following iterations.

tofumatt:
Hey there, updated the PR. Should have fixed everything. Let me know if that test in validator is still too much and we can take it out, but it should be less brittle now.

I liked testing against an actual file, but if it's not discrete enough let me know.

src/validator.js
```diff
@@ -24,6 +25,13 @@ export default class Validator {
     this.collector = new Collector();
   }
 
+  addToCollector(message) {
+    var messageType = message.severity.charAt(0).toUpperCase() +
+                      message.severity.slice(1);
+    console.log(message);
```
muffinresearch: Could probably remove this - though at some point it would be nice to have some configurable logging - we should add an issue to add something with debug levels.

src/validator.js
```diff
@@ -75,6 +83,68 @@ export default class Validator {
     return output;
   }
 
+  scan(_Xpi=Xpi) {
+    return new Promise((resolve, reject) => {
+      this.checkFileExists(this.packagePath)
+        .then(() => {
+          this.xpi = new _Xpi(this.packagePath);
+          return this.xpi.getMetaData();
```
muffinresearch: I reckon you should be ok to kill this and the following line.

src/validator.js
```diff
@@ -75,6 +83,68 @@ export default class Validator {
...
+  scanJSFiles(jsFiles) {
+    return new Promise((resolve, reject) => {
+      // Resolve once every file in the XPI has been checked.
+      var jsFilesPromises = [];
+
+      jsFiles.forEach((filename) => {
+        jsFilesPromises.push(this.scanJSFile(filename));
+      });
+
+      return Promise.all(jsFilesPromises).then(() => {
```
muffinresearch: Not sure if you like it but so far I've always been start `then()` and `catch()` on the next line. Makes it a bit easier to see the different blocks imho.

tests/test.javascript.js
```diff
@@ -0,0 +1,141 @@
...
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 1);
+        assert.equal(validationMessages[0].id,
+                     messages.MOZINDEXEDDB_PROPERTY.code);
+        assert.equal(validationMessages[0].severity, VALIDATION_WARNING);
+      });
+  });
+
+  // Not sure how to test for this one yet; it's pretty messy.
```
muffinresearch: Might be nice to use TODO and point at your issue re: obfuscation for this one.

tests/test.utils.js
```diff
@@ -1,6 +1,30 @@
...
+  it('returns false if a string does not end with a specified string', () => {
+    assert.notOk(utils.endsWith('Matthew', 'Mac'));
+  });
+
+  it('returns true if a string ends with itself', () => {
+    assert.ok(utils.endsWith('Matthew', 'Matthew'));
+  });
+
+  it('returns true if a strange character is present', () => {
+    assert.ok(utils.endsWith('G chord! 🎸', '🎸'));
```
muffinresearch: heh

tests/test.xpi.js
```diff
@@ -264,4 +277,70 @@ describe('Xpi.getFileAsStream()', function() {
...
+    this.openStub.yieldsAsync(null, this.fakeZipFile);
+
+    var rstream = new Readable();
+    rstream.push('line one\n');
+    rstream.push('line two');
+    rstream.push(null);
+
+    this.openReadStreamStub.yields(null, rstream);
+
+    myXpi.getFileAsString('install.rdf')
```
muffinresearch: Can this one just return a promise?

tests/test.xpi.js
```diff
@@ -264,4 +277,70 @@ describe('Xpi.getFileAsStream()', function() {
...
+  it('should return all JS files', (done) => {
+    var myXpi = new Xpi('foo/bar', this.fakeZipLib);
+    myXpi.metadata = {
+      'install.rdf': installRdfEntry,
+      'chrome.manifest': chromeManifestEntry,
+      'main.js': jsMainFileEntry,
+      'secondary.js': jsSecondaryFileEntry,
+    };
+
+    myXpi.getJSFiles()
```
muffinresearch: return the promise?

src/messages/javascript.js
```diff
@@ -0,0 +1,14 @@
+import { gettext } from 'utils';
```
muffinresearch: What no love for '_'  :)

muffinresearch:
Nice on! Couple of minor nits left but looks good r+wc.

src/validator.js
tofumatt: It’s typically not how I’ve written stuff but I can totally get on board, and I’m for consistency > all else :-)

I’ll see if I can wire up a JSCS rule to prevent me from doing that without getting yelled at by the linter :)

src/validator.js
tofumatt: Oops, yes, this is totally unintentional.

tofumatt:
Thanks! All the nits are fixed now; gonna squash a few commits then merge it.

PR #45, opened 2015-10-05
Fix issue with close not called when openReadStream is invoked

muffinresearch:
The close event wasn't being fired when getting the metadata for zips. I think it might be because when you call openReadStream you need to inhale the entire stream. If you call openReadStream and don't consume the stream it results in the close event not firing. Anyhoo, looking at our code it made me realise we don't really need to use `openReadStream` just to get the metadata that we already have.

Fixing the tests now that handleEntry is sync was a bit of a pain, but I found a better way to get the callback pass to `zipfile.on('entry', cb)`. This then means that the handleEntry code is called properly in the way it was registered rather than directly, which makes for a more consistent approach for all cases where we are adding metadata during testing.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@
     "spread": true,
   },
   "globals": {
+    "afterEach": false,
     "assert": false,
     "beforeEach": false,
     "describe": false,
--- a/src/xpi.js
+++ b/src/xpi.js
@@ -21,24 +21,6 @@ export default class Xpi {
     this.entries = [];
   }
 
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -17,6 +17,7 @@
     "spread": true,
   },
   "globals": {
+    "afterEach": false,
```
muffinresearch: I didn't end up needing this but I'm sure we will at some point.

tofumatt:
r+

PR #57, opened 2015-10-06
Add text output (fixes #51)

### LINKED ISSUE #51: Add text output processing.
We've got a start on outputting JSON - we need to add something to output the rules in a nice way as just text with nice colours assuming boring is false.

muffinresearch:
(fixes #51)

Added the text ouput - whilst it's not crucial it'll be nice for development as this is easier to grok than looking at a pile of JSON.

1__bash


src/validator.js
```diff
@@ -84,6 +100,59 @@ export default class Validator {
...
+    out.push(_('Validation Summary:'));
+    out.push('');
+    out.push(columnify(this.output.summary, {
+      showHeaders: false,
+      minWidth: 15,
+    }));
+    out.push('');
+
+    for (let type of constants.MESSAGE_TYPES) {
+      /*eslint-disable no-loop-func */
```
tofumatt: We're using `let` so this shouldn't be an issue: http://eslint.org/docs/rules/no-loop-func.html#rule-details

Was this giving you an error? :frowning:

tofumatt:
:+1: Very nice. Let me know about the eslint bit, but if it needs to be there that's fine. Maybe we even want to remove that rule (we can be smart about loops I think).

src/validator.js
muffinresearch: Yeah it's the callback funcs passed into then etc. I reckon you're right, we should probably just turn that rule off.

src/javascript.js
```diff
@@ -33,7 +33,8 @@ export default class JavaScriptScanner {
           id: message.ruleId.toUpperCase(),
           message: messages[message.ruleId.toUpperCase()].message,
           description: messages[message.ruleId.toUpperCase()].description,
-          code: message.source,
+          code: message.ruleId.toUpperCase(),
```
tofumatt: We already use `id` for this, so should we ditch the `id` property? I'm fine with that.

src/javascript.js
```diff
@@ -33,7 +33,8 @@ export default class JavaScriptScanner {
...
+          eslintCode: message.source,
```
tofumatt: This is the source code that caused the error, so a property name like `sourceCode` or `sourceText` would be better.

tests/test.messages.js
```diff
@@ -0,0 +1,16 @@
+import * as messages from 'messages';
+
+describe('Messages', function() {
+
+  it('should only have codes with a length < 25', () => {
```
tofumatt: That's a good test to have; prevents `E_TOO_LONG_ERROR_NAMES_THAT_ARE_A_BIT_OVERKILL` too. :+1:

tofumatt:
This looks good; we should stop using `id` at all and switch over to `code` everywhere. I can do that if you like in a separate PR as housekeeping.

`eslintCode` should be something like `source`, `sourceCode`, or `sourceText`.

r+wc

src/javascript.js
muffinresearch: Ok cool, I'll remove id.

src/javascript.js
muffinresearch: Sounds good I'll use that.

PR #58, opened 2015-10-07
Handle invalid JS; add an error when encountered

### LINKED ISSUE #52: Make sure ESLint doesn't fail with invalid JS (syntax_error rule)
This is a test case I hadn't considered in #44; I'll leave it here as a separate bug.

We should be sure bad JS like:

```js
var = foo:
```

doesn't crash the validator.

Turns out this is the `syntax_error` rule. This is a good one to implement early I think.

tofumatt:
Adds a test case for malformed JS and a bit of extra code in the
JS Scanner to make sure we handle SyntaxErrors, which are a bit
different from other errors created by ESLint.

Fixes #52.

### DIFF
--- a/src/javascript.js
+++ b/src/javascript.js
@@ -31,8 +31,9 @@ export default class JavaScriptScanner {
       for (let message of report.results[0].messages) {
         // Fatal error messages (like SyntaxErrors) are a bit different, we
         // need to handle them specially.
-        if (message.fatal) {
+        if (message.fatal === true) {
           message.ruleId = messages.JS_SYNTAX_ERROR.code;
+          // message.severity = VALIDATION_ERROR;
         }
 
         validatorMessages.push({
@@ -46,6 +47,8 @@ export default class JavaScriptScanner {
           type: ESLINT_TYPES[message.severity],
         });
       }
+
+      console.log(validatorMessages);
 
       resolve(validatorMessages);
     });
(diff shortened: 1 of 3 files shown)


tests/test.javascript.js
```diff
@@ -88,7 +88,7 @@ describe('JS Code Checker', function() {
     return jsScanner.scan()
       .then((validationMessages) => {
         assert.equal(validationMessages.length, 1);
-        assert.equal(validationMessages[0].id, 'mozIndexedDB_possible');
+        assert.equal(validationMessages[0].id, 'OBFUSCATION');
```
tofumatt: Just added this as a placeholder over the old identifier.

tests/test.javascript.js
```diff
@@ -121,11 +121,33 @@ describe('JS Code Checker', function() {
...
         assert.equal(validationMessages[0].severity, VALIDATION_WARNING);
```
muffinresearch: You might still need to re-base against master as `severity` is now `type`.

src/javascript.js
```diff
@@ -29,6 +29,12 @@ export default class JavaScriptScanner {
       var report = eslint.executeOnText(this.code, this.filename);
 
       for (let message of report.results[0].messages) {
+        // Fatal error messages (like SyntaxErrors) are a bit different, we
+        // need to handle them specially.
+        if (message.fatal === true) {
```
muffinresearch: Does eslint go no further when it hits a syntax error?

muffinresearch:
r+wc - just needs rebasing and updating `severity` -> `type`

src/javascript.js
tofumatt: Nope. It stops, as it can’t continue parsing the JS when it errors.

PR #60, opened 2015-10-07
Make scanning more generic

muffinresearch:
* Removed addToCollector as this was more aligned to JS Scanning specifics.
* Made `scanFile()` / `scanFiles()` so that we can just re-use for CSS etc. The expectation is now that you provide a list of raw data objects that messages can be created from. A type key is used to create the right message.
* `getJSFiles()` is now `getFilesByExt('.js')`
* Fixed the following tests as they weren't doing what they looked like they we doing.
    * https://github.com/mozilla/addons-validator/blob/master/tests/test.validator.js#L120
    * https://github.com/mozilla/addons-validator/blob/master/tests/test.validator.js#L135

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -1,3 +1,4 @@
+import { extname } from 'path';
 import * as fs from 'fs';
 
 import columnify from 'columnify';
@@ -9,8 +10,8 @@ import * as exceptions from 'exceptions';
 import * as messages from 'messages';
 import { gettext as _, singleLineString } from 'utils';
 
-import JavaScriptScanner from 'javascript';
 import Collector from 'collector';
+import JavaScriptScanner from 'javascript';
 import Xpi from 'xpi';
 
 export var lstat = promisify(fs.lstat);
@@ -25,12 +26,6 @@ export default class Validator {
     this.chalk = new chalk.constructor(
       {enabled: !this.config.boring});
     this.collector = new Collector();
-  }
-
-  addToCollector(message) {
-    var messageType = message.severity.charAt(0).toUpperCase() +
-                      message.severity.slice(1);
-    this.collector[`add${messageType}`](message);
   }
 
   colorize(type) {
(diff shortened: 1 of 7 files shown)


src/validator.js
```diff
@@ -9,8 +10,8 @@ import * as exceptions from 'exceptions';
 import * as messages from 'messages';
 import { gettext as _, singleLineString } from 'utils';
 
-import JavaScriptScanner from 'javascript';
 import Collector from 'collector';
+import JavaScriptScanner from 'javascript';
```
tofumatt: How did my OCD about alphabetising not catch this? :smile:

src/validator.js
```diff
@@ -171,15 +166,62 @@ export default class Validator {
...
+    });
+  }
+
   scan(_Xpi=Xpi) {
     return new Promise((resolve, reject) => {
       this.checkFileExists(this.packagePath)
         .then(() => {
           this.xpi = new _Xpi(this.packagePath);
-          return this.xpi.getJSFiles();
+          return this.xpi.getFilesByExt('.js');
```
tofumatt: Eventually we might need to refactor this to be `ofType(JS)` as we look for JS files without the extension.

Just something to keep in mind. This is way better than what we had before though, :+1:

tests/test.validator.js
```diff
@@ -117,33 +124,49 @@ describe('Validator', function() {
       });
   });
 
-  it('should throw an error if the JS scan fails (file not found)', () => {
+  it('should see an error if scanFiles() blows up', () => {
```
tofumatt: Disappointed there's no :bomb: emoji in this string :wink:

tofumatt:
r+ from me; @muffinresearch is off at a conf and said if it's good I should just merge it, so that's what I'ma do.

tofumatt:
Whoa, sweet job refactoring this too: #58 still merges cleanly. :+1:

src/validator.js
muffinresearch: Yeah good point - there's quite a number of initial assumptions here - but that's fine we can refactor as we go.

PR #63, opened 2015-10-08
Add an RDF parser and one rule (close #48)

### LINKED ISSUE #48: Hook up rdf parser and an RDF related rule.
* Add the rdf parser along similar lines to the rest
* Add an example rule and pass messages to the collector.

tofumatt:
This should implement `shouldnt_exist` and `obsolete` from
https://github.com/mozilla/addons-validator/blob/master/docs/rules.md#installrdf


src/javascript.js
```diff
@@ -50,6 +50,4 @@ export default class JavaScriptScanner {
       resolve(validatorMessages);
     });
   }
-
```
tofumatt: Dunno how these got in here :fearful:

src/messages/rdf.js
```diff
@@ -0,0 +1,41 @@
...
+
+
+export const TAG_NOT_ALLOWED_HIDDEN = {
+  code: 'TAG_NOT_ALLOWED_HIDDEN',
+  message: _('The <hidden> tag is not allowed'),
+  description: _(singleLineString`Your RDF file contains the <hidden> tag,
+    which is not allowed in an Add-on.`),
+};
+
+// NOTE: I didn't want to change the commonality of messages without discussion,
```
tofumatt: I can do this quickly before we merge this if you want; let me know what you think of it.

I imagine for something like RDF where we have a bunch of restricted identifiers easily found it might be the way to go. Possibly down the road with JS too.

tests/test.rdf.js
```diff
@@ -0,0 +1,128 @@
...
+  it('should blow up 💣 when handed malformed XML', () => {
+    var contents = validRDF('<em:hidden>true/em:hidden>');
+    var rdfScanner = new RDFScanner(contents, 'install.rdf');
+
+    return rdfScanner.scan()
+      .then(() => {
+        assert.fail(null, null, 'Unexpected success');
+      })
+      .catch((err) => {
+        console.log(err);
```
muffinresearch: Still needed?

tests/test.validator.js
```diff
@@ -112,15 +112,26 @@ describe('Validator', function() {
...
+  //
+  // Uses our test XPI, with the following file layout:
+  //
+  // - chrome.manifest
+  // - chrome/
+  // - components/
+  //   - main.js (has a mozIndexedDB assignment)
+  //   - secondary.js (nothing bad)
+  // - install.rdf
+  it('should scan all files', () => {
```
muffinresearch: Good call!

src/messages/rdf.js
muffinresearch: Yeah - that makes sense - seems like that would DRY things up. Would you generate them when needed or up front from a list?

tofumatt: I'd rather up-front from a list but I didn't think I export dynamic variable names in JS (eg: `export const TAG_NOT_ALLOWED_${tagName} = tagNotAllowed(tagName);`.

So I'd think we'd generate when needed.

tests/test.rdf.js
tofumatt: Oops, no. I need a pre-commit check for these! :wink:

src/messages/rdf.js
tofumatt: Haha, nevermind, I was thinking in Ruby and trying to metaprogram too much. I'll make them up-front, new commit incoming in a few :smile:

tofumatt:
Alrighty, all fixed up, rebased, and generating the messages dynamically. r?

src/messages/rdf.js
```diff
@@ -0,0 +1,26 @@
...
+export var _tagObsolete = (tagName) => {
+  return {
+    code: `TAG_OBSOLETE_${tagName.toUpperCase()}`,
+    message: _(`The <${tagName}> tag is obsolete`),
+    description: _(singleLineString`Your RDF file contains the <${tagName}> tag,
+      which is obsolete.`),
+  };
+};
+
+export const TAG_NOT_ALLOWED_HIDDEN = _tagNotAllowed('hidden');
```
muffinresearch: Nice!

muffinresearch: I wonder could you iterate over the list to generate the constants that way you've only got to add to a list?

src/validators/rdf.js
```diff
@@ -0,0 +1,105 @@
...
+    //
+    // this.namespace = this._xmlDoc.documentElement._nsMap.em;
+    this.namespace = 'http://www.mozilla.org/2004/em-rdf#';
+    this.validatorMessages = [];
+    this._xmlDoc = null;
+  }
+
+  scan() {
+    return new Promise((resolve, reject) => {
+      this.mustNotExist()
```
muffinresearch: Could the code in `mustNotExist()` be moved to `scan()` or are you expecting to need to expand this further, hence scan needs to wrap other functions?

tofumatt: I was thinking each “type” of rule check would be its own method, so we'd just have a bunch of calls in `scan()`:

```js
scan() {
    this.mustNotExist()
      .then(() => {
        return this.checkForBadValues();
      })
      .then(() => {
        resolve(this.validatorMessages);
      })
      .catch(reject);
  });
}
```
etc.

src/validators/rdf.js
```diff
@@ -0,0 +1,105 @@
...
+
+  mustNotExist() {
+    return new Promise((resolve, reject) => {
+      this.getXMLDoc()
+        .then((xmlDoc) => {
+          var bannedTags = RDF_UNALLOWED_TAGS;
+          var addonIsListed = (xmlDoc
+            .getElementsByTagNameNS(this.namespace, 'listed').length > 0);
+
+          if (addonIsListed) {
```
muffinresearch: Would be cool to add a test for this.

tofumatt: Oh, good call. Will do it now ^_^

muffinresearch:
r+wc

src/validators/rdf.js
tofumatt: Thanks; turns out I wasn't even exporting a message for this one! Yay for testing.

src/messages/rdf.js
tofumatt: I'd like that, but I didn't think one could do `export const $VARIABLE_VARIABLE_NAME = whatever()`.

It was my instinct coming from Ruby :smile: . But I feel like ES6 can't do it? We could make them keys off an object, but then messages become less consistent in their usage (from `TAG_NOT_ALLOWED_HIDDEN` to `TAG_NOT_ALLOWED.HIDDEN`).

PR #70, opened 2015-10-09
Add HTML rule checker (close #54)

### LINKED ISSUE #54: Add HTML parsing tests
There are [existing HTML/Markup tests](https://github.com/mozilla/addons-validator/blob/master/docs/rules.md#html) in the old validator, so we should hook up at least one to see what that looks like.

tofumatt:
WIP for HTML parsing; the cheerio API is possibly a bit nicer but I think it leaves out info we need as a validator, so I'll likely use `xmldom`/`DOMParser` for this as well, instead of `cheerio`. Might add a few convenience functions in though, for things like getting an attribute.

I'll look at this more over the weekend or Monday. No sense reviewing yet, just posting what I have thus far :smile:


muffinresearch:
@tofumatt  Seems travis sees PR branches from mozilla/addons-validator as reflecting the build status - which is odd as it's not master (we can also review the .travis.yml settings and make sure it's not something there). Perhaps if you push to a fork and make PRs from there it will workaround this until we can fix it.

tofumatt:
You mean the badge in the readme?

I'm on my phone now but I do know how to fix that. I'll do it in a few hours, it can take a branch name in the image URL to only display one branch.

Should be the default if you ask me!

- tofumatt (Sent from mobile)

tofumatt:
Redoing this with Cheerio, just using DOMParser. Fewer dependencies is nice and it really isn't more verbose anyway.

tofumatt:
@muffinresearch I think this is ready for r?

I am getting an error here: https://github.com/mozilla/addons-validator/pull/70/files#diff-babdc8cb26d4132818bf2339a71cf2a7R55 and can't quite figure out why. I think I'm missing something obvious. If not I'll comb through more.

muffinresearch:
As a workaround for the error you could look at the name property. I'm not quite sure what's going on - it seems like the getName util in chai is being passed undefined - maybe double check the thing it's being compared with is imported correctly?

The other thing as I mentioned yesterday - xmldom seems to have very minimal tests at 17% coverage 
whereas something like HTMLParser2 [1] has 96%.

[1] https://www.npmjs.com/package/htmlparser2

tofumatt:
Yeah, I’m taking a look at htmlparser today; was going to last night but got a bit sidetracked.

tofumatt:
Okay, this is ready for r? again.

Redid things with cheerio as the test coverage is good and the API is a bit nicer and reduces our own code/test footprint (I didn't have to write a helper to grab attributes for instance).

I'm still searching for a way to get cheerio to output line/column numbers and will add when found, but it might be a different PR if that isn't a huge deal.

Because htmlparser2 is _very_ forgiving there isn't a "bad HTML blows things up" test; instead there's a "bad HTML _doesn't_ blow things up" test. It seems to be built to scrape existing pages which have all kinds of errors.

We'll likely need to enable strict parsing with htmlparser or use XMLDom for strict parsing tests (of which there is one in the HTML parsing checks).

src/rules/html/ensureRequiredAttributes.js
```diff
@@ -0,0 +1,46 @@
...
+import * as messages from 'messages';
+
+
+export var ensureRequiredAttributes = ($, filename=null) => {
+  return new Promise((resolve, reject) => {
+    var promises = [];
+    var validatorMessages = [];
+
+    for (let tag in HTML_TAGS_WITH_REQUIRED_ATTRIBUTES) {
+      promises.push(_ensureAttributesInTag($, tag,
```
tofumatt: Come to think of it this can be simpler; it doesn't need to use Promises here. I'll clean it up a bit.

src/validators/rdf.js
```diff
@@ -40,7 +40,7 @@ export default class RDFScanner {
         errorHandler: (err) => {
           reject(new RDFParseError(`RDF Parse Error: ${err}`));
         },
-      }).parseFromString(this.contents);
+      }).parseFromString(this.contents, 'text/xml');
```
muffinresearch: Should changes to the rdf code be part of this?

tofumatt: Darn, nope, that crept in. It’s not even needed, was just being more explicit there. I’ll take it out :-)

src/rules/html/ensureRequiredAttributes.js
```diff
@@ -0,0 +1,46 @@
...
+
+        if ($(element).attr(attributeName) === undefined) {
+          validatorMessages.push({
+            code: errorCode,
+            message: messages[errorCode].message,
+            description: messages[errorCode].description,
+            sourceCode: `<${tag}>`,
+            file: filename,
+            // TODO: This is where we store severity. Make it nicer?
+            type: attributes[attributeName],
```
muffinresearch: Doesn't this need to be one of VALIDATION_ERROR/NOTICE/WARNING?

tofumatt: It is, but the way it’s stored is bad (see the TODO above). I’ll clean that up, good call.

muffinresearch: Oh I see that's looked-up - seems a bit hard to follow.

tofumatt: Yeah it’s bad. Will fix it before a merge :-)

tests/test.html.js
```diff
@@ -0,0 +1,128 @@
...
+      .then((validatorMessages) => {
+        assert.equal(validatorMessages.length, 0);
+      });
+  });
+
+  it('should not blow up when handed malformed HTML', () => {
+    var html = validHTML('<div>Howdy <!-- >');
+    var htmlScanner = new HTMLScanner(html, 'index.html');
+
+    return htmlScanner.scan()
```
muffinresearch: I think you could probably just do `return htmlScanner.scan();` rather than adding the catch. If the promise is rejected you should see a failure.

tofumatt: Ah, good point.

tests/test.html.js
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  it('should run all rules in rules/html', () => {
+    var contents = validHTML();
+    var htmlScanner = new HTMLScanner(contents, 'index.html');
+    var fakeRule = () => {
+      Promise.resolve([]);
+    };
+    var fakeRules = {
+      iAmAFakeRule: fakeRule,
```
muffinresearch: You could just assign each of these to `sinon.stub();` and then you wouldn't need the extra lines adding the spys.

tofumatt: I did not know that automatically spy’d them. Thanks!

muffinresearch: Yeah the interface for stubs is pretty much the same as spys on the wholle. Spys are generally more useful for wrapping a real func so you can know it was called. If you just want a stand-in then stubs are the better way to go.

muffinresearch:
I wonder does xmlMode make cheerio's use of HTMLParser2 any less forgiving? See https://github.com/cheeriojs/cheerio#loading

tofumatt:
I think I tried that to little effect, but let me look again.

muffinresearch:
Sounds like there's a few tweaks happening - I'll hold off further review - just give me a shout when you want me to take another gander.

tofumatt:
Agreed, sounds good.

tofumatt:
Just tried and `xmlMode` doesn't make htmlparser2 behave any stricter... or at least not strict enough to fail my bad HTML comment test: https://github.com/mozilla/addons-validator/pull/70/files#diff-3579c070636b18c9b5794f5eabe6f237R83

tofumatt:
I like the idea of using a better-tested parser for the majority of our checks, but htmlparser2 seems more forgiving by design, so checking for malformed HTML may need to be handled by a separate library like XMLDom/DOMParser (which is good about it).

src/rules/html/ensureRequiredAttributes.js
```diff
@@ -0,0 +1,46 @@
...
+        resolve(validatorMessages);
+      })
+      .catch(reject);
+  });
+};
+
+var _ensureAttributesInTag = (
+($, tag, attributes, validatorMessages, filename=null) => {
+  return new Promise((resolve) => {
+    $('prefwindow').each((i, element) => {
```
tofumatt: Haha, oops, fixed this (was testing!) :scream:

tofumatt:
Okay, fixed the PR comments and made my tweaks so this should be good for r? again now.

src/rules/html/ensureRequiredAttributes.js
```diff
@@ -0,0 +1,41 @@
...
+      var errorCode = `${tag}_REQUIRES_${attributeName}`.toUpperCase();
+
+      if ($(element).attr(attributeName) === undefined) {
+        validatorMessages.push({
+          code: errorCode,
+          message: messages[errorCode].message,
+          description: messages[errorCode].description,
+          sourceCode: `<${tag}>`,
+          file: filename,
+          type: HTML_TAGS_REQUIRED_ATTRIBUTE_TYPE[`${tag}_${attributeName}`],
```
muffinresearch: Do you expect the type to vary? If not setting this to VALIDATION_ERROR here is bit more straight-forward.

tofumatt: For now I guess I don't. I'm building for the future but a warning for a tag not having an attribute isn't around. Or if it does could be a separate rule. :+1:

src/validator.js
```diff
@@ -245,6 +248,12 @@ export default class Validator {
           return this.scanFiles(cssFiles);
         })
         .then(() => {
+          return this.xpi.getFilesByExt('.html');
```
tofumatt: It occurs to me that using multiple extensions, at least for HTML, would make sense.

I could extend the `getFilesByExt` to take multiple extensions or I could make the call several times to each file type. I feel like the latter is cleaner, but it probably adds more code. Any particular opinion? (I'd go for the first option, eg:

```js
this.xpi.getFilesByExt('.html', '.htm');
```

muffinresearch: Yeah that looks nice.

tofumatt: Cool, just added it along with a test.

src/xpi.js
```diff
@@ -118,11 +118,13 @@ export default class Xpi {
     });
   }
 
-  getFilesByExt(ext) {
+  getFilesByExt(...extensions) {
```
muffinresearch: Shiny :sparkles:

src/xpi.js
```diff
@@ -130,8 +132,10 @@ export default class Xpi {
           let files = [];
 
           for (let filename in metadata) {
-            if (filename.endsWith(ext)) {
-              files.push(filename);
+            for (let ext of extensions) {
+              if (filename.endsWith(ext)) {
```
muffinresearch: It's a shame endsWith doesn't take a list like python.

muffinresearch:
Lgtm r+

src/xpi.js
tofumatt: If it were Rubyists we’d just replace it in `String.prototype` ;-)

tofumatt:
Cheers! Just squashed most of the commits; will merge after I see a green badge :smile:

PR #72, opened 2015-10-10
Quick fix to silence output (fixes #69)

### LINKED ISSUE #69: Make tests less noisy
Output from the tests is a bit noisy because we have `console.log` calls inside the code path that aren't silenced for tests.

screenshot 2015-10-09 21 00 55

Realistically, we probably want an `validator.config.output` option that is `nothing` or `null`.

Not sure when the CLI would ever want it, but it's what we want for the tests and we should already have the infra for it via that `config.output` option.

Other `console.log` should be silenced by this option. We probably just want to use our own `log` method everywhere.

muffinresearch:
(See #69)

This just stubs the code paths that otherwise lead to output.

### DIFF
--- a/tests/test.validator.js
+++ b/tests/test.validator.js
@@ -103,6 +103,7 @@ describe('Validator', function() {
   // - install.rdf
   it('should send JSScanner messages to the collector', () => {
     var addonValidator = new Validator({_: ['tests/example.xpi']});
+    addonValidator.print = sinon.stub();
 
     assert.equal(addonValidator.collector.errors.length, 0);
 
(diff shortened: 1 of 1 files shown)


tests/test.validator.js
```diff
@@ -103,6 +103,7 @@ describe('Validator', function() {
   // - install.rdf
   it('should send JSScanner messages to the collector', () => {
     var addonValidator = new Validator({_: ['tests/example.xpi']});
+    addonValidator.print = sinon.stub();
```
tofumatt: A comment to explain why we're doing this would be handy for the future. Normally I assume stubbing methods is for testing purposes, but this is really just a flourish so it'd be good to clarify why it's here.

tofumatt:
r+wc

I'm actually not sure why we'd want to be more robust in the future (when would we need a validator that didn't output anything? ^_^). I'd say this should do for awhile. :+1:

tests/test.validator.js
muffinresearch: Yep - good call.

PR #76, opened 2015-10-12
Add css scanner (fixes #53, fixes #31)

### LINKED ISSUE #31: Carry out some basic CSS validation rules with the CSS parser
We're planning on using https://github.com/reworkcss/css for this. This bug is to stand-up some basic css checks with that parser.

### LINKED ISSUE #53: Handling CSS parser failures
Deal with inability to process CSS due to parser errors.

muffinresearch:
* Moved the validation tests under tests/validators
* Add CSS Parser
* Handle CSS errors.

Looks like this:

1__bash


src/main.js
```diff
@@ -1,6 +1,8 @@
 import cli from 'cli';
 import Validator from 'validator';
 
+import 'babel-core/polyfill';
```
tofumatt: What is this used for?

tofumatt: (I just would have thought webpack handles this for us or it would go in a config.)

src/messages/css.js
```diff
@@ -0,0 +1,16 @@
...
+  code: 'MOZ_BINDING_EXT_REFERENCE',
+  message: _('Illegal reference to external scripts'),
+  description: _(singleLineString`-moz-binding may not reference external
+    scripts in CSS. This is considered to be a security issue. The script
+    file must be placed in the /content/ directory of the package.`),
+};
+
+export const CSS_SYNTAX_ERROR = {
+  code: 'CSS_SYNTAX_ERRIOR',
+  description: _(singleLineString`An error was found in the CSS file being
```
tofumatt: This one is missing a message.

src/validators/css.js
```diff
@@ -0,0 +1,53 @@
...
+          messages.push(Object.assign({}, CSS_SYNTAX_ERROR, {
+            type: VALIDATION_ERROR,
+            // Use the reason for the error as the message.
+            message: e.reason,
+            column: e.column,
+            line: e.line,
+            file: e.filename,
+          }));
+        }
+        // A syntax erro has been encounted so it's game over.
```
tofumatt: Nit: "error" instead of "erro"

tests/rules/test.css.js
```diff
@@ -0,0 +1,31 @@
+import { BAD_URL_RX } from 'rules/css';
+import { singleLineString } from 'utils';
+
+
+describe('BAD_URL_RX', () => {
+
+  it('should match remote urls', () => {
+    assert.ok(BAD_URL_RX.test('url(http://foo/com)'));
+    assert.ok(BAD_URL_RX.test('url(https://foo/com)'));
```
tofumatt: What about `//foo.com`? Not sure they work in add-ons but if they do we should test for them.

Also why the `/` in the string instead of a `.`?

tests/test.validator.js
```diff
@@ -310,6 +239,119 @@ describe('Validator', function() {
       addonValidator.colorize('whatever');
     }, Error, /colorize passed invalid type/);
   });
+});
+
```
tofumatt: Nice tidying here. :+1:

tofumatt:
Curious about `//foo.com` type URLs, we should probably just test for them anyway; can't hurt.

Looks good to me; r+wc.

muffinresearch:
Ah yeah - good call - the original regex definitely work for protocol free urls but I'll double-check since I made some changes.

src/main.js
muffinresearch: The polyfills for things like Object.assign have to be required/imported.

tofumatt: Ah, cool. Thanks :)

muffinresearch:
Yep that needed updating nice catch!

PR #81, opened 2015-10-13
Require at least node 0.12.x

muffinresearch:
As we're potentially using streams the newer streaming apis are apparently unreliable in older node versions. So I'm making this whole thing bail if we're not using at least 0.12.0. Travis will use 0.12.7 via nvm but system node on osx is 0.12.4.

There's also an .nvmrc to specify 0.12.latest.


tofumatt:
Good idea! r+

PR #82, opened 2015-10-13
Add basic chrome.manifest parser

muffinresearch:
* Adds the basics of a chrome.manifest parser based on the original python code.
* I'll extend this as needed when implementing some rules in a validator tomorrow.

### DIFF
--- a/src/parsers/chromemanifest.js
+++ b/src/parsers/chromemanifest.js
@@ -1,0 +1,73 @@
+import readline from 'readline';
+
+
+export default class ChromeManifestParser {
+
+  constructor(stream, filename) {
+    this.stream = stream;
+    this.filename = filename;
+    this.triples = null;
+  }
+
+  filterTriples({subject, predicate, object} = {}) {
+    return this.parse()
+      .then((triples) => {
+        return new Promise((resolve) => {
+          var filteredTriples = triples.filter((triple) => {
+            var searches = {subject, predicate, object};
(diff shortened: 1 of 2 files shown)


muffinresearch:
r? tofumatt

src/parsers/chromemanifest.js
```diff
@@ -0,0 +1,71 @@
...
+    this.stream = stream;
+    this.filename = filename;
+    this.triples = null;
+  }
+
+  filterTriples({subject, predicate, object} = {}) {
+    return this.parse()
+      .then((triples) => {
+        return new Promise((resolve) => {
+          resolve(triples.filter((triple) => {
```
tofumatt: I'm used to seeing `resolve([args])` but not a function inside it.

It took me a second to make sense of this only because I don't expect it. I know it's slightly more verbose but what about:

```js
var filteredTriples = triples.filter((triple) => {
  var searches = {subject, predicate, object};
  for (let key in searches) {
    let val = searches[key];
    if (typeof val !== undefined && val === triple[key]) {
      return true;
    }
  }
});

resolve(filteredTriples);
```

tofumatt:
r+wc; curious about your thoughts on the promise `resolve()` call, otherwise looks good.

src/parsers/chromemanifest.js
muffinresearch: Nice, I like it - definitely reads better.

Issue #78, opened 2015-10-13
Quick review of existing structure before we move onto adding more rules.


muffinresearch:
Before we go too far and add lots of rules I think it might be worth taking a step back briefly to look at what we have in terms of parsers and how they are integrated and make sure we are being as consistent in our approach as possible for each parser so far.

Kind of questions I have in mind are:

* Use of rule functions vs class methods.
* How are the message objects setup and passed back to the collector.

We can then file issues for any things that need to be changed if necessary.

tofumatt:
Which do you prefer? I realised I was doing class methods in the HTML parser but even now I’m now sure I like it, and I could see it getting crazy out of hand (a file with a thousand lines? :-/) pretty fast.

Should we be moving to a list of rules in a folder that we import as functions? Something like

    - rules
      - html
        - mustNotExist.js
        - idIsRequired.js
      - js
        - eval.js
        - mozIndexedDB.js

etc.?

muffinresearch:
Yeah I went something along the lines of that direction with the CSS scanner as I thought the eslint style of rule funcs seemed quite a nice approach. 

A file per rule might be a bit much but depends it might be best if there are loads.

As a side note I noticed coverage  didn't seem to see the js rules but I suspect that's because they're not directly imported due to eslint.

muffinresearch:
Definitely, something in the back of our minds should be making it super easy for new rules to be added. This way it will make it much easier to get help with the not insignificant amount of rules that need to be implemented.

tofumatt:
Ah. That would be a problem. I could probably import them directly and feed them the kind of data ESLint expects, but the rules do get tested. I’d rather not tell the coverage checker to ignore that folder, but the rules are tested via eslint and not directly, true.

tofumatt:
Okay cool, I'll move the #70 rules to that "rule-per-file" style, as well as the RDF ones: #79.

tofumatt:
We assessed a fair bit of stuff and made it better... we can close this now I think. Thoughts @muffinresearch?

muffinresearch:
Yep lets close this and raise issues for specifics.

PR #84, opened 2015-10-14
Move RDF + JS rules to their own folders

tofumatt:
Related to #78; this moves the RDF rules into their own folder and the rule tests into a separate file.

Note: This PR also moves JS rules to their own folder, but in its own commit.


tofumatt:
Good to r? now

src/validators/javascript.js
```diff
@@ -20,7 +20,7 @@ export default class JavaScriptScanner {
       // pass it the entire source file as a string.
       let eslint = new ESLint.CLIEngine({
         ignore: false,
-        rulePaths: [path.join('src', 'rules')],
+        rulePaths: [path.join('src', 'rules', 'javascript')],
```
muffinresearch: Is the join needed? Won't this just be 'src/rules/javascript'

src/rules/rdf/mustNotExist.js
```diff
@@ -0,0 +1,57 @@
...
+    validatorMessages = validatorMessages.concat(
+      _checkForTags(xmlDoc, namespace, RDF_OBSOLETE_TAGS, VALIDATION_WARNING,
+                    'TAG_OBSOLETE_', filename));
+
+    resolve(validatorMessages);
+  });
+};
+
+export var _checkForTags = (
+(xmlDoc, namespace, tags, severity, errorCodePrefix, filename=null) => {
```
muffinresearch: Seems like a weird indent.

muffinresearch: Do you need the fat arrow function? 

Also perhaps use an object for the args - that will make it clearer in the caller what args you're passing.

e.g. 

```
function _checkForTags({xmlDoc, namespace, tags, severity, errorCodePrefix, filename=null} = {}) {
  // use any arg passed in the object directly here.
}

_checkForTags({severity: 'hai'});

```

Here's an example showing that sort of thing in action:

http://jsbin.com/cowaki/edit?js,console

muffinresearch:
Looks good to me aside from possibly tweaking the few things in the comments.

r+wc

src/rules/rdf/mustNotExist.js
tofumatt: Ah, that's a good idea. I too thought the function signature was too long.

src/validators/javascript.js
tofumatt: I always worry about Windows users, in case a module (like `ESLint.CLIEngine`) isn't normalising the paths itself.

PR #90, opened 2015-10-14
Use terminal width for CLI wrapping

tofumatt:
I was using a larger-than-usual terminal window and noticed the output was still wrapped to 78 characters.

This uses process.stdout.columns if available to set the CLI wrap value. (Minus two pixels to preserve a similar padding effect as using 78 over 80).

Kind of a superfluous tweak, but I added a bunch of tests for it and it's nice to be a good CLI citizen like this.

#### Before:
screenshot 2015-10-14 19 59 40

#### After:
screenshot 2015-10-14 20 16 53

### DIFF
--- a/tests/test.validator.js
+++ b/tests/test.validator.js
@@ -5,6 +5,7 @@ import * as messages from 'messages';
 import CSSScanner from 'validators/css';
 import { DuplicateZipEntryError } from 'exceptions';
 import { fakeMessageData } from './helpers';
+import { singleLineString} from 'utils';
 
 
 describe('Validator', function() {
@@ -362,6 +363,18 @@ describe('Validator.textOutput()', function() {
     return 1000;
   }
 
+  function mediumTerminalWidth() {
+    return 77;
+  }
+
+  function smallTerminalWidth() {
+    return 59;
+  }
+
+  function uselesslyTinyTerminalWidth() {
+    return 1;
+  }
+
   it('should have error in textOutput()', () => {
     var addonValidator = new Validator({_: ['bar']});
     addonValidator.collector.addError({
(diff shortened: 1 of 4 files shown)


muffinresearch:
You might want to check the text output too. I glossed over trying to make it work in small terminals because it got cramped fast. But if you can work out a solution there too that fits in with this that would be nice too.

tofumatt:
Oh, I figured they were rarely under 78. But I'll make the minumum size lower than 78. 10 should be reasonable.

- tofumatt (Sent from mobile)

muffinresearch:
We might be at cross-purposes I'm talking about the valdation error / warning text output.

tofumatt:
Oh my you’re right, that is weird.

I’ll check this out more and push when it’s all good, thanks!

muffinresearch:
We can always deal with that in a separate bug - it's all good :)

tofumatt:
Yeah, looks like it’s always been that way.

I fixed it here; now the validator output is dependant on the size of the terminal, so it looks better in smaller terminals.

muffinresearch:
One thing is the error-codes if the colorization is applied and the heading width is too narrow the colour will bleed into the rest of the content. I filed a bug with columnify [1] upstream to deal with this - but until then we'd need to watch that or just take the colours out.

https://github.com/timoxley/columnify/issues/26

tofumatt:
Yeah, I set the heading width to always be 35 regardless of window size, so that shouldn’t happen (in my testing it didn’t, we could write a test for it though combing through those codes to test is a PITA for an upstream bug).

muffinresearch:
r+wc

tofumatt:
I added a few more tests and removed certain columns in smaller terminals. I added tests for that as well as for terminals as small as 1 column. Tested it in my terminal too, everything seems better now.

tests/test.validator.js
```diff
@@ -393,11 +412,61 @@ describe('Validator.textOutput()', function() {
...
+    assert.equal(addonValidator.output.summary.errors, 1);
+    assert.notInclude(text, 'Description');
+    assert.notInclude(text, 'whatever error description');
+    assert.notInclude(text, 'Column');
+    assert.notInclude(text, '5');
+    assert.notInclude(text, 'Line');
+    assert.notInclude(text, '20');
+  });
+
+  it('should survive even a 1 column terminal', () => {
```
muffinresearch: That's a niche use-case :)

tofumatt: Indeed, but now we know it won’t crash! :)

muffinresearch:
Works as advertised. r+

PR #91, opened 2015-10-14
Add chrome.manifest scanner + rule

### LINKED ISSUE #47: Add a parser + single rule for chrome.manifest.
* Hook up something to read a chome manifest
* Connect a rule so that it's passing messages into the collector.

muffinresearch:
Fixes #47 

I'm not yet sure if there might be more than one chrome.manifest - but will look into that and file a further bug to handle that later.


src/validator.js
```diff
@@ -195,18 +201,16 @@ export default class Validator {
       case '.rdf':
         return RDFScanner;
       default:
-        throw new Error('No scanner available for ${filename}');
+        throw new Error(`No scanner available for ${filename}`);
```
tofumatt: Haha, didn't catch this. Whoops!

tofumatt:
Is there any reason not to just check for `.manifest` files in case there are multiple files the same way we do with `.js`, etc.?

src/rules/chromemanifest/categories.js
```diff
@@ -0,0 +1,35 @@
...
+  'JavaScript-DOM-class',
+  'JavaScript-DOM-interface',
+];
+
+
+export function checkCategories(triples) {
+  var validationMessages = [];
+
+  for (let triple of triples) {
+    if (triple.subject === 'category' &&
```
tofumatt: That's a helluva `if` statement!

src/validator.js
```diff
@@ -231,6 +235,18 @@ export default class Validator {
...
         .then(() => {
           this.xpi = new _Xpi(this.packagePath);
+          return this.xpi.getMetaData();
+        })
+        .then((metadata) => {
+          var chromeManifest = 'chrome.manifest';
+          if (Object.keys(metadata).indexOf(chromeManifest) > -1) {
+            return this.scanFile(chromeManifest, 'stream');
+          } else {
+            // TODO: Log this.
```
tofumatt: This should be an error, right? (Is a chrome manifest required?)

tests/validators/test.chromemanifest.js
```diff
@@ -0,0 +1,55 @@
...
+  it('should detect dangerous categories', () => {
+    var rstream = new Readable();
+    rstream.push('content  necko   jar:comm.jar!/content/necko/\n');
+    rstream.push('category JavaScript-DOM-class foo bar\n');
+    rstream.push(null);
+    var cmScanner = new ChromeManifestScanner(rstream, 'chrome.manifest');
+    return cmScanner.scan()
+      .then((messages) => {
+        assert.equal(messages.length, 1);
+        assert.equal(messages[0].code, 'DANGEROUS_CATEGORY');
```
tofumatt: I've been importing `messages` and doing `messages.DANGEROUS_CATEGORY.code` to reduce having to change things if things change... but now I wonder if that's just more verbose as we want the codes to be the same as the constant name.

Maybe I should go through tests I've written and change mine to this way?

src/validator.js
muffinresearch: I'm not yet sure about that, it doesn't look like it looking at some of the code in the original [1]

https://github.com/mozilla/amo-validator/blob/master/tests/test_chromemanifest_testcases.py#L19

tests/validators/test.chromemanifest.js
muffinresearch: I think your way is better - I should switch to that.

muffinresearch:
> Is there any reason not to just check for .manifest files in case there are multiple files the same way we do with .js, etc.?

I don't think so at the moment. Obviously if we needed to we could change it later. Apparently for Gecko 2.0 onwards only the root manifest is the one parsed [1] . We need to lock-down what versions we support but I'd be very surprised if we need to worry about pre 2.0.

[1] https://developer.mozilla.org/en/docs/Chrome_Registration

tests/validators/test.chromemanifest.js
muffinresearch: I should move these to rule tests too, I'll do that now.

src/validator.js
muffinresearch: I think we'll need to work this out for the layout tests following actually knowing what extension type we're validating it. Tracking down precisely what the rules are for this might be a pain. But we should probably use the existing validator a guide.

tofumatt:
All those answers make sense. :+1: to moving the rules into their own directory.

r+

PR #93, opened 2015-10-15
Use our grunt task for npm start

tofumatt:
Super-simple tweak; means webpack shouldn't need to be installed globally to run the `npm start` script, which is the node build tool-agnostic command for "get this app to work" to my mind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Mozilla addons validator",
   "main": "index.js",
   "scripts": {
-    "start": "webpack",
+    "start": "node -e \"require('grunt').cli()\" null webpack:build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Mozilla addons validator",
   "main": "index.js",
   "scripts": {
-    "start": "webpack",
+    "start": "node -e \"require('grunt').cli()\" null webpack:build",
```
muffinresearch: You could add watch too for rebuild on changes?

muffinresearch:
r+wc

package.json
tofumatt: To the same task? Good call.

PR #94, opened 2015-10-15
Pass back message list instead of passing it

### LINKED ISSUE #88: Refactor CSS rules to pass back a list
Following discussions we're going to have all rule funcs pass back a list of errors rather than passing the list of errors into each rule func.

muffinresearch:
Fixes #88


tofumatt:
r+

PR #98, opened 2015-10-15
Test for ESLint {allowInlineConfig: false} (close #7)

### LINKED ISSUE #7: Get ESLint to ignore /*eslint*/ comments
We need a way to either configure ESLint not to respect comments inside the code (usually in `/* eslint nocheckforbadthing */`-type comments).

If that's not possible, we'll need to strip out all comments when we parse code or something so the code can't tell the validator to back off :smile:

tofumatt:
Currently fails, but I'm on it :wink:

### DIFF
--- a/tests/scanners/test.javascript.js
+++ b/tests/scanners/test.javascript.js
@@ -141,7 +141,7 @@ describe('JavaScript Scanner', function() {
   });
 
   // Depends on: https://github.com/mozilla/addons-validator/issues/7
-  it.skip('ignores /*eslint-disable*/ comments', () => {
+  it('ignores /*eslint-disable*/ comments', () => {
     var code = '/*eslint-disable*/\n';
     code += 'var myDatabase = indexeddb || mozIndexedDB;';
     var jsScanner = new JavaScriptScanner(code, 'badcode.js');
@@ -149,7 +149,7 @@ describe('JavaScript Scanner', function() {
     return jsScanner.scan()
       .then((validationMessages) => {
         assert.equal(validationMessages.length, 1);
-        assert.equal(validationMessages[0].id, 'mozIndexedDB');
+        assert.equal(validationMessages[0].code, messages.MOZINDEXEDDB.code);
       });
   });
 
--- a/tests/scanners/test.javascript.js
+++ b/tests/scanners/test.javascript.js
@@ -1,3 +1,5 @@
+import ESLint from 'eslint';
+
 import { ESLINT_RULE_MAPPING, VALIDATION_ERROR,
          VALIDATION_WARNING } from 'const';
 import { MissingFilenameError } from 'exceptions';
@@ -5,7 +7,7 @@ import JavaScriptScanner from 'scanners/javascript';
 import * as messages from 'messages';
 import * as rules from 'rules/javascript';
 import { ignorePrivateFunctions, singleLineString } from 'utils';
-import { getRuleFiles, unexpectedSuccess } from '../helpers';
+import { getRuleFiles, getVariable, unexpectedSuccess } from '../helpers';
 
 
 describe('JavaScript Scanner', function() {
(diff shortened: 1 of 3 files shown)


tofumatt:
Wondering about this after #145, but I'll leave it open for now.

tofumatt:
Update on this: eslint won't see it merged until I can rebase onto an API fix for their own stuff... so I'm gonna add a "blocked" label to this for now. I may need to make said fix myself if it doesn't happen soon to push this along :unamused:

tofumatt:
ESLint has merged the patch to allow us this, so once a new version is out this'll work! Yay!

tofumatt:
Aaaaand finally r? :smile:

muffinresearch:
\o/

Does this kill all inline config including env and globals as well as the disabled lines? 

E.g. 

`/*eslint-env node, mocha */`

and 

`/*global var1, var2*/`

tofumatt:
It does, and while those tests exist in ESLint (I wrote them there—https://github.com/eslint/eslint/pull/4186/files#diff-d2c296e7287e8fa768390287cf1246a3R2290), I didn’t duplicate them all here.

But we certainly could if you like, I don’t mind :-)

muffinresearch:
Yeah it might be nice to - just so we will know if the dep changed beneath us at any point in the future.

tofumatt:
Okay, I can test for that but I need to use ESLint directly as we don't have rules that check for these kinds of things.

muffinresearch:
Yep - I can see that waiting until we use something that does something with globals makes sense.

For env comments though  could we try turning off es6 and then  using es6 syntax in some code. If that was allowed you should get a syntax error. E.g. the test should fail with old eslint version. 

Can always do this as a follow-up if it's easier - it's probably more "nice to have" than anything.

tofumatt:
Oh, good point. I'll try that!

tofumatt:
Added extra tests (the global one is cribbed from ESLint). r?

tests/scanners/test.javascript.js
```diff
@@ -141,18 +143,68 @@ describe('JavaScript Scanner', function() {
...
       .then((validationMessages) => {
         assert.equal(validationMessages.length, 1);
-        assert.equal(validationMessages[0].id, 'mozIndexedDB');
+        assert.equal(validationMessages[0].code, messages.MOZINDEXEDDB.code);
+      });
+  });
+
+  // This should not cause a syntax error; it should still be parsing code
+  // as ES6 because it ignores the env change.
+  it('ignores /*eslint-env*/ comments', () => {
```
muffinresearch: This one doesn't seem to fail with the older version of eslint. I'm not sure why that is.

muffinresearch:
r+wc

tests/scanners/test.javascript.js
tofumatt: Looks like it never worked, I’ll add a test that _adds_ an env then, which seems the only option.

muffinresearch: Ah - yeah, my suggestion was bad - I did see something saying you could only add not remove.

Still, leave this test just in-case that changes.

tofumatt: Haha okay. I just took it out but I’ll add it back just to be safe ^_^

- tofumatt

On 23 November 2015 at 12:45:38, Stuart Colville (notifications@github.com) wrote:

In tests/scanners/test.javascript.js:

>      var jsScanner = new JavaScriptScanner(code, 'badcode.js');
>   
>      return jsScanner.scan()
>        .then((validationMessages) => {
>          assert.equal(validationMessages.length, 1);
> -        assert.equal(validationMessages[0].id, 'mozIndexedDB');
> +        assert.equal(validationMessages[0].code, messages.MOZINDEXEDDB.code);
> +      });
> +  });
> +
> +  // This should not cause a syntax error; it should still be parsing code
> +  // as ES6 because it ignores the env change.
> +  it('ignores /*eslint-env*/ comments', () => {
Ah - yeah, my suggestion was bad - I did see something saying you could only add not remove.

Still, leave this test just in-case that changes.

PR #99, opened 2015-10-16
Add logger

muffinresearch:
This adds bunyan logging to the validator.

* By default logging is pretty much off (set to 'fatal') .
* Logging in tests can be enabled with `LOG_LEVEL=debug grunt test`
* Logging on the cli can be enabled with `--log-level [level]` as a parameter.
* Bunyan by default logs JSON. If you want the json to be pretty printed pipe anything that logs into `bunyan` e.g. `LOG_LEVEL=debug grunt test | node_modules/bunyan/bin/bunyan`

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,0 +1,7 @@
+import bunyan from 'bunyan';
+
+export default bunyan.createLogger({
+  name: 'AddonValidatorJS',
+  stream: process.stdout,
+  level: process.env.LOG_LEVEL || 'fatal',
+});
--- a/README.md
+++ b/README.md
@@ -30,3 +30,14 @@ Tests use `grunt` but don't require global `grunt`. Just run:
 ```
 npm test
 ```
+
+## Logging
+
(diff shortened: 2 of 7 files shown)


src/logger.js
```diff
@@ -0,0 +1,7 @@
+import bunyan from 'bunyan';
+
+export default bunyan.createLogger({
+  name: 'AddonValidatorJS',
+  stream: process.stdout,
+  level: process.env.LOG_LEVEL || 'fatal',
```
tofumatt: Could we have a test for this so a log level other than an acceptable one either errors or defaults to fatal? I like an error better, just in case someone types a log level in that's wrong and gets stumped by the output.

tofumatt:
r+wc, cool.

src/logger.js
muffinresearch: Sure, that sounds like a plan.

muffinresearch: I'll do this as a follow-up just to save context-switching. I've added https://github.com/mozilla/addons-validator/issues/100 so it doesn't get forgotten.

PR #105, opened 2015-10-19
Add Addon type detection

### LINKED ISSUE #75: Detect addon-type
We need to know what addon-type we are dealing with as some validation tests are type-specific.

muffinresearch:
Fixes #75

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -28,3 +28,34 @@ export const RDF_OBSOLETE_TAGS = ['file', 'requires', 'skin'];
 export const HTML_TAGS_WITH_REQUIRED_ATTRIBUTES = {
   prefwindow: ['id'],
 };
+
+// Package type constants.
+export const PACKAGE_ANY = 0;
+export const PACKAGE_EXTENSION = 1;
+export const PACKAGE_THEME = 2;
+export const PACKAGE_DICTIONARY = 3;
+export const PACKAGE_LANGPACK = 4;
+export const PACKAGE_SEARCHPROV = 5;
+export const PACKAGE_MULTI = 1; // A multi extension is an extension
+export const PACKAGE_SUBPACKAGE = 7;
+
+export const PACKAGE_TYPES = {
+  PACKAGE_ANY: PACKAGE_ANY,
+  PACKAGE_EXTENSION: PACKAGE_EXTENSION,
+  PACKAGE_THEME: PACKAGE_THEME,
+  PACKAGE_DICTIONARY: PACKAGE_DICTIONARY,
+  PACKAGE_LANGPACK: PACKAGE_LANGPACK,
+  PACKAGE_SEARCHPROV: PACKAGE_SEARCHPROV,
+  PACKAGE_MULTI: PACKAGE_MULTI,
+  PACKAGE_SUBPACKAGE: PACKAGE_SUBPACKAGE,
+};
+
+// Types in install.rdf don't match the types
+// we use internally. This provides a mapping.
+export const INSTALL_RDF_TYPE_MAP = {
+  2: PACKAGE_EXTENSION,
+  4: PACKAGE_THEME,
+  8: PACKAGE_LANGPACK,
+  32: PACKAGE_MULTI,
+  64: PACKAGE_DICTIONARY,
+};
--- a/src/messages/layout.js
+++ b/src/messages/layout.js
@@ -8,3 +8,36 @@ export const DUPLICATE_XPI_ENTRY = {
     with the same name. This practice has been banned. Try unzipping
     and re-zipping your add-on package and try again.`),
 };
+
+export const TYPE_NO_INSTALL_RDF = {
+  code: 'TYPE_NO_INSTALL_RDF',
+  message: _('install.rdf was not found'),
+  description: _(singleLineString`The type should be determined by
+    install.rdf if present. As there's no install.rdf, type detection
+    will be attempted to be inferred by package layout.`),
+};
+
+export const TYPE_INVALID = {
+  code: 'TYPE_INVALID',
+  message: _('Invalid <em:type> value'),
+  description: _(singleLineString`The only valid values for <em:type>
+    are 2, 4, 8, and 32. Any other values are either invalid or
+    deprecated.`),
+  filename: 'install.rdf',
+};
+
+export const TYPE_MISSING = {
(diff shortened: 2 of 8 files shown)


src/const.js
```diff
@@ -28,3 +28,34 @@ export const RDF_OBSOLETE_TAGS = ['file', 'requires', 'skin'];
...
 };
+
+// Package type constants.
+export const PACKAGE_ANY = 0;
+export const PACKAGE_EXTENSION = 1;
+export const PACKAGE_THEME = 2;
+export const PACKAGE_DICTIONARY = 3;
+export const PACKAGE_LANGPACK = 4;
+export const PACKAGE_SEARCHPROV = 5;
+export const PACKAGE_MULTI = 1; // A multi extension is an extension
```
tofumatt: Heh, that confused me at first! :smile:

src/const.js
```diff
@@ -28,3 +28,34 @@ export const RDF_OBSOLETE_TAGS = ['file', 'requires', 'skin'];
...
+  PACKAGE_EXTENSION: PACKAGE_EXTENSION,
+  PACKAGE_THEME: PACKAGE_THEME,
+  PACKAGE_DICTIONARY: PACKAGE_DICTIONARY,
+  PACKAGE_LANGPACK: PACKAGE_LANGPACK,
+  PACKAGE_SEARCHPROV: PACKAGE_SEARCHPROV,
+  PACKAGE_MULTI: PACKAGE_MULTI,
+  PACKAGE_SUBPACKAGE: PACKAGE_SUBPACKAGE,
+};
+
+// Typs in install.rdf don't match the types
```
tofumatt: Nitpick here, but that should say "Types" not "Typs"

src/messages/layout.js
```diff
@@ -8,3 +8,36 @@ export const DUPLICATE_XPI_ENTRY = {
     with the same name. This practice has been banned. Try unzipping
     and re-zipping your add-on package and try again.`),
 };
+
+export const TYPE_NO_INSTALL_RDF = {
+  code: 'TYPE_NO_INSTALL_RDF',
+  message: _('install.rdf was not found'),
+  description: _(singleLineString`The type should be determined by
```
tofumatt: Sorry, I can't quite make sense of this description. It's saying "No install.rdf was found so we're guessing your add-on type, but we can't be sure.", right?

src/messages/layout.js
```diff
@@ -8,3 +8,36 @@ export const DUPLICATE_XPI_ENTRY = {
...
+    install.rdf if present. As it isn't falling back to a best-guess
+    approach.`),
+};
+
+export const TYPE_INVALID = {
+  code: 'TYPE_INVALID',
+  message: _('Invalid <em:type> value'),
+  description: _(singleLineString`The only valid values for <em:type>
+    are 2, 4, 8, and 32. Any other values are either invalid or
```
tofumatt: I remember seeing this on MDN and thinking "Why no love for 16"?!

src/validator.js
```diff
@@ -47,6 +50,92 @@ export default class Validator {
...
+              }
+
+              // TODO: This follows the amo-validator approach.
+              // There's no type element, so the spec says that it's either a
+              // theme or an extension. At this point, we know that it isn't
+              // a dictionary, language pack, or multiple extension pack.
+              log.warn('Falling back to decide type from package extension');
+
+              var extensions = {
+                '.jar': '4',
```
tofumatt: These should use the constants for extension type defined above instead of strings, I think.

src/validator.js
```diff
@@ -271,14 +360,22 @@ export default class Validator {
         })
         .then(() => {
           this.xpi = new _Xpi(this.packagePath);
+          return this.detectPackageType();
+        })
+        .then((packageType) => {
+          this.packageType = packageType;
+          if (packageType) {
+            log.debug('Package type detected as %s',
```
tofumatt: While in this particular case the formatting probably _does_ look nicer than having to include `singleLineString`, we should be consistent and use template strings everywhere rather than `%s`.

tests/test.validator.js
```diff
@@ -476,3 +477,199 @@ describe('Validator.textOutput()', function() {
...
+
+
+describe('Validator.detectType()', function() {
+
+  it('should reject on multiple em:type nodes', () => {
+    var addonValidator = new Validator({_: ['bar']});
+    addonValidator.xpi = {
+      getFileAsString: () => {
+        return Promise.resolve(singleLineString`
+          <?xml version='1.0' encoding='utf-8'?>
```
tofumatt: There's a `validRDF()` helper in [`tests/helpers.js`](https://github.com/mozilla/addons-validator/blob/master/tests/helpers.js#L22) that reduces the boilerplate in these tests. Instead of this whole string you just need:

```js
validRDF(singleLineString`
<em:type>2</em:type>
<em:type>2</em:type>
`)
```

tofumatt:
r+wc

I think the message for `TYPE_NO_INSTALL_RDF` could be a bit clearer, but otherwise good once comments are addressed. I like all the logging!

src/messages/layout.js
muffinresearch: Yeah I'll make it clearer.

src/validator.js
muffinresearch: Yeah good idea. Copied this pretty much verbatim from the python code but that would save some look-ups.

src/validator.js
muffinresearch: Could do, this is a feature of the specific logging stuff we're using though.

tests/test.validator.js
muffinresearch: Ah nice, I shall use it :)

src/const.js
muffinresearch: Thanks - call me Mr Typo.

src/validator.js
tofumatt: If there’s an extra benefit to doing the `%s` stuff for logging purposes I’m cool with it. Otherwise just seems easier to be consistent with string usage.

src/const.js
tofumatt: I’d have gone with "Mr. Typ” 😉

src/validator.js
muffinresearch: Yep - that's a lot clearer now.

PR #121, opened 2015-10-20
Add architecture docs

### LINKED ISSUE #107: Document rule functions

muffinresearch:
Fixes #107 

TODO: Will add the architecture flow image once that's landed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,15 @@
 
 The Add-ons Validator, JS edition.
 
-## Required node version
+## Development
+
+### Required node version
 
 Node v0.12.x or greater is required. Using nvm is probably the easiest way
 to manage multiple node versions side by side. See
 [nvm on github](https://github.com/creationix/nvm) for more details.
 
-## Install dependencies
+### Install dependencies
 
 Install dependencies with [npm](http://nodejs.org/):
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,3 +43,45 @@ We use [bunyan](https://github.com/trentm/node-bunyan) for logging:
...
+### Scanners
+
+Generally each file-type has a scanner. For example we have a CSS and a
+JavaScript scanner amongst others. Each scanner looks at relevant
+files and passes each file through a parser which then hands off to
+various rules that look for specific things.
+
+### Rules
+
+Rules are generally a single function in a file organized by directories
```
tofumatt: No need for "generally" here, as our aim is to always have a one-to-one rule-to-file-to-function ratio. Private functions don't count, so what we're getting at is more:

> Rules are exported via a single function in a single file. A rule can have private functions it uses internally, but rule code should not depend on another rule file and only one function should be exported per file.

That seems oddly worded too, but it's what we want to get at I think?

tofumatt:
This is really handy stuff, especially as new folks come on-board!

r+wc, maybe I'm just being too weird about the rule text, but feel free to comment on it/tweak it if you feel I was on the right track.

README.md
muffinresearch: Ok cool :+1:

PR #131, opened 2015-10-21
Add done? field and code refs to rules table

muffinresearch:
Press "view" on the markdown file to get a better idea of what this looks like.


muffinresearch:
A thought with this is that we could add a test that looks in the markdown to see if any new message codes are missing. That would then remind us to keep this up to date.

tofumatt:
Looks good; and yeah, a test could be neat. Oh, I just saw the other PR, I'll look there too :+1: 

r+

PR #133, opened 2015-10-21
Create base class for CSS/HTML/RDF validators

### LINKED ISSUE #83: Refactor HTML/RDF scanner classes
Looking at the HTML and RDF scanner classes, they're nearly the same now that I've started on #79. Might be nice to define a "markup" scanner class and have them simply extend it (essentially the only difference is the parser used and the rules they're using).

### LINKED ISSUE #127: Add check to HTML and RDF scanners to check the rule is a function before trying to run it.
This is in-case we do export other things for tests. See https://github.com/mozilla/addons-validator/blob/master/src/validators/html.js#L21

### LINKED ISSUE #128: Exclude rule functions that start with '_' from being run.
The rdf scanner has this feature it would be nice to make that common across the board.

https://github.com/mozilla/addons-validator/blob/master/src/validators/rdf.js#L31

tofumatt:
This creates a `BaseScanner` class that several scanners with very similar code can extend. DRYs up a fair bit of code and tests.

I was starting on #128, but realised a good opportunity to do a refactor of
some of the validators.

The CSS validator could be DRY'd further but this should still consolidate
functionality.

This moves tests for rules into the base validator tests and out of the
CSS/HTML/RDF parser tests; all classes that extend the Base class implement.

* Closes #83
* Closes #127
* Closes #128

This also renames the `validators` folder to `scanners`, keeping things a bit more consistent.


tofumatt:
I was hoping this would be more red than green, darn.

src/utils.js
```diff
@@ -64,3 +64,26 @@ export function getPackageTypeAsString(numericPackageType) {
   }
   throw new Error(`Invalid package type constant "${numericPackageType}"`);
 }
+
+/*
+ * Looks throuh all exported functions and returns only
```
muffinresearch: Typo

src/utils.js
```diff
@@ -64,3 +64,26 @@ export function getPackageTypeAsString(numericPackageType) {
...
+ * Rules can have private functions we don't run; anything that
+ * starts with an "_" shouldn't be returned.
+ *
+ * This exists because we export private functions in rule files
+ * for testing.
+ */
+export function ignorePrivateFunctions(list) {
+  var filteredList = {};
+
+  for (let functionName in list) {
```
muffinresearch: No love for Array.filter?

tests/scanners/test.base.js
```diff
@@ -0,0 +1,74 @@
...
+
+  it('should reject when _getContents is not implemented', () => {
+    var baseScanner = new BaseScanner('', 'index.html');
+
+    return baseScanner.scan()
+      .then(() => {
+        assert.fail(null, null, 'Unexpected success');
+      })
+      .catch((err) => {
+        assert.instanceOf(err, NotImplentedError);
```
muffinresearch: Nice!

muffinresearch:
> This also renames the validators folder to scanners, keeping things a bit more consistent.

Good call

src/scanners/css.js
```diff
@@ -0,0 +1,67 @@
+import BaseScanner from 'scanners/base';
+import * as cssParser from 'css';
+import * as rules from 'rules/css';
+import { CSS_SYNTAX_ERROR } from 'messages';
+import { VALIDATION_ERROR } from 'const';
+import { ignorePrivateFunctions } from 'utils';
+
+
+export default class CSSScanner extends BaseScanner {
```
muffinresearch: Shame extending didn't dry things up more for this CSS scanner. I wonder if we could re-think and maybe optionally pre-process the descriptors somehow to avoid repeating most of the scan function.

muffinresearch:
Could the chrome manifest scanner not be moved to this too?

muffinresearch:
Like this a lot - there's a few questions in there but I'm otherwise r+wc

src/scanners/css.js
tofumatt: Yeah I think we could, I just didn’t get to it right away. I’ll spend some more time today on that and see what I can get.

src/utils.js
tofumatt: Problem is, it’s actually an object (key/value) because they’re imported using `* as rules from ‘rules/html’`.

I’ll change the name of the argument to something that makes that clear and/or add a comment.

I actually used .filter at first then wondered why my tests failed ;-)

muffinresearch: Oh crap yeah I didn't spot that. :) It's funny I always get tripped up by in/of too.

src/scanners/css.js
muffinresearch: Sounds like a plan to look at that separately to this patch too.

tofumatt: Agreed; I've created #139 for that.

tofumatt:
Good call, I'll see about moving the ChromeManifest to this in the same PR.

muffinresearch:
@tofumatt probably be good to land this without the chrome.manifest stuff and make that a separate patch unless you'd prefer to include it?

tofumatt:
Yeah I can do that; haven’t finished the patch yet so I’ll submit what I have and do the manifest bit as a separate branch. This one’s already pretty busy!

PR #136, opened 2015-10-22
Add docs for end users

### LINKED ISSUE #115: Add docs to README for end-users
Add docs about `npm install` and usage for end-users.

Even though we aren't at rule-parity with the Python validator, this will let adventurous users install it and try it out!

I'm working on this but filing for posterity and so no one else wastes their time :smile:

tofumatt:
Closes #115.

Adds basic installation and usage docs.

I added a simple `CONTRIBUTING.md` file while I was at it; figured it couldn't hurt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,38 @@
 
 The Add-ons Validator, JS edition.
 
+## Usage
+
+### Command Line
+
+You need node.js to use the add-ons validator.
+
+To validate your add-on locally, install the validator from
+[npm](http://nodejs.org/):
+
+```
+# Install locally so you can use the validator from anywhere on your machine.
+npm install -g addons-validator
+```
+
+After installation, run the validator and direct it to your add-on file:
+
+```
+addons-validator my-addon.xpi
+```
+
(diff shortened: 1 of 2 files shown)


muffinresearch:
Nice - I like all the changes/additions  r+

README.md
```diff
@@ -109,14 +136,13 @@ it('should reject because of x', () => {
 
 #### Assertions and testing APIs
 
-`assert`, `describe`, `it`, `beforeEach` and `afterEach` are
-available in tests by default - you don't need to import
-anything for those to work.
+`assert`, `describe`, `it`, `beforeEach`, and `afterEach` are
```
tofumatt: Oh no, would you prefer them to not be here?

Mom’s style guide actually doesn’t use them but I tend to for developer docs.

But consistent beats my preference, so let me know ;-)

muffinresearch: I'm not worried either way to be honest. I've never used them myself and I'm not likely to, but I'm fine if this is left as is.

PR #142, opened 2015-10-22
Invert regex for urls

### LINKED ISSUE #138: Invert bad protocol detection
We have a gnarly regex that detects protocols / and patterns we don't allow, we should invert that to only detect the allowed protocols instead.

See https://github.com/mozilla/addons-validator/blob/master/src/rules/css/detectBadMozBinding.js#L4

Also this should be moved to a regex file as there are similar patterns needed for other rules.

muffinresearch:
Fixes #138


tofumatt:
r+ This is nicer, good call :+1:

PR #151, opened 2015-10-23
Pre-process eslint rules

### LINKED ISSUE #149: Make it possible to use imports in eslint rules
Because eslint rules are dug up via the file-system any imports are going to find the es6 code rather than the pre-transpiled code.

To fix this we either need to find a way to hand eslint functions ourself or worst-case pre-process the rules into a different dir so we get the benefits of sharing code e.g. regexes, message constants etc.

muffinresearch:
Fixes #149

This runs webpack over the eslint rules and puts them in dist/eslint and tells eslint to look for them there.

This allows us to use es6 in the eslint rules and we can now import anything we want which means sharing things between rules is now possible.

I re-jigged the commands so the watching still works and both the rules and/or the main lib should be processed correctly. The readme updates have the details.

### DIFF
--- a/tasks/webpack.js
+++ b/tasks/webpack.js
@@ -1,5 +1,6 @@
-var webpackConfig = require('../webpack.config.js');
+var grunt = require('grunt');
 var path = require('path');
+var webpackConfig = require('../webpack.config.js');
 
 var defaultResolve = webpackConfig.resolve;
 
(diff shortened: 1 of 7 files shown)


tasks/webpack.js
```diff
@@ -13,31 +14,64 @@ buildResolve.modulesDirectories.push('src/');
...
     keepalive: true,
     resolve: buildResolve,
   },
   coverage: {
-    entry: './tests/runner.js',
-    output: {
-      path: path.join(__dirname, '../dist'),
-      filename: 'tests.js',
-    },
+    entry: testConfig.entry,
```
muffinresearch: Really wish we had `Object.assign` here would make this a bit nicer.

tofumatt:
Nice. r+

PR #152, opened 2015-10-23
Create a common API for rules

### LINKED ISSUE #134: Create a common API for rules used by any BaseValidator
After/if #133 gets merged this becomes relevant :smile: 

I'm thinking that any rule function should have this method signature:

```js
aRuleFunction(parsedContent, filename=null, {} = {}) {

}
```

with the last options object used for anything rule-specific (the RDF rules, for instance, want a `namespace` variable).

This should DRY up all `scan` functions; it should hopefully make more rules easily understood.

Nolski:
resolves #134

### DIFF
--- a/src/rules/css/detectBadMozBinding.js
+++ b/src/rules/css/detectBadMozBinding.js
@@ -2,16 +2,17 @@ import * as messages from 'messages';
 import { isLocalCSSUri } from 'utils';
 
 
-export function detectBadMozBindingURL(rule) {
+export function detectBadMozBindingURL(cssContent, filename=null,
+                                       {startLine, startColumn}={}) {
   var messageList = [];
-  for (let declaration of rule.declarations) {
+  for (let declaration of cssContent.declarations) {
     if (declaration.property === '-moz-binding') {
       if (isLocalCSSUri(declaration.value) === false) {
         messageList.push(Object.assign({}, messages.MOZ_BINDING_EXT_REFERENCE, {
           type: 'error',
-          line: rule.position.start.line,
-          column: rule.position.start.column,
-          file: rule.position.source,
+          line: startLine,
+          column: startColumn,
+          file: filename,
         }));
       }
     }
--- a/src/rules/rdf/mustNotExist.js
+++ b/src/rules/rdf/mustNotExist.js
@@ -4,7 +4,7 @@ import * as messages from 'messages';
 
 
 export function mustNotExist(xmlDoc, filename=null,
-                             namespace='http://www.mozilla.org/2004/em-rdf#') {
+    {namespace='http://www.mozilla.org/2004/em-rdf#'}={}) {
   return new Promise((resolve) => {
     var bannedTags = RDF_UNALLOWED_TAGS;
     var validatorMessages = [];
--- a/src/rules/rdf/mustNotExist.js
+++ b/src/rules/rdf/mustNotExist.js
@@ -4,7 +4,7 @@ import * as messages from 'messages';
 
 
 export function mustNotExist(xmlDoc, filename=null,
-    {namespace='http://www.mozilla.org/2004/em-rdf#'}={}) {
+    {namespace}={}) {
   return new Promise((resolve) => {
     var bannedTags = RDF_UNALLOWED_TAGS;
     var validatorMessages = [];
--- a/src/scanners/css.js
+++ b/src/scanners/css.js
@@ -17,16 +17,20 @@ export default class CSSScanner extends BaseScanner {
           if (ast && ast.stylesheet && ast.stylesheet.rules) {
             var rules = ignorePrivateFunctions(_rules);
 
-            for (let cssRule in rules) {
+            for (let cssInstruction in rules) {
               this._rulesProcessed++;
 
-              for (let rule of ast.stylesheet.rules) {
-                if (rule.type === 'comment') {
+              for (let cssCode of ast.stylesheet.rules) {
+                if (cssCode.type === 'comment') {
                   continue;
                 }
 
                 this.validatorMessages = this.validatorMessages.concat(
-                  rules[cssRule](rule));
+                  rules[cssInstruction](cssCode, cssCode.position.source,{
+                    startLine: cssCode.position.start.line,
+                    startColumn: cssCode.position.start.column,
+                  })
+                );
               }
             }
           }
--- a/src/scanners/rdf.js
+++ b/src/scanners/rdf.js
@@ -8,13 +8,17 @@ import * as rules from 'rules/rdf';
 export default class RDFScanner extends BaseScanner {
 
   _defaultRules = rules;
-  // I don't think this ever needs to be different, but if it does we can
-  // extract the em namespace using:
-  //
-  // this.namespace = this._xmlDoc.documentElement._nsMap.em;
-  //
-  // Inside _getContents.
-  namespace = 'http://www.mozilla.org/2004/em-rdf#';
+
+  constructor(contents, filename) {
+    super(contents, filename);
+    // I don't think this ever needs to be different, but if it does we can
+    // extract the em namespace using:
+    //
+    // this.namespace = this._xmlDoc.documentElement._nsMap.em;
+    //
+    // Inside _getContents.
+    this.options.namespace = 'http://www.mozilla.org/2004/em-rdf#';
+  }
 
   _getContents() {
     return new Promise((resolve, reject) => {
@@ -24,7 +28,6 @@ export default class RDFScanner extends BaseScanner {
         },
       }).parseFromString(this.contents, 'text/xml');
 
-      this.options.namespace = this.namespace;
       resolve(xmlDoc);
     });
   }
--- a/src/scanners/rdf.js
+++ b/src/scanners/rdf.js
@@ -24,6 +24,7 @@ export default class RDFScanner extends BaseScanner {
         },
       }).parseFromString(this.contents, 'text/xml');
 
+      this.options.namespace = this.namespace;
       resolve(xmlDoc);
     });
   }
--- a/tests/rules/test.rdf.js
+++ b/tests/rules/test.rdf.js
@@ -7,14 +7,16 @@ import { validRDF } from '../helpers';
 
 
 describe('RDF: mustNotExist', () => {
+  var filename = 'install.rdf';
 
   it('should not allow <hidden> tag', () => {
     var contents = validRDF('<em:hidden>true</em:hidden>');
-    var rdfScanner = new RDFScanner(contents, 'install.rdf');
+    var rdfScanner = new RDFScanner(contents, filename);
 
     return rdfScanner.getContents()
       .then((xmlDoc) => {
-        return rules.mustNotExist(xmlDoc, 'install.rdf');
+        return rules.mustNotExist(xmlDoc, filename,
+                                  rdfScanner.options);
       })
       .then((validatorMessages) => {
         assert.equal(validatorMessages.length, 1);
@@ -27,11 +29,12 @@ describe('RDF: mustNotExist', () => {
   it('should not blow up when multiple bad tags are found', () => {
     var contents = validRDF(singleLineString`<em:hidden>true</em:hidden>
       <em:hidden>false</em:hidden>`);
-    var rdfScanner = new RDFScanner(contents, 'install.rdf');
+    var rdfScanner = new RDFScanner(contents, filename);
 
     return rdfScanner.getContents()
       .then((xmlDoc) => {
-        return rules.mustNotExist(xmlDoc, 'install.rdf');
+        return rules.mustNotExist(xmlDoc, filename,
+                                 rdfScanner.options);
       })
       .then((validatorMessages) => {
         assert.equal(validatorMessages.length, 2);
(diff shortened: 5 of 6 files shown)


src/scanners/rdf.js
```diff
@@ -24,7 +24,7 @@ export default class RDFScanner extends BaseScanner {
         },
       }).parseFromString(this.contents, 'text/xml');
 
-      resolve(xmlDoc);
+      resolve(xmlDoc, null, {namespace: this.namespace});
```
tofumatt: We do have the filename available via `this.filename` here (it's passed into all classes that extend `BaseScanner`) so this should be:

```js
        resolve(xmlDoc, this.filename, {namespace: this.namespace});
```

tofumatt: Wait, no, `_getContents` only returns the parsed contents to `scan`, so you shouldn't need to change this line at all.

```js
resolve(xmlDoc);
```
should be fine.

tofumatt: Seems I already partially set this up; in `BaseScanner` calls to rule functions are made like this: https://github.com/mozilla/addons-validator/blob/master/src/scanners/base.js#L31

So really we want to be setting `this.options.namespace = this.namespace`;

We should define `this.options = {}` by default in `BaseScanner` too.

src/scanners/css.js
```diff
@@ -26,7 +26,15 @@ export default class CSSScanner extends BaseScanner {
                 }
 
                 this.validatorMessages = this.validatorMessages.concat(
-                  rules[cssRule](rule));
+                  rules[cssRule](
+                    rule.declarations,
```
muffinresearch: I think I'd rather pass the whole cssRule object and then the rule function has access to everything. For example we need access to selectors to be able to reproduce one of the existing tests see https://github.com/mozilla/amo-validator/search?utf8=%E2%9C%93&q=%23identity-box

Nolski: Isn't this what the `{}={}` is for in the rule function?

tofumatt: The `cssRule` object in this case is a notion of a "CSS Scannable Document", more/less right?

In that case I agree; we pass an HTML/RDF document to those rules.

muffinresearch: > Isn't this what the {}={} is for in the rule function?

Could do but it seems better to provide the entire cssRule object to all validator rule functions than selectively provide specific data for a subset of the rules.  To clarify I'm only talking about the first arg. Passing other more specific bits of data via other args is fine - especially along the lines of making the API more generic. 

You could use a destructuring assignment [1] to pull out the other bits that you're explicitly passing - but it might be simplest to send the whole object verbatim despite there being some duplication in what you send as other args.

[1] http://jsbin.com/rohegudeno/edit?js,console

muffinresearch:
This still needs to pass the cssRule object as the first arg to css rule functions.

muffinresearch:
@Nolski any updates on this? Would be nice to not let hang around - once we start doing more than JS rules this will bit-rot quite quickly.

Nolski:
@muffinresearch should be good to look at now.

src/rules/rdf/mustNotExist.js
```diff
@@ -3,8 +3,7 @@ import { RDF_OBSOLETE_TAGS, RDF_UNALLOWED_TAGS, RDF_UNALLOWED_IF_LISTED_TAGS,
 import * as messages from 'messages';
 
 
-export function mustNotExist(xmlDoc, filename=null,
-                             namespace='http://www.mozilla.org/2004/em-rdf#') {
+export function mustNotExist(xmlDoc, filename=null, {namespace}={}) {
```
muffinresearch: Shouldn't this default to the namspace previously set as a default?

src/rules/css/detectBadMozBinding.js
```diff
@@ -2,16 +2,17 @@ import * as messages from 'messages';
 import { LOCAL_CSS_URL } from 'regex';
 
 
-export function detectBadMozBindingURL(rule) {
+export function detectBadMozBindingURL(rule, filename=null,
```
muffinresearch: Can we take this opportunity to rename `rule` to `cssRule`? This'll help save some potential confusion about what we're talking about since there are cssRules and rule functions.

src/scanners/css.js
```diff
@@ -26,7 +26,15 @@ export default class CSSScanner extends BaseScanner {
                 }
 
                 this.validatorMessages = this.validatorMessages.concat(
-                  rules[cssRule](rule));
```
muffinresearch: Again here - it gets confusing - (totally my fault) but it'll make sense to clean this up now. We need to differentiate between cssRules as in bits of CSS and the rule functions. To be consistent with the previous suggestion we need to make `cssRule` -> `cssRuleFunc` and `rule` -> `cssRule` I think.

src/scanners/css.js
```diff
@@ -26,7 +26,15 @@ export default class CSSScanner extends BaseScanner {
...
+                  rules[cssRule](
```
muffinresearch: Would be nice to try and indent this so it's more obviously a function - maybe see about moving the first two args up onto the first line?

tests/rules/test.rdf.js
```diff
@@ -14,7 +14,8 @@ describe('RDF: mustNotExist', () => {
 
     return rdfScanner.getContents()
       .then((xmlDoc) => {
-        return rules.mustNotExist(xmlDoc, rdfScanner.namespace);
+        return rules.mustNotExist(xmlDoc, null,
+                                  {namespace: rdfScanner.namespace});
```
muffinresearch: If you make the default sane, you shouldn't need to pass this at all.

src/rules/css/detectBadMozBinding.js
```diff
@@ -2,16 +2,17 @@ import * as messages from 'messages';
 import { LOCAL_CSS_URL } from 'regex';
 
 
-export function detectBadMozBindingURL(rule) {
+export function detectBadMozBindingURL(cssRule, filename= null,
```
tofumatt: Nitpick here, but the space after the `filename=` should be removed.

src/rules/css/detectBadMozBinding.js
```diff
@@ -2,16 +2,17 @@ import * as messages from 'messages';
...
+                                       {startLine, startColumn}={}) {
   var messageList = [];
-  for (let declaration of rule.declarations) {
+  for (let declaration of cssRule.declarations) {
```
tofumatt: This is probably unrelated to this PR, but it'd be nice to rename these from `cssRule` to `cssInstruction` or _something_ without rule in the name. I know it's because they're a CSS Rule, but it's conceptually hard to separate that from our own validator rules and I get confused.

src/rules/rdf/mustNotExist.js
```diff
@@ -16,6 +16,7 @@ export function mustNotExist(xmlDoc, filename=null,
       bannedTags = bannedTags.concat(RDF_UNALLOWED_IF_LISTED_TAGS);
     }
 
+
```
tofumatt: Another nitpick, but this extra newline isn't needed.

src/scanners/css.js
tofumatt: Even better would be something without the word "rule" entirely.

`rule` stays `rule` and `cssRule`/`cssRuleFunc` becomes `cssInstruction` or even `cssCode` perhaps? Getting the word "Rule" out of there entirely would really help scalability even if the alternate names are a bit weird.

muffinresearch: Yep - sounds good. Much better than my suggestion!

src/rules/css/detectBadMozBinding.js
muffinresearch: Well spotted. Is this something jscs should catch?

src/rules/css/detectBadMozBinding.js
muffinresearch: I think it make sense to clean this up now whilst we're in all these files. 

@Nolski - sorry for adding this to your plate but I think now's a good time since you're touching all the bits that use these vars.

src/rules/css/detectBadMozBinding.js
tofumatt: Yeah, I’ll look for a rule and add a quick PR.

tests/rules/test.rdf.js
```diff
@@ -14,7 +14,7 @@ describe('RDF: mustNotExist', () => {
 
     return rdfScanner.getContents()
       .then((xmlDoc) => {
-        return rules.mustNotExist(xmlDoc, rdfScanner.namespace);
+        return rules.mustNotExist(xmlDoc, null);
```
muffinresearch: Elsewhere we've used a dummy filename. Make it a bit easier to understand what this is.

src/rules/css/detectBadMozBinding.js
tofumatt: Actually I can't find a rule for this one. :unamused:

src/scanners/css.js
```diff
@@ -17,16 +17,22 @@ export default class CSSScanner extends BaseScanner {
...
-              for (let rule of ast.stylesheet.rules) {
-                if (rule.type === 'comment') {
+              for (let cssRule of ast.stylesheet.rules) {
+                if (cssRule.type === 'comment') {
                   continue;
                 }
 
                 this.validatorMessages = this.validatorMessages.concat(
-                  rules[cssRule](rule));
+                  rules[cssRuleFunc](cssRule, cssRule.position.source,
```
muffinresearch: I still think this should be compacted more.

src/scanners/css.js
```diff
@@ -17,16 +17,22 @@ export default class CSSScanner extends BaseScanner {
...
+                    {
+                      startLine: cssRule.position.start.line,
+                      startColumn: cssRule.position.start.column,
+                    }
```
muffinresearch: You're missing the default object here. Calling this without passing the optional args object would result in an error.

Contrast: http://jsbin.com/gumifo/edit?js,console,output 

with this: http://jsbin.com/pimibi/edit?js,console,output

muffinresearch:
@Nolski I'll make a new bug for dealing with passing the type so that can at least be handled separately.

src/scanners/css.js
muffinresearch: Ah ignore me, this is calling the rule funcs not defining them. 

*/me gets more coffee*

src/rules/css/detectBadMozBinding.js
Nolski: Won't this add extra confusion because the _rule_ here is the argument being passed and the _instruction_ is the actual rule function? Or am I mis-understanding your previous comment?

muffinresearch: So how about calling the arg that's passed in simply `css`. Then `css.declarations` looks ok?

muffinresearch: I know we're bike-shedding on naming here, but getting this right will help alot further along.

Nolski: How about `cssContent`? That way it kind of fits the original `parsedContent` idea for that first variable.

Nolski:
@muffinresearch @tofumatt  should be good for review again.

src/rules/css/detectBadMozBinding.js
tofumatt: I like that the most, nice 👍

src/rules/css/detectBadMozBinding.js
```diff
@@ -2,16 +2,17 @@ import * as messages from 'messages';
 import { isLocalCSSUri } from 'utils';
 
 
-export function detectBadMozBindingURL(rule) {
+export function detectBadMozBindingURL(cssContent, filename=null,
+                                       {startLine, startColumn}={}) {
   var messageList = [];
-  for (let declaration of rule.declarations) {
+  for (let declaration of cssContent.declarations) {
```
tofumatt: As mentioned in the outdated diff: I'm :+1: for this. Thanks and sorry for the :bike: :house:

src/rules/rdf/mustNotExist.js
```diff
@@ -4,7 +4,7 @@ import * as messages from 'messages';
 
 
 export function mustNotExist(xmlDoc, filename=null,
-                             namespace='http://www.mozilla.org/2004/em-rdf#') {
+    {namespace='http://www.mozilla.org/2004/em-rdf#'}={}) {
```
muffinresearch: Doesn't this get passed by default in the options? If we do this can be {namespace}={}

muffinresearch:
Aside from that last comment looks good. Nice work!

Fix that up, squash the commits and :shipit:

r+wc

src/rules/rdf/mustNotExist.js
Nolski: It does, however in a previous comment you said 
> Shouldn't this default to the namspace previously set as a default?

so I assumed that namespace should have a default value?

tofumatt: I think when I originally coded this it was a bit messy. It should live canonically in the scanner itself and be passed to the rules from the scanners. The rules shouldn't assume anything about the document.

Sorry about that, my fault :astonished:

Nolski:
That last change caused a larger than average delta so @tofumatt  it might be good to take one last look before I merge. I'm not quite sure why the tests aren't running `rdfScanner.scan()` alone since it seems like would call `getContents` on it's own and it would simplify out unit tests quite a bit in the process.

muffinresearch:
r+  Could you file an issue for the API nit (re: scan() vs getContents) please! Thanks!

PR #153, opened 2015-10-23
Provide a VERY clear quickstart for devs

tofumatt:
All this info is in the README but it's (appropriately) split up into different sections.

Sometimes, though, I get frustrated not seeing _just_ the commands I want to build/run the program I wanna hack on.

Plus we aren't even on npm yet and when @andymckay checked out the lib to run it it became clear to me we didn't provide up-front installation instructions to run the development version.

I find doing this is how I start on things _I_ hack on: let me just see it working locally at all, then I'll see how to run the tests, etc.


muffinresearch:
r+

PR #157, opened 2015-10-26
Add openDialog checks

### LINKED ISSUE #126: Rule: %s_remote_uri
Implement the `%s_remote_uri` rule for JS validator.

muffinresearch:
Fixes #126 

Made the files and rule names from eslints PoV all lowercase rather than a mix of lower + camel + underscores. This matches closer to typical eslint naming.

* add tests for non-literal uri arg to openDialog.
* add checks for remote uri passed to openDialog.
* It kills off all the regex - negating matching data: but allowing other protocols started to make the regexes too gnarly. Instead I've replaced the regex with simpler utils that actually parse the urls. 
* css uris are now extracted using via regex in utils. 

TODO: this doesn't do anything to check whether CSS uri values are escaped correctly e.g. ' " and () in un-quoted CSS uris [1] and also we could use something to check css quotes match etc too. I'll file another issue for that.

[1] https://drafts.csswg.org/css-values-3/#urls


tofumatt:
Looks like the existing JS rules weren’t moved so would still have weird filenames.

Getting rid of those weird names is several 👍 from me!

muffinresearch:
Case preserving filesystems FTL. :/

src/rules/css/detectBadMozBinding.js
```diff
@@ -1,12 +1,12 @@
 import * as messages from 'messages';
```
tofumatt: I like that our JS files now aren't `camelCase` or whatever that's called, can we do the same for CSS?

src/utils.js
```diff
@@ -88,3 +90,36 @@ export function ignorePrivateFunctions(list) {
 
   return filteredList;
 }
+
+
+export function isLocalUrl(url_) {
```
tofumatt: `url_` is a weird variable name, did you mean `_url`?

tests/rules/javascript/test.opendialog_remote_uri.js
```diff
@@ -0,0 +1,41 @@
...
+                foo.openDialog("ftps://foo.com/bar/");
+                foo.openDialog("ftp://foo.com/bar/");
+                foo.openDialog("//foo.com/bar/");
+                foo.openDialog("data:whatever");`;
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 6);
+        for (let msg of validationMessages) {
```
tofumatt: Nitpick here, but I always like full variable names when we can get them, so `message` over `msg` if it's possible.

tofumatt:
I like that functions over wild regexs and the prettier filenames :+1: 

r+wc, if the CSS filenames can be made more like `detect_bad_moz_binding` that'd be great and there are a few variable name nitpicks.

src/utils.js
muffinresearch: Hmm, I should probably use es6 modules import syntax to modify the url lib or just import parse and then I can use url here.

tests/rules/javascript/test.opendialog_remote_uri.js
muffinresearch: Yeah fair enough. I tend to be little bit less worried about that for tests. But I suppose a consistent rule here is a bit more important than anything else.

muffinresearch:
I'll file a bug for rule file name changes elsewhere.

PR #160, opened 2015-10-27
Add test for eslint rule keys + files

### LINKED ISSUE #159: Add test on the eslint rule key and filename
The key must match the filname 100% we should look at the keys in rules/javascript/index.js and check they match up.

muffinresearch:
Fixes #159

Detects if key missing to match and file and vice versa.


tofumatt:
Nice! r+

PR #163, opened 2015-10-27
Create amend JS rule

### LINKED ISSUE #111: Rule: eval
Implement the `eval` rule for JS validator.

Nolski:
Resolves #111

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -17,12 +17,11 @@ To update it edit `docs/rules.md` in the
 | :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_PROCESSNEXTEVENT |
 | :x: | warning | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
 | :x: | warning | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
-| :x: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | | | |
+| :white_check_mark: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/javascript/actions.py#L488) | | EVAL_STRING_ARG |
 | :x: | warning | low_level_module (not from src) | | Usage of low-level or non-SDK interface | | | | |
 | :x: | warning | widget | | Use of deprecated SDK module | | | | |
 | :negative_squared_cross_mark: | notice |  \_readonly_top | | window.top is a reserved variable | | | ('testcases_javascript_actions', '_readonly_top' | **Removed** |
 | :x: | warning | global_overwrite | | Global variable overwrite | | | | |
-
 
 ### Call definitions
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -100,3 +100,10 @@ export function _bannedIdentifier(name) {
 
 export const BANNED_NEWTHREAD = _bannedIdentifier('newThread');
 export const BANNED_PROCESSNEXTEVENT = _bannedIdentifier('processNextEvent');
+export const EVAL_STRING_ARG = {
+  code: 'EVAL_STRING_ARG',
+  message: _('setTimeout or setInterval must have function as 1st arg'),
+  description: _(singleLineString`setTimeout and setInterval functions should be
+    called only with function expressions as their first argument`),
+  legacyCode: ['javascript', 'dangerous_global', 'eval'],
+};
--- a/src/rules/javascript/eval_string_arg.js
+++ b/src/rules/javascript/eval_string_arg.js
@@ -1,0 +1,33 @@
+import { EVAL_STRING_ARG } from 'messages/javascript';
+import { getVariable } from 'utils';
+
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      // Check if what's being called is setTimeout or setInterval
+      if (typeof node.callee.property !== 'undefined' &&
+          node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        // If the first arg is a variable check if it's a function
+        if (node.arguments.length > 0 &&
+            node.arguments[0].type === 'Identifier') {
+          var originalDef = getVariable(context, node.arguments[0].name);
+          if (typeof originalDef !== 'undefined' &&
+            (originalDef.type !== 'FunctionExpression' &&
+             originalDef.type !== 'ArrowFunctionExpression')) {
+            return context.report({node: node, message: EVAL_STRING_ARG.code});
+          } else {
+            return;
+          }
+        // Otherwise just check the type of the variable
+        } else if (node.arguments.length > 0 &&
+            (node.arguments[0].type !== 'FunctionExpression' &&
+             node.arguments[0].type !== 'ArrowFunctionExpression')) {
+          return context.report({node: node, message: EVAL_STRING_ARG.code});
+        }
+      }
+    },
+  };
+}
--- a/src/rules/javascript/index.js
+++ b/src/rules/javascript/index.js
@@ -7,4 +7,5 @@ export default {
   opendialog_remote_uri: ESLINT_WARNING,
   opendialog_nonlit_uri: ESLINT_WARNING,
   shallow_wrapper: ESLINT_WARNING,
+  eval_string_arg: ESLINT_ERROR,
 };
--- a/src/utils.js
+++ b/src/utils.js
@@ -33,6 +33,21 @@ export function singleLineString(strings, ...vars) {
 }
 
 /*
+ * Get a variable from a eslint context object if it exists, otherwise
+ * undefined.
+ */
+export function getVariable(context, name) {
+  var variables = context.getScope().variables;
+  var result;
+  variables.forEach(function(variable) {
+    if (variable.name === name) {
+      result = variable.defs[0].name.parent.init;
+    }
+  });
+  return result;
+}
+
+/*
  * Gettext utils. No-op until we have proper
  * a proper l10n solution.
  *
--- a/tests/test.utils.js
+++ b/tests/test.utils.js
@@ -37,6 +37,42 @@ describe('utils.singleLineString()', function() {
 });
 
 
+describe('utils.getVariable()', function() {
+  // This is the expected schema from eslint
+  var context = {
+    getScope: function() {
+      return {
+        variables: [{
+          name: 'foo',
+          defs: [{
+            type: 'Variable',
+            name: {
+              parent: {
+                init: {
+                  type: 'Literal',
+                  value: 'bar',
+                },
+              },
+            },
+          }],
+        }],
+      };
+    },
+  };
+
+  it('should return the correct variable in the given context.', () => {
+    var foo = utils.getVariable(context, 'foo');
+    assert.equal(foo.type, 'Literal');
+    assert.equal(foo.value, 'bar');
+  });
+
+  it("should return undefined if the variable doesn't exist.", () => {
+    var undef = utils.getVariable(context, 'doesNotExist');
+    assert.equal(typeof undef, 'undefined');
+  });
+});
+
+
 describe('utils.checkMinNodeVersion()', function() {
 
   it('should reject if version is not high enough', () => {
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -21,6 +21,7 @@ export default class JavaScriptScanner {
         rulePaths: ['dist/eslint'],
         rules: ESLintRules,
         useEslintrc: false,
+        envs: ['es6'],
       });
 
       var validatorMessages = [];
--- a/tests/rules/javascript/test.eval_string_arg.js
+++ b/tests/rules/javascript/test.eval_string_arg.js
@@ -1,0 +1,74 @@
+import { VALIDATION_ERROR } from 'const';
+import { singleLineString } from 'utils';
+import JavaScriptScanner from 'scanners/javascript';
+import * as messages from 'messages';
+
+describe('eval_string_arg', () => {
+  it('should allow the use of a function for setTimeout', () => {
+    var code = 'window.setTimeout(function () { console.log("foo"); }, 1)';
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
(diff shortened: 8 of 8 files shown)


docs/rules.md
```diff
@@ -13,7 +13,7 @@ We should try and list out all the validation rules we know of by file type. E.g
 | :x: | banned_identifier | | Banned or deprecated JavaScript Identifier | | | | |
 | :x: | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
 | :x: | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
-| :x: | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | | | |
+| :x: | EVAL | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/javascript/actions.py#L488) | | |
```
muffinresearch: EVAL  should go in the last column.

src/messages/javascript.js
```diff
@@ -42,3 +42,10 @@ export function _methodPassedRemoteUri(method) {
 
 export const OPENDIALOG_REMOTE_URI = _methodPassedRemoteUri('openDialog');
 export const OPENDIALOG_NONLIT_URI = _nonLiteralUri('openDialog');
+
+export const EVAL = {
+  code: 'EVAL',
```
muffinresearch: I wonder if this code should be narrowed down to something relating to strings as args as opposed to just `EVAL` e.g. `EVAL_STRING_ARG` or something.

src/messages/javascript.js
```diff
@@ -42,3 +42,10 @@ export function _methodPassedRemoteUri(method) {
...
+  message: _('setTimeout or setInterval must have function as 1st arg'),
+  description: _(singleLineString`setTimeout and setInterval functions should be
+             called only with function expressions as their first argument`)
```
tofumatt: This is an odd indentation; see the rest of the file for a guide.

src/rules/javascript/eval.js
```diff
@@ -0,0 +1,16 @@
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            (node.arguments[0].type !== 'FunctionExpression') &&
+             node.arguments[0].type !== 'ArrowFunctionExpression') {
```
muffinresearch: shiny

tests/rules/javascript/test.eval.js
```diff
@@ -0,0 +1,51 @@
+import { VALIDATION_ERROR } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+import * as messages from 'messages';
+
+describe('eval', () => {
+  it('should allow the use of a function for setTimeout', () => {
```
muffinresearch: What about vars assigned to functions?

src/rules/javascript/eval.js
```diff
@@ -0,0 +1,16 @@
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            (node.arguments[0].type !== 'FunctionExpression') &&
```
muffinresearch: I was thinking about using `indexOf` here but this is probably more readable.

src/rules/javascript/eval_string_arg.js
```diff
@@ -0,0 +1,16 @@
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            (node.arguments[0].type !== 'FunctionExpression') &&
```
muffinresearch: I was thinking about using `indexOf` here against an Array of both types but this is probably more readable.

*I'm making the same comment again because the previous one was buried by changes*

tofumatt: Yeah I like this way more too; if we started checking for more than two either a helper or `indexOf` might be nicer, but this is best for now.

src/messages/javascript.js
```diff
@@ -42,3 +42,10 @@ export function _methodPassedRemoteUri(method) {
 
 export const OPENDIALOG_REMOTE_URI = _methodPassedRemoteUri('openDialog');
 export const OPENDIALOG_NONLIT_URI = _nonLiteralUri('openDialog');
+
+export const EVAL_STRING_ARG = {
+  code: 'EVAL_STRING_ARG',
```
muffinresearch: If you could add a `legacyCode` prop here with the contents of whatever the original err_id tuple has that will save this failing upcoming tests. 

We want to try and be able to correlate the new validation rules with the old ones as much as possible.

src/rules/javascript/eval_string_arg.js
```diff
@@ -0,0 +1,17 @@
...
+    CallExpression: function(node) {
+      var acceptedArgs = ['FunctionExpression', 'ArrowFunctionExpression'];
+
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            acceptedArgs.indexOf(node.arguments[0].type) === -1) {
+          return context.report({node: node, message: 'eval_string_arg'});
```
muffinresearch: This should import the message you're using and use the code prop in place of the `'eval_string_arg'`. We're going to add a test for this soon so all new js rules could start doing that now to save us needing to refactor this later.

Nolski:
It looks like ESLint does not currently support any ES6 or JSX in their production version [1]. I'm fairly sure that this is why the code that I'm passing in the two failing tests is failing. Looking at the cause of the failures, it is ESLint saying the code contains syntax errors (that was about as descriptive as it got unfortunately.) I'm not sure how we want to go about resolving this. @muffinresearch, @tofumatt said that he thought you might have worked on getting ES6 code to run in ESLint?

[1] http://eslint.org/blog/2014/11/es6-jsx-support/

muffinresearch:
@Nolski That's an old post from 2014. The es6 setting exists in the config - we're using it on our own code.

Nolski:
@muffinresearch Do you have any idea why 

```javascript
window.setTimeout(() => { console.log("foo"); }, 1);
```

Would throw a syntax error when passed to ESLint?:

muffinresearch:
Have you updated the eslint config in the js validator to allow es6? Probably needs something here https://github.com/mozilla/addons-validator/blob/master/src/scanners/javascript.js#L19 (Note eslinting of our own code is completely separate to how it's used for validation). 

If you start by adding an env object with `{ es6: true }` and see if that makes a difference. We might need to change the parser to babel-eslint too, but try just enabling es6 first.

Nolski:
So after some further digging (and some asking ESLint devs) we can't configure ESLint to simply work with ES6. CLIEngine does take [configuration parameters](http://eslint.org/docs/developer-guide/nodejs-api.html#cliengine) and perhaps the issue here is with the parser we are using. By default eslint uses [espree](https://github.com/eslint/espree) which looks like it only supports ES5.1? This part I'm a bit confused by because while it says it fully supports 5.1, it also implements EST for ES5 and ES6.

muffinresearch:
@Nolski I said in my comment above you might need to change the parser.

muffinresearch:
Ok so checking out your branch I see how to fix it - use this patch:

```diff
diff --git a/src/scanners/javascript.js b/src/scanners/javascript.js
index e64eb1e..4a03c83 100644
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -21,6 +21,7 @@ export default class JavaScriptScanner {
         rulePaths: ['dist/eslint'],
         rules: ESLintRules,
         useEslintrc: false,
+        envs: ['es6'],
       });
```

That makes the arrow func test pass.

muffinresearch:
For future reference all the info you need was here - http://eslint.org/docs/developer-guide/nodejs-api.html#cliengine - I suggested an object for env - but in the CLIEngine docs the key is `envs` and it's an Array. 

No parser changes necessary.

muffinresearch:
Also I used this http://eslint.org/demo/ and repro'd the error and also saw it was fixed by ticking 'es6'. That will provide a quick way to validate what's possible if we hit similar problems in future.

Nolski:
Alright @muffinresearch @tofumatt ES6 arrow functions and having previously declared variables should be working now. Feel free to take a look and if it looks good and you want to r+ I can rebase and squash.

docs/rules.md
```diff
@@ -13,13 +13,12 @@ We should try and list out all the validation rules we know of by file type. E.g
 | :x: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | | |
 | :x: | warning | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
 | :x: | warning | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
-| :x: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | | | |
+| :x: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/javascript/actions.py#L488) | | EVAL_STRING_ARG |
```
muffinresearch: You need to change the emoji to  :white_check_mark:

src/rules/javascript/eval_string_arg.js
```diff
@@ -0,0 +1,31 @@
...
+      var acceptedArgs = ['FunctionExpression', 'ArrowFunctionExpression'];
+
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            node.arguments[0].type === 'Identifier') {
+          var originalDef = getVariable(context, node.arguments[0].name);
+          if (originalDef !== undefined &&
```
muffinresearch: I guess this is ok - I've tended to use `typeof originalDef !== 'undefined'` old habit from when undefined could get redefined. Old habits die-hard.

We should decide on which way we want to go with this and make jscs enforce it either way. I like consistency more than anything else.

1__node

src/rules/javascript/index.js
```diff
@@ -5,4 +5,5 @@ export default {
   mozindexeddb_property: ESLINT_WARNING,
   opendialog_remote_uri: ESLINT_WARNING,
   opendialog_nonlit_uri: ESLINT_WARNING,
+  eval_string_arg: ESLINT_ERROR,
```
muffinresearch: This is fine for now - but I'm completely unclear with this rule what message type it is. We should see what the current validator makes of this to know.

src/rules/javascript/eval_string_arg.js
```diff
@@ -0,0 +1,31 @@
+import { EVAL_STRING_ARG } from 'messages/javascript';
+import { getVariable } from 'utils';
+
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      var acceptedArgs = ['FunctionExpression', 'ArrowFunctionExpression'];
```
muffinresearch: We said before we were ok with long-hand for the if statement e.g. like you still have for setTimeout/setInterval. Despite being a bit more concise `indexOf` is less instantly readable. Even more so if you've got to find the Array var up here.

src/utils.js
```diff
@@ -32,6 +32,17 @@ export function singleLineString(strings, ...vars) {
   }).join(' ').trim();
 }
 
+export function getVariable(context, name) {
```
muffinresearch: I'm surprised why this isn't getting executed implicitly by the tests?

src/rules/javascript/eval_string_arg.js
```diff
@@ -0,0 +1,31 @@
...
+
+      if (node.callee.property.type === 'Identifier' &&
+          (node.callee.property.name === 'setTimeout' ||
+           node.callee.property.name === 'setInterval')) {
+
+        if (node.arguments.length > 0 &&
+            node.arguments[0].type === 'Identifier') {
+          var originalDef = getVariable(context, node.arguments[0].name);
```
muffinresearch: Let's add some comments to this explaining what's going on. It will help massively in the future. We should all comment more in the rules.

src/utils.js
```diff
@@ -32,6 +32,17 @@ export function singleLineString(strings, ...vars) {
...
+  var variables = context.getScope().variables;
+  var result;
+  variables.forEach(function(variable) {
```
muffinresearch: Comments here to explain the eslint apis being used and what it's doing are particularly important here.

tests/rules/javascript/test.eval.js
```diff
@@ -0,0 +1,74 @@
+import { VALIDATION_ERROR } from 'const';
+import { singleLineString } from 'utils';
+import JavaScriptScanner from 'scanners/javascript';
+import * as messages from 'messages';
+
+describe('eval', () => {
```
muffinresearch: Should update the description to eval_string_arg

src/rules/javascript/eval_string_arg.js
muffinresearch: So I've discussed this with @tofumatt. We came to a consensus that given we've already got `typeof foo === 'undefined'` elsewhere it makes sense to stick to that for now.

src/utils.js
muffinresearch: Ah ok - it's because whilst the utils is instrumented the rules aren't due to how they are run indirectly by eslint. We'll need to independently unit test this is isolation. More direct testing will be a good thing given util funcs like this will likely be used quite a lot. 

Mocking out the eslint APIs is fine if that makes it easier.

muffinresearch:
So there's a few nits and you're missing some coverage. Fix that and we should be close to being able to land this.

Nolski:
@muffinresearch This should be good to look over again when you get the chance.

tests/test.utils.js
```diff
@@ -37,6 +37,42 @@ describe('utils.singleLineString()', function() {
...
+    },
+  };
+
+  it('should return the correct variable in the given context.', () => {
+    var foo = utils.getVariable(context, 'foo');
+    assert.equal(foo.type, 'Literal');
+    assert.equal(foo.value, 'bar');
+  });
+
+  it('should return undefined if the variable doesn\'t exist.', () => {
```
muffinresearch: Use double quotes or a template string here to avoid  the need to escape the `'` in doesn't

muffinresearch:
Has conflicts but I suspect that's just the index.js with the rules object. 

r+wc

PR #165, opened 2015-10-27
Add references for old rules

### LINKED ISSUE #122: Add a reference to the old rules
When we get to a point of referencing the old rules we'll want to be able to relate our message codes back to the old messages we should probably add something to the message objects to make that easier.

muffinresearch:
Fixes #122 

1__bash

A null value is an explicit way to say the old `err_id` tuple is not useful e.g. a matching code doesn't exist or it's not unique. In which case any comparisons done will need to be based on other things or dealt with manually.

### DIFF
--- a/src/messages/chromemanifest.js
+++ b/src/messages/chromemanifest.js
@@ -3,6 +3,11 @@ import { gettext as _, singleLineString } from 'utils';
 
 export const DANGEROUS_CATEGORY = {
   code: 'DANGEROUS_CATEGORY',
+  legacyCode: [
+    'testcases_chromemanifest',
+    'test_resourcemodules',
+    'resource_modules',
+  ],
   message: 'Potentially dangerous category entry',
   description: _(singleLineString`Add-ons defining global properties via
     category entries require careful review by an administrative reviewer.`),
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -3,6 +3,11 @@ import { gettext as _, singleLineString } from 'utils';
 
 export const JS_SYNTAX_ERROR = {
   code: 'JS_SYNTAX_ERROR',
+  legacyCode: [
+    'testcases_scripting',
+    'test_js_file',
+    'syntax_error',
+  ],
   message: _('JavaScript syntax error'),
   description: _(singleLineString`There is a JavaScript syntax error in your
     code; validation cannot continue on this file.`),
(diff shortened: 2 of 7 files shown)


src/messages/chromemanifest.js
```diff
@@ -3,6 +3,11 @@ import { gettext as _, singleLineString } from 'utils';
 
 export const DANGEROUS_CATEGORY = {
   code: 'DANGEROUS_CATEGORY',
+  legacyCode: [
```
tofumatt: Semantics here, but shouldn't this be `legacyCodes`?

src/messages/javascript.js
```diff
@@ -10,12 +15,18 @@ export const JS_SYNTAX_ERROR = {
...
+  // Original code appeared to have a non-unique err_id which is no
+  // use for comparsions. ('testcases_regex', 'generic', '_generated')
+  legacyCode: null,
   message: _('mozIndexedDB has been removed; use indexedDB instead'),
   description: _('mozIndexedDB has been removed; use indexedDB instead.'),
 };
 
 export const MOZINDEXEDDB_PROPERTY = {
   code: 'MOZINDEXEDDB_PROPERTY',
+  // Original code appeared to have a non-unique err_id which is no
```
tofumatt: Thanks for the notes :+1:

src/messages/chromemanifest.js
tofumatt: Oh, nevermind, I see. It's not really an array of values it's like a struct.

tests/test.messages.js
```diff
@@ -42,6 +44,19 @@ describe('Messages', function() {
...
 
+  it('should have a legacyCode property in every message', () => {
+    for (let message in messages) {
+      if (typeof messages[message] === 'object') {
+        var legacyCode = messages[message].legacyCode;
+        if ((legacyCode instanceof Array && legacyCode.length !== 3) ||
+            (!(legacyCode instanceof Array) && legacyCode !== null)) {
+          assert.fail(null, null, singleLineString`A valide legacyCode could
+            not be found for code: "${messages[message].code}". Should be
+            an Array with 3 values based on the amo-validator err_id or null.`);
```
tofumatt: This is the canonical place to add that note about reasons why we'd use a `null` value.

tofumatt:
r+wc

PR #167, opened 2015-10-28
Test for Shallow XPCOM wrapper usage

### LINKED ISSUE #124: Rule: shallow
Implement the `shallow` rule for JS validator.

tofumatt:
Fixes #124.

I believe it's proper to allow this with only one argument.

### DIFF
--- a/src/rules/javascript/shallow_wrapper.js
+++ b/src/rules/javascript/shallow_wrapper.js
@@ -1,0 +1,24 @@
+import * as messages from 'messages/javascript';
+
+// This rule is used for older add-ons (non-Jetpack I believe) but honestly
+// it is a bit unclear as the MDN docs on XPCNativeWrapper seem to have moved
+// to XRay: https://developer.mozilla.org/en-US/docs/Xray_vision
+//
+// TODO: Find out more about this rule.
+export default function(context) {
+  function _testForShallowWrapper(node) {
+    if (node.callee.name === 'XPCNativeWrapper' &&
+        node.callee.type === 'Identifier') {
+
+      if (node.arguments[0].type === 'Identifier' ||
+          node.arguments[0].type === 'MemberExpression') {
+        return context.report(node, messages.SHALLOW_WRAPPER.code);
+      }
+    }
+  }
+
+  return {
+    CallExpression: _testForShallowWrapper,
+    NewExpression: _testForShallowWrapper,
+  };
+}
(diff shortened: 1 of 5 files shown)


tests/rules/javascript/test.shallow_wrapper.js
```diff
@@ -0,0 +1,62 @@
...
+        assert.equal(validationMessages.length, 1);
+        assert.equal(validationMessages[0].code, messages.SHALLOW_WRAPPER.code);
+        assert.equal(validationMessages[0].type, VALIDATION_WARNING);
+
+        code = `var wrap = new XPCNativeWrapper(unsafeWindow, 'doc', 'a');`;
+        jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+        return jsScanner.scan();
+      })
+      .then((validationMessages) => {
```
muffinresearch: You could roll these up into a single bit of code to test and then check that you get 2 errors and then iterate over the validationMessages and check the code is as expected in each case.

tests/rules/javascript/test.shallow_wrapper.js
```diff
@@ -0,0 +1,62 @@
...
+        assert.equal(validationMessages[0].type, VALIDATION_WARNING);
+      });
+  });
+
+  it('should not allow use of shallow wrapper with a member variable', () => {
+    var code = `XPCNativeWrapper(window._content, 'foo');`;
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
```
muffinresearch: As before roll this up into a single scan call.

src/rules/javascript/shallow_wrapper.js
```diff
@@ -0,0 +1,35 @@
+import * as messages from 'messages/javascript';
+
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      if (node.callee.name === 'XPCNativeWrapper' &&
+          node.callee.type === 'Identifier' &&
+          // If only one argument is used it's okay;
```
muffinresearch: Moving this to the top of the rule with supporting links would save on the duplication - it's not so spread out as to be an issue.

muffinresearch:
So not sure who to ask about this - I feel like it would be good to properly understand and document the context of this rule.

muffinresearch:
This has conflicts - because I moved things around in the rules file. Sorry!

Given the existing rule is about it having more than one arg we can land this and deal with any issues when it comes to testing this validator next to the old one.

muffinresearch:
r+wc

PR #173, opened 2015-10-29
Add ability to publish rules to gh-pages

muffinresearch:
Needed a break from JS rule yak-shaving and wanted to make the rule list more usable.

Running `grunt publish-rules` puts the rules on gh-pages.

See http://mozilla.github.io/addons-validator/

I'll set this up to be automated on pushes to master in due course.


tofumatt:
> I'll set this up to be automated on pushes to master in due course.

That was going to be my only comment.

r+ :+1:

Gruntfile.js
```diff
@@ -11,6 +17,16 @@ module.exports = function(grunt) {
 
   grunt.initConfig(configs);
 
+  grunt.registerTask('buildrules', 'Build the rules', function() {
+    md.use(emoji);
```
tofumatt: :tada:

PR #174, opened 2015-10-29
Add header links

muffinresearch:
Makes it possible to link to sections of rules - e.g. http://mozilla.github.io/addons-validator/#javascript


tofumatt:
Good idea! r+

PR #176, opened 2015-10-29
Add rule for banned_identifiers

### LINKED ISSUE #175: Add rule for banned_identifier
See https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py

muffinresearch:
Fixes #175

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,7 +1,7 @@
 
 # Validator Rules
 
-This document is automatically published on [github pages](http://mozilla.github.io/addons-validator/). 
+This document is automatically published on [github pages](http://mozilla.github.io/addons-validator/).
 To update it edit `docs/rules.md` in the
 [github repo](https://github.com/mozilla/addons-validator).
 
@@ -13,7 +13,8 @@ To update it edit `docs/rules.md` in the
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
 | ----- | ------- | --------- | ---------- | ----------- | --------- | ---------- | -------- | -------- |
 | :x: | warning | dangerous_contract | | Dangerous XPCOM contract ID |  | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py)|
-| :x: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | | |
+| :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_NEWTHREAD |
+| :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_PROCESSNEXTEVENT |
 | :x: | warning | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
 | :x: | warning | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
 | :x: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | | | |
--- a/src/rules/javascript/banned_identifiers.js
+++ b/src/rules/javascript/banned_identifiers.js
@@ -1,0 +1,17 @@
+import { BANNED_IDENTIFIERS } from 'const';
+
+export default function(context) {
+  return {
+    Identifier: function(node) {
+      if (BANNED_IDENTIFIERS.indexOf(node.name) > -1) {
+        return context.report(node, `BANNED_${node.name.toUpperCase()}`);
+      }
+    },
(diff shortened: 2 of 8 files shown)


docs/rules.md
```diff
@@ -10,7 +10,8 @@ We should try and list out all the validation rules we know of by file type. E.g
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
 | ----- | ------- | --------- | ---------- | ----------- | --------- | ---------- | -------- | -------- |
 | :x: | warning | dangerous_contract | | Dangerous XPCOM contract ID |  | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py)|
-| :x: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | | |
+| :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | | BANNED_NEWTHREAD |
```
tofumatt: No rule ID for these ones in the old validator I'm guessing?

tofumatt: Oh, just looked in the code below, seems there is an old code. Should it be added in here?

muffinresearch: Yeah there is but it's sort of almalgamated under one id.

muffinresearch: I wonder actually if our current approach makes sense for this. It would be nice to just be able to add to the list of BANNED_IDENTIFERS rather than add a whole load of other things. But then that's where const bites us because you can't dynamically define constants. Maybe we should change that. Let's chat about it tomorrow.

src/scanners/javascript.js
```diff
@@ -33,13 +33,23 @@ export default class JavaScriptScanner {
           message.ruleId = messages.JS_SYNTAX_ERROR.code;
         }
 
+        var code = message.ruleId.toUpperCase();
+        var messageObj = messages[code];
+
+        // Fallback to looking up the message object by the
+        // constant string passed in context.report()
+        if (!messageObj && message.message) {
```
tofumatt: It would be good to add a test for this bit of code. It's implicitly tested by this rule I guess.

docs/rules.md
muffinresearch: >  Oh, just looked in the code below, seems there is an old code. Should it be added in here?

Yeah I've not done that yet (I presume you're referring to the tuple stuff which is legacyCode?)

src/scanners/javascript.js
muffinresearch: Yeah it blows up if description isn't found.

docs/rules.md
tofumatt: > Yeah I've not done that yet (I presume you're referring to the tuple stuff which is legacyCode?)

Yup, that was it. Cool, sounds good.

src/scanners/javascript.js
muffinresearch: It might make more sense to make the fallback the default then we'd be less reliant on the files?

src/rules/javascript/banned_identifiers.js
```diff
@@ -0,0 +1,17 @@
+import { BANNED_IDENTIFIERS } from 'const';
+
+export default function(context) {
+  return {
+    Identifier: function(node) {
```
muffinresearch: I notice my checks are slightly different to say the mozindexeddb prop ones  - we should maybe see if we can DRY these type of checks up - assuming it does make sense.

src/scanners/javascript.js
tofumatt: `message.message` is the string passed as the second argument? Yeah, I'd say we may that the default instead, it makes the argument seem less superfluous and we can use error/warnings/notices that way. I'd say make it the default/only way we fetch the `messageObject`.

src/rules/javascript/banned_identifiers.js
tofumatt: That makes sense. Possible that it just gets put into the banned identifiers list anyway. :smile:

tofumatt:
Looks good; happy to chat about the `BANNED_IDENTIFIERS` tomorrow.

If you wanted to make the fallback the only way of getting the `messageObj` that'd be great. I'm not sure ESLint considers that second message argument optional, but if it does we should probably have a test to make sure we require that argument.

I like the level of DRY for the banned identifier check but I agree I can see opportunities to do more of it. I'm certainly open to being flexible with our approach if it keeps our code footprint low as long as the rules are still followable (which these certainly are).

src/scanners/javascript.js
muffinresearch: I'll make that the only way in a separate patch it'll make for easier review. Can see about using message imports at the same time.

muffinresearch:
Ok this adds the smallest fixes to this rule - the rest can come in follow-ups.

muffinresearch:
Oh - I just need a test to tie the map and the list of banned_identifiers together.

muffinresearch:
@tofumatt ok - added a couple of tests. Looks like this if you just add a key to the list.

1__bash

tofumatt:
Cool, this looks good now. All r+

PR #177, opened 2015-10-29
Fix code display in rules.md

muffinresearch:
* Fixes emoji displayed in `<em:optionsType>`
* Fixes broken column.


tofumatt:
I don't see the emoji fix but otherwise seems good! r+bwefdym*

*bwefdym = But what emoji fix did you mean?

muffinresearch:
See http://mozilla.github.io/addons-validator/#installrdf  `<em:optionType>` is no longer: 

validator_rules

tofumatt:
Gotcha! In that case r+ without abbreviated caveats.

PR #184, opened 2015-11-02
Pass message code in context.report()

### LINKED ISSUE #164: Use a constant in JS rules report() calls
I realised that we're using a string here but could really be using our message codes: https://github.com/mozilla/addons-validator/blob/master/src/rules/javascript/mozindexeddb_property.js#L14

It's not a huge deal but it would be nice.

### LINKED ISSUE #179: Pass message code in context.report directly and remove the fallback
This is a bit more direct that look-ups based on the ruleId. We're doing this already but still falling back to the ruleId lookup. Let's move to passing the message.code 100%.

muffinresearch:
Fixes #179 and fixes #164

### DIFF
--- a/tests/scanners/test.javascript.js
+++ b/tests/scanners/test.javascript.js
@@ -3,7 +3,7 @@ import JavaScriptScanner from 'scanners/javascript';
 import * as messages from 'messages';
 import * as rules from 'rules/javascript';
 import { singleLineString } from 'utils';
-import { getRuleFiles } from '../helpers';
+import { getRuleFiles, unexpectedSuccess } from '../helpers';
 
 
 describe('JavaScript', function() {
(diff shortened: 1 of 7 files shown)


tests/scanners/test.javascript.js
```diff
@@ -96,6 +96,36 @@ describe('JavaScript', function() {
...
+    }
+
+    var FakeESLint = {
+      CLIEngine: FakeCLIEngine,
+    };
+
+    var jsScanner = new JavaScriptScanner('whatever', 'badcode.js');
+
+    return jsScanner.scan(FakeESLint)
+      .then(() => {
```
tofumatt: This could be shortened to:

```js
return jsScanner.scan(FakeESLint)
  .then(unexpectedSuccess)
  .catch((err) => {
    [...]
  });
```

tofumatt:
r+wc, coolio!

tests/scanners/test.javascript.js
muffinresearch: Ah yes, good catch!

Issue #187, opened 2015-11-02
unexpectedSuccess should be imported from tests/helpers not src/utils


tofumatt:
~~The `unexpectedSuccess` helper is only used in tests but is located in `src/utils.js`.~~

~~It should be moved to `tests/helpers.js` to keep things consistent/organised.~~

Some imports for `unexpectedSuccess` are using it from `utils` not `tests/helpers`. As this isn't ever called it went unnoticed that the import will be undefined.

muffinresearch:
Is it? It's here: https://github.com/mozilla/addons-validator/blob/master/tests/helpers.js#L62

tofumatt:
I didn’t get an error when I imported it from utils! I am surprised!

muffinresearch:
Yeah, presumably it's just undefined. You'd only get an error when you use it.

tofumatt:
Which _should_ never happen. Interesting! :-)

tofumatt:
Rather, it should be *imported* from `tests/helper.js`. It's not in README, at least. I'm reopening and I'll fix it to close this issue.

muffinresearch:
Oops - yeah must have missed that it crept in here https://github.com/mozilla/addons-validator/commit/c042f075fdf0f85bb5f66b61f7ac3bcf1b1ca29d

Also the readme is wrong.

muffinresearch:
> Rather, it should be imported from tests/helper.js. It's not in README, at least. I'm reopening and I'll fix it to close this issue.

Cool - I've renamed this appropriately.

PR #188, opened 2015-11-02
Refactor ChromeManifestScanner: extend BaseScanner

### LINKED ISSUE #144: ChromeManifestScanner should extend BaseScanner
As mentioned in #133, we can have the `ChromeManifestScanner` extend `BaseScanner` as well with minimal effort, but we'll do it in a separate issue.

tofumatt:
Closes #144.

### DIFF
--- a/src/scanners/chromemanifest.js
+++ b/src/scanners/chromemanifest.js
@@ -1,32 +1,15 @@
+import BaseScanner from 'scanners/base';
 import ChromeManifestParser from 'parsers/chromemanifest';
 import * as rules from 'rules/chromemanifest';
-import { ignorePrivateFunctions } from 'utils';
 
 
-export default class ChromeManifestScanner {
+export default class ChromeManifestScanner extends BaseScanner {
 
-  constructor(stream, filename) {
-    this.stream = stream;
-    this.filename = filename;
-    this._rulesProcessed = 0;
+  _defaultRules = rules;
+
+  _getContents(_ChromeManifestParser=ChromeManifestParser) {
+    var cmParser = new _ChromeManifestParser(this.contents, this.filename);
+    return cmParser.parse();
   }
 
-  scan(_ChromeManifestParser=ChromeManifestParser, _rules=rules) {
-    return new Promise((resolve) => {
-      var validatorMessages = [];
-      var cmParser = new _ChromeManifestParser(this.stream, this.filename);
-      return cmParser.parse()
-        .then((triples) => {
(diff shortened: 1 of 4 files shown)


tofumatt:
I cleaned up a few classes while I was at it; hope you don't mind it being mixed in with this commit.

tests/scanners/test.chromemanifest.js
```diff
@@ -1,7 +1,7 @@
 import ChromeManifestScanner from 'scanners/chromemanifest';
 import * as rules from 'rules/chromemanifest';
 import { getRuleFiles, validChromeManifest } from '../helpers';
-import { ignorePrivateFunctions } from 'utils';
+import { ignorePrivateFunctions, unexpectedSuccess } from 'utils';
```
muffinresearch: This shouldn't work as it *is* in helpers. Re: #187

tofumatt: Oh weird, I saw it imported this way elsewhere.

muffinresearch: oh, ok - that's a definite bug/mistake then.

src/scanners/chromemanifest.js
```diff
@@ -1,32 +1,21 @@
...
 
-  scan(_ChromeManifestParser=ChromeManifestParser, _rules=rules) {
-    return new Promise((resolve) => {
-      var validatorMessages = [];
-      var cmParser = new _ChromeManifestParser(this.stream, this.filename);
-      return cmParser.parse()
+  _getContents(_ChromeManifestParser=ChromeManifestParser) {
+    return new Promise((resolve, reject) => {
+      var cmParser = new _ChromeManifestParser(this.contents, this.filename);
+      cmParser.parse()
```
muffinresearch: Can't you just return this and not create a new Promise?

muffinresearch:
I think you can kill the unnecessary Promise. Otherwise r+wc

src/scanners/chromemanifest.js
tofumatt: Could do; we won't catch a parse error but I suppose it doesn't even implement `reject` right now anyway.

tofumatt:
Sweet. screenshot 2015-11-03 16 21 54

src/scanners/chromemanifest.js
muffinresearch: But if `parse()` blows up that'll implicitly reject the returned promise won't it?

muffinresearch:
more-red-than-green-award

PR #193, opened 2015-11-03
Update eslint to 1.8.0

muffinresearch:
Did this to see if we could hold off using babel-eslint, but looks to be still necessary for modules (in our code, not the rules).


muffinresearch:
Hopefully we can get greenkeeper hooked up again.

tofumatt:
r+

PR #204, opened 2015-11-04
Remove args that are noop

### LINKED ISSUE #203: Remove CLI args not currently no-op
Given we're going to be moving ahead with using the app sooner for metadata extraction let's remove the CLI args that aren't in active use and we can re-instate them at a later date as they become usable.

It could be confusing to have a whole bunch of no-op args.

muffinresearch:
Fixes #203

Individual commits so they're easy to re-instate from history later.

### DIFF
--- a/tests/test.cli.js
+++ b/tests/test.cli.js
@@ -35,11 +35,6 @@ describe('Basic CLI tests', function() {
     assert.equal(args.pretty, false);
   });
 
-  it('should default determined to false', () => {
-    var args = cli.parse(['foo/bar.xpi']);
-    assert.equal(args.determined, false);
-  });
-
   it('should default boring to false', () => {
     var args = cli.parse(['foo/bar.xpi']);
     assert.equal(args.boring, false);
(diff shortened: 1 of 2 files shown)


tofumatt:
> Individual commits so they're easy to re-instate from history later.

You understand `git revert`? Witch!

tests/test.cli.js
```diff
@@ -62,13 +46,6 @@ describe('Basic CLI tests', function() {
       'Not enough non-option arguments'));
   });
 
-  it('should show error if incorrect type', () => {
```
tofumatt: Makes more sense to remove than to skip?

I wonder about even just commenting things out with a TODO that says "nix these if we still aren't using them by v1" or something; technically less git magic but it seems easier to manage in my brain.

tofumatt:
r+wc; I leave the logistics up to you :smile:

tests/test.cli.js
muffinresearch: Yeah initially I was just going to comment out the args and skip the tests. But that is pretty clunky. It seemed if we could use history to bring back the arg it might as well have the test there too. 

Less clutter that way.

tofumatt: Fair enough!

PR #206, opened 2015-11-04
Create rule for complex defaults code

### LINKED ISSUE #109: Rule: complex_prefs_defaults_code
Implement the `complex_prefs_defaults_code` rule for JS validator.

Nolski:
Resolves #109


src/rules/javascript/complex_prefs_defaults.js
```diff
@@ -0,0 +1,26 @@
+import { COMPLEX_PREFS_DEFAULTS } from 'messages/javascript';
+
+export default function(context) {
+  var filename = context.getFilename();
+
+  if (filename.indexOf('defaults/preferences/') === 0) {
```
tofumatt: Explaining why we do this inside the rule would be handy. I don't mind it being used outside the event function `CallExpression` but just explaining what purpose it serves would be nice.

Also: we should have a test to see what happens if no filename is supplied. We don't usually fail to pass a filename but in theory ESLint can take stuff from `stdin` and I just like covering bases. A simple check to see if `filename` is truthy in here along with the `indexOf` makes sense to me.

src/rules/javascript/complex_prefs_defaults.js
```diff
@@ -0,0 +1,26 @@
...
+    return {
+      CallExpression: function(node) {
+        var callee = node.callee;
+        var root;
+
+        if (callee.type === 'Identifier') {
+          root = callee;
+        } else if (callee.type === 'MemberExpression') {
```
tofumatt: Can neither of these be true? (`root` would be undefined then and error out below)

src/messages/javascript.js
```diff
@@ -107,3 +107,15 @@ export const EVAL_STRING_ARG = {
     called only with function expressions as their first argument`),
   legacyCode: ['javascript', 'dangerous_global', 'eval'],
 };
+
+export const COMPLEX_PREFS_DEFAULTS = {
+  code: 'COMPLEX_PREFS_DEFAULTS',
+  message: _('Complex code should not appear in preference defaults files'),
+  description: _(singleLineString`Code in preference defaults files should
```
tofumatt: I would change this description to something like: "Preference files are only allowed to call the pref() and user_pref() functions. Please move other code outside preference defaults file." (Feel free to remove the second second if it's TOO wordy)

tests/rules/javascript/test.complex_prefs_defaults.js
```diff
@@ -0,0 +1,42 @@
+import JavaScriptScanner from 'scanners/javascript';
+import { VALIDATION_WARNING } from 'const';
+import { COMPLEX_PREFS_DEFAULTS } from 'messages/javascript';
+
+describe('complex_prefs_defaults', () => {
+  it('should pass when only calling pref', () => {
+    var code = 'pref();';
+    var jsScanner = new JavaScriptScanner(code,
```
tofumatt:
```js
    var jsScanner = new JavaScriptScanner(code, 'defaults/preferences/pref.js');
```
Should fit (just) in 80 chars, no?

tests/rules/javascript/test.complex_prefs_defaults.js
```diff
@@ -0,0 +1,42 @@
...
+                                          'defaults/preferences/pref.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 0);
+      });
+  });
+
+  it('should fail due to complex code', () => {
```
tofumatt: The message is complex code, but really it's just not calling the global `pref()`; something like:

```js
it('should fail if pref() is not called as a global')
```
or something like that, as we're essentially making sure pref() as a method of anything but the global object isn't okay.

tests/rules/javascript/test.complex_prefs_defaults.js
```diff
@@ -0,0 +1,42 @@
+import JavaScriptScanner from 'scanners/javascript';
+import { VALIDATION_WARNING } from 'const';
+import { COMPLEX_PREFS_DEFAULTS } from 'messages/javascript';
+
+describe('complex_prefs_defaults', () => {
```
tofumatt: You never check for `user_pref`; I'd add that check for completeness.

docs/rules.md
```diff
@@ -15,7 +15,7 @@ To update it edit `docs/rules.md` in the
 | :negative_squared_cross_mark: | warning | dangerous_contract | | Dangerous XPCOM contract ID |  | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py)| | ('js', 'actions', 'dangerous_contract') | **Removed** |
 | :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_NEWTHREAD |
 | :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_PROCESSNEXTEVENT |
-| :x: | warning | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
+| :white_check_mark: | warning | complex_prefs_defaults | | Complex code should not appear in preference defaults files | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py#L427) | ('testcases_javascript_actions', '_call_expression', 'complex_prefs_defaults_code')| COMPLEX_PREFS_DEFAULTS |
```
tofumatt: Do we want to change our code name for this? "Complexity" is a misnomer here; we literally allow only two methods with very specific functionality :smile: 

Maybe our code should be `ONLY_PREFS_IN_DEFAULTS`

src/rules/javascript/complex_prefs_defaults.js
tofumatt: Come to think of it, I believe the `JavaScriptScanner` class won't fail without a `filename` (or with an `undefined`/`null` one).

We should either demand a filename or we should check for `filename` existing (that'd be a separate issue). Thoughts?

src/rules/javascript/complex_prefs_defaults.js
```diff
@@ -0,0 +1,26 @@
...
+          root = callee.object;
+        }
+
+        if (root.name !== 'pref' && root.name !== 'user_pref') {
```
tofumatt: I'd at least check to see if `root` isn't `undefined`/`null` here.

src/rules/javascript/complex_prefs_defaults.js
```diff
@@ -0,0 +1,26 @@
+import { COMPLEX_PREFS_DEFAULTS } from 'messages/javascript';
+
+export default function(context) {
+  var filename = context.getFilename();
+
+  if (filename.indexOf('defaults/preferences/') === 0) {
+    return {
+      CallExpression: function(node) {
+        var callee = node.callee;
```
tofumatt: This seems a bit superfluous; doesn't help with readability much and doesn't shorten any lines; you can probably just use `node.callee` directly below.

tofumatt:
All looked over for now; I left a few questions and there are some things that need tweaking. Let me know when you want me to have another look, but overall patch is :+1:

src/rules/javascript/complex_prefs_defaults.js
muffinresearch: Yeah I think it sounds like we need a separate issue for making file name passing more robust especially if we're relying on it for narrowing tests down.

muffinresearch: I've filed #207

Nolski:
@tofumatt I had to kind of go back and rethink about how I was parsing for that first symbol in a chained expression but the newly added utility and updated tests seem to work so this should be good to take another look at.

tofumatt:
Looks like you're missing a test: coveralls should tell you which lines need coverage but ping me again once there's another test in place :smile:

docs/rules.md
```diff
@@ -15,7 +15,7 @@ To update it edit `docs/rules.md` in the
 | :negative_squared_cross_mark: | warning | dangerous_contract | | Dangerous XPCOM contract ID |  | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py) | ('js', 'actions', 'dangerous_contract') | **Removed** |
 | :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_NEWTHREAD |
 | :white_check_mark: | warning |  banned_identifier | | Banned or deprecated JavaScript Identifier | | | ('js', 'actions', 'banned_identifier') | BANNED_PROCESSNEXTEVENT |
-| :x: | warning | complex_prefs_defaults_code | | Complex code should not appear in preference defaults files | | | | |
+| :white_check_mark: | warning | complex_prefs_defaults | | Complex code should not appear in preference defaults files | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py#L427) | ('testcases_javascript_actions', '_call_expression', 'complex_prefs_defaults_code')| ONLY_PREFS_IN_DEFAULTS |
```
tofumatt: Nice putting the line number in.

tests/test.utils.js
```diff
@@ -37,6 +37,41 @@ describe('utils.singleLineString()', function() {
 
 });
 
+describe('utils.getRootExpression()', function() {
+  var node = {
```
tofumatt: I can imagine us needing node creation helpers for tests eventually :smile:

tofumatt:
Looks good to me, sorry for the late review! r+

PR #208, opened 2015-11-04
Detect require() of deprecated widget module

### LINKED ISSUE #113: Rule: widget
Implement the `widget` rule for JS validator.

muffinresearch:
Fixes #113

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -19,9 +19,10 @@ To update it edit `docs/rules.md` in the
 | :x: | warning | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
 | :white_check_mark: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/javascript/actions.py#L488) | | EVAL_STRING_ARG |
 | :x: | warning | low_level_module (not from src) | | Usage of low-level or non-SDK interface | | | | |
-| :x: | warning | widget | | Use of deprecated SDK module | | | | |
+| :white_check_mark: | warning | widget | | Use of deprecated SDK module | | | null | DEPREC_SDK_MOD_WIDGET |
 | :negative_squared_cross_mark: | notice |  \_readonly_top | | window.top is a reserved variable | | | ('testcases_javascript_actions', '_readonly_top' | **Removed** |
 | :x: | warning | global_overwrite | | Global variable overwrite | | | | |
+| :white_check_mark: | warning | unexpected_global_arg [NEW] | | Unexpected global passed as an argument | | | null | UNEXPECTED_GLOGAL_ARG |
 
 ### Call definitions
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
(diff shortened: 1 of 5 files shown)


docs/rules.md
```diff
@@ -19,9 +19,10 @@ To update it edit `docs/rules.md` in the
 | :x: | warning | called_dangerous_global | | `%s` called in potentially dangerous manner' | | | | |
 | :white_check_mark: | error? | eval | | In order to prevent vulnerabilities, the `setTimeout` 'and `setInterval` functions should be called only with function expressions as their first argument. | | [testcases/javascript/actions.py](https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/javascript/actions.py#L488) | | EVAL_STRING_ARG |
 | :x: | warning | low_level_module (not from src) | | Usage of low-level or non-SDK interface | | | | |
-| :x: | warning | widget | | Use of deprecated SDK module | | | | |
+| :white_check_mark: | warning | widget | | Use of deprecated SDK module | | | | DEPREC_SDK_MOD_WIDGET |
```
tofumatt: No old code for this one? Adding `null` like you did below is nice to indicate that at least :smile:

tests/rules/javascript/test.widget_module.js
```diff
@@ -0,0 +1,61 @@
...
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 1);
+        assert.equal(validationMessages[0].code,
+                     messages.DEPREC_SDK_MOD_WIDGET.code);
+        assert.equal(validationMessages[0].type, VALIDATION_WARNING);
+      });
+  });
+
+  it('should catch require() first arg being a global', () => {
```
tofumatt: Can you add a test of it not raising an error when it's `var widgetPath` just for completeness? I think it's good if we have examples of what TO do as well as what not to do.

tofumatt:
r+wc

tests/rules/javascript/test.widget_module.js
muffinresearch: The var test above this deals with that, e.g. there only one message and it's not the UNEXPECTED_GLOBAL one?

docs/rules.md
muffinresearch: Yep, - I couldn't see it unless it's somewhere else. Well spotted!

tests/rules/javascript/test.widget_module.js
tofumatt: Hmm, true.

I think I'm starting to like the idea of having tests with zero failures so people have an idea of what code is acceptable from the tests as well when they go back and reference this repo (I often don't find them in the old codebase and it's frustrating).

I like that idea.

But if it's too noisy I'll stand down.

muffinresearch: Oh ok. I see what you mean - I'll see if I can add something along those lines. 

I don't want to overlap tests too much but I *do* like the idea of making our tests more of a an explanation as to what's possible.

muffinresearch: Yep - so looking at the code I don't see a good way to explain this rule further that isn't overlapping with the way the other two tests are working as they're already very explicit. This specific case it doesn't make much sense to put some different test code and assert nothing is caught - because that's too open-ended imho - but maybe I'm misunderstanding something as to how you see we could enhance the story the tests tell?

If you have an idea how to do this feel free to add an issue and we can tackle it separately.

tofumatt: Fair enough. Sounds like I’m overcomplicating things, all good! :)

PR #209, opened 2015-11-04
Remove `deprec` rule; close #123

### LINKED ISSUE #123: Rule: deprec
Implement the `deprec` rule for JS validator.

tofumatt:
http://people.mozilla.org/~amckay/out.html indicates this is rarely
encountered and is from single-digit-Gecko version days. Bye bye! 🙋


muffinresearch:
r+

PR #215, opened 2015-11-05
Add low-level module rule

### LINKED ISSUE #112: Rule low_level_module
Implement the `low_level_module` rule for JS validator.

muffinresearch:
Fixes #112

* This splits the UNEXPECTED_GLOBAL out into its own rule.
* Covers the basic behaviour of the low_level checks.

@kmaglione - could you clarify the rationale for stripping of 'sdk/' in the old validator here? https://github.com/mozilla/amo-validator/blob/master/validator/testcases/javascript/actions.py#L533 it doesn't appear to tie up with the regex tests that were checking this stuff.  Are there requires that might be `sdk/sdk/[module]` or is everything needed already in that list? 

I've left the sdk prefix stripping out at the moment because it makes things harder to read and grok what's actually being looked for.

### DIFF
--- a/src/rules/javascript/low_level_module.js
+++ b/src/rules/javascript/low_level_module.js
@@ -1,20 +1,21 @@
-import { DEPREC_SDK_MOD_WIDGET } from 'messages';
+import { LOW_LEVEL_MODULES } from 'const';
+import { LOW_LEVEL_MODULE } from 'messages';
 import { getVariable } from 'utils';
 
-const WIDGET_PATH = 'sdk/widget';
 
 /*
  * This rule will detect use of `require()` with the first arg being either
- * a literal that matches the widget module or var pointing at a literal.
+ * a literal that matches a restricted module or a var pointing to a literal
+ * that matches a restricted value.
  *
- * TODO: This rule looks to be related to compat - based on the old tests.
  * TODO: This rule should only be run for jetpack.
+ * TODO: Check what the requires_chrome feature does in the old code.
  *
  */
 export default function(context) {
   return {
     CallExpression: function(node) {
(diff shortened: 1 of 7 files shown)


muffinresearch:
@tofumatt r? - rather than block on this we can ask the questions later.

src/rules/javascript/low_level_module.js
```diff
@@ -0,0 +1,48 @@
...
+      if (node.callee.name === 'require' &&
+          node.arguments &&
+          node.arguments.length) {
+
+        var firstArg = node.arguments[0];
+
+        // Find a literal string value passed to the
+        // the require function.
+        if (firstArg.type === 'Literal' &&
+            LOW_LEVEL_MODULES.indexOf(firstArg.value) > -1) {
```
tofumatt: Does ES6 have `contains()` or is that a mozilla-specific thing that never went anywhere?

tofumatt:
r+

src/rules/javascript/low_level_module.js
muffinresearch: Do you mean `Array.prototype.includes()` - that's in es2016 - but that should mean we have it - I'll check.

tofumatt: Oops, yeah, I think contains was a Moz-only thing. If we have it we should use it (just clearer) but otherwise the PR is r+ from me!

muffinresearch: Yep - we do have it and it works. Nicely spotted. Apparently it was previously called `contains` but the spec got changed. https://github.com/tc39/Array.prototype.includes

PR #216, opened 2015-11-05
Increase mocha test timeout

johannhof:
The default timeout is 2000ms, which tests like "Base Scanner Class
should run all rules" sometimes exceed, even on my fast MacBook.

If you look at the Travis output you can see that that test is getting dangerously close to 2s, last value was 1886ms and I suppose we want to avoid intermittent failures.

### DIFF
--- a/tasks/mochaTest.js
+++ b/tasks/mochaTest.js
@@ -1,5 +1,6 @@
 module.exports = {
   options: {
+    timeout: 5000,
     require: [
       function() {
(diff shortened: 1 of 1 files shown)


tofumatt:
Works for me. Mind adding a tiny comment about why we do this? :smile:

tofumatt:
Never mind actually, I'm just being crazy. r+!

PR #228, opened 2015-11-10
Refactor manifest parsing

### LINKED ISSUE #191: Parse manifest.json
This is the equivalent of install.rdf parsing for the new Web Extensions.

Supports #33

### LINKED ISSUE #226: Create install.rdf specific parser.
This will extract the specific fields we're interested in for metadata purposes.

muffinresearch:
Fixes #226 
Fixes #191 

I ended up moving the install.rdf parsing bits out to its own class and similar for manifest.json. Fallbacks to type detection based on layout remain in the validator for now.

I've made some assumptions about `manifest_version` for the `manifest.json`. Those will need to be checked later but this was a way to get us off the blocks with parsing of both types of manifest integrated together.

### DIFF
--- a/src/parsers/installrdf.js
+++ b/src/parsers/installrdf.js
@@ -1,0 +1,54 @@
+import { ADDON_TYPE_MAP, INSTALL_RDF } from 'const';
+import { TYPE_INVALID, TYPE_MISSING } from 'messages';
+import log from 'logger';
+import RDFScanner from 'scanners/rdf';
+
+
+export default class InstallRdfParser {
+
+  constructor(rdfString, collector) {
+    this.rdfString = rdfString;
+    // Provides ability to directly add messages to
+    // the collector.
+    this.collector = collector;
+  }
+
+  parseDoc() {
+    var rdfScanner = new RDFScanner(this.rdfString, INSTALL_RDF);
+    return rdfScanner.getContents();
+  }
+
+  getMetaData() {
+    return this.parseDoc()
+      .then((xmlDoc) => {
+        var addonMetaData = {};
+        // Sync metadata property lookups.
+        addonMetaData.type = this._getAddonType(xmlDoc);
+        return Promise.resolve(addonMetaData);
+      });
+  }
+
+  _getAddonType(xmlDoc) {
+    var addonType = null;
+    var typeNodes = xmlDoc.getElementsByTagName('em:type');
+    if (typeNodes.length > 1) {
+      throw new Error('Multiple <em:type> elements found');
+    }
+    var node = typeNodes[0];
+    if (node && node.firstChild && node.firstChild.nodeValue) {
+      var typeValue = node.firstChild.nodeValue;
+      if (!ADDON_TYPE_MAP.hasOwnProperty(typeValue)) {
+        log.debug('Invalid type value "%s"', typeValue);
+        this.collector.addError(TYPE_INVALID);
+      } else {
+        var addonType = ADDON_TYPE_MAP[typeValue];
+        log.debug('Mapping original <em:type> value "%s" -> "%s"',
+                  typeValue, addonType);
+      }
+    } else {
+      log.warn('<em:type> was not found in install.rdf');
+      this.collector.addNotice(TYPE_MISSING);
+    }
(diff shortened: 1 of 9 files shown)


src/parsers/installrdf.js
```diff
@@ -0,0 +1,56 @@
+import { ADDON_TYPE_MAP, INSTALL_RDF } from 'const';
+import { TYPE_INVALID,
+         TYPE_MISSING } from 'messages';
```
tofumatt: Nitpick, but this could fit on one line.

src/parsers/installrdf.js
```diff
@@ -0,0 +1,56 @@
...
+import RDFScanner from 'scanners/rdf';
+
+
+export default class InstallRdfParser {
+
+  constructor(rdfString, collector) {
+    this.rdfString = rdfString;
+    // Provides ability to directly add messages to
+    // the collector.
+    this.collector = collector;
```
tofumatt: I thought of parsers as simply parsing data for the scanners to look through, and with the exception of things like parse errors they wouldn't handle errors. Things like the `TYPE_INVALID` and `TYPE_MISSING` errors seem like they'd go in the scanner, while the parser is just the code that converts a format (in this case RDF) into a JS API/object for us to look through.

src/validator.js
```diff
@@ -7,14 +7,16 @@ import promisify from 'es6-promisify';
...
 import * as exceptions from 'exceptions';
 import * as messages from 'messages';
-import { getPackageTypeAsString,
-         checkMinNodeVersion,
+import { checkMinNodeVersion,
          gettext as _, singleLineString } from 'utils';
 
 import log from 'logger';
 import Collector from 'collector';
+import InstallRdfParser from 'parsers/installrdf';
```
tofumatt: Again, as above, I'd think the `Validator` would be dealing only with `Scanner` classes, while the `Scanner` classes would use a `Parser` class internally to process files.

Maybe it's just a semantic thing and these should be called `Scanner`s and not `Parser`s.

tofumatt:
I think this is good but I wonder about at least the semantics of calling these new files "parsers". Largely the code is good, maybe it's just semantics.

src/parsers/installrdf.js
tofumatt: Scratch this comment. I'm looking over this again as I feel like my brain wasn't working this evening. I get what's going on now and this makes perfect sense. Sorry! :disappointed:

src/validator.js
```diff
@@ -272,6 +189,70 @@ export default class Validator {
     return output;
   }
 
+  getAddonMetaData() {
+    return this.xpi.getMetaData()
```
tofumatt: Should this logic go into the `XPI` class so the `Validator` class isn't directly loading Parsers? I like the idea of stripping the validator down in terms of the logic in it.

tofumatt:
I was dumb about my earlier review; this weekend has been hard to focus and my brain is finally working.

r+wc; let me know what you think about moving the add-on type into the XPI class, but if you think that's not a good idea feel free to merge!

src/validator.js
muffinresearch: Yeah - I know what you mean - I'd like to separate concerns as much as possible too. One problem is the collector is needed in the manifest parsers. So to move this would require passing the collector to the xpi class and then from there to the manifest parsers. So there's a bit of a balancing act there.

Also if we move the metadata parsing into the xpi class there's some potential confusion around the xpi meta + the meta from the manifest files. Currently the xpi class has remained pretty dumb in that it only knows and cares about xpi handling which feels right to me.

I think I'll land this now and work with it a bit. But, if we see further scope for refactoring (I feel like there's still a bit of that we can do) then we should do so. One nice thing is it does feel possible to refactor fairly easily since the tests should catch breakage.

Something we could think about is removing the error/notice collection from the metadata parsing and move it up into the validator? I like the idea of the classes only needing to do one job - but then again passing the collector was easy to do and keeps the error collection close to the code it cares about. Another thing with that though is we could de-dupe some of the errors. However there might be a corner to paint ourselves into there as the fields differ quite a bit.

PR #230, opened 2015-11-10
Detect jpm add-on type

tofumatt:
Pretty sure this is what we were supposed to do here; we just want a way to see if an add-on is a Jetpack or not.

This expands the `metadata` property of an `XPI` to include the XPI's architecture (eg Extension, Jetpack, or Web Extension) but could be expanded to include other types.

This is how the original AMO validator does it so it's how I've implemented it, but if it looks good I'll add another bug for checking the contents of `bootstrap.js` against what `cfx`/`jpm` generate so only valid `bootstrap.js` files trigger the change.


src/xpi.js
```diff
@@ -68,6 +73,18 @@ export default class Xpi {
           // after the last entry event is emitted and streams
           // may still be being read with openReadStream.
           zipfile.on('close', () => {
+            this.metadata._processed = true;
+
+            // If we find a file named bootstrap.js this is assumed to be a
+            // Jetpack add-on: https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/jetpack.py#L154
+            // TODO: Check against file contents to make this more robust.
+            if (this.entries.includes('bootstrap.js') && (
```
muffinresearch: This might work better with the `(` after the `&&` on the next line.

src/xpi.js
```diff
@@ -68,6 +73,18 @@ export default class Xpi {
...
+              this.entries.includes('harness-options.json') ||
+              this.entries.includes('package.json')
+            )) {
+              this.metadata.architecture = ARCH_JETPACK;
```
muffinresearch: Would be good to add logging here.

PR #231, opened 2015-11-11
Require filenames in scanners (fix #207)

### LINKED ISSUE #207: Require filenames to be passed to each scanner.
We should blow up if this isn't the case.

tofumatt:
Simple fix to make sure we always have filenames in our scanners.

### DIFF
--- a/src/exceptions.js
+++ b/src/exceptions.js
@@ -12,5 +12,6 @@ export class ExtensibleError extends Error {
 }
 
 export class DuplicateZipEntryError extends ExtensibleError {}
-export class RDFParseError extends ExtensibleError {}
+export class MissingFilenameError extends ExtensibleError {}
 export class NotImplentedError extends ExtensibleError {}
+export class RDFParseError extends ExtensibleError {}
--- a/src/scanners/base.js
+++ b/src/scanners/base.js
@@ -1,5 +1,5 @@
 import { NotImplentedError } from 'exceptions';
-import { ignorePrivateFunctions } from 'utils';
+import { ensureFilenameExists, ignorePrivateFunctions } from 'utils';
 
 
 export default class BaseScanner {
@@ -12,6 +12,8 @@ export default class BaseScanner {
     this._parsedContent = null;
     this._rulesProcessed = 0;
     this.options = {};
+
+    ensureFilenameExists(this.filename);
   }
 
   scan(_rules=this._defaultRules) {
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -5,7 +5,7 @@ import ESLint from 'eslint';
 import { ESLINT_TYPES } from 'const';
 import * as messages from 'messages';
 import ESLintRules from 'rules/javascript';
-import { singleLineString } from 'utils';
+import { ensureFilenameExists, singleLineString } from 'utils';
 
 
 export default class JavaScriptScanner {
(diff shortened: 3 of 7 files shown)


src/scanners/base.js
```diff
@@ -12,6 +12,12 @@ export default class BaseScanner {
     this._parsedContent = null;
     this._rulesProcessed = 0;
     this.options = {};
+
+    if (this.filename === undefined || this.filename === null ||
```
muffinresearch: Presumably if this.filename was `0` e.g. the number zero, then that would be invalid too because it should be a string. In which case this might be a good case to use a falsey check.

However, if you prefer this more explicit route  it would be good to use `typeof` for the undefined check.

src/scanners/javascript.js
```diff
@@ -13,6 +14,12 @@ export default class JavaScriptScanner {
   constructor(code, filename) {
     this.code = code;
     this.filename = filename;
+
+    if (this.filename === undefined || this.filename === null ||
```
muffinresearch: Seeing this twice makes me think a util func might be worth doing.

tofumatt: Good point, and yeah I forgot about typeof, oops!

I'll clean it up a bit. Testing for a string makes sense too... Maybe it should just be typeof === string and length > 0

- tofumatt

tofumatt:
Okay, fixed things up a bit, ready for another r?

src/utils.js
```diff
@@ -131,6 +132,17 @@ export function ignorePrivateFunctions(list) {
 }
 
 
+/*
+ * Check a filename to make sure it's valid; used by scanners so we never
+ * accept new scanners that don't specify which file they're referencing.
+ */
+export function ensureFilenameExists(filename) {
+  if (typeof(filename) !== 'string' || filename.length < 1) {
```
muffinresearch: `typeof` is an operator - I think this works because it's being interpreted as  `typeof (filename)`.

tests/test.utils.js
```diff
@@ -109,6 +110,45 @@ describe('utils.getVariable()', function() {
 });
 
 
+describe('utils.ensureFilenameExists()', function() {
+
+  it('should throw error when filename is not a string', () => {
+    assert.throws(() => {
+      // An empty filename doesn't count.
```
muffinresearch: These comments are probably superfluous as the test description covers it.

src/exceptions.js
```diff
@@ -12,5 +12,9 @@ export class ExtensibleError extends Error {
 }
 
 export class DuplicateZipEntryError extends ExtensibleError {}
-export class RDFParseError extends ExtensibleError {}
+
```
muffinresearch: I'd got rid of this space as it seemed unecessary - but can add that back if you prefer.

muffinresearch:
Couple of minor nits but otherwise r+wc

src/exceptions.js
tofumatt: No you're totally right, _I_ thought you liked them. I'll remove :smile:

src/utils.js
tofumatt: Yeah this is a bad habit of mine, sorry. Thanks for catching it.

tests/test.utils.js
tofumatt: They are copypasta too, whoops. :open_mouth:

Issue #195, opened 2015-11-03
Extract metadata from manifests


muffinresearch:
Here’s the information that is needed at the moment:
- [x] guid
- [x] ~~type: extension, dictionary, langpack, search tool…~~ 
- [x] name
- [x] applications
- [x] version
- [x] homepage (optional)
- [x] summary (optional)
- [x] no restart (optional)
- [x] strict compatibility (optional)
- [x] target applications (eg Firefox, Thunderbird…)
- [x] is multi package

muffinresearch:
This can be split out into more than one issue if it makes sense.

muffinresearch:
Looks like this is resolved. Closing

PR #232, opened 2015-11-11
Test for max GUID size (fix #214)

### LINKED ISSUE #214: Check max guid size
The max guid size that we can store in olympia is 255 chars. Let's throw a hard failure if the guid size is bigger than that. That should be enough for anyone right?

See https://bugzilla.mozilla.org/show_bug.cgi?id=1210032 for more information.

tofumatt:
Assuming this will also need to check ID length in other metadata (`package.json`) so we can wait until #228 lands and I can build on top of that.

### DIFF
--- a/src/parsers/installrdf.js
+++ b/src/parsers/installrdf.js
@@ -4,6 +4,9 @@ import log from 'logger';
 import RDFScanner from 'scanners/rdf';
 
 
+// HACK: Remove before merging ^_^
+var namespace = 'http://www.mozilla.org/2004/em-rdf#';
+
 export default class InstallRdfParser {
 
   constructor(rdfString, collector) {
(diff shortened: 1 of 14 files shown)


tofumatt:
It could also make sense to test for this by using normalised metadata in #195 rather than checking inside each parser.

muffinresearch:
Yep, this is something that can probably happen in the newly landed `Validator.getAddonMetaData()` - as that's a centralized place which will be able to look at metadata that came from either `install.rdf` or `metadata.json`.

tofumatt:
I think this is good for review now.

src/parsers/installrdf.js
```diff
@@ -4,6 +4,9 @@ import log from 'logger';
 import RDFScanner from 'scanners/rdf';
 
 
+// HACK: Remove before merging ^_^
+var namespace = 'http://www.mozilla.org/2004/em-rdf#';
```
muffinresearch: Is this still needed?

muffinresearch: Or maybe should be a `const`

tofumatt: Oops, crud, I was supposed to fix that. Thanks!

muffinresearch:
Lgtm r+wc

PR #236, opened 2015-11-11
Remove () for typeof

tofumatt:
As mentioned in #231, these aren't needed around `typeof`, so I cleared them out to be consistent.


muffinresearch:
We should add a custom eslint/jscs rule.

r+

tofumatt:
Good call; if I can do that I’ll include it in this PR before I merge.

tofumatt:
Couldn't find an existing rule so yeah, we should add a style rule for this eventually ^_^

PR #237, opened 2015-11-11
Created rules for numerous deprecated entities

### LINKED ISSUE #235: Rule: evil
Implement the `evil` rule.

https://github.com/mozilla/addons-validator/blob/master/docs/rules.md#entity-values

Nolski:
Resolves #235


tests/rules/javascript/test.no_document_write.js
```diff
@@ -0,0 +1,38 @@
+import { VALIDATION_WARNING } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+import * as messages from 'messages';
+
+describe('no_document_write', () => {
+  it('it should warn about using document.write()', () => {
```
muffinresearch: This should read naturally so you can remove the 2nd 'it'.

tests/rules/javascript/test.no_document_write.js
```diff
@@ -0,0 +1,38 @@
...
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 0);
+      });
+  });
+
+  it('it should not warn about calling write in general', () => {
+    var code = 'foo.write("<h1>foobarbaz</h1>");';
```
muffinresearch: What happens if `foo` is a reference to `document` ?

src/rules/javascript/no_document_write.js
```diff
@@ -0,0 +1,16 @@
+import { NO_DOCUMENT_WRITE } from 'messages/javascript';
+
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      // Check if what's being called is setTimeout or setInterval
```
muffinresearch: Old comment.

muffinresearch:
The `rules.md`  docs says: `Deprecated entities are checked with a generated set of rules` - so should this be checking a list of deprecated entities with the same code - you could probably kill a big chunk of the rules if that would work.

Nolski:
@muffinresearch Sure, I was going to create each entity as a separate rule and pull request. Are you saying I should pull them all into one rule and one pull request?

muffinresearch:
> @muffinresearch Sure, I was going to create each entity as a separate rule and pull request. Are you saying I should pull them all into one rule and one pull request?

I've not looked at the code but if this would lend itself to being one rule and iterating over a data structure instead of multiple rules then it would make sense to do this in one rule. You'd need some way to generate the messages but we have examples of that in the messages file.

If adding to a data structure requires additional messages then we just add a test for that so that someone will know to update both things.

See https://github.com/mozilla/addons-validator/blob/master/src/rules/javascript/banned_identifiers.js and  https://github.com/mozilla/addons-validator/blob/master/tests/test.messages.js#L73 as an example of a similar thing.

Nolski:
@muffinresearch I'm not convinced that this would be easier if done in one rule. Since many of the entities vary quite a bit (for instance we have `document.write()` and `eval()`). I think it'd be possible to make a few hoops but it might be a bit easier to just have separate rules for each of them (I think there's only 5 or 6). Perhaps we can make this a bit easier to write by consolidating the testing?

muffinresearch:
> What happens if foo is a reference to document ?

What I was driving at with this was more a test that catches this:

```javascript
var foo = document;
foo.write('<h1>hai</h1>');
```

muffinresearch:
> @muffinresearch I'm not convinced that this would be easier if done in one rule. Since many of the entities vary quite a bit (for instance we have document.write() and eval()). I think it'd be possible to make a few hoops but it might be a bit easier to just have separate rules for each of them (I think there's only 5 or 6). Perhaps we can make this a bit easier to write by consolidating the testing?

Ok - let's keep going as you are. But if you find there's repetition then it'll be better to refactor this than have a bunch of rules all doing the same logic in multiple rules.

Nolski:
@muffinresearch This should be good to take another look at when you get the chance.

tests/test.utils.js
```diff
@@ -74,6 +74,39 @@ describe('utils.getRootExpression()', function() {
   });
 });
 
+describe('utils.getNodeReferenceName()', () => {
+  var context = {
```
Nolski: This is starting to get bothersome but I'm not sure if there is a good way to start building and mocking parts of the ESLint API since these AST nodes are so unwieldy.

src/rules/javascript/no_document_write.js
```diff
@@ -0,0 +1,29 @@
...
+import { getNodeReferenceName } from 'utils';
+
+export default function(context) {
+  return {
+    CallExpression: function(node) {
+      var objectName;
+      if (typeof node.callee.object !== 'undefined' &&
+          node.callee.object.type === 'Identifier') {
+
+        objectName = getNodeReferenceName(context, node.callee.object);
```
muffinresearch: Can you comment bits like this - just to make it clearer what's going on? Just helps with immediacy later on.

src/rules/javascript/no_document_write.js
```diff
@@ -0,0 +1,29 @@
...
+        if (typeof objectName === 'undefined') {
+          return;
+        }
+      }
+
+
+      // check to see if the node is document.write()
+      if (typeof node.callee.property !== 'undefined' &&
```
muffinresearch: I think you can probably DRY up these ifs a bit. The first two checks are the same as the if starting line 7.

muffinresearch: Oh actually no they're not. I'm miss-reading.

muffinresearch:
There's a couple of minor nits and I think the two rule if statements might be consolidated. Otherwise r+wc.

muffinresearch:
@nolski Not sure I understand the change of direction here. This was ready to land albeit with a couple of minor fixes needed.

Nolski:
Oops. That's my mistake on deprecated. (not sure how vim spellcheck never caught that). If you'd like I can move this into a separate PR. We discussed in the beginning of this PR the viability of catching all instances of a deprecated entity in a single rule. At the time looking at the rules doc, it looked like they were significantly different that the method for doing so wouldn't be worth it. After looking at the python implementation of the rule, it turned out that the docs around the rules were just a bit misleading.

If you think that shouldn't be the case, I can back out the second commit and submit the original changes that were r+'d

tests/rules/javascript/test.deprecated_entities.js
```diff
@@ -0,0 +1,77 @@
...
+      var code = `foo.${entity.property}();`;
+      var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+      return jsScanner.scan()
+        .then((validationMessages) => {
+          assert.equal(validationMessages.length, 0);
+        });
+    });
+
+    it(singleLineString`should not warn about calling ${entity.property} to a
```
Nolski: Not quite sure why this is throwing jscs errors for indentation. Any idea @muffinresearch?

muffinresearch: Seems to be getting confused by the template string's 2nd line indent. Maybe you could try using vars in the loop (before the test-cases) to shorten the template subs vars? It would be nice to avoid needing `singleLineString` for a test description.

src/const.js
```diff
@@ -87,6 +93,16 @@ export const LOW_LEVEL_MODULES = [
   'system/events', 'sdk/system/events',
 ];
 
+export const DEPRECATED_ENTITIES = [
+  { object: 'document', property: 'write', error: NO_DOCUMENT_WRITE },
```
muffinresearch: Let's make the object format consistent with the middle object in the list.

src/const.js
```diff
@@ -87,6 +93,16 @@ export const LOW_LEVEL_MODULES = [
...
+  {
+    object: 'nsIDNSService',
```
muffinresearch: If you can use alpha sort for keys here, it's a bit nicer to scan.

muffinresearch: Should `object` be `global` or is it not always a global?

src/utils.js
```diff
@@ -58,6 +58,33 @@ export function getRootExpression(node) {
...
+  var variables = context.getScope().variables;
+  var scopeVar;
+
+  variables.forEach((variable) => {
+    if (variable.name === node.name) {
+      scopeVar = variable;
+    }
+  });
+
+  if (typeof scopeVar === 'undefined' ||
```
muffinresearch: I suspect falsey checks would work for prop checks here unless you know there's some potential ambiguity in terms of values.

muffinresearch: That said, this is missing some comments as to what this API represents.

src/utils.js
```diff
@@ -58,6 +58,33 @@ export function getRootExpression(node) {
...
+      typeof scopeVar.defs === 'undefined' ||
+      typeof scopeVar.defs[0] === 'undefined' ||
+      typeof scopeVar.defs[0].node.init === 'undefined' ||
```
muffinresearch: What about node? That wasn't checked in previous lines. Is that guaranteed to be the object you expect?

Nolski: Hard to say. I haven't seen an AST node with a defs field that didn't have a node as part of one of those objects but it's been difficult to nail down the schema of any of this. I can throw that check in there just to be sure. This is already looking pretty nasty though, is there a better way of doing this?

src/utils.js
```diff
@@ -58,6 +58,33 @@ export function getRootExpression(node) {
...
+ * returns the name of the reference node passed.
+ *
+ * example: var foo = document;
+ *  The node for foo will return 'document'
+ */
+export function getNodeReferenceName(context, node) {
+  var variables = context.getScope().variables;
+  var scopeVar;
+
+  variables.forEach((variable) => {
```
muffinresearch: What's this loop doing? Does this need to iterate over everything? Or could you short-circuit when you find something?

src/const.js
```diff
@@ -87,6 +93,22 @@ export const LOW_LEVEL_MODULES = [
   'system/events', 'sdk/system/events',
 ];
 
+export const DEPRECATED_ENTITIES = [
+  {
```
muffinresearch: This could come after `[`

src/utils.js
muffinresearch: I would invert the checks, maybe something like this:

```javascript
if (scopeVar && scopeVar.defs && scopeVar.defs[0] && 
    scopeVar.defs[0].node && scopeVar.defs[0].node.init) {
  return scopeVar.defs[0].node.init.name; 
} else {
  return node.name;
}
```

Of course this assumes non-falsey values.

src/utils.js
```diff
@@ -58,6 +58,33 @@ export function getRootExpression(node) {
...
+    if (variable.name === node.name) {
+      scopeVar = variable;
+    }
+  });
+
+  if (typeof scopeVar === 'undefined' ||
+      typeof scopeVar.defs === 'undefined' ||
+      typeof scopeVar.defs[0] === 'undefined' ||
```
muffinresearch: Do further list items in defs mean anything in this context? Definitely would help to have some comments so that looking at this code it's clearer without needing to look up eslint docs.

src/const.js
Nolski: This is causing issues with jscs if I have
```javascript
[{
...
}];
```
Due to conflicting rules. It for some reason jscs requires a comma be after the `{` but not before the `[`.

muffinresearch: Hmm. that's annoying. Can you provide a paste or commit it like that so we can see the error + affected rules?

src/utils.js
Nolski:
```javascript

Running "jscs:js" (jscs) task
Missing comma before closing bracket at src/const.js :
   106 |  object: 'nsISound',
   107 |  property: 'play',
   108 |}];
---------^
   109 |
   110 |export const INSTALL_RDF = 'install.rdf';
>> 1 code style errors found!
Warning: Task "jscs:js" failed. Use --force to continue.
```
was generated in according to
```javascript
export const DEPRECATED_ENTITIES = [{
  error: NO_DOCUMENT_WRITE,
  object: 'document',
  property: 'write',
},{
  error: NSI_DNS_SERVICE_RESOLVE,
  object: 'nsIDNSService',
  property: 'resolve',
},{
  error: NSI_SOUND_PLAY,
  object: 'nsISound',
  property: 'play',
}];
```

src/utils.js
Nolski: Not sure what you mean? Is the block comment at the top not enough?

src/utils.js
Nolski: Honestly I'm not sure, I _think_ that `defs` holds all definitions of the variable starting at the most locally scoped version but again, I don't know. I haven't been able to find docs talking about this.

src/const.js
Nolski:
```javascript

Running "jscs:js" (jscs) task
Missing comma before closing bracket at src/const.js :
   106 |  object: 'nsISound',
   107 |  property: 'play',
   108 |}];
---------^
   109 |
   110 |export const INSTALL_RDF = 'install.rdf';
>> 1 code style errors found!
Warning: Task "jscs:js" failed. Use --force to continue.
```
was generated in according to
```javascript
export const DEPRECATED_ENTITIES = [{
  error: NO_DOCUMENT_WRITE,
  object: 'document',
  property: 'write',
},{
  error: NSI_DNS_SERVICE_RESOLVE,
  object: 'nsIDNSService',
  property: 'resolve',
},{
  error: NSI_SOUND_PLAY,
  object: 'nsISound',
  property: 'play',
}];
```

Nolski:
@muffinresearch this should be good to take another look at.

Nolski:
@muffinresearch This should be alright to take another look at.

src/utils.js
```diff
@@ -58,6 +58,36 @@ export function getRootExpression(node) {
...
+      scopeVar = variable;
+      break;
+    }
+  }
+
+  if (scopeVar && scopeVar.defs && scopeVar.defs[0] &&
+      scopeVar.defs[0].node && scopeVar.defs[0].node.init &&
+      scopeVar.defs[0].node.init !== null) {
+    // Return the name of the first definition of the variable which
+    // corresponds to the node passed in.
```
muffinresearch: This is definitely clearer.

src/rules/javascript/deprecated_entities.js
```diff
@@ -0,0 +1,25 @@
...
+      var objectName;
+      // We're only looking for calls that look like `foo.bar()`
+      if (typeof node.callee.object !== 'undefined' &&
+          node.callee.property.type === 'Identifier' &&
+          node.callee.object.type === 'Identifier') {
+
+        objectName = getNodeReferenceName(context, node.callee.object);
+
+        for (let entity of DEPRECATED_ENTITIES) {
+          // check to see if the node is matches a deprecated entity
```
muffinresearch: Minor nit - all comments should be sentences starting with a capital letter and ending with a period.

Also there's a typo 'is matches'

src/utils.js
```diff
@@ -58,6 +58,36 @@ export function getRootExpression(node) {
...
+    return scopeVar.defs[0].node.init.name;
+  } else {
+    // If that variable doesn't exist in scope, then just return the node
+    return node.name;
```
muffinresearch: Should this be `return the node's name`?

src/utils.js
muffinresearch: How do we know the first definition is what we want?

Nolski: The first item in defs has always shown to be the first use of the variable (and therefore the definition). I suppose if we want to be safe I can check each item in `defs` and try and see which ones are variable assignments and return the last variable assignment used. (which would be the last in that list? docs aren't really clear but that's what it looks like to me).

muffinresearch: So what would happen if you re-defined the var? Obviously there's limits to what lengths we need to go to here - but, I feel it's important to understand what those limits are and document them so it's clear.

src/utils.js
```diff
@@ -58,6 +58,57 @@ export function getRootExpression(node) {
...
+
+  // Finds variable reference in current scope.
+  for (let variable of variables) {
+    if (variable.name === node.name) {
+      scopeVar = variable;
+      break;
+    }
+  }
+
+  if (scopeVar && scopeVar.defs && scopeVar.defs[0] &&
```
Nolski: This bit is a bit complex and required a bit of work due to the structure that eslint gives us AST nodes. I tried to comment it pretty well to make everything clear but let me know if you have any questions here.

src/const.js
```diff
@@ -1,3 +1,9 @@
+import {
```
tofumatt: I'd love to find a JSCS rule for this, but for now I'll comment here: our import style for lines longer than one line is:

```js
import { NO_DOCUMENT_WRITE, NSI_DNS_SERVICE_RESOLVE,
          NSI_SOUND_PLAY } from 'messages';
```

src/const.js
tofumatt: We should fix our JSCS then, because that's silly :smile:

src/const.js
```diff
@@ -101,6 +108,22 @@ export const LOW_LEVEL_MODULES = [
   'system/events', 'sdk/system/events',
 ];
 
+export const DEPRECATED_ENTITIES = [
+  {
+    error: NO_DOCUMENT_WRITE,
+    object: 'document',
+    property: 'write',
+  },{
```
tofumatt: We should have a JSCS rule to force a space after these kinds of commas.

src/messages/javascript.js
```diff
@@ -141,6 +141,40 @@ export const ONLY_PREFS_IN_DEFAULTS = {
   ],
 };
 
+export const NO_DOCUMENT_WRITE = {
+  code: 'NO_DOCUMENT_WRITE',
+  message: _('Use of document.write strongly discouraged.'),
```
tofumatt: I'm leaving a comment here because separate to this PR I'd like to propose some copy-edits here, but it can be a separate PR that doesn't need to block this. IGNORE ME. ^_^

src/rules/javascript/deprecated_entities.js
```diff
@@ -0,0 +1,25 @@
+import { DEPRECATED_ENTITIES } from 'const';
+import { getNodeReferenceName } from 'utils';
+
+export function deprecated_entities(context) {
+  return {
+    CallExpression: function(node) {
+      var objectName;
```
tofumatt: Why is this instantiated here when it's only used below?

src/utils.js
```diff
@@ -58,6 +58,57 @@ export function getRootExpression(node) {
 }
 
 /*
+ * Returns the name of the reference node passed.
+ *
+ * example: var foo = document;
+ *  The node for foo will return 'document'
```
tofumatt: We can get this via the `Identifier` event using ESLint, so why all the context lookup again? https://github.com/mozilla/addons-validator/blob/master/src/rules/javascript/mozindexeddb.js#L5

tests/rules/javascript/test.deprecated_entities.js
```diff
@@ -0,0 +1,77 @@
+import { VALIDATION_WARNING, DEPRECATED_ENTITIES } from 'const';
+import { singleLineString } from 'utils';
+import JavaScriptScanner from 'scanners/javascript';
+
+describe('no_document_write', () => {
+  for (let entity of DEPRECATED_ENTITIES) {
+    let obj = entity.object;
```
tofumatt: I dislike short variable names like this; they're especially hard to read for non-native English speakers, but also we don't need to use them; we have space for `object`, `property`, etc.

Though in this instance we shouldn't really need them at all, you're only using them twice and using as many lines to define the variable.

tests/rules/javascript/test.deprecated_entities.js
```diff
@@ -0,0 +1,77 @@
...
+    let prop = entity.property;
```
tofumatt: A newline after this to separate things would help readability.

src/const.js
tofumatt: The constants file shouldn't import anything else; I worry about circular dependencies (even if they work, which I forget if they do or not), it makes things really hard to follow.

Constants for us are largely named variables to avoid [magic constants](https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants).

Can this be refactored to avoid the `import`?

tofumatt:
I like all the tests in `tests/rules/javascript/test.deprecated_entities.js`.

Sorry for nitpicking this apart. This is making sense to me, but it does mean that multiple "rules" now need to be traced part out of the rules folder. I wonder if there is a place to put these deprecated entities that's more obvious. Even directly in the rule file might make sense. I like the idea of making it very obvious how the rule works.

I'm thinking out loud here, let me know if I make no sense.

I think I'm finished review for now but the code here seems solid. Let me know what you think of my comments. Sorry for making this PR take ages.

src/const.js
muffinresearch: Yeah - I looked into what we can change. Comma omission appears to only be allowed when the whole thing is on one line. So it doesn't seem like there's a great way to change it. We need a rule that says if you have `}];` you don't need `},];`

src/const.js
Nolski: I had this discussion with @muffinresearch. It's possible but we'd be moving messages out of messages which seemed like it would also seemed like a poor idea.

src/utils.js
Nolski: I think the docs are poorly worded. This get's the _most recent_ reference to the node's value. So for insance
```javascript
var foo = window;
foo = document;
```
will return `document` rather than `window`.

Nolski:
So I've hem and haw'd over this issue too. I personally think that `const.js` is one of the better choices to put the entities. The entities themselves are mostly static and it makes more sense to import them from the list of constants. I wouldn't be averse to moving it into the rule itself, I just don't see all that much benefit to it.

tofumatt:
If moving them into the rule means it's easier to grok the rule from a single file, they aren't imported from anywhere else (except maybe tests), and it avoids the need to do imports inside the `const.js` file (which I don't like, as I don't find that as constant), I think they're best moved.

src/rules/javascript/deprecated_entities.js
```diff
@@ -0,0 +1,42 @@
...
+    property: 'play',
+  },
+];
+
+
+export function deprecated_entities(context) {
+  return {
+    CallExpression: function(node) {
+      // We're only looking for calls that look like `foo.bar()`.
+      if (typeof node.callee.object !== 'undefined' &&
```
tofumatt: I assume `node.callee` is always defined? (I forget, AST spec is hard to remember ^_^)

src/utils.js
tofumatt: Oh, cool. Could you add that to the doc? The extra line adds a lot of context :smile:

tests/test.utils.js
```diff
@@ -74,6 +74,58 @@ describe('utils.getRootExpression()', function() {
   });
 });
 
+describe('utils.getNodeReferenceName()', () => {
```
tofumatt: Wouldn't it be easier to just run this code into something that creates an AST tree for us? That would be a handy utility; we care about testing code using AST, but we shouldn't really need to create the AST directly.

Can we use ESPrima to just create an AST quickly?

If this is complex enough for a separate PR I'm okay with that if you add a TODO here. But this seems cumbersome at best; brittle at worst.

tofumatt:
r+wc, if you want to tackle the AST comment in another PR that's probably sane; we've been at this awhile and eventually you'll think you're Sisyphus :sob:

src/rules/javascript/deprecated_entities.js
Nolski: If it's a call expression, I believe so.

PR #245, opened 2015-11-12
Fix type in web extensions + export manifest_version

### LINKED ISSUE #239: Check manifest_version equates to type
In the `manifest.json` `manifest_version` is currently only allowed to be 2. This looks like it maps to 2 as the type in `install.rdf`. We need to confirm this is the case.

muffinresearch:
Fix #239

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -1,0 +1,9 @@
+import { gettext as _ } from 'utils';
+
+export const MANIFEST_VERSION_INVALID = {
+  code: 'MANIFEST_VERSION_INVALID',
+  legacyCode: null,
+  message: _('"manifest_version" in the manifest.json is not a valid value'),
+  description: _('See http://mzl.la/20PenXl (MDN Docs) for more information.'),
+};
+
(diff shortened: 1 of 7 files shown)


src/messages/manifestjson.js
```diff
@@ -0,0 +1,9 @@
+import { gettext as _ } from 'utils';
+
+export const MANIFEST_VERSION_INVALID = {
+  code: 'MANIFEST_VERSION_INVALID',
+  legacyCode: null,
+  message: _('"manifest_version" in the manifest.json is not a valid value'),
+  description: _('See https://goo.gl/WyioRC (MDN Docs) for more information.'),
```
tofumatt: Can we use a Mozilla short URL here? :wink:

tofumatt:
r+

src/messages/manifestjson.js
muffinresearch: Yeah, do you have a link?

tofumatt: I don’t, and as discussed in IRC it didn’t seem easy to find. I say ship this, it’s cool.

PR #246, opened 2015-11-12
Remove some unnecessary Promise wrappers

### LINKED ISSUE #240: Look for any other unnecessary Promise wrappers
If code calls something that returns a Promise then generally wrapping that up in a `new Promise()` incantation should not be required.

This issue is to look for them and remove them as needed. Incidental use of resolve/reject can be replaced with `return Promise.resolve()` etc as needed.

muffinresearch:
Fixes #240

I've cleaned up some of the Promisey code to try and get rid of some unnecessary bits.

* I've consolidated on using throw instead of reject as long as we're inside a Promise constructor callback, a `then` or a `catch` (though this means then next `catch` up gets the error).
* We must always return something from `then()` but this can just be a value.
* It's unnecessary to wrap Promisey code in a promise just to use `resolve`/`reject`.

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -212,7 +212,7 @@ export default class Validator {
                    was found in the package metadata`);
           // TODO: Add notice for missing manifest.json?
           this.collector.addNotice(messages.TYPE_NO_INSTALL_RDF);
-          return Promise.resolve({});
+          return {};
         }
       })
       .then((addonMetaData) => {
@@ -221,10 +221,10 @@ export default class Validator {
           return this.detectTypeFromLayout()
             .then((addonType) => {
               addonMetaData.type = addonType;
-              return Promise.resolve(addonMetaData);
+              return addonMetaData;
             });
         } else {
-          return Promise.resolve(addonMetaData);
+          return addonMetaData;
         }
       });
   }
@@ -243,7 +243,7 @@ export default class Validator {
       .then((xpiMeta) => {
         for (let path_ of Object.keys(xpiMeta)) {
           if (path_.startsWith('dictionaries/')) {
-            return Promise.resolve(constants.PACKAGE_DICTIONARY);
+            return constants.PACKAGE_DICTIONARY;
           }
         }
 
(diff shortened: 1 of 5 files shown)


muffinresearch:
Oh also more red than green :)

tofumatt:
red-celebrate

SO MUCH RED!

src/validator.js
```diff
@@ -221,10 +221,10 @@ export default class Validator {
           return this.detectTypeFromLayout()
             .then((addonType) => {
               addonMetaData.type = addonType;
-              return Promise.resolve(addonMetaData);
+              return addonMetaData;
```
tofumatt: Hmm...

The `Promise.resolve()` isn't needed but it clears up what is eventually returned... and what happens if we chain something else after this promise--do we need to add `Promise.resolve()` back?

src/validator.js
```diff
@@ -256,49 +256,41 @@ export default class Validator {
...
-            reject(err);
-          } else {
-            reject(invalidMessage);
-          }
-        });
-    });
+    return _lstat(filepath)
+      .then((stats) => {
+        if (stats.isFile() === true) {
+          return;
```
tofumatt: Again, now I just feel like it's harder to quickly grasp this returns a Promise. Given nearly everything we do does it's not too bad, but I feel like something is lost here. More code feels more clear in this case.

src/validator.js
```diff
@@ -256,49 +256,41 @@ export default class Validator {
...
-      return Promise.all(promises)
-        .then(() => {
-          resolve();
-        }).catch(reject);
-    });
+    var promises = [];
+    for (let filename of files) {
+      promises.push(this.scanFile(filename));
+    }
+    return Promise.all(promises);
```
tofumatt: This is _definitely_ better though.

src/validator.js
```diff
@@ -409,7 +398,7 @@ export default class Validator {
         } else {
           this.handleError(err);
         }
-        return Promise.reject(err);
+        throw err;
```
tofumatt: I will say that I _do_ like throwing the errors rather than rejecting the Promises. Nobody uses rejections directly anyway ^_^

src/xpi.js
```diff
@@ -134,28 +134,27 @@ export default class Xpi {
...
 
-      for (let ext of extensions) {
-        if (ext.indexOf('.') !== 0) {
-          throw new Error("File extension must start with '.'");
-        }
+    for (let ext of extensions) {
+      if (ext.indexOf('.') !== 0) {
+        // Use Promise.reject as outside a then a throw
+        // wouldn't be caught.
+        return Promise.reject(new Error("File extension must start with '.'"));
```
tofumatt: Why not throw an error here?

tofumatt:
Didn't even realise you could do some of this stuff! Let me know what you think of my comments, some of it feels harder to read but I'd like your thoughts.

src/xpi.js
muffinresearch: As per the comment. We're not in the Promise constructor callback or a then so this is not part of the Promise chain at this point. I can improve the comment.

tofumatt: Whoops, I totally missed that. Maybe it’s time for dinner ;-)

src/validator.js
muffinresearch: I don't know if this clears it up: http://jsbin.com/delatenike/edit?html,js,console 

In a nutshell there's three things you can do in a `then`:

* return a promise. 
* return a value
* throw an error

This is a brilliant reference: http://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html - I've been re-reading it as I'd sniffed out the Promise-wrapping stuff but then realized that a lot of the times we use `Promise.resolve()` unnecessarily as returning a value is equivalent.

That said if you want a promise when there isn't one the `return Promise.resolve()` *is* the right thing to do. We do alot of that in the tests for example where we're stubbing promisey code.

I like the author's rule which is to always `return` or `throw` in a `then()` and everyhing will be fine.

src/xpi.js
tofumatt: Reading the comment again, yeah it doesn't quite stand out.

```js
// We use Promise.reject as we're outside a `then` block.
// Using `throw` directly wouldn't be caught.
```
Might work better? I try to use markdown in comments to help highlight references to code.

tofumatt:
Alrighty, I'm r+wc (just see that comment here: https://github.com/mozilla/addons-validator/pull/246#discussion_r44731464)

Very nice, I've been won over after another look :+1:

PR #248, opened 2015-11-12
Detect jetpack add-on (close #192)

### LINKED ISSUE #192: Detect JPM addons
Needed to provide more metadata.

tofumatt:
Closes #230 as well; this is essentially a second pass at that.


tests/test.validator.js
```diff
@@ -125,7 +125,7 @@ describe('Validator', function() {
   });
 
   it('should throw when message.type is undefined', () => {
-    var addonValidator = new Validator({_: ['tests/example.xpi']});
+    var addonValidator = new Validator({_: ['tests/fixtures/example.xpi']});
```
muffinresearch: Nice, I like this path change :+1:

src/validator.js
```diff
@@ -36,6 +37,7 @@ export default class Validator {
     this.chalk = new chalk.constructor(
       {enabled: !this.config.boring});
     this.collector = new Collector();
+    this.addonMetaData = {};
```
muffinresearch: Good idea.

src/validator.js
```diff
@@ -190,8 +192,13 @@ export default class Validator {
   }
 
   getAddonMetaData() {
+    var _xpiMeta;
+
     return this.xpi.getMetaData()
       .then((xpiMeta) => {
+        // Used later to get add-on architecture.
+        _xpiMeta = xpiMeta;
```
muffinresearch: This will need a rename - but I'm sure you know that already :)

src/validator.js
```diff
@@ -420,4 +429,19 @@ export default class Validator {
       return this.scan(_Xpi);
     }
   }
+
+  _getAddonArchitecture(xpiMetaData) {
+    // If we find a file named bootstrap.js this is assumed to be a
+    // Jetpack add-on: https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/jetpack.py#L154
```
muffinresearch: Does eslint/jscs not whine about the length of this line?

src/validator.js
muffinresearch: I wonder if this should be null, then we could more easily do a check and return it if it's already defined. Can be a todo or new issue.

muffinresearch:
Going to need to renames following #251 but otherwise r+wc

src/validator.js
tofumatt: Nope. It's smart enough to realise it's a URL and not complain.

- tofumatt (Sent from mobile)

muffinresearch: Oh shiny. Cool

src/validator.js
tofumatt: I can integrate that into this PR no problem. Have to do a rebase anyway after #251 lands :smiley:

tofumatt: Actually, it involves extra tests and such so I'll make a new issue for it, just to keep this lean. #257

Issue #243, opened 2015-11-12
Add support for validation on a directory.


muffinresearch:
This would allow reviewers to not need a zip to carry out validation.

PR #251, opened 2015-11-13
Rename XPI metadata to files (close #234)

### LINKED ISSUE #234: Rename XPI "metadata" property to "files"
The metadata property is a bit confusing as we also deal with add-on metadata (#195). We should rename this to `files`.

tofumatt:
I think this clears things up a bit; I really consider this to just be "files" and don't think of it as metadata. And now that we're dealing with Add-on metadata, let's keep the terminology distant :smile:

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -190,18 +190,18 @@ export default class Validator {
   }
 
   getAddonMetaData() {
-    return this.xpi.getMetaData()
-      .then((xpiMeta) => {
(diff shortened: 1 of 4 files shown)


muffinresearch:
This is great R+ :sparkles:

muffinresearch:
Also will make it easier when we come to doing #243

tofumatt:
Thanks. I'll rebase this onto master then merge it in.

tofumatt:
I just realised: red and green exactly the same:
red-and-green-the-same

PR #262, opened 2015-11-13
Refactor metadata extraction and expose name

### LINKED ISSUE #255: Extract name from manifests

### LINKED ISSUE #258: Pass the install.rdf parse an xmlDoc directly
Removes the need to pass around the xmlDoc all the time.

muffinresearch:
Fixes #258 
Fixes #255

* Bit more refactoring to make manifest parsers simpler and easier to test each field in isolation.
* Expose name + validation of it in each case.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -267,8 +267,9 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
 | ----- | ------- | --------- | ---------- | ----------- | --------- | ---------- | -------- | -------- |
 | :white_check_mark: | notice | missing_install_rdf | | Add-on missing install.rdf for type detection | | [22](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L22) | ('typedetection', 'detect_type', 'missing_install_rdf') | TYPE_NO_INSTALL_RDF |
-| :white_check_mark: | error | invalid_em_type | | The only valid values for `<em:type>` are 2, 4, 8, and 32 | | [46](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L46) | ('typedetection', 'detect_type', 'invalid_em_type') | TYPE_INVALID |
-| :white_check_mark: | notice | no_em:type | | No `<em:type>` element found in install.rdf | | [66](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L66) | ('typedetection', 'detect_type', 'no_em:type') | TYPE_MISSING |
+| :white_check_mark: | error | invalid_em_type | | The only valid values for `<em:type>` are 2, 4, 8, and 32 | | [46](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L46) | ('typedetection', 'detect_type', 'invalid_em_type') | RDF_TYPE_INVALID |
+| :white_check_mark: | notice | no_em:type | | No `<em:type>` element found in install.rdf | | [66](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L66) | ('typedetection', 'detect_type', 'no_em:type') | RDF_TYPE_MISSING |
+| :white_check_mark: | notice | no_em:name | | No `<em:name>` element found in install.rdf | | [66](https://github.com/mozilla/amo-validator/blob/master/validator/typedetection.py#L66) | new rule | RDF_NAME_MISSING |
 | :white_check_mark: | error | undeterminable_type | | Unable to determine add-on type | | [195](https://github.com/mozilla/amo-validator/blob/master/validator/submain.py#L195) | ('main', 'test_package', 'undeterminable_type') | TYPE_NOT_DETERMINED |
 
 
@@ -339,4 +340,5 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
 | ----- | ------- | --------- | ---------- | ----------- | --------- | ---------- | -------- |
-| :white_check_mark: | warning | N/A |  | manifest_version in manifest.json is not valid. | | null | MANIFEST_VERSION_INVALID |
+| :white_check_mark: | error | N/A |  | manifest_version in manifest.json is not valid. | | null | MANIFEST_VERSION_INVALID |
+| :white_check_mark: | error | N/A |  | name in manifest.json is not valid. | | null | MANIFEST_NAME_INVALID |
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -7,3 +7,10 @@ export const MANIFEST_VERSION_INVALID = {
   description: _('See http://mzl.la/20PenXl (MDN Docs) for more information.'),
 };
 
+export const MANIFEST_NAME_INVALID = {
+  code: 'MANIFEST_NAME_INVALID',
+  legacyCode: null,
+  message: _('"name" in the manifest.json is not a valid value'),
+  description: _('See http://mzl.la/1STmr48 (MDN Docs) for more information.'),
+};
+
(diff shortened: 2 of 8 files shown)


src/messages/manifestjson.js
```diff
@@ -7,3 +7,10 @@ export const MANIFEST_VERSION_INVALID = {
   description: _('See http://mzl.la/20PenXl (MDN Docs) for more information.'),
 };
 
+export const MANIFEST_NAME_INVALID = {
+  code: 'MANIFEST_NAME_INVALID',
+  legacyCode: null,
+  message: _('"name" in the manifest.json is not a valid value'),
+  description: _('See https://goo.gl/lSAJmB (MDN Docs) for more information.'),
```
tofumatt: I thought the bit.ly to Mozilla/mdn links worked 😉

src/parsers/installrdf.js
```diff
@@ -1,66 +1,77 @@
...
   }
 
-  _getAddonType(xmlDoc) {
-    var addonType = null;
-    var typeNodes = xmlDoc.getElementsByTagName('em:type');
+  _getNode(tag) {
+    var typeNodes = this.xmlDoc.getElementsByTagName(tag);
     if (typeNodes.length > 1) {
-      throw new Error('Multiple <em:type> elements found');
+      throw new Error(`Multiple <${tag}> elements found`);
```
tofumatt: Are there no scenarios where multiple tags of the same type are allowed? I think there are instances where the same tag appears multiple times and it's okay.

docs/rules.md
```diff
@@ -340,3 +341,4 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | Done? | MsgType | Rule name | Addon type | Description | File Type | Source ref | Old Code | New Code |
 | ----- | ------- | --------- | ---------- | ----------- | --------- | ---------- | -------- |
 | :white_check_mark: | warning | N/A |  | manifest_version in manifest.json is not valid. | | null | MANIFEST_VERSION_INVALID |
+| :white_check_mark: | warning | N/A |  | name in manifest.json is not valid. | | null | MANIFEST_NAME_INVALID |
```
tofumatt: Why is this a warning and not an error? Seems like something is either valid or not, so it should be an error if it isn't valid.

src/parsers/installrdf.js
```diff
@@ -1,66 +1,77 @@
...
   }
 
-  parseDoc() {
-    var rdfScanner = new RDFScanner(this.rdfString, INSTALL_RDF);
-    return rdfScanner.getContents();
-  }
-
   getMetaData() {
-    return this.parseDoc()
-      .then((xmlDoc) => {
```
tofumatt: I like this simplification, but it introduces a required order of operations in the code now. `parseDoc` should cache the document and return it if it's already parsed (for performance), but always requesting the document means we never have to worry about when we call code. We do it elsewhere and it's handy for testing and other things.

src/parsers/installrdf.js
```diff
@@ -1,66 +1,77 @@
...
+      this.collector.addNotice(RDF_TYPE_MISSING);
     }
     return addonType;
   }
 
-  _getGUID(xmlDoc) {
-    if (xmlDoc.getElementsByTagNameNS(this.namespace, 'id').length > 0) {
-      var idNode = xmlDoc.getElementsByTagNameNS(this.namespace, 'id').item(0);
+  _getGUID() {
+    // Note: this has shared validiation in the `Validator.getAddonMetaData`
```
tofumatt: This caught me off-guard at first. I think the comment needs a bit more clarity:

```js
// NOTE: We validate this rule in one place: `Validator.getAddonMetaData`.
// This is because both `install.rdf` and `manifest.json` share the same requirements
// for guid.
```

Wordier, but clearer?

tests/parsers/test.installrdf.js
```diff
@@ -22,45 +28,125 @@ describe('InstallRdfParser.getMetaData()', function() {
...
+        assert.equal(type, constants.PACKAGE_EXTENSION);
       });
   });
 
   it('should collect a notice if type is missing', () => {
     var addonValidator = new Validator({_: ['bar']});
-    var installRdfParser = new InstallRdfParser(validRDF(''),
-                                                addonValidator.collector);
-    return installRdfParser.getMetaData()
+    var rdfScanner = new RDFScanner(validRDF(''), INSTALL_RDF);
```
tofumatt: `validRDF()` should be fine, you don't need the empty string argument. (Though it's not really a `validRDF` anymore, is it :wink: )

src/parsers/manifestjson.js
```diff
@@ -1,28 +1,38 @@
 import { PACKAGE_EXTENSION, VALID_MANIFEST_VERSION } from 'const';
 import log from 'logger';
-import { MANIFEST_VERSION_INVALID } from 'messages';
+import { MANIFEST_VERSION_INVALID, MANIFEST_NAME_INVALID } from 'messages';
```
tofumatt: This is a nitpick, and possibly a thing we should clean up later, but can these be alphabetical? It makes stuff _way_ easier to scan for me that way.

tests/parsers/test.installrdf.js
```diff
@@ -22,45 +28,125 @@ describe('InstallRdfParser.getMetaData()', function() {
...
+      .then((xmlDoc) => {
+        var installRdfParser = new InstallRdfParser(xmlDoc);
+        return installRdfParser._getGUID();
+      })
+      .then((guid) => {
+        assert.equal(guid, 'myid');
+      });
+  });
+
+  it('should return null for guid if not defined', () => {
```
tofumatt: A note on why we do this would be handy; I could imagine stumbling in new and it taking a bit to track the "why" of this down.

tofumatt:
Read over everything and it largely looks good, but see my comments, especially around using `this.xmlDoc` directly.

Let me know when you want me to have another look.

src/parsers/installrdf.js
muffinresearch: It felt wrong to me to need to keep passing the xmlDoc around. This meant this wasn't necessary. The `installrdf` metadata class' *only* job is to look things up from the XMLdoc. Returning a promise in a constructor is a no-no, otherwise that might have been a better solution. This meant that the doc is always available on the isntance and there's not a dependency on calling a specific method before calling another one as that happens externally.

Alot of our other code is doing explicit get something and pass it on down the chain. E.g. the files (metadata) from the xpi class.

tests/parsers/test.installrdf.js
muffinresearch: Not sure I follow.

src/parsers/installrdf.js
tofumatt: Fair enough, I’m convinced :-)

tests/parsers/test.installrdf.js
tofumatt: Oh wait, sorry, I confused this with something else entirely. Never mind!

src/parsers/installrdf.js
muffinresearch: In which case we'd have a different helper like `getNodes()`

tofumatt: Ah, good call :)

docs/rules.md
muffinresearch: Yeah, I think you're right - this is a required field after all.

tofumatt:
I am now r+wc as my only concerns with discussing have been addressed :smile:

src/messages/manifestjson.js
muffinresearch: bitly was down :/

muffinresearch: Ok it's working again today - I shall update. Not a great sign for a link shortener though.

src/parsers/installrdf.js
muffinresearch: Yep, perfect I'll use that.

src/parsers/manifestjson.js
muffinresearch: :+1:

docs/rules.md
muffinresearch: Oh - it's just the docs that are wrong.

PR #263, opened 2015-11-13
stub console

### LINKED ISSUE #254: Use print() inside extractMetaData
We're using `console.log` in https://github.com/mozilla/addons-validator/blob/master/src/validator.js#L347, meaning the test output includes a stray "undefined", which I thought was my fault from testing in a PR :wink: 

We should make this use the Validator's `print` method so we can easily silence it in tests.

screenshot 2015-11-13 18 32 51

muffinresearch:
Fixes #254 - took a slightly different route as `Validator.print()` isn't a great fit for the metadata output use-case.

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -435,9 +435,9 @@ export default class Validator {
       });
   }
 
-  run(_Xpi=Xpi) {
+  run(_Xpi=Xpi, _console=console) {
     if (this.config.metadata === true) {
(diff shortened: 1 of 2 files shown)


tofumatt:
I figured this might be the case. I'd have done the same when I noticed it but didn't want to introduce it in the same PR.

Cool, r+

PR #293, opened 2015-11-17
Metadata is one word

### LINKED ISSUE #261: Settle on metadata or metaData
Re: camelCase or not of the word metadata.

I prefer the first (it's one word), but we should just be consistent in the code, which we aren't right now.

muffinresearch:
Fixes #261


tofumatt:
BUT WHERE IS THE TEST?! :wink: 

Seriously though thanks. r+

red-and-green-the-same

PR #307, opened 2015-11-17
Fix return value of getGUID

### LINKED ISSUE #306: Return value of _getGUID in install.rdf parser is a node.
Sneakily this gets co-erced to a String so even assert.equal didn't catch it.

muffinresearch:
Fixed #306


tofumatt:
Nice catch! r+

PR #309, opened 2015-11-17
Ensure metadata scanner is run for all metadata extraction

### LINKED ISSUE #308: metadata scanning doesn't happen when --metadata is used.
Suggest we roll the metadata scanning into the getAddonMetadata func.

muffinresearch:
Fixes #308

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -369,15 +369,15 @@ export default class Validator {
           if (typeof message.type === 'undefined') {
             throw new Error('message.type must be defined');
           }
-
           this.collector._addMessage(message.type, message);
         }
 
-        return;
+        return metadata;
       });
   }
 
-  extractMetadata(_Xpi=Xpi, _console=console) {
+  extractMetadata(_Xpi=Xpi, _console=console,
+                  _MetadataScanner=MetadataScanner) {
     return checkMinNodeVersion()
       .then(() => {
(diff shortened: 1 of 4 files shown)


src/validator.js
```diff
@@ -463,7 +460,7 @@ export default class Validator {
 
   _getAddonArchitecture(xpiMetadata) {
     // If we find a file named bootstrap.js this is assumed to be a
-    // Jetpack add-on: https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/jetpack.py#L154
+    // Jetpack add-on: https://github.com/mozilla/amo-validator/blob/7a8011a/validator/testcases/jetpack.py#L154
```
tofumatt: Heh, I always worry eventually something will repeat the short hash, but I assume Git then picks the first hash in the history that uses it?

I should get over this fear! ^_^

tofumatt:
Good call, r+

src/validator.js
muffinresearch: Yeah - this article sums it up quite nicely [1]  - I think in our small repos we should be ok with 7chars. But even if that broke it wouldn't be the end of the world.


[1] http://blog.cuviper.com/2013/11/10/how-short-can-git-abbreviate/

Issue #190, opened 2015-11-03
Pass metadata to rule functions


muffinresearch:
When #152 lands let's work out how to pass the metadata like add-on type. I think there's going to be quite a lot of it so maybe passing the validation instance might start to make more sense.

tofumatt:
My simple solution was to attach it to the XPI in https://github.com/mozilla/addons-validator/pull/230; I think of the XPI object as the "add-on", but maybe that's a bit off.

The XPI already has a `metadata` properly so I put it in there. Determining the add-on type and metadata requires scanners and other things too, but putting that info in the XPI makes sense to me. It seems the simplest solution so I gravitated toward it, but I haven't actually had to use the data yet, so maybe there are obvious pitfalls I don't see.

tofumatt:
Right now we're assigning all the metadata to an object at `validatorInstance.metadata`, which I think makes sense. Passing in just that object, rather than the whole `validatorInstance` helps us pass around too much global scope which I think is good.

muffinresearch:
It's `Validator.addonMetaData` https://github.com/mozilla/addons-validator/blob/master/src/validator.js#L356 - we're taking about addon metadata as opposed to files here.

tofumatt:
Yup, that’s what I’m thinking :-)

PR #312, opened 2015-11-17
Add options to scanners (fix #311)

### LINKED ISSUE #311: Normalise "options" usage in scanners
This will help with #190 and keeps our method of supplying scanner classes with arbitrary, extra info (eg things other than contents or filename) consistent.

tofumatt:
Started this while doing #190, realised we needed a consistent way to pass in extra data to scanners and this seemed the best way; we can pass all options right onto the rules with this.

### DIFF
--- a/src/scanners/css.js
+++ b/src/scanners/css.js
@@ -31,24 +31,22 @@ export default class CSSScanner extends BaseScanner {
     }
 
     var file = cssCode.position.source;
-    var startLine = cssCode.position.start.line;
-    var startColumn = cssCode.position.start.column;
+    var cssOptions = Object.assign({}, this.options, {
+      startLine: cssCode.position.start.line,
+      startColumn: cssCode.position.start.column,
+    });
 
     log.debug('Passing CSS code to rule function "%s"',
       cssInstruction, {
         cssCode: cssCode,
         file: file,
-        startLine: startLine,
-        startColumn: startColumn,
+        startLine: cssOptions.startLine,
+        startColumn: cssOptions.startColumn,
       });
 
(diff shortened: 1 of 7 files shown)


src/scanners/css.js
```diff
@@ -31,25 +31,23 @@ export default class CSSScanner extends BaseScanner {
     }
 
     var file = cssCode.position.source;
-    var startLine = cssCode.position.start.line;
-    var startColumn = cssCode.position.start.column;
+    var options = Object.assign(this.options, {
```
muffinresearch: This is mutating an instance level options object each time - I'm not too sure about that.

muffinresearch: It might be better to call this something else and the use `Object.assign({}, this.options, { blah })`

tofumatt: Oh, you're right, my bad. That was a total oversight. Thanks! :+1:

src/scanners/css.js
```diff
@@ -31,25 +31,23 @@ export default class CSSScanner extends BaseScanner {
     }
 
     var file = cssCode.position.source;
-    var startLine = cssCode.position.start.line;
-    var startColumn = cssCode.position.start.column;
+    var cssInfo = Object.assign({}, this.options, {
```
muffinresearch: Maybe `opts` or `cssOptions` would be better as this is called options in the actual rules?

muffinresearch:
r+wc

src/scanners/css.js
tofumatt: Sure thing; I’ll fix and merge.

PR #319, opened 2015-11-18
Import and define rules directly (fix #317)

### LINKED ISSUE #317: Import ESLint rules directly instead of precompiling with webpack
This is a pretty major yak to shave, and I've done it. Filing for posterity, but we can include our tests in code coverage as well as simplify our build process this way.

We do lose out on having all our ESLint rules compiled in a `dist/` folder on npm. But that's fine.

tofumatt:
We can define the ESLint rules directly using the `ESLint.linter.defineRule` API and using `ESLint.linter.verify()` method to scan each file.

This allows our code coverage tools to scan our rules (it caught one skipped test!) and simplifies our build process a lot.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,12 +31,10 @@ module.exports = function(grunt) {
   });
 
   grunt.registerTask('start', [
-    'webpack:eslintwatch',
     'webpack:buildwatch',
   ]);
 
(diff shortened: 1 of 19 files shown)


tofumatt:
Bahahaha that GIF. :smile:

PR #321, opened 2015-11-19
Update rules.md

andymckay:
Missing table column caused a broken table.


tofumatt:
Oops, good catch.

r+

PR #324, opened 2015-11-19
Pass metadata to rule functions (fix #190)

### LINKED ISSUE #190: Pass metadata to rule functions

tofumatt:
Allows rules to check on the metadata of an add-on, for contextual validation.

### DIFF
--- a/src/rules/chromemanifest/categories.js
+++ b/src/rules/chromemanifest/categories.js
@@ -14,7 +14,7 @@ export const DANGEROUS_CATEGORIES = [
 ];
 
 
-export function checkCategories(triples) {
+export function checkCategories(triples, filename) {
   var validationMessages = [];
 
   for (let triple of triples) {
@@ -24,7 +24,7 @@ export function checkCategories(triples) {
           triple.object.startsWith('global ') ||
           triple.object.startsWith('DOM ')))) {
       var message = Object.assign({}, DANGEROUS_CATEGORY, {
-        file: triple.filename,
+        file: filename,
         line: triple.line,
         type: VALIDATION_WARNING,
       });
--- a/src/scanners/css.js
+++ b/src/scanners/css.js
@@ -12,7 +12,7 @@ export default class CSSScanner extends BaseScanner {
 
   _defaultRules = rules;
 
-  processCode(cssCode, cssInstruction) {
+  processCode(cssCode, cssInstruction, _rules=this._defaultRules) {
     if (cssCode.type === 'comment') {
       log.debug('Found CSS comment. Skipping');
       return;
@@ -22,7 +22,7 @@ export default class CSSScanner extends BaseScanner {
       log.debug('Processing media rules');
       if (cssCode.rules.length) {
         for (let mediaCssCode of cssCode.rules) {
-          this.processCode(mediaCssCode, cssInstruction);
+          this.processCode(mediaCssCode, cssInstruction, _rules);
         }
       } else {
         log.debug('No media rules found');
@@ -45,7 +45,7 @@ export default class CSSScanner extends BaseScanner {
       });
 
     this.validatorMessages = this.validatorMessages.concat(
-      rules[cssInstruction](cssCode, file, cssOptions));
+      _rules[cssInstruction](cssCode, file, cssOptions));
   }
 
   scan(_rules=this._defaultRules) {
@@ -58,7 +58,7 @@ export default class CSSScanner extends BaseScanner {
             for (let cssInstruction in rules) {
               this._rulesProcessed++;
               for (let cssCode of ast.stylesheet.rules) {
-                this.processCode(cssCode, cssInstruction);
+                this.processCode(cssCode, cssInstruction, rules);
               }
             }
           }
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -9,6 +9,8 @@ import { ensureFilenameExists, ignorePrivateFunctions,
 
 export default class JavaScriptScanner {
 
+  _defaultRules = rules;
+
   constructor(code, filename, options={}) {
     this.code = code;
     this.filename = filename;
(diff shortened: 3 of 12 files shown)


src/rules/chromemanifest/categories.js
```diff
@@ -14,7 +14,7 @@ export const DANGEROUS_CATEGORIES = [
 ];
 
 
-export function checkCategories(triples) {
+export function checkCategories(triples, filename) {
```
tofumatt: Changing this, even though filename is passed in the `triples` object, so we have a consistent rule function signature for every rule type other than ESLint/JS.

tofumatt:
Alright, have all this working, but I'd love thoughts on if it seems a bit too... sloppy? I dunno, maybe it was just me but I feel like it was a bit of a "cram a bunch of extra stuff" into the scanner to get the tests working.

src/scanners/javascript.js
```diff
@@ -19,22 +21,26 @@ export default class JavaScriptScanner {
     ensureFilenameExists(this.filename);
   }
 
-  scan(_ESLint=ESLint) {
+  scan(_ESLint=ESLint, _rules=this._defaultRules,
```
muffinresearch: I think once we get to this it's probably time to consider using an object. Makes it super-obvious from the calling side too.

tofumatt: Agreed. I’ll move to that 👍

src/rules/chromemanifest/categories.js
muffinresearch: Makes sense - I :heart: consistency.

tests/scanners/test.metadata.js
```diff
@@ -1,11 +1,24 @@
...
 
+  it('should pass addon metadata to rules', () => {
+    var fakeRules = { fakeRule1: fakeMetadataRuleCheck };
+
+    var metadata = validMetadata();
+    var scanner = new MetadataScanner(metadata, 'fake.xpi', fakeMetadata);
+
+    return scanner.scan(fakeRules)
+      .then((validatorMessages) => {
+        assert.equal(validatorMessages.length, 0);
```
muffinresearch: What is this testing wrt passing of the metadata?

muffinresearch: Ah I see it's all in the helper.

muffinresearch: I wonder if it might be more direct to have this be a sinon stub and check for what was passed in the assertions?

muffinresearch:
Looks good on the whole - I think using a rule stub might make the tests easier to grok at a glance.

tofumatt:
Sounds good, I'll make those tweaks after I'm back from the gym and ask for another quick look. 

- tofumatt

tofumatt:
@muffinresearch This has been tweaked; the CSS Scanner needed a way to load custom rules so I added that for the tests. (I tested it separately from other classes that extend `BaseScanner` because it has its own `scan()` method).

Let me know what you think.

src/scanners/css.js
```diff
@@ -12,7 +12,7 @@ export default class CSSScanner extends BaseScanner {
 
   _defaultRules = rules;
 
-  processCode(cssCode, cssInstruction) {
+  processCode(cssCode, cssInstruction, rules) {
```
muffinresearch: Is this for injection? Might be better to do `_rules=rules` if that's the case to make the distinction.

tofumatt: It is; good call. (I thought that myself but I think I was being lazy; no excuse!)

muffinresearch: Actually should this use _defaultRules like this: https://github.com/mozilla/addons-validator/blob/pass-metadata-to-rules-190/src/scanners/css.js#L51

tofumatt: :+1:

tests/scanners/test.css.js
```diff
@@ -25,6 +25,26 @@ describe('CSSScanner', () => {
...
+    // This rule calls assert.fail() if no metadata is passed to it.
+    sinon.stub(fakeRules, 'metadataPassCheck', metadataPassCheck);
+
+    var scanner = new CSSScanner(code, 'fake.css', {
+      addonMetadata: validMetadata({guid: 'snowflake'}),
+    });
+
+    return scanner.scan(fakeRules)
+      .then((validatorMessages) => {
+        assert.ok(fakeRules.metadataPassCheck.called);
```
muffinresearch: Nice - this is easier to grok.

tests/scanners/test.javascript.js
```diff
@@ -227,6 +228,45 @@ describe('JavaScript Scanner', function() {
...
 
+  it('should pass addon metadata to rules', () => {
+    var fakeRules = { metadata_not_passed: () => {} };
+
+    sinon.stub(fakeRules, 'metadata_not_passed', (context) => {
+      return {
+        Identifier: (node) => {
+          var metadata = context.settings.addonMetadata;
+
+          if (!metadata || typeof metadata.guid === 'undefined') {
```
muffinresearch: Is guid guaranteed - it's required but would this mean we'd not be able to carry on with validation even if we wanted to.

Maybe check metadata is an object?

muffinresearch:
r+wc

tests/scanners/test.javascript.js
tofumatt: GUID is guaranteed here as it's only running this rule in this test, where we pass in a metadata object with a GUID.

I guess I could either check for an object (less explicit) or be even _more_ explicit and test for the exact GUID value. I'll do the latter if that's cool.

muffinresearch: Duh, I don't know why but I didn't spot this was a test  - ignore me!

dont-panic-and-carry-on1

tofumatt: Lol. Well I cleaned it up a bit anyway ^_^

PR #333, opened 2015-11-23
Process a directory too

### LINKED ISSUE #243: Add support for validation on a directory.

muffinresearch:
Fixes #243 

TL;DR - this patch means you can process a directory in the same way you can pointing at an Xpi. Tthe path is introspected and if it's a directory the files backend is used and it's not we use the Xpi class instead.

* Base class for IO operations (provides common elements between Xpi and dir processing)
* Move to object for dep injection since we'd got to a point where there was starting to get too many args
* Fixed up some tests that had issues + added error event handlers in stream reading functions.

Testing on a xpi/dir with errors shows the same results. However the layout fallback used to determine the extension type from the file ext, obviously won't work.


src/io/base.js
```diff
@@ -0,0 +1,47 @@
...
+    for (let ext of extensions) {
+      if (ext.indexOf('.') !== 0) {
+        // We use Promise.reject as we're not inside a `then()` or a
+        // Promise constructor callback.
+        // If we throw here it won't be caught.
+        return Promise.reject(new Error("File extension must start with '.'"));
+      }
+    }
+
+    return this.getFiles()
```
tofumatt: In `BaseScanner` we define `_getContents()` (which other scanners must implement) as a simple function that throws a `NotImplentedError`.

Doing so here would be consistent and I just like outlining what the class should look like inside it without needing to look at a separate file.

src/io/directory.js
```diff
@@ -0,0 +1,83 @@
...
+
+import log from 'logger';
+
+
+export class Directory extends IOBase {
+
+  constructor(dirPath) {
+    super();
+    this.dirPath = dirPath;
+    this.files = {};
```
tofumatt: `this.files` and `this.entries` are initialised the same way in each constructor, so they should be moved into the base class.

And if you considered `path` to be the first argument (it's either a path to a file or a path to a directory) you could define it as such in the `IOBase` class and `Directory` wouldn't even need to change the constructor.

src/io/directory.js
```diff
@@ -0,0 +1,83 @@
...
+
+  getFiles(_walkPromise=walkPromise) {
+    // If we have already processed this directory and have data
+    // on this instance return that.
+    if (Object.keys(this.files).length) {
+      log.info(singleLineString`Files already exist for directory
+               "${this.dirPath}" returning cached data`);
+      return Promise.resolve(this.files);
+    }
+    return _walkPromise(this.dirPath)
```
tofumatt: A newline here might help things breathe better, but that's just me. I love whitespace :wink:

src/io/directory.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  getFileAsStream(relativeFilePath) {
+
+    if (!this.files.hasOwnProperty(relativeFilePath)) {
+      return Promise.reject(
+        new Error(`Path "${relativeFilePath}" does not exist in this dir.`));
+    }
+
+    var absDirPath = path.resolve(this.dirPath);
```
tofumatt: Nitpick, but we use `relativeFilePath` above, can we use `absoluteDirPath` here for clarity?

src/io/directory.js
```diff
@@ -0,0 +1,83 @@
...
+            .on('readable', () => {
+              var chunk;
+              while (null !== (chunk = readStream.read())) {
+                content += chunk.toString();
+              }
+            })
+            .on('end', () => {
+              resolve(content);
+            })
+            .on('error', (err) => {
```
tofumatt:
```js
.on('error', reject);
```
should suffice here.

src/io/xpi.js
```diff
@@ -128,35 +118,11 @@ export default class Xpi {
...
-        for (let filename in xpiFiles) {
-          for (let ext of extensions) {
-            if (filename.endsWith(ext)) {
-              files.push(filename);
-            }
-          }
-        }
-
-        return files;
+          fileStream.on('error', (err) => {
```
tofumatt: Nitpick, but `fileStream.on('error', reject);` would be sufficient here.

src/validator.js
```diff
@@ -376,14 +373,21 @@ export default class Validator {
       });
   }
 
-  extractMetadata(_Xpi=Xpi, _console=console,
-                  _MetadataScanner=MetadataScanner) {
+  extractMetadata({ _Xpi=Xpi, _console=console,
+                    _MetadataScanner=MetadataScanner,
+                    _Directory=Directory } = {}) { // jscs:ignore
```
tofumatt: What are we ignoring here, no trailing comma? (Can we ignore just the one rule, just curious.)

tofumatt:
Pretty cool! :+1: 

I've flagged some nitpicks and such but overall this is very nice. I'm r+wc, though ping me if you want to talk about the constructor refactor I mentioned or if you want me to have a quick glance at that before you merge.

Very nice.

tofumatt:
Oh! Might want to add some docs (even a few lines to the README) about how to use this to scan a directory too.

Or just file an issue for that, either's fine.

muffinresearch:
> Oh! Might want to add some docs (even a few lines to the README) about how to use this to scan a directory too.

Yeah I should update the CLI - that should work.

src/validator.js
muffinresearch: Yes we can but then we go over 80 chars :/

src/io/base.js
muffinresearch: Yep - good call, that's done.

muffinresearch:
I updated the arg + added something to the readme.

PR #341, opened 2015-11-24
Created init rule

### LINKED ISSUE #325: Rule: init
Disallow `nsITransferrable.init(null)`

https://github.com/mozilla/addons-validator/blob/master/docs/rules.md#entity-values

Nolski:
Resolves #325

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -9,6 +9,7 @@ export const ESLINT_RULE_MAPPING = {
   deprecated_entities: ESLINT_WARNING,
   eval_string_arg: ESLINT_ERROR,
   global_require_arg: ESLINT_WARNING,
+  init_null_arg: ESLINT_WARNING,
   low_level_module: ESLINT_WARNING,
   mozindexeddb: ESLINT_ERROR,
   mozindexeddb_property: ESLINT_WARNING,
--- a/src/rules/javascript/deprecated_entities.js
+++ b/src/rules/javascript/deprecated_entities.js
@@ -1,6 +1,6 @@
 import { NO_DOCUMENT_WRITE, NSI_DNS_SERVICE_RESOLVE,
          NSI_SOUND_PLAY, TAB_URL_OVERRIDE } from 'messages';
-import { getNodeReferenceName } from 'utils';
+import { getNodeReference } from 'utils';
 
 export const DEPRECATED_ENTITIES = [
   {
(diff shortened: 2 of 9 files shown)


src/rules/javascript/init.js
```diff
@@ -0,0 +1,24 @@
+import { INIT } from 'messages/javascript';
+import { getNodeReference } from 'utils';
+
+export function init(context) {
+  return {
+    CallExpression: function(node) {
+      // Check if what's being called is setTimeout or setInterval
+      if (typeof node.callee.object !== 'undefined' &&
```
tofumatt: This is pretty hard to parse. What exactly does it pass/fail for?

src/rules/javascript/init.js
```diff
@@ -0,0 +1,24 @@
+import { INIT } from 'messages/javascript';
+import { getNodeReference } from 'utils';
+
+export function init(context) {
+  return {
+    CallExpression: function(node) {
+      // Check if what's being called is nsiTransferable.init()
```
Nolski: Sorry, some of these comments were out of date.

src/const.js
```diff
@@ -9,6 +9,7 @@ export const ESLINT_RULE_MAPPING = {
   deprecated_entities: ESLINT_WARNING,
   eval_string_arg: ESLINT_ERROR,
   global_require_arg: ESLINT_WARNING,
+  init: ESLINT_WARNING,
```
muffinresearch: Again this is too vague - please add some more context to the warning. It would be nice to try and infer a little bit more info from just the error code. `INIT_NULL_FIRST_ARG` or something like that.

src/rules/javascript/deprecated_entities.js
```diff
@@ -27,11 +27,11 @@ export function deprecated_entities(context) {
           node.callee.property.type === 'Identifier' &&
           node.callee.object.type === 'Identifier') {
 
-        let objectName = getNodeReferenceName(context, node.callee.object);
+        let object = getNodeReference(context, node.callee.object);
```
muffinresearch: Something like this it's probably nice to add something to the var name so we know what object we're talking about later on in the code.

tests/rules/javascript/test.init.js
```diff
@@ -0,0 +1,40 @@
+import JavaScriptScanner from 'scanners/javascript';
+import { VALIDATION_WARNING } from 'const';
+import * as messages from 'messages';
+
+describe('init', () => {
+  it('should not allow init called with null', () => {
```
muffinresearch: Does this rule handle aliasing e.g:

```javascript
var foo = nsITransferable.init;
foo(null);
```

and

```javascript
var foo = nsITransferable;
foo.init(null);
```

tests/rules/javascript/test.init_null_arg.js
```diff
@@ -0,0 +1,61 @@
...
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 0);
+      });
+  });
+
+  it('should not allow init called with an aliased varaible and null', () => {
+    var code = 'var foo = nsITransferable; foo.init(null);';
```
muffinresearch: Nice, can you add one more test to show it catches this?
```
var foo = nsITransferable.init;
foo(null);
```

muffinresearch:
r+wc

PR #342, opened 2015-11-24
Make RDF node lookup a bit more robust

### LINKED ISSUE #318: Have more direct way to look-up rdf metadata
Currently some things are looked up and we just return the first item. This doesn't check that we're definitely a child of the `<Description>` node that's a direct descendant of `<RDF>`.

muffinresearch:
Fixes #318

Allows for checking topLevel tags only, as opposed to using getElementsByTagName which would get results all the way down into the tree.


tofumatt:
Very nice :+1: 

r+

PR #343, opened 2015-11-25
Extract version and remove metadata scanner

### LINKED ISSUE #294: Extract version from metadata files

muffinresearch:
Fixes #294

* Remove metadata Scanner
* Extract version
* Do all error collection in the metadata parsers.
* Add check for filename being passed in the message object when it should be file.

TODO as a follow-up will be to expose errors in for `--metadata` JSON as part of the data structure.

I've also added further issues to cover more metadata validation for version and type checking for when pulling data from JSON.


src/message.js
```diff
@@ -23,6 +23,11 @@ export default class Message {
 
   constructor(type, data={}) {
     this.type = type;
+
+    if (data.hasOwnProperty('filename')) {
+      throw new Error('The key for the file is "file" not "filename"');
```
tofumatt: Nice :smile:

tofumatt:
red-is-winning
The red lines are killing it!

r+

PR #350, opened 2015-11-26
Add metadata errors to --metadata output

### LINKED ISSUE #344: Expose errors as a list in the --metadata option JSON
We should make the data structure something like this when invalid.

```
{ 
   valid: false,
   metadata: {
      // whatever
   },
   errors: [ 
      // List of error objects here.
   ]
}
```

And this for a valid set of metadata:

```
{ 
   valid: true,
   metadata: {
      // whatever
   },
}
```

muffinresearch:
Fixes #344


muffinresearch:
In the end I decided on `hasErrors` instead of `valid`.

tofumatt:
r+

PR #359, opened 2015-11-29
Flag empty files inside a package (fix #334)

### LINKED ISSUE #334: Mark empty files so olympia source viewer can ignore them
The old amo-validator ignores files that are on our checksum list. They get greyed out in the olympia source viewer and a note about the recognized library shows in the validation report.

For empty files, we also want the addons-validator to recognize such files so the source viewer can grey them as well, but the HTML validation report on olympia should *not* list the empty file, because that's irrelevant.

tofumatt:
Internally flags all empty files inside addonMetadata without
throwing an error/raising any notices to the user, as requested
in #334.


src/validator.js
```diff
@@ -471,4 +474,34 @@ export default class Validator {
...
+        addonMetadata.emptyFiles = emptyFiles;
+        return addonMetadata;
+      });
+  }
+
+  _isFileEmpty(filename, emptyFiles) {
+    return this.io.getFile(filename)
+      .then((contents) => {
+        // If this file is empty, we want to flag it as such.
+        if (contents.length === 0) {
```
muffinresearch: Can we determine from size metadata in both the XPI + file system cases? I also wonder if we also need a check to see if a file is just whitespace too?

tofumatt: That’s possible, yeah. This was my first-pass-simplest-thing try, but I suppose checking the unpacked size would work. I’ll try that.

muffinresearch: Yep yauzl iirc exposes unpacked size in the metadata object and the Directory class has just size which is from a stat object.

tofumatt: Cool; that is probably a bit faster too as we should already have the data. 👍

tofumatt:
Using `uncompressedSize` now, should be a bit leaner/better!

src/validator.js
```diff
@@ -471,4 +474,21 @@ export default class Validator {
...
     }
   }
+
+  _markEmptyFiles(addonMetadata) {
+    var emptyFiles = [];
+
+    return this.io.getFiles()
+      .then((files) => {
+        for (let filename in files) {
+          if (files[filename].uncompressedSize === 0) {
```
muffinresearch: This API is different for the directory class [1].

Maybe we should add uncompressedSize as an alias? Either that or you need to check size too.

[1] https://github.com/mozilla/addons-validator/blob/master/src/io/utils.js#L21

tofumatt: I’d go for adding uncompressedSize as an alias; also means I should add an extra test for directories then. Thanks!

tofumatt: Actually, scrap that, seems cleaner to check for uncompressedSize then just size. I’ll have a patch in a few.

tofumatt:
Pushed again; should be all set now.

muffinresearch:
Lgtm. r+

PR #365, opened 2015-11-30
Switch to postcss for CSS parsing

### LINKED ISSUE #364: Replace CSS parser
The CSS parser we are using can't handle certain syntax. In addition it's primarily regex based which means it can be prone to breaking. 

We need to switch to something that does more robust parsing.

muffinresearch:
Fixes #364


tofumatt:
r+

Very nice.

PR #366, opened 2015-11-30
Created event_listener_fourth rule

### LINKED ISSUE #353: Rule: addEventListener_fourth
https://github.com/mozilla/addons-validator/blob/master/docs/rules.md#instanceactions

Nolski:
Resolves #353

### DIFF
--- a/tests/rules/javascript/test.event_listener_fourth.js
+++ b/tests/rules/javascript/test.event_listener_fourth.js
@@ -1,0 +1,128 @@
+import JavaScriptScanner from 'scanners/javascript';
+import { VALIDATION_WARNING } from 'const';
+import { singleLineString } from 'utils';
+import * as messages from 'messages';
+
+describe('event_listener_fourth', () => {
+  it('should not allow a true literal', () => {
+    var code = 'window.addEventListener("click", function(){}, false, true);';
+    var jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    return jsScanner.scan()
+      .then((validationMessages) => {
+        assert.equal(validationMessages.length, 1);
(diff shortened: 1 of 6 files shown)


tests/rules/javascript/test.event_listener_fourth.js
```diff
@@ -0,0 +1,128 @@
+import JavaScriptScanner from 'scanners/javascript';
+import { VALIDATION_WARNING } from 'const';
+import { singleLineString } from 'utils';
+import * as messages from 'messages';
+
+describe('event_listener_fourth', () => {
+  it('should not allow a true literal', () => {
+    var code = 'window.addEventListener("click", function(){}, false, true);';
```
muffinresearch: Nice comprehensive tests.

I wonder if we could add something to someone using `window.addEventListener.call` and `window.addEventListener.apply`. Presumably the old code doesn't go that far?

muffinresearch:
r+wc

tests/rules/javascript/test.event_listener_fourth.js
Nolski: I agree, in fact, I'm beginning to wonder if there may be a better way to wrap ESLint here. For instance when we are looking for call expressions for something like `foo()` we normally are looking for `window.foo()` or `var a = foo; a();` or other sort of things. And I'm seeing that the rules are often having to contain this logic individually and some of it is starting to get repeated.

muffinresearch: Yep the nicest thing I think we can do here is keep the many rules but encapsulate more functionality into the utils helpers. That way we can keep things dry but also retain as much of the eslint way of doing things as possible.

PR #367, opened 2015-11-30
Covers edge case for deprecated entities

Nolski:
This commit now covers the edge case for aliasing a variable to inner
properties such as `var foo = document.write; foo()`.


tofumatt:
wfm, r+

PR #375, opened 2015-12-02
Fixed issue with inconsistent eslint type

Nolski:
So turns out that `scopeVar.defs[0].parent.parent.body` can sometimes return an array and sometimes return an object. Since for-of cannot iterate over javascript objects it would throw a type error when that happened.

This should fix that.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -87,16 +87,18 @@ export function getNodeReference(context, node) {
     let occurances = scopeVar.defs[0].parent.parent.body;
     let lastAssignment;
 
-    for (let occurance of occurances) {
-      if (occurance.type === 'VariableDeclaration' &&
-          occurance.declarations[0].init !== null) {
-        // Get what the name of what it was assigned to or the raw
-        // value depending on the initalization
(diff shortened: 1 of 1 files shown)


tofumatt:
Looks good to me, but what is the cause of it returning an object, and do we want to use that data too?

Nolski:
@tofumatt It looks like one of those things where someone decided to return an array when there are !multiple items and just an object when there is one item (which is infuriating). If there is just one item, then we know there isn't any other reference and should just return the node anyways (which we do on line 114).

muffinresearch:
Sounds like it would be reasonable to file a bug upstream if you know where the code is that causes that.

tofumatt:
Fair enough, makes sense.

Yeah, if you want to file a bug upstream that’d be cool, though who knows if it’s implicitly part of the spec at this point :-/

Cool, I’m r+ on this then.

muffinresearch:
Would be nice to add some explicit tests for this.

r+wc

muffinresearch:
I think we should just get this in an worry about updates later.

PR #378, opened 2015-12-02
Add limit of 100mb for a single file

### LINKED ISSUE #376: Implement size limit for individual files opened by zip or from a directory.
Relates to #14

This is for the reason that we want to protect against unpacking a file that is too large.

Creating a large zip can be done as follows - this creates a zip containing a 2gb install.rdf using a named pipe.

```
mkfifo install.rdf ;  dd if=/dev/zero bs=1024 count=2097152 > install.rdf & zip output.zip -FI install.rdf ; rm install.rdf
```

We also need to so the same limit for the `io.Directory` class too so that users will see the same limits locally.

muffinresearch:
Fixes  #376

You can test this by creating a zip with a file larger than 100mb and it should catch it.

e.g to create a too large install.rdf (with no useful content) you can use:

```
dd if=/dev/zero bs=1024 count=104448 > ../install.rdf
```

### DIFF
--- a/src/io/base.js
+++ b/src/io/base.js
@@ -1,5 +1,6 @@
 import { singleLineString } from '../utils';
 import { NotImplentedError } from 'exceptions';
+import { MAX_FILE_SIZE_MB } from 'const';
 
 /*
  * Base class for io operations for both an Xpi or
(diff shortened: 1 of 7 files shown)


src/io/base.js
```diff
@@ -11,6 +11,10 @@ export class IOBase {
     this.path = packageOrDirPath;
     this.files = {};
     this.entries = [];
+    // If this is too large the node process will hit a RangeError
+    // when it runs out of memory.
+    let maxMegabytes = 100;
```
tofumatt: Seems like this should be a default we set in constants.

Might also make sense as a CLI flag but that can be a later thing :smile:

tofumatt:
r+wc

Issue #384, opened 2015-12-04
Build our own list of JS libs


tofumatt:
We should have a node way of [building the JS lib hashes](https://github.com/mozilla/amo-validator/blob/master/extras/build_whitelist.py).

@muffinresearch and I talked about this and the whitelist might make sense as a separate repo we could:

* update independently
* share with others

tofumatt:
In progress at https://github.com/mozilla/dispensary

tofumatt:
Complete.

tofumatt:
We had a version on NPM at https://github.com/mozilla/dispensary/commit/c98fb2b0bfe3ad1ba1ba631156a013484ae0c9a6

Issue #66, opened 2015-10-09
Deal with known JS libs 


muffinresearch:
* [x] Check against known library hashes #383
* [x] Build our own hash list #384
* [ ] Hook up Dispensary inside the linter

We need a way to skip over validation of known safe libs.

muffinresearch:
This is potentially quite big so we could make this a stretch goal.

PR #385, opened 2015-12-04
Compare JS files against known JS libs (fix #383)

### LINKED ISSUE #383: Compare JS files against hash whitelist
First step to #66 is to compare all JS files against a collection of known JS libs. We're using hashes the same way the old validator did it, as it seems the sane way to go about it.

tofumatt:
Part of #66, this checks all JS files against a hash of known JS
libraries.

We're borrowing the current list from the old [amo-validator](https://github.com/mozilla/amo-validator/blob/master/validator/testcases/hashes.txt)
but in the future (#384) we'll have our own list.

Named filehasher and avoiding the name whitelist because of @andymckay's
[thoughtful blog post](http://mckay.pub/2015-01-07-allowed/).


muffinresearch:
> By a miss, do you mean it not flagging when there are no matches? 

Yep exactly - just showing that a non-matching file isn't in the list.

muffinresearch:
r+wc

PR #388, opened 2015-12-04
Parser now retrieves application metadata

### LINKED ISSUE #296: Extract "applications" from metadata
Related to #195, cross-linking because synergy! :sparkles:

Nolski:
Resolves #296

### DIFF
--- a/src/parsers/installrdf.js
+++ b/src/parsers/installrdf.js
@@ -3,6 +3,9 @@ import * as messages from 'messages';
 import log from 'logger';
 import { singleLineString } from 'utils';
 
+const TARGET_APPLICATION_FIELDS = [
+  'id', 'minVersion', 'maxVersion',
+];
 
 export default class InstallRdfParser {
 
@@ -21,6 +24,7 @@ export default class InstallRdfParser {
       type: this._getAddonType(),
       version: this._getVersion(),
       restartless: this._getIsBootstrapped(),
+      application: this._getApplication(),
     });
   }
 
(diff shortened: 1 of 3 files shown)


src/parsers/installrdf.js
```diff
@@ -105,6 +109,40 @@ export default class InstallRdfParser {
...
   }
 
+  _getApplication() {
+    var targetApplications = this._getTopLevelNodesByTag(
+      'em:targetApplication');
+    var applications = [];
+
+    for (let targetApplication of targetApplications) {
+      let application = {};
+      let descriptionNodes = Array.prototype.slice.call(
```
muffinresearch: Anywhere you're doing this you can use `this._makeArray` I see there's a couple of other places that could be used in this file too.

src/parsers/installrdf.js
```diff
@@ -3,6 +3,9 @@ import * as messages from 'messages';
 import log from 'logger';
 import { singleLineString } from 'utils';
 
+const TARGET_APPLICATION_FIELDS = [
```
muffinresearch: Maybe move to `const` and add something to the name to reflect this is to do with RDF?

src/parsers/installrdf.js
```diff
@@ -105,6 +109,40 @@ export default class InstallRdfParser {
...
+          .filter((node) => node.nodeName === 'Description');
+
+      if (descriptionNodes.length !== 1) {
+        throw new Error(singleLineString`<em:targetApplication> node should only
+          have a single descendant <Description>`);
+      }
+
+      for (var i in descriptionNodes[0].childNodes) {
+        let childNode = descriptionNodes[0].childNodes[i];
+        if (TARGET_APPLICATION_FIELDS.indexOf(childNode.localName) !== -1) {
```
muffinresearch: Should be able to use `Array.prototype.includes` [1] here.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/parsers/installrdf.js
```diff
@@ -105,6 +109,40 @@ export default class InstallRdfParser {
...
+          application[childNode.localName] = this._getNodeValue(childNode);
+        }
+      }
+
+      if (application.id && application.maxVersion && application.minVersion) {
+        applications.push(application);
+      } else {
+        throw new Error(singleLineString`targetApplication must contain an id,
```
muffinresearch: Shouldn't this be an error message pushed onto the collector?

Also I think doing something a little more granular will help developers e.g. flag each field that's invalid separately so that it's clearer when just one thing is broken.

Nolski:
Since it looks like https://github.com/mozilla/addons-linter/issues/195 is going to be closed, should I worry about this PR as well or should we just suspend work on this as well?

muffinresearch:
Yeah I guess we can put this on hold for now. I'll close it for now and we can always re-open.

PR #399, opened 2015-12-15
Rename bin/addons-validator -> bin/addons-linter

muffinresearch:


tofumatt:
Oops, good catch, thanks.

PR #407, opened 2015-12-15
Fix path for hashes in package

muffinresearch:
This adds back the global appRoot hack so we can determine paths in two different ways one for test and one for when we're calling via the binary.

### DIFF
--- a/src/filehasher.js
+++ b/src/filehasher.js
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import path from 'path';
 import readline from 'readline';
 
 import createHash from 'sha.js';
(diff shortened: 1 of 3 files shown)


src/filehasher.js
```diff
@@ -6,10 +7,14 @@ import createHash from 'sha.js';
 import log from 'logger';
 import { lstatPromise } from 'io/utils';
 
+// TODO: This will be removed when hashes are coming
+// from the dispensary module..
```
tofumatt: Ever so minor, but that should be one period for a sentence, or three for an ellipsis...

tofumatt:
r+wc

Are the coveralls tests running now?

muffinresearch:
>  Are the coveralls tests running now? 

Hmm they don't appear to be. I'll check what's up.

muffinresearch:
Ah yes they are just a little delayed.

PR #410, opened 2015-12-16
Remove old validator rules (fix #406)

### LINKED ISSUE #406: Flag all old add-on rules and mark as "won't fix"
Check through `rules.md` and make sure we flag old rules we've decided not to do.

tofumatt:
This is a first pass at this but I think closes out everything it should and nothing it shouldn't. Sooo...

r?


muffinresearch:
Good start - I think it might make sense to try and re-org once we have a rough plan of all the new things we need to deal with and then we can move out the deprecated stuff to a separate doc.

muffinresearch:
R+

tofumatt:
Agreed; this eventually doesn't need to be our reference doc for web extensions. It's handy to keep around just as a reference, as we never had anything like it before.

PR #425, opened 2016-01-04
Add exit codes on exit

### LINKED ISSUE #355: Provide exit codes on exit
When we exit we should provide a valid exit code. 

E.g. 0 if there were no problems and something no-zero otherwise.

muffinresearch:
Fixes #355


src/linter.js
```diff
@@ -438,7 +438,10 @@ export default class Linter {
       })
       .then(() => {
         this.print();
-        return;
+        /* istanbul ignore if  */
```
muffinresearch: This is skipped in the code coverage because the test runs against un-instrumented code.

tofumatt: Makes sense. Can you add this in as an inline comment?

tests/test.cli-process.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Tests running the process via the CLI.
```
tofumatt: Yay! This is awesome and I was thinking we should be doing this :+1:

tofumatt:
r+wc, :+1:

PR #426, opened 2016-01-04
Remove unused scanner

muffinresearch:


tofumatt:
Do any tests need to be removed for this? If not r+

muffinresearch:
> Do any tests need to be removed for this? If not r+ 

No, they already got removed and it's not part of the instrumented/compiled tests hence the lack of coverage changes.

PR #429, opened 2016-01-05
Remove type detection fallback

### LINKED ISSUE #411: Rip out the dictionary detection fall-back "feature"
Dictionary detection falls-back to detecting based on directories. This feels like something we should be stricter on. 

It looks like people are not identifying some dictionaries by install.rdf and as a result FF thinks they should be signed as it uses the install.rdf to know what the type is.

I don't know the history of the fallback but it feels like being explicit here is pretty important so we know exactly what thing we're dealing with. Especially in a new world where the old validator needs to handle legacy extensions and the new one certain other types.

muffinresearch:
Fixes #411


tofumatt:
r+

PR #430, opened 2016-01-05
Speed up grunt tasks

muffinresearch:
* Saves instumenting files on every test run.
* Runs jshint and eslint only when src files have changed.

Hopefully at somepoint we can do something to make webpack not do a full rebuild each time too when the src is unchanged. Though currently that requires keeping a caching object around between runs if using webpacks own cache (as used in webpack's own watch).


tofumatt:
Right on, r+

PR #480, opened 2016-02-04
Make var look-ups more defensive

### LINKED ISSUE #469: “Cannot read property 'type' of null” for the eval_string_arg rule when declaring variable without value
Minimal steps to reproduce:

1. Create a directory. In this example I’ll call it “addon”.
2. Add a valid install.rdf inside it.
3. Add a bootstrap.js file with the following code:

   ```js
   var a
   b.setTimeout(a)
   ```

4. Run `addons-linter addon/`.

Result: The message “Cannot read property 'type' of null” is printed in red.
Expected: At least it shouldn’t crash :)

I ran into this while working on a program in written in CoffeeScript. I had code looking something like this:

```coffee
fn = ->
window.setTimeout(fn, 100)
```

which compiles to:

```js
var fn;

fn = function() {};

window.setTimeout(fn);
```

The problem seems to have to do with declaring a variable without giving it a value at the same time, because the following does not result in a crash:

```js
var fn = function() {};

window.setTimeout(fn);
```

addons-linter version: v0.0.11

muffinresearch:
Fixes #469

### DIFF
--- a/src/rules/javascript/eval_string_arg.js
+++ b/src/rules/javascript/eval_string_arg.js
@@ -14,9 +14,9 @@ export function eval_string_arg(context) {
         if (node.arguments.length > 0 &&
             node.arguments[0].type === 'Identifier') {
           var originalDef = getVariable(context, node.arguments[0].name);
-          if (typeof originalDef !== 'undefined' &&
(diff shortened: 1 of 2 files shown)


tofumatt:
r+

PR #483, opened 2016-02-05
Only run coverage for 0.12

### LINKED ISSUE #482: Fix node 4.0 slowness
https://github.com/mozilla/addons-linter/pull/481 shows us that we need an over `10s` timeout limit to run our tests in node 4.0 or things like our `XPI.open()` test will fail.

We need to find out why things are suddenly so slow when they weren't before.

muffinresearch:
Fixes #482

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ notifications:
     on_success: change
     on_failure: always
 after_script:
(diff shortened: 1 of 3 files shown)


tofumatt:
r+, nice!

PR #494, opened 2016-02-09
Validate manifest.json via JSON schema.

### LINKED ISSUE #400: Switch to using JSON schema for manifest.json
The rules for the manifest.json are quite structured. Whilst it's just JSON we should move to a schema-based approach and split the schema into its own repo to benefit other tools that can use it.

Using a versioned schema should scale much better as we add increasingly more complex rules as to what settings are allowed in a particular context.

muffinresearch:
Fixes #400

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -1,6 +1,6 @@
-import { PACKAGE_EXTENSION, VALID_MANIFEST_VERSION } from 'const';
+import { PACKAGE_EXTENSION } from 'const';
 import log from 'logger';
-import * as messages from 'messages';
+import validate from 'mozilla-web-extension-manifest-schema';
 
 
 export default class ManifestJSONParser {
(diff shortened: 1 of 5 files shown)


src/parsers/manifestjson.js
```diff
@@ -12,54 +12,43 @@ export default class ManifestJSONParser {
...
-      log.debug('Invalid version "%s"', version);
-      this.collector.addError(messages.PROP_VERSION_INVALID);
-      version = null;
-    }
-    return version;
+  getMetadata() {
+    var isValid = this.checkSchema();
+    return {
+      isValid: isValid,
+      metadata: {
```
kumar303: this key name seems redundant. Is it necessary to make a sub-grouping for this?

kumar303:
Pretty sweet! I'll definitely be using the manifest-schema package. I didn't go through all the tests here but are there some duplicates that could be moved to the manifest-schema's test suite? Just seemed like a lot of tests for what is really just an adapter.

r+wc

muffinresearch:
Yeah, this will cover us for changes to the lib underneath and since they were already there it seemed to make sense to just update them. The majority of testing of the schema will indeed live in the schema project as it's updated.

src/parsers/manifestjson.js
muffinresearch: Yeah I know what you mean. The problem is I don't want `isValid` mixed with the data.

muffinresearch: Perhaps an improvement would be to separate out the metadata from the validation rather than trying to have getMetaData do two things.

PR #497, opened 2016-02-10
Separate validation from metadata retrieval

### LINKED ISSUE #496: Validate Schema separately to getMetaData

muffinresearch:
Fixes #496

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -223,9 +223,9 @@ export default class Linter {
           log.info('Retrieving metadata from manifest.json');
           return this.io.getFileAsString(MANIFEST_JSON)
             .then((json) => {
-              var result = new ManifestJSONParser(
-                json, this.collector).getMetadata();
-              return result.metadata;
+              var manifestParser = new ManifestJSONParser(json, this.collector);
+              manifestParser.validate();
+              return manifestParser.getMetadata();
             });
         } else {
           log.warn(singleLineString`No ${INSTALL_RDF} or ${MANIFEST_JSON}
--- a/src/linter.js
+++ b/src/linter.js
@@ -224,7 +224,6 @@ export default class Linter {
           return this.io.getFileAsString(MANIFEST_JSON)
             .then((json) => {
               var manifestParser = new ManifestJSONParser(json, this.collector);
-              manifestParser.validate();
               return manifestParser.getMetadata();
             });
         } else {
--- a/tests/parsers/test.manifestjson.js
+++ b/tests/parsers/test.manifestjson.js
@@ -6,6 +6,16 @@ import { PACKAGE_EXTENSION, VALID_MANIFEST_VERSION } from 'const';
 import { validManifestJSON } from '../helpers';
 
 
+describe('ManifestJSONParser', () => {
+  it("should throw if validate() isn't called before getMetadata()", () => {
+    assert.throws(() => {
+      var json = validManifestJSON({manifest_version: 'whatever'});
+      var manifestJSONParser = new ManifestJSONParser(json);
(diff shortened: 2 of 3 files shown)


kumar303:
I would be ok with this but a method that depends on another is always a sign of clunkiness to me. The question to ask is: would you ever need to get metadata without validating the manifest first? I think the answer is no. If that's true, you could change the class to this:
```` 
export default class ManifestJSONParser {

  constructor(jsonString, collector) {
    this.parsedJSON = JSON.parse(jsonString);
    this.collector = collector;
    this.isValid = this._validate();
  }
  ...
}
````
The only downside is it conceals the fact that some potentially costly validation is taking place inside the constructor. I think the upside of making the `getMetaData()` function more easy to work with is worth the trade-off though.

r+wc

**Comment by muffinresearch** [comment]
It's all swings and roundabouts!

I was working on the basis that the constructor shouldn't do too much, but I see what you're saying and I think you're right -  it's probably a touch cleaner your way on balance.

**Comment by muffinresearch** [comment]
@kumar303 ok this is ready for another look.

File: src/linter.js [hunk 1]
```diff
@@ -223,9 +223,8 @@ export default class Linter {
           log.info('Retrieving metadata from manifest.json');
           return this.io.getFileAsString(MANIFEST_JSON)
             .then((json) => {
-              var result = new ManifestJSONParser(
-                json, this.collector).getMetadata();
-              return result.metadata;
+              var manifestParser = new ManifestJSONParser(json, this.collector);
```
kumar303: could be `let manifestParser...`

tests/parsers/test.manifestjson.js
```diff
@@ -13,8 +13,8 @@ describe('ManifestJSONParser manifestVersion', function() {
     var json = validManifestJSON({manifest_version: 'whatever'});
     var manifestJSONParser = new ManifestJSONParser(json,
                                                     addonLinter.collector);
-    var result = manifestJSONParser.getMetadata();
-    assert.equal(result.isValid, false);
+    assert.equal(manifestJSONParser.isValid, false);
+    manifestJSONParser.getMetadata();
```
kumar303: does this have some side effect? Why does it need to be called here?

tests/parsers/test.manifestjson.js
```diff
@@ -26,8 +26,9 @@ describe('ManifestJSONParser manifestVersion', function() {
     var json = validManifestJSON({manifest_version: '1'});
     var manifestJSONParser = new ManifestJSONParser(json,
                                                     addonLinter.collector);
-    var result = manifestJSONParser.getMetadata();
-    assert.equal(result.isValid, false);
+    assert.equal(manifestJSONParser.isValid, false);
+
+    manifestJSONParser.getMetadata();
```
kumar303: I can't spot any side effects that would happen from calling this so it looks unnecessary.

kumar303:
r+wc

tests/parsers/test.manifestjson.js
muffinresearch: Oh duh - that's a leftover - well spotted.

src/linter.js
muffinresearch: I'm tending to use `let` when needed. Inside a function like this there's no difference between `var` and `let`.

muffinresearch: ...and even then I forget to sometimes - old habits die hard.

kumar303: Whatever works. I've been trying to always use let unless var is needed. My rationale is that a code refactor might introduce a new block (a line down from your closure, for example) that could leak the variable.

PR #564, opened 2016-03-15
Add metadata to scan output JSON

### LINKED ISSUE #557: Include metadata in scans as well as when called with the --metadata option.
Not sure how to call this but having the addon metadata alongside the validation result is helpful for showing more details and extracting `architecture`.

Maybe report `metadata` by default (and keep have `--metadata` to only see the metadata) or add a new `--full` option that merges both.

muffinresearch:
Fixes #557


tofumatt:
r+

PR #569, opened 2016-03-16
Catch end of central directory record not found

### LINKED ISSUE #556: Detect and fail properly for invalid files
Example error output:

```bash
$ addons-linter ./src/olympia/devhub/tests/addons/unopenable.xpi
end of central directory record signature not found
```

The file isn't an actual zip: https://github.com/mozilla/addons-server/blob/master/src/olympia/devhub/tests/addons/unopenable.xpi

Expected behavior:

```json
"errors": [{
    "_type": "error",
    "code": "BAD_ZIPFILE",
    "message" : "Corrupt ZIP file",
    "description" : "We were unable to decompress the zip file.",
    "file": null
}]
```

the `code`, `message` and `description` can be anythhing, that's just the error message that `amo-validator` has.

muffinresearch:
Fixes #556


tofumatt:
r+

PR #607, opened 2016-04-20
Strip BOM from files

### LINKED ISSUE #606: Handle BOM encoding
See https://github.com/mozilla/addons-server/issues/2038 for an example that currently fails and maybe https://github.com/mozilla/amo-validator/blob/master/validator/unicodehelper.py for an idea on a fix.

muffinresearch:
Fixes #606 

Removes the BOM when pulling from a directory or the zip.

Also fixes a small bug in the CSS parser which was preventing manual testing of the extension mentioned in the related bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "semver": "5.1.0",
     "sha.js": "2.4.5",
     "source-map-support": "0.4.0",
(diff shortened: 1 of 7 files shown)


EnTeQuAk:
r+

Thanks a lot for fixing this so quickly!

PR #610, opened 2016-04-20
cope more nicely with badly formed manifest.json

andymckay:
note the evil trailing comma.

screenshot 2016-04-20 13 08 34

before:

screenshot 2016-04-20 13 11 33

after:

screenshot 2016-04-20 13 11 12

it feels like a) i should be reading some bugs although I couldn't spot one 
and b) there should be a general JSON linting library that gives awesome errors I should be running this through, maybe jsonlint?


muffinresearch:
This looks great to me. R+

PR #611, opened 2016-04-20
show a warning if content_security_policy defined

### LINKED ISSUE #562: Check for CSP changes in the manifest and flag them for reviewers
We should warn for all CSP changes made to the manifest so reviewers know to check and make sure there are appropriate.

Certain CSP changes should be straight-up not allowed; others should warn developers and reviewers.

This will solve a lot of security issues.

andymckay:
fixes #562, but...
* I'm not sure if we should be doing something more fancy using the schema in web-extension-manifest-schema
* even raising a warning is the right thing to do here

This doesn't do anything fancy like validate the contents of the CSP policy, that would be a nice next bug.

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -10,6 +10,14 @@ export const MANIFEST_VERSION_INVALID = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_CSP = {
+  code: 'MANIFEST_CSP',
+  legacyCode: null,
+  message: _('"content_security_policy" is defined in the manifest.json'),
+  description: _('A custom content_security_policy needs additional review.'),
+  file: MANIFEST_CSP,
+};
+
 export const PROP_NAME_INVALID = {
   code: 'PROP_NAME_INVALID',
   legacyCode: null,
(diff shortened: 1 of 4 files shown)


muffinresearch:
This seems like the right place and is a good start. I like the idea of doing more on this later but for now flagging it up seems like a good approach.

There's some parsers for looking at the CSP we could look at using later on.

src/messages/manifestjson.js
```diff
@@ -10,6 +10,14 @@ export const MANIFEST_VERSION_INVALID = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_CSP = {
+  code: 'MANIFEST_CSP',
+  legacyCode: null,
+  message: _('"content_security_policy" is defined in the manifest.json'),
+  description: _('A custom content_security_policy will get more review.'),
```
muffinresearch: s/will get more review/need additional review/

muffinresearch:
Fix the conflicts - if tests are green + coverage is good r+wc

src/messages/manifestjson.js
andymckay: k

andymckay:
I did briefly play with csp-parse https://www.npmjs.com/package/csp-parse, but what I really wanted it to do was tell me when I'd done an invalid CSP - which it doesn't. But I think there's more chance for improvement long term here.

muffinresearch:
Yeah the definition of "valid" needs to be defined in the context of web-ext. E.g. we could parse the rules and look for things that might point to insecurity e.g. the use of "unsafe-inline".

PR #612, opened 2016-04-20
warn if an update_url is defined

### LINKED ISSUE #565: Fail for listed webextensions that have an `update_url`property
As with traditional extensions, webextensions hosted on AMO are not allowed to use an external `update_url`. Therefore, the linter should fail if the manifest.json of a listed webextension makes use of `update_url`.

andymckay:
fixes #565, but...
* like #611 I'm not sure if this is the right way to do it
* its an AMO problem and isn't something relevant to say Chrome add-ons, is there a special way to run this stuff?
* should this be something in mozilla/web-extension-manifest-schema

advice please...

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -357,3 +357,4 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | :white_check_mark: | error | Web extension | version is invalid in manifest.json | manifest.json | | null | PROP_VERSION_INVALID |
 | :white_check_mark: | error | Web extension | install.rdf and manifest.json present | manifest.json | | null | MULITPLE_MANIFESTS |
 | :white_check_mark: | warning | Web extension | content_security_policy in manifest.json means more review | manifest.json | | null | MANIFEST_CSP |
+| :white_check_mark: | error | Web extension | update_url not allowed in manifest.json | manifest.json | | null | MANIFEST_UPDATE_URL |
--- a/src/cli.js
+++ b/src/cli.js
@@ -55,6 +55,11 @@ export default argv
     type: 'boolean',
     default: false,
   })
+  .option('self-hosted', {
+    describe: 'Disables messages related to hosting on addons.mozilla.org.',
+    type: 'boolean',
+    default: false,
+  })
   // Require one non-option.
   .demand(1)
   .help('help')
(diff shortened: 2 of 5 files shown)


muffinresearch:
If this was added in the schema (which looks possible e.g: http://dolphm.com/how-to-disallow-an-attribute-in-json-schema-draft-3/) then it would apply all the time, which looking at #565, isn't what we want. 

Looks like your approach is good but we need a new option here [1] to match the --self-hosted option and *not* do this check when that is set.

[1] https://github.com/mozilla/addons-linter/blob/master/src/cli.js

andymckay:
Looking for feedback again @muffinresearch and @tofumatt. I wasn't sure how to add in support for that cli arg, so I did it this way. I really didn't want to pass the cli arg around to the code because you end up passing that all over the place.

Basically this moves responsibility to the collector which can then chose to put the message in or not. I started to think of complicated scenarios there and there's probably many ways we could make it more complicated (e.g. the collector moves their status).

This felt relatively clean. Any opinion before I write tests?

muffinresearch:
I'm not sure about this. It's hard to understand what `ignoreWhen` really represents and returning a boolean value when adding messages is a little bit woolly.

I think it would be nicer to work have the more direct connection between the cli option and whether you call addMessage or not.

You don't necessarily need to pass anything around. Can you not just query yargs where you need that data? You could use DI here [1] passing the original lib as a default so you can modify the yargs data for a test with a fake one as needed.

[1] https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L9

andymckay:
Well that's simpler, so I'm not going to argue with that :)

r? @muffinresearch ^

muffinresearch:
@andy since selfHosted is an option the signature will be clearer if you use an object e.g:

```js
constructor(jsonString, collector, {selfHosted=cli.argv.selfHosted}={}) {
  // do something here.
}
```
Then calling it in tests will be more self-explanatory:

```js
ManifestJSONParser(json, addonLinter.collector, {selfHosted: true});
```
as opposed to:

```js
ManifestJSONParser(json, addonLinter.collector, true);
```

See this example showing the values in more detail: http://jsbin.com/bekuyataju/edit?js,console

muffinresearch:
r+

PR #613, opened 2016-04-20
fail if both install.rdf and manifest.json defined

### LINKED ISSUE #605: Raise an error if there's another metadata-file if manifest.json exists
We should throw an exception if there's a `install.rdf` or `package.json` once there's a `manifest.json` to make sure we block duplicated metadata early enough.

Tracking: https://github.com/mozilla/addons-server/issues/2450

andymckay:
* fixes #605 sorta

before:
screenshot 2016-04-20 15 03 56

after:
screenshot 2016-04-20 15 05 14

It feels like i'm changing some fundamental decision about the linter was written that if it just fails with some key files it throws a system error. I felt that this way there's a clear and consistent error @EnTeQuAk can pass back to the user who uploaded an addon. But I'm probably wrong :)

### DIFF
--- a/src/messages/layout.js
+++ b/src/messages/layout.js
@@ -47,6 +47,19 @@ export const TYPE_NO_MANIFEST_JSON = {
     will be attempted to be inferred by package layout.`),
 };
 
+export const MULITPLE_MANIFESTS = {
+  code: 'MULITPLE_MANIFESTS',
+  legacyCode: [
+    'typedetection',
+    'detect_type',
+    'install_rdf_and_manifest_json',
+  ],
+  message: _('Both install_rdf and manifest.json found'),
+  description: _(singleLineString`The type should be determined by
+    manifest.json if present. Both install_rdf and manifest_json
+    are defined.`),
+};
+
 export const TYPE_NOT_DETERMINED = {
   code: 'TYPE_NOT_DETERMINED',
   legacyCode: [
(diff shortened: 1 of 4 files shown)


src/messages/layout.js
```diff
@@ -47,6 +47,19 @@ export const TYPE_NO_MANIFEST_JSON = {
     will be attempted to be inferred by package layout.`),
 };
 
+export const TYPE_MULTIPLE_FILES = {
```
muffinresearch: I think this might be better as `MULTIPLE_MANIFEST_FILES` or `MULTIPLE_MANIFESTS`

src/messages/layout.js
```diff
@@ -47,6 +47,19 @@ export const TYPE_NO_MANIFEST_JSON = {
...
+  code: 'TYPE_MULTIPLE_FILES',
+  legacyCode: [
+    'typedetection',
+    'detect_type',
+    'install_rdf_andy_manifest_json',
```
muffinresearch: Typo: install_rdf_**andy**_manifest_json

src/messages/layout.js
muffinresearch: That said if this matches the previous validator I guess we might be better to keep it like this for easier comparisons.

src/messages/layout.js
andymckay: hehe

src/messages/layout.js
andymckay: There wasn't anything in the previous validator about this, so I'm easy.

muffinresearch:
r+wc

Issue #617, opened 2016-04-21
Flag known hidden files


muffinresearch:
References: https://github.com/mozilla/addons-server/issues/2187

Examples:

* `__MACOSX` - a hidden directory found in zips created by OSX's archive util.
*  ~~`.DS_STORE`~~  `.DS_Store` - aka mac droppings https://en.wikipedia.org/wiki/.DS_Store
* `Thumbs.db` - windows thumbnail cache 

Check the old code on amo-validator for any other things we should look for and how the errors should be surfaced.

muffinresearch:
Fwiw: dealing with the `__MACOSX` dir is the highest priority since it breaks validation.

andymckay:
Fixed I think.

PR #619, opened 2016-04-21
create a hidden scanner that triggers for hidden files and throws a w…

andymckay:
…arning, if asked to scan the file

* support mozilla/addons-server#2187, but doesn't fix it as there's more things going there.
* this just adds in a hidden scanner, so if we are asked to scan a file and it meets the regex as "hidden" it will throw a warning
* note this doesn't list every hidden file because we currently don't scan every file, but we kinda need to and I think that's a seperate bug

screenshot 2016-04-21 16 38 15

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -114,4 +114,7 @@ export const VALID_MANIFEST_VERSION = 2;
 // io classes will open as strings or streams.
 export const MAX_FILE_SIZE_MB = 100;
 
-export const HIDDEN_FILE_REGEX = /^__MACOSX/;
+export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
+export const FLAGGED_FILE_REGEX = /Thumbs.db$|.DS_STORE$|.orig$|.old$|\~$/;
+export const HIDDEN_FILE_SCANNER_REGEX = new RegExp(
+  `${FLAGGED_FILE_REGEX.source}|${HIDDEN_FILE_REGEX.source}`);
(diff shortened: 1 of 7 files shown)


andymckay:
This will help with #617 all it needs to do is add some more smarts to HiddenScanner to and the regexes to cope with all the other things the validator coped with.

muffinresearch:
I don't think getScanner is the right place for the hidden file scanner. You could simply run the hidden scanner on every file by running it over the output of getFiles.

muffinresearch:
Aha but this makes a lot more sense in context of #621

muffinresearch:
Would be good to see this after #621 lands.

src/const.js
```diff
@@ -113,3 +113,8 @@ export const VALID_MANIFEST_VERSION = 2;
 // The max file size in MB that the
 // io classes will open as strings or streams.
 export const MAX_FILE_SIZE_MB = 100;
+
+export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
+export const FLAGGED_FILE_REGEX = /Thumbs.db$|.DS_STORE$|.orig$|.old$|\~$/;
```
muffinresearch: I think it would be an idea to make this case insensitive so that you'd catch `thumbs.db` too:

muffinresearch: I think it would be a good to add a simple test case covering all of the various patterns we're looking to catch.

muffinresearch:
r+wc

andymckay:
It has https://github.com/mozilla/addons-linter/pull/619/files#diff-6e755214ba9ab18ef118959411a4767aR42 for the regexs.

src/const.js
andymckay: I don't know if .DS_STORE would ever be lower case? It might make sense just do (t|T) or something.

PR #621, opened 2016-04-22
scan every file and add a null pass through, also remove loop

### LINKED ISSUE #620: Remove extra loops in linter
Currently linter.js in scan it has the following pattern:

```
      .then(() => {
        return this.io.getFilesByExt('.rdf');
      })
      .then((rdfFiles) => {
        return this.scanFiles(rdfFiles);
      })
```

The scanFiles then looks at the filename and hits a switch statement inside `getScanner` which then finds the appropriate scanner. So it doesn't really matter what set of files you pass to `scanFiles` it will scan them anyway with the right scanner. All this does is:
1) cause a bunch more loops (admittedly v. fast ones)
2) meant that not every file gets passed to scanned files, meaning hidden, binary and other files never get a chance to be hit with a scanner
3) each type of scanner we add means adding in another copy of this pattern

I think we can remove a bit of code, make things a bit faster and get more functionality.

andymckay:
* fixes #620 

Means that every file will get passed to `scanFile` and we can do some more stuff there. This will work nicely for when we look for hidden files and binary files.

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -20,6 +20,7 @@ import ChromeManifestScanner from 'scanners/chromemanifest';
 import CSSScanner from 'scanners/css';
 import HTMLScanner from 'scanners/html';
 import JavaScriptScanner from 'scanners/javascript';
+import NullScanner from 'scanners/null';
 import RDFScanner from 'scanners/rdf';
 import { Directory, Xpi } from 'io';
 
@@ -282,7 +283,11 @@ export default class Linter {
   scanFiles(files) {
     var promises = [];
     for (let filename of files) {
-      promises.push(this.scanFile(filename));
+      if (filename === CHROME_MANIFEST) {
+        promises.push(this.scanFile(filename, 'stream'));
+      } else {
+        promises.push(this.scanFile(filename));
+      }
     }
     return Promise.all(promises);
   }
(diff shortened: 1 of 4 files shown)


muffinresearch:
This is excellent! It should make things a lot faster as we're doing more work than we need to.

I think this would probably mean `getFilesByExt` can be removed assuming nothing else is using it.

Might be nice to have a test that shows what happens when a scan blows up.

muffinresearch:
r+wc

tofumatt:
Very cool 👍

andymckay:
Do you mean when a scanner blows up as opposed to when `scanFiles` blows up? Since `scanFiles` has no explicit throw anymore the test for triggering that seemed pretty pointless.

If we can remove `getFilesExt` that would be great, but its currently still in use by `_markJSLibs`, for `dispensary`, so @tofumatt probably needs that.

Sounds like this is ok, apart from the test, so I'll merge so I can move on with more hidden files. If I can get more ideas about that test you want, I'll file a bug on that and can add that in later.

muffinresearch:
> Do you mean when a scanner blows up as opposed to when `scanFiles` blows up? Since `scanFiles` has no explicit throw anymore the test for triggering that seemed pretty pointless.

I was thinking more about the scanning of a file blowing up, and making sure the scanFiles rejects ok. This might already be covered.

PR #628, opened 2016-04-25
Remove jscs

### LINKED ISSUE #608: Remove JSCS
JSCS has been folded into eslint now so we should remove jscs and enable more of the eslint stylistic rules as required.

muffinresearch:
Fixes #608

* Remove jscs
* Add rules that match jscs rules in eslint.
* Moves to use strings instead of magic numbers for eslint conf
* Fix a few nits.


.eslintrc
```diff
@@ -28,45 +28,70 @@
...
-    "no-unused-vars": 2,
-    "no-with": 2,
-    "quotes": [2, "single", "avoid-escape"],
-    "quote-props": [1, "consistent-as-needed"],
-    "semi": [2, "always"],
-    "space-infix-ops": 0,
-    "strict": [2, "never"],
-    "valid-typeof": 2,
-    "wrap-iife": [2, "inside"]
+    "block-scoped-var": "off",
```
tofumatt: 👍🏻 I didn't know you could do that.

tofumatt:
r+

.eslintrc
muffinresearch: Yeah I noticed it in their docs and thought "I like that". It's much clearer as I always forget what the numbers mean.

PR #632, opened 2016-04-29
allow getting the first chunks as a buffer

### LINKED ISSUE #627: Ability to read the first chunk of a file
The `src/io` libraries allow to read the whole file as a string, or stream the whole file. To support #56 we need to be able to just get the first 260 odd bytes. So we could just wrap the `getFileAsStream` in a convenience method, or add in a `getFileAsChunk` or something (although the latter means changing the calling methods). Then we can pass this chunk through to something like https://www.npmjs.com/package/file-type and get the file type.

andymckay:
* fixes #627 to help with the binary parser (which will be next)
* i confess to bit of copy and paste on the tests, wondering if the could be a bit more concise

### DIFF
--- a/src/io/directory.js
+++ b/src/io/directory.js
@@ -1,5 +1,6 @@
 import * as path from 'path';
 import { createReadStream } from 'fs';
+import firstChunkStream from 'first-chunk-stream';
 import stripBomStream from 'strip-bom-stream';
 
 import { IOBase } from 'io/base';
@@ -31,8 +32,7 @@ export class Directory extends IOBase {
       });
   }
 
-  getFileAsStream(relativeFilePath) {
-
+  getPath(relativeFilePath) {
     if (!this.files.hasOwnProperty(relativeFilePath)) {
       return Promise.reject(
         new Error(`Path "${relativeFilePath}" does not exist in this dir.`));
@@ -54,11 +54,18 @@ export class Directory extends IOBase {
         new Error(`Path argument must be relative to ${this.path}`));
     }
 
-    return Promise.resolve(createReadStream(filePath, {
-      flags: 'r',
-      encoding: 'utf8',
-      autoClose: true,
-    }).pipe(stripBomStream()));
+    return Promise.resolve(filePath);
+  }
+
+  getFileAsStream(relativeFilePath) {
+    return this.getPath(relativeFilePath)
+      .then((filePath) => {
+        return Promise.resolve(createReadStream(filePath, {
+          flags: 'r',
+          encoding: 'utf8',
+          autoClose: true,
+        }).pipe(stripBomStream()));
+      });
   }
 
   getFileAsString(path) {
@@ -82,4 +89,25 @@ export class Directory extends IOBase {
       });
   }
 
+  getChunkAsBuffer(relativeFilePath, chunkLength) {
+    return this.getPath(relativeFilePath)
+      .then((filePath) => {
+        return new Promise((resolve) => {
+          createReadStream(filePath, {
+            flags: 'r',
+            // This is important because you don't want to encode the
+            // bytes if you are doing a binary check.
+            encoding: null,
+            autoClose: true,
+          })
+          .pipe(
+            firstChunkStream({chunkLength: chunkLength},
+              function(_, enc) {
+                resolve(enc);
+              }
+            )
+          );
+        });
+      });
+  }
 }
--- a/tests/io/test.directory.js
+++ b/tests/io/test.directory.js
@@ -39,6 +39,47 @@ describe('Directory.getFiles()', function() {
 
 });
 
+describe('Directory._getPath()', function() {
+
+  it('should reject if not a file that exists', () => {
+    var myDirectory = new Directory('tests/fixtures/io/');
+    return myDirectory.getFiles()
+      .then(() => {
+        return myDirectory.getPath('whatever')
+          .then(unexpectedSuccess)
+          .catch((err) => {
+            assert.include(
+              err.message, '"whatever" does not exist in this dir.');
+          });
+      });
+  });
+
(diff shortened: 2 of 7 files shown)


src/io/directory.js
```diff
@@ -82,4 +89,25 @@ export class Directory extends IOBase {
...
+          })
+          .pipe(
+            firstChunkStream({chunkLength: chunkLength},
+              function(_, enc) {
+                resolve(enc);
+              }
+            )
+          );
+        });
+      });
```
tofumatt: I think we want to add a `catch` here to catch any errors inside this code, then a test to make sure the errors bubble up. We do that for most (all?) of our methods that have code inside a `then()` block.

tofumatt: Oh wait, no. This should be fine.

Plus coveralls thinks we're coverd 👍

tests/io/test.xpi.js
```diff
@@ -230,20 +287,61 @@ describe('Xpi.getFileAsStream()', function() {
...
+        assert.equal(buffer.toString(), '12');
       });
   });
 
+});
+
+
+describe('Xpi.getFileAsStream()', function() {
+
+  beforeEach(() => {
```
tofumatt: If these were copy + pasted we could turn this into a helper and `import` it rather than redeclaring it, but this amount of copy/pasting doesn't bother me.

tests/io/test.directory.js
```diff
@@ -39,6 +39,49 @@ describe('Directory.getFiles()', function() {
...
+  it("should reject if path starts with '/'", () => {
+    var myDirectory = new Directory('tests/fixtures/io/');
+    myDirectory.files = {
+      '/file1.txt': {},
+    };
+    return myDirectory.getPath('/file1.txt')
+      .then(unexpectedSuccess)
+      .catch((err) => {
+        assert.include(
+          err.message, 'Path argument must be relative');
```
tofumatt: This should fit onto the previous line and be under 80 characters long.

tests/io/test.directory.js
```diff
@@ -39,6 +39,49 @@ describe('Directory.getFiles()', function() {
...
+  it('should reject if path does not start with base', () => {
+    var myDirectory = new Directory('tests/fixtures/io/');
+    myDirectory.files = {
+      '../file1.txt': {},
+    };
+    return myDirectory.getPath('../file1.txt')
+      .then(unexpectedSuccess)
+      .catch((err) => {
+        assert.include(
+          err.message, 'Path argument must be relative');
```
tofumatt: Can be on the line above.

tests/io/test.directory.js
```diff
@@ -198,3 +201,19 @@ describe('Directory.getFileAsString()', function() {
...
+describe('Directory.getChunkAsBuffer()', function() {
+
+  it('should get a buffer', () => {
+    var myDirectory = new Directory('tests/fixtures/io/');
+    return myDirectory.getFiles()
+      .then(() => {
+        return myDirectory.getChunkAsBuffer('dir2/dir3/file3.txt', 2);
+      })
+      .then((buffer) => {
+        assert.equal(buffer.toString(), '12');
```
muffinresearch: Nice test

tofumatt: I don't quite get this one, it's too clever for me!

tests/io/test.directory.js
```diff
@@ -198,3 +201,19 @@ describe('Directory.getFileAsString()', function() {
...
 });
+
+
+describe('Directory.getChunkAsBuffer()', function() {
+
+  it('should get a buffer', () => {
+    var myDirectory = new Directory('tests/fixtures/io/');
+    return myDirectory.getFiles()
+      .then(() => {
+        return myDirectory.getChunkAsBuffer('dir2/dir3/file3.txt', 2);
```
muffinresearch: I suppose chunks could be part of an options object then it would be a more obvious signature.

muffinresearch: As an aside - with fat arrow function if it's a single line return like this you can forgo the braces and the return and just have the expression. However, this is clear and explicit so it's fine as is.

tofumatt: I like the explicit and consistent way, so I say we avoid the implicit returns. It's definitely harder for a newbie to grok and it's harder to visually scan as well.

tests/io/test.directory.js
muffinresearch: The file contains `123` the first 2 bytes are `12`. 

Also:

```
$ head -c 2 tests/fixtures/io/dir2/dir3/file3.txt
12
```

muffinresearch:
So the chunk grabbing code won't remove a BOM - presumably that's by design?

tests/io/test.directory.js
tofumatt: Ah, shoot, okay.

In my tests with fixture data (XPIs) I referenced what was inside the fixture data so the tests would be more immediately clear without having to reference the fixtures (eg https://github.com/mozilla/addons-linter/blob/master/tests/test.linter.js#L749). We should do that here just so it's immediately apparent.

tofumatt: Though yes, now that I understand it: this is a cool test 👍

muffinresearch:
r+wc

tofumatt:
r+wc from me

tests/io/test.directory.js
andymckay: I will add some more comments into these two tests so its more obvious.

andymckay:
A BOM is for a text stream and the libraries I've seen basically assume there is no BOM in the file for determining its file type, so I left it out as well.

tests/io/test.directory.js
andymckay: Both updated.

PR #633, opened 2016-04-29
Make flagged file regex case insensitive

muffinresearch:
r? @andymckay

I realise I messed-up the original description of #617 since I put  `.DS_STORE` instead of `.DS_Store` however I think there's no harm in making that entire regex case insensitive.

The regexes are used individually because you can't have part of a regex case insensitive. That said maybe we should consider making the __MACOSX one the same since I don't see it would create false positives.

The other question I had is I see we don't match a directory `foo.old` or `foo.orig` and instead only match files ending with `.orig`/`.old`. I wondered it that was intended?

Should we also check for .git directories etc? Although I guess this could be a endless list...

Lastly - the regex tests are now created via loops as that will make it easier to add strings later and this way it's much easier to know what string caused a problem:

### DIFF
--- a/tests/scanners/test.hidden.js
+++ b/tests/scanners/test.hidden.js
@@ -1,5 +1,5 @@
 import HiddenScanner from 'scanners/hidden';
-import {HIDDEN_FILE_SCANNER_REGEX} from 'const';
+import { FLAGGED_FILE_REGEX, HIDDEN_FILE_REGEX } from 'const';
 
 describe('HiddenScanner', function() {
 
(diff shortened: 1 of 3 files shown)


andymckay:
All it does in the end is a warning, its not a blocker, its just trying to help out the developer and the reviewer, so if we are little too aggressive and catch too many things that's probably cool. Likewise if we miss a few things, that's probably fine, so that's why I didn't bother looking at directories.

tests/scanners/test.hidden.js
```diff
@@ -37,16 +37,59 @@ describe('HiddenScanner', function() {
...
+    'whatever.old/something',
+  ];
 
+  for (const filePath of nonMatchingFlaggedFiles) {
+    it(`should not match ${filePath} as a flagged file`, () => {
+      assert.isNotOk(filePath.match(FLAGGED_FILE_REGEX),
+        `${filePath} should not match flagged file regex`);
+    });
+  }
 });
```
andymckay: Much clearer tests, thanks.

andymckay:
r+

PR #635, opened 2016-04-29
allow scanner to tell what style of stream its getting

andymckay:


tofumatt:
Looks like you've got a test failure in the style: https://travis-ci.org/mozilla/addons-linter/jobs/126754123#L1167

src/scanners/base.js
```diff
@@ -11,7 +26,7 @@ export default class BaseScanner {
     this._defaultRules = [];
     this._parsedContent = null;
     this._rulesProcessed = 0;
-
+
```
tofumatt: This'd be your style nit; trailing whitespace.

tofumatt:
Don't see any tests here; would be good to add one. Even if it's just to verify the method exists.

andymckay:
I fail in style in so many ways.

andymckay:
Fixed style test and added two tests.

muffinresearch:
I think `fileStreamType` being `string` or `stream` is confusing. The old var name was clearer.

muffinresearch:
Fwiw: It doesn't have to be `streamOrString` but it should be something that is closer to representing what a `stream` or `string` arg represents.

andymckay:
`streamOrStringOrChunk` gets awkward.

muffinresearch:
> streamOrStringOrChunk gets awkward.

hah I agree. How about `fileResultType`?

andymckay:
Works for me.

andymckay:
Updated to `fileResultType`.

tofumatt:
r+

Issue #56, opened 2015-10-06
Add binary file tests


muffinresearch:
There are a set of tests that read the first few bytes of a file to identify them.

muffinresearch:
Related to https://github.com/mozilla/amo-validator/blob/7a8011aba8bf8c665aef2b51eb26d0697b3e19c3/validator/testcases/packagelayout.py

PR #638, opened 2016-05-01
Binary scanner

andymckay:
* requires #635 
* look for bad file extensions
* look for bad magic numbers
* since the null scanner wasn't being used, but instead was the fallback to search for binary files, changed its name to binary scanner
* since the hidden scanner was basically just scanning by filename, i thought filename scanner would be more appropriate


andymckay:
Will need rebasing with #635 when that lands.

src/const.js
```diff
@@ -116,3 +116,21 @@ export const MAX_FILE_SIZE_MB = 100;
 
 export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs.db$|.DS_Store$|.orig$|.old$|\~$/i;
+
+export const FLAGGED_FILE_EXTENSION_REGEX =
+  /.exe$|.dll$|.dylib$|.so$|.sh$|.class$|.swf$/i;
```
tofumatt: Do the dots need escaping (eg `\.exe`)?

src/scanners/binary.js
```diff
@@ -0,0 +1,34 @@
...
+    this.linterMessages.push(
+      Object.assign({}, messages.FLAGGED_FILE_TYPE, {
+        type: constants.VALIDATION_NOTICE,
+        file: this.filename,
+      })
+    );
+  }
+
+  scan() {
+    return new Promise((resolve) => {
```
tofumatt: You can instead return `return Promise.resolve(this.linterMessages);` at the bottom and avoid returning the function. We've been trying to do that in this repo.

tests/scanners/test.binary.js
```diff
@@ -0,0 +1,35 @@
+import BinaryScanner from 'scanners/binary';
+import {FLAGGED_FILE_MAGIC_NUMBERS} from 'const';
```
tofumatt: This is a major nitpick but we usually put spaces around grouped imports like this:

```js
import { FLAGGED_FILE_MAGIC_NUMBERS } from 'const';
```

tofumatt:
r+wc (and obviously #635 being merged 😄 ).

src/const.js
andymckay: Actually they should, interestingly in FLAGGED_FILE_REGEX too.

src/scanners/binary.js
andymckay: Ah cool, thanks

tests/scanners/test.binary.js
andymckay: Rightio :)

andymckay:
fixes #56

PR #652, opened 2016-05-05
drop all css errors down to warnings

### LINKED ISSUE #623: Make CSS parse error clearer
Currently comes out as "Unknown Word" mozilla/addons-server#2187

andymckay:
* fixes #623, maybe it will be a warning in more context now
* but the more important point is that all CSS errors are now warnings

Shame I can't assign more people but a look over from @wagnerand would be nice.


muffinresearch:
r+

PR #656, opened 2016-05-06
drop eval down to a warning

andymckay:
* fixes mozilla/addons-server#2560


muffinresearch:
Do the docs need any updates to reflect this? 

r+wc

andymckay:
Docs updated, for that field it was `error?` 😄

PR #667, opened 2016-05-12
warn if already signed

### LINKED ISSUE #666: Already signed
Lets warn if the add-on is already signed, eg:

screenshot 2016-05-12 15 52 11

andymckay:
* fixes #666 
* there's also multi-package extensions tests in the current validator, i'm not sure what we do there if webextensions

### DIFF
--- a/tests/test.linter.js
+++ b/tests/test.linter.js
@@ -260,17 +260,20 @@ describe('Linter.getScanner()', function() {
     assert.deepEqual(Scanner, CSSScanner);
   });
 
-  it('should return FilenameScanner', function() {
-    var addonLinter = new Linter({_: ['foo']});
-    var Scanner = addonLinter.getScanner('__MACOSX/foo.txt');
-    assert.deepEqual(Scanner, FilenameScanner);
-  });
+  var shouldBeFilenameScanned = [
+    '__MACOSX/foo.txt',
+    'wat.dll',
+    'META-INF/manifest.mf',
+  ];
 
-  it('should return FilenameScanner', function() {
-    var addonLinter = new Linter({_: ['foo']});
-    var Scanner = addonLinter.getScanner('wat.dll');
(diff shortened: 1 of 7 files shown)


tests/test.linter.js
```diff
@@ -260,17 +260,20 @@ describe('Linter.getScanner()', function() {
...
-    var Scanner = addonLinter.getScanner('__MACOSX/foo.txt');
-    assert.deepEqual(Scanner, FilenameScanner);
-  });
+  const shouldBeFilenameScanned = [
+    '__MACOSX/foo.txt',
+    'wat.dll',
+    'META-INF/manifest.mf',
+  ];
+
+  for (const filename of shouldBeFilenameScanned) {
```
tofumatt: Why is `filename` a `const` here, I don't get it.

tofumatt:
Not sure about that `const` bit so maybe a comment or just explaining to me would help.

Otherwise it's good! r+wc

tests/test.linter.js
andymckay: Absolutely no reason, will change to a var.

andymckay:
By the way, I learned that multipackage extensions are signalled by having a particular id (32) in the install.rdf. Since we don't have an install.rdf in WebExtensions, it means that there are no similar multipackage extensions and hence the tests for this are pointless.

See https://developer.mozilla.org/en-US/docs/Multiple_Item_Packaging#install.rdf for more.

tests/test.linter.js
andymckay: Oh wait let's not change it to a var, let's change it to `let` or a `const`, both of those will rebind the variable on each iteration. If you make it `var` then you'll get one value of the loop repeated each iteration and because those are valid every time, the tests still pass! Yay, but they aren't testing the real thing.

I've put in a pull request to fix this, sorry for the numpty mistake.

https://github.com/mozilla/addons-linter/pull/674

Issue #668, opened 2016-05-13
Conflate multiple schema errors down


andymckay:
Which is accurate, but it should really be 1 error, not 4.

<img width="1222" alt="screenshot 2016-05-12 17 31 33" src="https://cloud.githubusercontent.com/assets/74699/15234523/7acb97c4-1867-11e6-8fee-ea7b7ec07157.png">

Its detailed though! 😄

andymckay:
```
ack permissions -A 2 ~/sandboxes/webextensions-examples/tabs-tabs-tabs/manifest.json
  "permissions": [
      "wat"
  ]
```

andymckay:
A few other things to note, it shouldn't be an error (instead a warning) and the description and code should be set. @muffinresearch and @tofumatt and guidance on how to do this, is the intention to leave the web-manifest-whatever-it-is purely a parser and do this work in the validator?

I'm feeling the best way is to use the `/permissions/...` to figure out where the error is and work off that?

andymckay:
Warning in firefox:

1463104032482	addons.webextension.tabs-tabs-tabs@mozilla.org	WARN	Loading extension 'tabs-tabs-tabs@mozilla.org': Reading manifest: Error processing permissions.0: Unknown permission "wat"

PR #669, opened 2016-05-13
wip

andymckay:
* this is a wip for #668 
* tries to differentiate the errors coming back to seperate out required errors vs array errors and so will hopefully cope with some of the more complicated ones we haven't implemented yet, like content scripts

screenshot 2016-05-12 19 51 06

* feedback gratefully received and yeah no tests yet

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -37,28 +37,48 @@ export default class ManifestJSONParser {
     this.isValid = this._validate();
   }
 
+  errorLookup(error) {
+    var result = {
+      code: 'MANIFEST_JSON_INVALID',
+      description: 'MANIFEST_ERROR',
+      file: 'manifest.json',
+      level: 'error',
+      message: `${error.dataPath}: ${error.message}`,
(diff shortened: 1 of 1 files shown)


andymckay:
... and if its not an array, we get a fatal error

screenshot 2016-05-12 19 53 33

src/parsers/manifestjson.js
```diff
@@ -37,28 +37,48 @@ export default class ManifestJSONParser {
     this.isValid = this._validate();
   }
 
+  errorLookup(error) {
+    var result = {
```
tofumatt: This is odd as it appears to be a message that isn't in the `message.js` file. We kept everything there so we'd be able to trace where all messages came from. I'd start with a message there and add your dynamically built `message` attribute to it here.

tofumatt:
I had a quibble with the location of the message data but I'm 👍🏻 on the idea! I like the output a *lot* more with your implementation.

andymckay:
I ended with a new pull request, oops.

PR #671, opened 2016-05-13
Create warning and error list for third-party JS libraries

### LINKED ISSUE #598: Create a warning list for third party libraries
We need to warn reviewers of potentially dangerous or forbidden third party libraries.

My proposal would be to create a list of third party libraries we always warn about, similar to what we already have with dispensary for whitelisting.

tofumatt:
Fix #598

This won't work until https://github.com/mozilla/dispensary/pull/142 is merged.


tofumatt:
r?

wagnerand:
@tofumatt could you update this list with the recent angular additions?

wagnerand:
We'd also need to get https://github.com/mozilla/addons-linter/blob/master/docs/third-party-libraries.md up and running before this lands on prod.

src/linter.js
```diff
@@ -9,6 +9,7 @@ import { terminalWidth } from 'cli';
 import * as constants from 'const';
 import { ARCH_DEFAULT, ARCH_JETPACK, CHROME_MANIFEST, INSTALL_RDF,
          MANIFEST_JSON } from 'const';
+import { BANNED_LIBRARIES, UNADVISED_LIBRARIES } from 'libraries';
```
muffinresearch: `UNADVISED` reads weird to me, but that's probably more because it's not commonly used.

muffinresearch:
lgtm although there's a tiny bit of missing coverage.

r+wc (code)

src/linter.js
tofumatt: Yeah, I didn't know what to call it. They're like... "warning" or "caution"... I couldn't think of a better word 😞

tofumatt:
We have one uncovered line, but that was the same as before, odd.

PR #674, opened 2016-05-14
fix let mistake in test

andymckay:
...also spotted an errant `console.log` in there.


tofumatt:
r+, in the future `let` is preferred declaring a variable inside a loop constructor anyway 👍🏻

Sorry about that, should have said so when reviewing the other patch!

PR #705, opened 2016-05-25
Run coverage on node 4+ if not on travis (fix #703)

### LINKED ISSUE #703: Run coverage on newer versions of node when not on Travis
Our code that [doesn't run test coverage on newer versions of node on Travis](https://github.com/mozilla/addons-linter/blob/master/Gruntfile.js#L57) (because it's slow) means that on my Mac running node 5 I never run the coverage without explicitly remembering to do so.

This is bad because new contributors will run `npm test` locally, could pass, then push and Travis would fail.

So let's test for Travis env variables and only opt-out of higher node versions if we're on Travis.

tofumatt:
I didn't like that `npm test` wasn't running coverage.

Alternatively we can rip out that check entirely, though I'd imagine that would ever-so-slightly increase the run time of Travis tests. So I'm fine not to, though if you said to do it I would 😄


muffinresearch:
Since we're deploying on a more recent node we should probably switch the .nvmrc to that and try again with having coverage run on all versions and see if it's still slow.

Maybe the coverage is not slow any more. 

If not you could land this and file some issues to fix up the coverage I used a different solution in addons-frontend and iirc it was noticeably faster.

muffinresearch:
Hmm, yeah that test fail looks familar.

muffinresearch:
Seeing if the coverage lib changes have any effect: https://github.com/mozilla/addons-linter/pull/709

muffinresearch:
Superceded by #709

PR #706, opened 2016-05-26
Update unadvised libs with more robust test

wagnerand:
~~still WIP.~~


tests/test.linter.js
```diff
@@ -899,21 +899,24 @@ describe('Linter.extractMetadata()', function() {
     var addonLinter = new Linter({
       _: ['tests/fixtures/angular-unadvised-library.xpi'],
```
muffinresearch: You should probably change this line to 'fake.xpi' or something since you're not actually looking into the fixture xpi ata all any more. Also all the comments above can be removed.

src/linter.js
```diff
@@ -448,7 +448,7 @@ export default class Linter {
     }
   }
 
-  _markBannedLibs(addonMetadata) {
+  _markBannedLibs(addonMetadata, unadvised_libraries = UNADVISED_LIBRARIES) {
```
muffinresearch: We generally have a pattern of a leading underscore to denote something used for DI. E.g: https://github.com/mozilla/addons-linter/blob/master/src/linter.js#L340

tests/test.linter.js
muffinresearch: As mentioned on irc the fixture data can be deleted if it's not used any where else.

tests/test.linter.js
```diff
@@ -883,37 +883,28 @@ describe('Linter.extractMetadata()', function() {
...
-  // - install.rdf
-  // - package.json
-  // - README.md
   it('should flag unadvised JS libraries in an XPI.', () => {
     var addonLinter = new Linter({
-      _: ['tests/fixtures/angular-unadvised-library.xpi'],
+      _: ['fake.xpi'],
     });
-    var markUnadvisedSpy = sinon.spy(addonLinter, '_markBannedLibs');
+    var unadvised_libs = ['test_unadvised_fake_lib.js'];
```
muffinresearch: vars should be camelCased s/unadvised_libs/fakeUnadvisedLibs/

src/linter.js
```diff
@@ -448,7 +448,7 @@ export default class Linter {
     }
   }
 
-  _markBannedLibs(addonMetadata) {
+  _markBannedLibs(addonMetadata, _unadvised_libraries = UNADVISED_LIBRARIES) {
```
muffinresearch: Same re camelCasing `_unadvisedLibraries` or `_unadvisedLibs` would be better.

muffinresearch:
Couple of minor stylistic nits otherwise this looks good.

r+wc

tests/test.linter.js
```diff
@@ -883,37 +883,28 @@ describe('Linter.extractMetadata()', function() {
...
-      .then((metadata) => {
-        assert.ok(markUnadvisedSpy.called);
-        assert.equal(Object.keys(metadata.jsLibs).length, 2);
-        assert.deepEqual(metadata.jsLibs, {
-          'data/angular-1.5.3.js': 'angularjs.1.5.3.angular.js',
-          'data/jquery-2.1.4.min.js': 'jquery.2.1.4.jquery.min.js',
-        });
+    var result = addonLinter._markBannedLibs(metadata, unadvised_libs);
+    assert.equal(Object.keys(result.jsLibs).length, 2);
+    assert.deepEqual(result.jsLibs, {
```
muffinresearch: You set this above, maybe you can just point to metadata.jsLibs instead of duplicating it?

wagnerand:
@muffinresearch thank you for your mentoring!

PR #709, opened 2016-05-26
Refactor coverage

muffinresearch:
Based on #705 

* Refactors coverage to use better libs.


muffinresearch:
I'll clean this up a bit more tomorrow but it looks to do the job.

muffinresearch:
So the slow test is still there. I think this is overall better but didn't solve the root problem. 

I added another 5 secs to the timeout to try and ensure the node 4 tests aren't flakey. What we need to try and do is get to the bottom of the slow test. 

What is odd is if you skip the slow tests the slowness moves elsewhere. The test that comes up as slow really isn't doing that much so it's unclear why it's a problem - I don't think it's quite as it looks in the test output, e.g. the cause of the slowness is the test that looks slow.

To dig into this more I think it would make sense to profile the test suite and see if anything odd is happening.

muffinresearch:
The downside of this approach is that it makes the coverage slower for repeat runs since it moves the coverage into the webpack build and that means the instrumentation is never skipped if it's already been run.

muffinresearch:
New version:

```
time npm run test
...
real	0m33.234s
user	0m33.617s
sys	0m1.397s
```

```
npm run test-no-coverage:
...
real	0m15.390s
user	0m15.853s
sys	0m1.037s
```

Old:

With primed instrumenation cache:

```
npm run test
...
real	0m15.191s
user	0m15.666s
sys	0m1.046s
```

```
npm test-no-coverage
...
real	0m15.344s
user	0m15.791s
sys	0m1.027s
```

muffinresearch:
@tofumatt, so what do you think - we could just do your original patch and bump the timeout and that should fix most of the trouble. At least in lieu of finding a direct cause for the slowness in instrumented tests. I'm not sure this refactor bought us much unfortunately.

tasks/mochaTest.js
```diff
@@ -1,25 +1,47 @@
+const path = require('path');
+
+const defaultOptions = {
+  // Timeout is long due to tests that check *all* rules have been run.
+  timeout: 15000,
```
tofumatt: This is the only bit that seems worse than before; I presume for my patch we could just up the one slow test's timeout? Or no?

tofumatt:
The refactor seems a bit cleaner to me. It's more in line with what we have at addons-frontend, right?

I'm fine with this even if it doesn't buy us much, but honestly I'd be more interested in the results; I didn't find the old coverage tasks bad and they seem to be the ones largely changed.

If you squashed this and we didn't need to set the timeout high for everything I'm r+wc on this--especially if we need a globally high timeout in my patch too.

If we can do the timeout just for the slow test on my old PR I'd go with that just because I like not setting a high timeout everywhere.

tofumatt:
We discussed this on vidyo on Friday but I forgot to comment.

r+

tofumatt:
@muffinresearch Ping, good to merge this?

muffinresearch:
Yeah let's try it, this is closer to web-exts setup and the addons-frontend too. Any probs let me know.

PR #724, opened 2016-05-31
Ignore comments in manifest.json (fix #723)

### LINKED ISSUE #723: Ignore comments in manifest.json
Supports http://github.com/mozilla/addons-server/issues/2734

We need to allow manifest.json files with comments (at least single-line comments) to be parsed without an error.

tofumatt:
Supports https://github.com/mozilla/addons-server/issues/2734

Code borrowed from: https://github.com/abarreir/crx2ff/blob/d2b882056f902d751ad05e329efda7eddcb9d268/libs/ext-converter.js#L19-L37

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -1,13 +1,18 @@
+import esprima from 'esprima';
+import validate from 'mozilla-web-extension-manifest-schema';
+
+import cli from 'cli';
 import { PACKAGE_EXTENSION } from 'const';
 import log from 'logger';
-import validate from 'mozilla-web-extension-manifest-schema';
 import * as messages from 'messages';
-import cli from 'cli';
 import { singleLineString } from 'utils';
 
 export default class ManifestJSONParser {
 
   constructor(jsonString, collector, {selfHosted=cli.argv.selfHosted}={}) {
+    // Add the JSON string to the object; we'll use this for testing.
+    this._jsonString = jsonString;
+
     // Provides ability to directly add messages to
     // the collector.
     this.collector = collector;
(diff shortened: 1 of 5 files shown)


src/parsers/manifestjson.js
```diff
@@ -22,16 +24,33 @@ export default class ManifestJSONParser {
...
+      // them out and see if we can parse the JSON.
+      // If not it's just garbage JSON and we error.
+      //
+      // Originally from https://github.com/abarreir/crx2ff/blob/d2b882056f902d751ad05e329efda7eddcb9d268/libs/ext-converter.js#L19-L37
+      var manifestString = `var o = ${jsonString}`;
+      try {
+        var tokens = esprima.tokenize(manifestString).slice(3);
+        var cleanedJSONString = tokens.reduce((json, token) => {
+          return `${json}${token.value}`;
+        }, '');
```
EnTeQuAk: We should totally document what the hell this does, I looked up `esprima` and wondered that this is an actual JavaScript parser, then noticed, yeah, JSON is probably just a subset of JavaScript so it'll work, parse it and then you're fetching out only relevant tokens.

On the other hand I also found 

 * http://stackoverflow.com/a/23753148
 * and https://github.com/judofyr/timeless/issues/57#issuecomment-31872462

I don't know about ES6 but it appears there are very very minimal differences. We should test them and at least expect them to fail or make sure that they succeed by some magic.

tests/parsers/test.manifestjson.js
```diff
@@ -279,3 +279,26 @@ describe('ManifestJSONParser update_url', function() {
...
+    var addonLinter = new Linter({_: ['bar']});
+    var json = `{"something": true,\n// I am a JSON comment, sigh\nblah}`;
+    var manifestJSONParser = new ManifestJSONParser(json,
+                                                    addonLinter.collector);
+    assert.equal(manifestJSONParser.isValid, false);
+    var errors = addonLinter.collector.errors;
+    assert.equal(errors[0].code, messages.MANIFEST_JSON_INVALID.code);
+    assert.include(errors[0].message, 'Invalid JSON in manifest file.');
+  });
+});
```
EnTeQuAk: Since we are "cleaning" this with a fully capable ECMAScript parser I would love to see examples of trying to inject some JavaScript code in there, e.g do some `var bla = "foo"` and make sure they don't appear in the final output.

Also, an actual valid example like the docs from https://developer.chrome.com/extensions/manifest (I just assume they're very valid) would be nice

```javascript
{
  // Required
  "manifest_version": 2,
  "name": "My Extension",
  // Make the hell sure to use semvar.org if increasing this
  "version": "0.0.1"
}
```

…would be something I'd do, maybe.

tofumatt:
Yeah I copied this from some stuff referenced in bugs... but you're right, this is basically stripping out comments by using a parser. I mean, ultimately we're still using `JSON.parse` so it _should_ be safe, but I'll see about making sure this is a bit saner.

And even if it is it needs more docs. Thanks @EnTeQuAk 👍

EnTeQuAk:
I don't see where we're using json.parse, correct me if I'm wrong but to
me it seems were just making the json string a Javascript object, parse
that and collect the tokens we want (I actually don't even see that were
filtering the tokens)
 
 
On Tue, May 31, 2016, at 06:14 AM, tofumatt ☕️ wrote:
> Yeah I copied this from some stuff referenced in bugs... but you're
> right, this is basically stripping out comments by using a parser. I
> mean, ultimately we're still using JSON.parse so it *should* be safe,
> but I'll see about making sure this is a bit saner.
> And even if it is it needs more docs. Thanks @EnTeQuAk[1] 👍
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub[2], or mute the
> thread[3].
>
 

Links:

  1. https://github.com/EnTeQuAk
  2. https://github.com/mozilla/addons-linter/pull/724#issuecomment-222588877
  3. https://github.com/notifications/unsubscribe/AAIfGZLmjUbHEdTNNaQ9hWcuzWFaKP77ks5qG7W9gaJpZM4IqHwu

tofumatt:
https://github.com/mozilla/addons-linter/pull/724/files#diff-baa432d9354e3de899d28a2638d1499eR40 is where we call `JSON.parse` again.

EnTeQuAk:
ah, okay, I didn't see that part :see_no_evil:

src/parsers/manifestjson.js
tofumatt: Hmm, okay. So I will document that in the code, but these seem like quite extreme edge cases. I can't imagine we'll have users with these kinds of characters in their manifests.

I'll see about catching those very specific errors and if not I'll at least add those links and some extra comments to the code. Thanks!

tests/parsers/test.manifestjson.js
```diff
@@ -279,3 +279,57 @@ describe('ManifestJSONParser update_url', function() {
...
+                                                    addonLinter.collector);
+    assert.equal(manifestJSONParser.isValid, false);
+    var errors = addonLinter.collector.errors;
+    assert.equal(errors[0].code, messages.MANIFEST_JSON_INVALID.code);
+    assert.include(errors[0].message, 'Invalid JSON in manifest file.');
+  });
+
+  it("doesn't evaluate JS code even though esprima is used", () => {
+    var addonLinter = new Linter({_: ['bar']});
+    var json = '// eval("");\n{"something": true}\nvar bla = "foo";';
```
EnTeQuAk: and… `'{"something": eval("")}'` is invalid too? I would assume that `eval()` in a comment doesn't work but outside could be a more realistic test maybe

EnTeQuAk:
Thanks for all the great improvements! Looks a lot better for me, just one improvement for te test maybe.

r+wc

PR #727, opened 2016-06-02
Don't error on update_url (fix #726)

### LINKED ISSUE #726: Strip update_url and warn developers instead of erroring
We banned the usage of `update_url` in #565 if a Web Extension is on AMO, but it might be more constructive to strip it out and alert developers that extensions hosted on AMO cannot use `update_url`... and that they need to upload new versions of their extensions to AMO for users to get updates.

But this way they could use the same manifest as their Chrome add-ons.

Supports https://github.com/mozilla/addons-server/issues/2579

tofumatt:
Modify how we deal with update_url.


tofumatt:
r?

tofumatt:
Ready for r?

EnTeQuAk:
Looks good to me. r+

PR #731, opened 2016-06-03
Flag large files (fix #730)

### LINKED ISSUE #730: Flag very large files
We should warn when we detect very large source code files in the linter:

* They may crash the linter
* They are a possible marker of bad code quality/bad add-on quality

Supports https://github.com/mozilla/addons-server/issues/2699

tofumatt:
Will fix https://github.com/mozilla/addons-server/issues/2699.

I figured we only need to flag/warn on large JS files, as:

1. it's unlikely other files would be as large
2. I _assume_ large CSS/JSON files won't blow up the parsers the same way JS does because ESLint ends up doing the whole tree-building statement assembling stuff because JS is a real language

If that seems too naive let me know; I thought it was nicer (and it ends up being slightly less complicated code as well, I think?)

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -113,6 +113,11 @@ export const VALID_MANIFEST_VERSION = 2;
 // The max file size in MB that the
 // io classes will open as strings or streams.
 export const MAX_FILE_SIZE_MB = 100;
+// This is the limit in megabytes of a file we will parse (eg. CSS, JS, etc.)
+// A singular CSS/JS file over 2MB seems bad and may actually be full of data
+// best stored in JSON/some other data format rather than code.
+// https://github.com/mozilla/addons-linter/issues/730
+export const MAX_FILE_SIZE_TO_PARSE_MB = 2;
 
 export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|\~$/i;
--- a/src/linter.js
+++ b/src/linter.js
@@ -321,8 +321,26 @@ export default class Linter {
     var ScannerClass = this.getScanner(filename);
     return this.io.getFile(filename, ScannerClass.fileResultType)
       .then((fileData) => {
-        let scanner = new ScannerClass(
-          fileData, filename, {addonMetadata: this.addonMetadata});
+        // First: check that this file is under our 2MB parsing limit. Otherwise
+        // it will be very slow and may crash the lint with an out-of-memory
+        // error.
+        let fileSize = this.io.files[filename].size !== undefined ?
+          this.io.files[filename].size :
+          this.io.files[filename].uncompressedSize;
+        var maxSize = 1024 * 1024 * constants.MAX_FILE_SIZE_TO_PARSE_MB;
+
+        if (ScannerClass !== BinaryScanner && fileSize >= maxSize) {
+          let filesizeError = Object.assign({}, messages.FILE_TOO_LARGE, {
+            file: filename,
+            type: constants.VALIDATION_ERROR,
+          });
+          return Promise.resolve([filesizeError]);
+        }
+
+        let scanner = new ScannerClass(fileData, filename, {
+          addonMetadata: this.addonMetadata,
+        });
+
         return scanner.scan();
       })
       // messages should be a list of raw message data objects.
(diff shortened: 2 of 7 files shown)


src/messages/layout.js
```diff
@@ -72,6 +73,16 @@ export const TYPE_NOT_DETERMINED = {
...
 };
 
+export const FILE_TOO_LARGE = {
+  code: 'FILE_TOO_LARGE',
+  legacyCode: null,
+  message: _('File is too large to parse.'),
+  description: _(singleLineString`This file is not binary and is too large to
+    parse. Files larger than ${MAX_FILE_SIZE_TO_PARSE_MB}MB will not be
+    parsed. If your JavaScript file has a large list, consider removing the
+    list and loadng it as a separate JSON file instead.`),
```
EnTeQuAk: loading

EnTeQuAk:
Two things:

* Where did you get the 8MB limit from, experimentation or guessing? Would be nice to get some actual tests where things break
* I'm not so sure about not applying the limits to other files as well. Sure, JS files can be large but CSS can be too and CSS is quite complex and the parser builds a tree as well afaik. JSON is a bit simpler but has some strict limits most of the time too, e.g deep nesting and sorts can break.

muffinresearch:
The limit of 8mb seems quite high - it would be nice to try and look at what are the larger files in current xpis even if that's by looking at legacy extensions - those files could be used for testing.

Does this work ok for the file-based IO as opposed to via a zip?

muffinresearch:
I agree with @EnTeQuAk - the file limits should apply across the board rather than just JS.

tofumatt:
Yeah I am also convinced, I'll change it around, shouldn't be hard.

- tofumatt

tofumatt:
@muffinresearch This works for file-based IO because the code is in `scanFile()`, so it's called for any IO class. Tests would just be stubbing out a different IO class to not much effect, so I didn't add any. But yeah, it works everywhere.

The limit is just random really; I was imagining a ton of libs/JS code combined into one add-on JS file and I wanted to be permissive with file limits.

Shall we say 2MB as a max? I just figure maybe someone including a few libraries and their own code could be around/over 1MB without being unreasonable.

tofumatt:
Changed to flag all files not run through `BinaryScanner` and use a 2MB limit. r?

EnTeQuAk:
Changes look good to me. r+

We might have to adapt the limits in case they're too low but that should be reasonable effort.

src/const.js
```diff
@@ -113,6 +113,11 @@ export const VALID_MANIFEST_VERSION = 2;
 // The max file size in MB that the
 // io classes will open as strings or streams.
 export const MAX_FILE_SIZE_MB = 100;
```
muffinresearch: Unrelated to this change but this limit seems prettty big.

tofumatt:
> We might have to adapt the limits [..]
This made me realise there's a constant for the limit but the numbers are hard-coded in the tests, so I changed the tests to be the constant value + a certain amount, at least makes things easier to change.

src/linter.js
```diff
@@ -321,8 +321,26 @@ export default class Linter {
     var ScannerClass = this.getScanner(filename);
     return this.io.getFile(filename, ScannerClass.fileResultType)
       .then((fileData) => {
-        let scanner = new ScannerClass(
-          fileData, filename, {addonMetadata: this.addonMetadata});
+        // First: check that this file is under our 2MB parsing limit. Otherwise
+        // it will be very slow and may crash the lint with an out-of-memory
+        // error.
+        let fileSize = this.io.files[filename].size !== undefined ?
```
muffinresearch: Whilst it's not strictly necessary I think we've generally used `typeof` for undefined equality checks.

tests/io/test.base.js
```diff
@@ -15,7 +15,7 @@ describe('io.IOBase()', function() {
     assert.equal(io.maxSizeBytes, 104857600);
   });
 
-  it('should should reject calling getFiles()', () => {
```
muffinresearch: parisinthespring

muffinresearch:
r+wc

src/messages/layout.js
tofumatt: Oh shoot, missed this. I'll fix it.

src/const.js
tofumatt: A movie or something could tip the scales... but yeah it's wildly large.

PR #733, opened 2016-06-03
Improve message for unknown keys (fix #732)

### LINKED ISSUE #732: "Should not have additional properties" message isn't super-clear
As pointed out in https://github.com/mozilla/addons-server/issues/2846#issuecomment-223596538, this message isn't super-clear; maybe we should improve it.

Meant to open this here, accidentally opened https://github.com/mozilla/addons-server/issues/2849. Oops!

tofumatt:
This is just a bit of an improvement on the message, IMO. The original one seemed very generic and unhelpful; hopefully my new one isn't too specific.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -91,8 +91,18 @@ export default class ManifestJSONParser {
   errorLookup(error) {
     // This is the default message.
     var baseObject = messages.MANIFEST_JSON_INVALID;
+
+    // This is the default from webextension-manifest-schema, but it's not a
+    // super helpful error. We'll tidy it up a bit:
+    if (error && error.message) {
+      let lowerCaseMessage = error.message.toLowerCase();
+      if (lowerCaseMessage === 'should not have additional properties') {
+        error.message = 'is not a valid key or has invalid extra properties';
+      }
+    }
+
     var overrides = {
-      message: `${error.dataPath} ${error.message}`,
+      message: `"${error.dataPath}" ${error.message}`,
     };
 
     if (error.keyword === 'required') {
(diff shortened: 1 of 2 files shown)


src/parsers/manifestjson.js
```diff
@@ -91,8 +91,15 @@ export default class ManifestJSONParser {
   errorLookup(error) {
     // This is the default message.
     var baseObject = messages.MANIFEST_JSON_INVALID;
+
+    // This is the default from webextension-manifest-schema, but it's not a
+    // super helpful error. We'll tidy it up a bit:
+    if (error.message === 'should NOT have additional properties') {
+      error.message = 'is not a valid key or has invalid extra properties';
+    }
```
EnTeQuAk: Shouldn't that actually be in `web-extension-manifest-schema`? I mean, once some decides to write the `NOT` in regular lower case characters this line is invalid an we won't even notice it. Well, the test might fail but still. Or, we should have error ids in the schema too to make targeting error message rewrites like that better but I'm more for just changing the error message in the schema than overwriting it here.

EnTeQuAk:
I don't see why this can't be changed in the schema directly and would prefer that instead of overwriting error messages in the linter.

tofumatt:
Fair enough; I didn't see an obvious way to change it there but I think I was being lazy on Friday. I'll look into that.

tofumatt:
I just spent awhile looking through the schema and there isn't a way (I can tell) to have custom error messages yet. It's coming in a new version but there are no provisions for it and the proposal for custom errors is six months old.

Especially because we call `validate()` on the schema in the linter and not the schema lib, we can't modify the messages until we're here from what I can tell.

So yeah, this is the best I can do. But if you don't think it's worth it we can scrap it. I at least did a `lowercase()` call so your point of "changing of case breaks it" is fixed 😉.

EnTeQuAk:
> 23:58 <@cgrebs> tofumatt: bloody hell, I just noticed that the error message you're changing is actually not coming from the schema but from `ajv` :-/

Well, r+ then :laughing: Sorry, I somehow thought the message was coming directly from the schema

PR #736, opened 2016-06-07
Validate a CRX file (close #696)

### LINKED ISSUE #696: Unpack and validate a CRX file
Supports https://github.com/mozilla/addons-server/issues/2579

We should allow the linter to unpack and validate a CRX file. The CRX format is just a fancy ZIP, but we may even be able to use https://github.com/shyiko/node-crx-parser rather than unpacking it ourselves.

tofumatt:
This is just us being fancy really, but now we can validate a CRX file.

Most of the code is already there in the `Xpi` class, so all I really had to do was strip the CRX header info, then it uses the `Xpi` class and underlying Yauzl stuff.


src/linter.js
```diff
@@ -337,16 +337,21 @@ export default class Linter {
...
       })
       .then((stats) => {
         if (stats.isFile() === true) {
-          log.info('Package is a file. Attempting to parse as an .xpi/.zip');
-          this.io = new _Xpi(this.packagePath);
+          if (this.packagePath.endsWith('.crx')) {
+            log.info('Package is a file ending in .crx; parsing as a CRX');
+            this.io = new _Crx(this.packagePath);
+          } else {
+            log.info('Package is a file. Attempting to parse as an .xpi/.zip');
```
EnTeQuAk: "parse as **a** .xpi/.zip" am I wrong?

EnTeQuAk:
Looks good to me, r+wc

src/linter.js
tofumatt: English is super weird, oddly enough it's "an Xpi" the same as "an X-Ray". 🙃

- tofumatt

EnTeQuAk: okay, I'll just accept that :laughing:

PR #745, opened 2016-06-09
Do not scan known third party libraries

### LINKED ISSUE #744: Linter does not ignore known libs
The linter marks known third party libraries using dispensary but does not ignore them during scanning. It should though, as this is the main point about marking them known.

wagnerand:
Fixes #744


src/linter.js
```diff
@@ -408,7 +408,14 @@ export default class Linter {
         return this.io.getFiles();
       })
       .then((files) => {
-        return this.scanFiles(Object.keys(files));
+
+        function excludeKnownLibsFilter(file) {
```
tofumatt: Avoid inline functions; they will be harder to test but also are harder to follow.

Additionally, I would return a new list of files, then pass that to `scanFiles`. More lines will make things clearer, I think.

tofumatt:
You'll need to fix the style issues in your PR: https://travis-ci.org/mozilla/addons-linter/jobs/136506550#L1435

tests/test.linter.js
```diff
@@ -884,6 +884,47 @@ describe('Linter.extractMetadata()', function() {
...
+        return Promise.resolve(Object.keys(fakeFiles));
+      }
+    }
+
+    return addonLinter.extractMetadata({
+      _console: fakeConsole,
+      _Xpi: FakeXpi,
+    }).then((metadata) => {
+      var warnings = addonLinter.collector.warnings;
+      var filtered = warnings.map((element) => {
```
tofumatt: This is a good test but a comment above it, just to explain to someone a bit less familiar with the code, would be great.

src/linter.js
wagnerand: Ah, I did not to that bc that new list of files was only used once :)

wagnerand:
Updated.

src/linter.js
```diff
@@ -408,7 +408,11 @@ export default class Linter {
         return this.io.getFiles();
       })
       .then((files) => {
-        return this.scanFiles(Object.keys(files));
+        // Known libraries do not need to be scanned
+        let unknownFiles = Object.keys(files).filter((file) => {
```
tofumatt: This is much better, but can we rename this to something a bit more clear? How about `filesWithoutJSLibraries` (or `filesWithoutJSLibs` if it's too long).

tofumatt:
r+wc

wagnerand:
Thank you!

PR #751, opened 2016-06-17
Fold web-extension-manifest-schema into addons-linter

### LINKED ISSUE #750: Fold web-extension-manifest-schema into this project.
This will help save us the need to bump one less dep all the way up to addons-server.

muffinresearch:
Fixes #750

This PR moves the schema into the linter so we have one less package to deal with when making updates. 

Once this passes review we can deprecate web-extension-manifest-schema repo and move all the bugs.


EnTeQuAk:
r+, this looks great and makes things way easier!

Issue #787, opened 2016-06-28
Don't allow http update_urls


andymckay:
Firefox only allows https update urls. This was talked about on IRC in relation to this bug:

https://bugzilla.mozilla.org/show_bug.cgi?id=1281030#c6

andymckay:
This has been done in #788.

PR #788, opened 2016-06-28
https only update_url

andymckay:
* only allows https urls as per #787 and https://bugzilla.mozilla.org/show_bug.cgi?id=1281030#c6
* not sure if we need the "uri" format check or just https:// is enough?


tofumatt:
I guess leading with `https://` means any URL there is fine... there might be restricted characters? But then we have unicode domains 😄 

r+

EnTeQuAk:
r+

PR #791, opened 2016-06-28
Flag non-string permissions as errors (fix #789)

### LINKED ISSUE #789: Flag bad permissions (or have a whitelist of permissions)
Supports https://github.com/mozilla/addons-server/issues/2726, we need to make sure only known permissions are allowed, or we'll end up accepting Chrome "app" extensions that don't work in Firefox.

tofumatt:
Whoever r?s it first 😄

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -35,6 +35,16 @@ export const MANIFEST_FIELD_INVALID = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_BAD_PERMISSION = {
+  code: 'MANIFEST_BAD_PERMISSION',
+  legacyCode: null,
+  message: _('The permission type is unsupported.'),
+  description: _(singleLineString`Permissions must be strings. If a permission
+    is an object, it's likely from a Chrome App Extension and will not work in
+    Firefox. See https://mzl.la/1R1n1t0 (MDN Docs) for more information.`),
+  file: MANIFEST_JSON,
+};
+
 export const MANIFEST_PERMISSIONS = {
   code: 'MANIFEST_PERMISSIONS',
   legacyCode: null,
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -107,6 +107,13 @@ export default class ManifestJSONParser {
 
     if (error.keyword === 'required') {
       baseObject = messages.MANIFEST_FIELD_REQUIRED;
+    } else if (error.dataPath.startsWith('/permissions') &&
+               typeof error.data !== 'undefined' &&
+               typeof error.data !== 'string') {
+      // Check for non-strings in the manifest permissions; these indicate
+      // a Chrome app extension. This means an error.
+      baseObject = messages.MANIFEST_BAD_PERMISSION;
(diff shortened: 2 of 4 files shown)


tests/parsers/test.manifestjson.js
```diff
@@ -100,6 +100,29 @@ describe('ManifestJSONParser manifestVersion', function() {
 
 });
 
+describe('ManifestJSONParser bad permissions', function() {
+
+  it('should error if permission is not a string', () => {
```
muffinresearch: A success case is presumably tested implicitly it might be nice to add a specific test in any case?

src/parsers/manifestjson.js
```diff
@@ -107,6 +107,13 @@ export default class ManifestJSONParser {
 
     if (error.keyword === 'required') {
       baseObject = messages.MANIFEST_FIELD_REQUIRED;
+    } else if (error.dataPath.startsWith('/permissions') &&
```
muffinresearch: Is the expectation that you'd get this error for every permission in the list that's not a string?

src/messages/manifestjson.js
```diff
@@ -35,6 +35,16 @@ export const MANIFEST_FIELD_INVALID = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_BAD_PERMISSION = {
+  code: 'MANIFEST_BAD_PERMISSION',
+  legacyCode: null,
+  message: _('The permission type is unsupported.'),
+  description: _(singleLineString`Permissions must be strings; object
```
muffinresearch: minor nitpick, it might not be an object.

muffinresearch: Looking through https://developer.chrome.com/extensions/declare_permissions I don't see any reference to non-string values - is that a historical thing?

muffinresearch:
r+wc

src/messages/manifestjson.js
tofumatt: I believe it's for apps, not extensions. See https://github.com/mozilla/addons-server/issues/2726#issuecomment-228899614

tofumatt: Amended the message to be `Permissions must be strings. If a permission is an object, it's likely from a Chrome App Extension and will not work in Firefox. [...]`

src/parsers/manifestjson.js
tofumatt: Indeed, as all permissions are a string, as I understand it: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/permissions

> This key is an array of strings, and each string is a request for a permission.

src/messages/manifestjson.js
muffinresearch: Ah ok - makes sense re the app thing. Are there any other indicators of an App versus ext? This still seems worth adding but I wonder if there's a more solid thing that we could also be checking for?

tofumatt: Don’t know of anything else yet, and it otherwise passed validation okay.
I’d like to get this merged into today’s tag, but if anything else comes up
I’ll look for a more definitive way to flag something as an app.

PR #797, opened 2016-06-29
Usage of 'mozIndexedDB' should be a warning instead of an error

### LINKED ISSUE #796: The use of mozIndexedDB should not be an error but a warning
Libraries and frameworks still use mozIndexedDB when assessing what's available in the browser, even https://github.com/mozilla/localForage/. The rules doc also says this should be a warning, but it is currently an error which for example, makes it impossible to submit webextensions that use localForage.

wagnerand:
Fixes #796

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -322,7 +322,7 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | :x: | warning | unsafe_template_escape |  | Potentially unsafe template escape sequence | | | | |
 | :x: | warning | js_protoype_extension_dissallowed |  | JS prototype extension not allowed  | | | | |
 | :white_check_mark: | warning | mozindexdb_removed |  | mozIndexedDB has been removed |  | | | MOZINDEXEDDB |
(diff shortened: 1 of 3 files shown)


tofumatt:
Please update `docs/rules.md` to clarify that this is now a warning and not an error.

r+wc

wagnerand:
I get some (seemlingly) unrelated test failures, let's wait for travis...

wagnerand:
Afaics, the rules document is correct, just the implementation was wrong.

tofumatt:
Right on. A nitpick would be to change the rule from "not allowed" to "discouraged".

wagnerand:
I am totally at loss about the test failures. Could you give me a hint?

Also, I don't see where it says "not allowed".

wagnerand:
Ok, I updated the rule description.

tofumatt:
After reviewing this and the tests failures this is r- (not safe to merge) until #798 is fixed. After that it's r+ but I'll have a look at making sure it's all okay and merge it in myself 👍

PR #804, opened 2016-07-01
Usage of 'mozIndexedDB' should be a warning instead of an error (updated)

### LINKED ISSUE #796: The use of mozIndexedDB should not be an error but a warning

tofumatt:
Fixes #796 and closes #797.

@wagnerand quick r? to make sure this is alright?


wagnerand:
r+

PR #807, opened 2016-07-02
Don't allow duplicate keys in manifest (fix #803)

### LINKED ISSUE #803: Error about duplicate properties in manifest.json
The linter should show a warning (or error?) if a manifest has duplicate properties in it. This will lead to confusing results when the JSON is parsed by Firefox (I think the last one wins). Example:
````json
{
  "name": "My Extension",
  "name": "My Extension 2",
  "version": "1.0"
}
````

### PR DESCRIPTION
**Pull request by tofumatt**
I tried to do this with [`JSON.parse`'s reviver parameter](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Using_the_reviver_parameter), but unfortunately it doesn't behave as I hoped and actually doesn't report dupe keys and is overall a bit weird.

So the easiest solution was to include a parser solely to check for dupe keys.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -1,15 +1,18 @@
 import esprima from 'esprima';
+import RJSON from 'relaxed-json';
 import validate from 'schema/validator';
 
 import cli from 'cli';
-import { PACKAGE_EXTENSION } from 'const';
+import { MANIFEST_JSON, PACKAGE_EXTENSION } from 'const';
 import log from 'logger';
 import * as messages from 'messages';
 import { singleLineString } from 'utils';
 
 export default class ManifestJSONParser {
 
-  constructor(jsonString, collector, {selfHosted=cli.argv.selfHosted}={}) {
+  constructor(jsonString, collector, {
+    rJSON=RJSON, selfHosted=cli.argv.selfHosted,
+  }={}) {
     // Add the JSON string to the object; we'll use this for testing.
     this._jsonString = jsonString;
 
@@ -86,6 +89,9 @@ export default class ManifestJSONParser {
 
     this.selfHosted = selfHosted;
     this.isValid = this._validate();
+
+    // Check for duplicate keys, which renders the manifest invalid.
+    this._checkForDuplicateKeys(rJSON);
   }
 
   errorLookup(error) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/parsers/manifestjson.js [hunk 1]
```diff
@@ -203,4 +207,19 @@ export default class ManifestJSONParser {
       version: this.parsedJSON.version,
     };
   }
+
+  _checkForDuplicateKeys() {
+    try {
+      RJSON.parse(this._jsonString, { duplicate: true, warnings: true });
+    } catch (err) {
+      if (err.message.includes('Duplicate key')) {
```
muffinresearch: This feels like it might be brittle.

src/parsers/manifestjson.js
```diff
@@ -203,4 +207,19 @@ export default class ManifestJSONParser {
...
+        let key = err.message.match(/Duplicate key: (.*)/)[1];
```
muffinresearch: If there's more errors after the duplicate key one, what do you end up with `key` representing?

tofumatt: The first key we hit. I suppose I could try to go through each one, if the
parser will allow it.

src/parsers/manifestjson.js
tofumatt: That’s why we have a test for it, if the underlying message changed we’d
know :-)

src/parsers/manifestjson.js
tofumatt: Turns out we can run the parser in a `tolerant` mode and collect all the errors. I'll do that instead.

tofumatt:
Okay, updated things a bit. I think it's a bit less fragile and it handles multiple duplicates now.

muffinresearch:
Does this parser produce other errors? If so it would be good to test everything works as expected if other errors are found in the middle of checking for duplicate keys.

tofumatt:
It could, although normally JSON.parse would hit them first. I think dupes
are the only things this one would hit.

But I’ll add a check. I’ll add an explicit check for “Duplicate key”. 👍🏻

tofumatt:
Okay, added a test for other errors too. Thanks for all the catches on this one.

src/parsers/manifestjson.js
```diff
@@ -203,4 +209,24 @@ export default class ManifestJSONParser {
       version: this.parsedJSON.version,
     };
   }
+
+  _checkForDuplicateKeys(_rJSON=RJSON) {
+    try {
+      _rJSON.parse(this._jsonString, { duplicate: true, tolerant: true });
+    } catch (err) {
+      if (err.warnings && err.warnings.length > 0) {
+        for (let error of err.warnings) {
```
muffinresearch: Nice, this looks a lot more robust.

tests/parsers/test.manifestjson.js
```diff
@@ -100,6 +101,121 @@ describe('ManifestJSONParser manifestVersion', function() {
...
+    var parseStub = sinon.stub(fakeRJSON, 'parse', () => {
+      throw {
+        warnings: [
+          {
+            line: 1,
+            message: 'Duplicate key: not actually found but this is a test',
+          },
+          {
+            line: 1,
+            message: 'DifferentError: Who cares',
```
muffinresearch: 👍

muffinresearch:
r+ lgtm

PR #835, opened 2016-07-24
Allow toolkit version (fix #764)

### LINKED ISSUE #764: Beta string versions should be allowed
@raphaelfaria previously filed this bug here: mozilla/web-extension-manifest-schema#144

The schema currently supports only integer numbers on version parts and prevents the signing of beta versions of web extensions.
It should allow this match `(a|alpha|b|beta|pre|rc)\d*$` at the last version part as described in: https://developer.mozilla.org/en-US/docs/Toolkit_version_format and https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance

Rob--W:


src/schema/formats.js
```diff
@@ -8,6 +8,12 @@ export function isValidVersionString(version) {
   if (typeof version !== 'string') {
     return false;
   }
+  // Firefox's version format is laxer than Chrome's, it accepts:
+  // https://developer.mozilla.org/en-US/docs/Toolkit_version_format
+  // We choose a slightly restricted version of that format (but still more
+  // permissive than Chrome) to allow Beta addons, per:
+  // https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance
+  version = version.replace(/(a|alpha|b|beta|pre|rc)\d*$/, '');
```
tofumatt: If we're more permissive than Chrome, we should add a warning/notice about using version strings with a "toolkit" format; that way WebExtension developers won't think their add-on will work unmodified in Chrome (cross-browser help is a goal of the linter).

Rob--W: Done. The MDN article should also be updated to mention https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance; I'll do that later once this PR is merged.

tests/parsers/test.manifestjson.js
```diff
@@ -415,6 +415,17 @@ describe('ManifestJSONParser version', function() {
...
   });
 
+  it('should collect a notice on toolkit version values', () => {
+    var addonLinter = new Linter({_: ['bar']});
+    var json = validManifestJSON({version: '1.0.0.0pre0'});
+    var manifestJSONParser = new ManifestJSONParser(json,
+                                                    addonLinter.collector);
+    assert.equal(manifestJSONParser.isValid, true);
+    var notices = addonLinter.collector.notices;
+    assert.equal(notices[0].code, 'PROP_VERSION_TOOLKIT_ONLY');
```
tofumatt: Sorry to be a pain, but can you use `constants.PROP_VERSION_TOOLKIT_ONLY.code` here instead? It's just a bit more future-proof (and what we try our best to do) in case the text of the code ever changes.

tofumatt:
r+wc, let me know when that one tweak is made and I'll get this merged. Thanks! 👍

Rob--W:
@tofumatt Done. I fixed it in all test files for consistency.

src/schema/formats.js
```diff
@@ -2,12 +2,23 @@ import parse from 'url-parse';
 
 const VALIDNUMRX = /^[0-9]{1,5}$/;
 
+// Firefox's version format is laxer than Chrome's, it accepts:
+// https://developer.mozilla.org/en-US/docs/Toolkit_version_format
+// We choose a slightly restricted version of that format (but still more
+// permissive than Chrome) to allow Beta addons, per:
+// https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance
+const TOOLKITVERSIONRX = /(a|alpha|b|beta|pre|rc)\d*$/;
```
tofumatt: This is just a note to myself to change this to `TOOLKIT_VERSION_REGEX` before I merge. I didn't notice this until now, but no sense in more back and forth, as the patch is perfect otherwise.

tofumatt:
Nice! You rock! 👍🏻😊

tofumatt:
Did a quick change to that variable name and merged in 2488841578d555af1d55907e89348caa5ade36c5.

PR #843, opened 2016-07-27
Add support for default Node version

kumar303:
It would be nice to support the default version of Node (like, what ships with OS X). I don't know how hard it would be so I wanted to kick off a build to see.

My main motivation is that `web-ext` aims to support the default Node (more importantly, npm 2.x) and we got this error when trying to upgrade to the latest `addons-linter`: https://travis-ci.org/mozilla/web-ext/jobs/147481973

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 sudo: false
 node_js:
+- '0.12'
 - '4'
 - '5'
 script: npm test


tofumatt:
That's a good point. I know what's causing it to fail and we thought "oh node 4 is good enough" but we can make it compatible easily. I'll get on this when back from PTO

muffinresearch:
I'm not sure about this one. Is there actually a default node version shipped with OSX?

The general plan for node support from the services in production perspective is that we want to target our support around Active LTS versions of node. And as newer versions hit Active LTS we'll want to move to that being the primary supported version. LTS versions overlap so whilst we'll want to start supporting node 6 sooner, 4 is still going to be in Active LTS for a bit longer until April 2017 at which point it will be in maintenance mode.

schedule

What we do for node support wrt add-on developers wouldn't necessarily have to tie-up 100% with what we're trying to do for production, however 0.12 is getting quite old now - its out of Active LTS already and will be out of maintenance in December 2016.

All in all, I feel like we'd want to be careful about trying to continue to support outgoing old node versions. If we can decide to support just Active LTS versions (and maybe LTS versions in maintainence too - either way that should only ever be a rolling 2 versions) that will be a lot easier to maintain and communicate.

For developers if you're installing this to use locally, managing versions of node is straight-forward with the likes of nvm so I don't think that will be a huge problem.

kumar303:
> For developers if you're installing this to use locally, managing versions of node is straight-forward with the likes of nvm so I don't think that will be a huge problem.

I don't think it's so straight forward on Windows. I know Node isn't installed by default on Windows but I got a lot of bug reports on web-ext from people running older versions of NPM which proved that they had installed something pre-4.0. 

I think 0.12 is still quite common for people who are not "power users." In the add-on world there seems to be a lot of novice developers so I think supporting a wide range of versions is a good idea, unless it becomes too difficult. 

If there's an easy fix for this then let's try it. If not, maybe we'll have to cut support. I'll also need to cut support in web-ext then.

kumar303:
Right, but I'm just saying if we can make addons-linter work out of the box for someone (if they already have Node) then it just reduces friction.

The LTS support for 0.12 only ended two months ago so I think it's still going to be used quite a bit by people who don't upgrade often. If it's too hard to support, let's ditch it, but if it's easy, why not?

muffinresearch:
> Right, but I'm just saying if we can make addons-linter work out of the box for someone (if they already have Node) then it just reduces friction.

I'm not saying we shouldn't make backwards compatibility changes if the effort is small.

However, I do think we need to find a balance somewhere around the node versions we actively support. Engineering resource is finite and in the addons-linter case particularly we're going need to make sure we target the right versions for use as a service.

The more we can guide people towards using the LTS versions the better. We could for example say if it works for you with node version X great, but we only support node >= Y.

tofumatt:
Turns out the only bit of code in the linter that doesn't run on `0.12` is in dispensary: https://github.com/mozilla/dispensary/pull/196/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R8

And we explicitly went to node 4.x because of it. If we didn't use `String.prototype.includes` we can work with older versions. If node `0.12` is what ships on OS X, out-of-the-box ability to install would be worth giving up one instance of `String.prototype.includes`.

muffinresearch:
@tofumatt - Unless I've missed it I'm not sure OSX ships with any default node does it?

tofumatt:
Oh, maybe not. Fair enough!

muffinresearch:
> Oh, maybe not. Fair enough!

I thought it did, probably because nvm talks about "system" node. But then I realised "system" in that sense was the one installed via brew. When I removed all the brew installed node to see what was left there didn't appear to be a node belonging to the system. ¯\_(ツ)_/¯

kumar303:
Looks like I was mistaken about OS X shipping with 0.12. I have seen a lot of bug reports in web-ext about users on 0.12 though. I'm not sure exactly why. Maybe they are using Linux distro defaults or something. Just based on the in-flux of bugs I think it's worth supporting 0.12 but, if it's a hassle, just let me know and I will update all the web-ext docs with red warnings about it.

kumar303:
Because my original assumptions about 0.12 being the default are not correct, I think it is not worth the effort to try and support 0.12. I think it will be harder than just updating dispesary (https://github.com/mozilla/addons-linter/pull/843#issuecomment-236576606) because the eslint toolchain seems to be incompatible with 0.12 and that's used internally by the linter. I'll close this and just deprecate 0.12 support in web-ext when I update its linter.

Issue #849, opened 2016-07-29
Provide a way to ignore node_modules and other artifacts


kumar303:
When linting the source directory for an extension, it's typical to have some artifacts that won't actually make it into the final extension, like `node_modules`. The linter will likely die when scanning this dir so there should be a way to ignore them. The error I see is:

````
Error: ENFILE: file table overflow, open '/Users/kumar/tmp/blok/node_modules/nyc/node_modules/lodash/debounce.js'
    at Error (native)
````

I think the [Directory class](https://github.com/mozilla/addons-linter/blob/master/src/io/directory.js) just needs to accept some filtering.

I am mostly interested in providing programattic support for `web-ext lint`. See https://github.com/mozilla/web-ext/issues/397

kumar303:
I can take a stab at this

muffinresearch:
Just a thought; node does walk backwards looking for `node_modules`, I wonder if it might be better to tell people to move their node_modules above the extension dir?

Supporting them leaving node_modules in the directory increases the risk that they're accidentally packaged doesn't it or does web-ext automatically exclude them when it packages too?

Also, what happens if you're bundling some deps in an extension is it ok to use node_modules then or would the expectation be that you'd copy the lib?

kumar303:
web-ext automatically excludes `node_modules`. I haven't seen anyone bundling their extension with webpack (or similar). That would probably require some additional build steps to point the manifest at the right source.

pdehaan:
Not sure if relevant here, but in some projects (like min-vid) we ignore the node_modules directory, but in some cases want to pull in a specific npm module or files: https://github.com/meandavejustice/min-vid/blob/master/.jpmignore

Ideally we'd have a way of specifying files to add/ignore so the user can control which files can get added to their web-extension.

kumar303:
We'd like to support an [ignore file](https://github.com/mozilla/web-ext/issues/131) in web-ext which could override the defaults. For the purpose of `web-ext lint`, this ignore file would apply to the linter as well.

PR #862, opened 2016-08-02
Add callback to control which files can be scanned

kumar303:
Fixes https://github.com/mozilla/addons-linter/issues/849

This satisfies what web-ext needs for https://github.com/mozilla/web-ext/issues/397 I'm not sure if the addons-linter command itself needs to support filtering or not.

### DIFF
--- a/src/io/base.js
+++ b/src/io/base.js
@@ -14,6 +14,15 @@ export class IOBase {
     // If this is too large the node process will hit a RangeError
     // when it runs out of memory.
     this.maxSizeBytes = 1024 * 1024 * MAX_FILE_SIZE_MB;
+    // A callback that accepts a relative file path and returns
+    // true if the path should be included in results for scanning.
+    this.shouldScanFile = () => true;
+  }
+
+  setScanFileCallback(callback) {
+    if (typeof callback === 'function') {
+      this.shouldScanFile = callback;
+    }
   }
 
   getFile(path, fileStreamType='string') {
--- a/src/io/directory.js
+++ b/src/io/directory.js
@@ -24,7 +24,10 @@ export class Directory extends IOBase {
       return Promise.resolve(this.files);
     }
 
-    return _walkPromise(this.path)
+    return _walkPromise(
+      this.path, {
+        shouldIncludePath: this.shouldScanFile.bind(this),
+      })
       .then((files) => {
         this.files = files;
         this.entries = Object.keys(files);
--- a/src/io/directory.js
+++ b/src/io/directory.js
@@ -26,7 +26,7 @@ export class Directory extends IOBase {
 
     return _walkPromise(
       this.path, {
-        shouldIncludePath: this.shouldScanFile.bind(this),
+        shouldIncludePath: (...args) => this.shouldScanFile(...args),
       })
       .then((files) => {
         this.files = files;
--- a/src/io/utils.js
+++ b/src/io/utils.js
@@ -1,12 +1,13 @@
 import { lstat, readdir } from 'fs';
 import * as path from 'path';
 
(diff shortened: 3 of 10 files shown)


src/io/xpi.js
```diff
@@ -39,6 +39,10 @@ export class Xpi extends IOBase {
     if (/\/$/.test(entry.fileName)) {
       return;
     }
+    if (!this.shouldScanFile(entry.fileName)) {
+      log.debug(`not including this for scanning: ${entry.fileName}`);
```
tofumatt: This is a serious nitpick but can we remove "this" from the sentence or rephrase it?

```js
`Not included for scanning: ${entry.fileName}`
```

would read easier to me were I a dev debugging stuff.

tests/test.linter.js
```diff
@@ -26,6 +26,20 @@ var fakeCheckFileExists = () => {
   });
 };
 
+class FakeIOBase {
```
tofumatt: 👏🏻👏🏻👏🏻

src/io/xpi.js
kumar303: sure

src/io/base.js
```diff
@@ -14,6 +14,15 @@ export class IOBase {
     // If this is too large the node process will hit a RangeError
     // when it runs out of memory.
     this.maxSizeBytes = 1024 * 1024 * MAX_FILE_SIZE_MB;
+    // A callback that accepts a relative file path and returns
+    // true if the path should be included in results for scanning.
+    this.shouldScanFile = () => true;
+  }
+
+  setScanFileCallback(callback) {
+    if (callback) {
```
tofumatt: Should we also test for `typeof callback === 'function'` just to make sure it's not passed garbage?

tofumatt:
This looks good to me, r+wc

src/io/directory.js
```diff
@@ -24,7 +24,10 @@ export class Directory extends IOBase {
       return Promise.resolve(this.files);
     }
 
-    return _walkPromise(this.path)
+    return _walkPromise(
+      this.path, {
+        shouldIncludePath: this.shouldScanFile.bind(this),
```
muffinresearch: Can't you use an arrow function for this?

src/io/utils.js
```diff
@@ -17,16 +18,26 @@ export function walkPromise(curPath) {
...
+          }
         } else if (stat.isDirectory()) {
           return readdirPromise(curPath)
             .then((files) => {
               // Map the list of files and make a list of readdir
               // promises to pass to Promise.all so we can recursively
               // get the data on all the files in the directory.
               return Promise.all(files.map((fileName) => {
-                return walk(path.join(curPath, fileName));
+                if (shouldIncludePath(fileName)) {
```
muffinresearch: At this point you're looking at the files in the directory, not the directory. If you're wanting to make the decision to actually skip the current directory shouldn't this be earlier?

tests/test.linter.js
muffinresearch: Heh nice! 🎉

muffinresearch:
r+wc

src/io/utils.js
kumar303: Mann, the recursion was hurting my brain on this. You're right, I just pushed a change that only calls `shouldIncludePath()` once, right at the top.

tests/test.linter.js
kumar303: I'm tempted to generalize and extract [fake()](https://github.com/mozilla/web-ext/blob/master/tests/helpers.js#L108-L136) because it has been helpful to preserve the original interface in a fake object.

Issue #864, opened 2016-08-02
Using linter as a lib: Error: only one instance of babel-polyfill is allowed


kumar303:
`web-ext` does a babel polyfill and uses the linter as a library. Since the linter also does a babel polyfill, this is causing an epic stack trace (below).

````
>> Mocha exploded!
>> Error: only one instance of babel-polyfill is allowed
>>     at Object.<anonymous> (/Users/kumar/dev/addons-linter/node_modules/babel-polyfill/lib/index.js:12:9)
>>     at Module._compile (module.js:410:26)
>>     at Object.Module._extensions..js (module.js:417:10)
>>     at Module.load (/Users/kumar/dev/web-ext/node_modules/coffee-script/lib/coffee-script/register.js:45:36)
>>     at Function.Module._load (module.js:301:12)
>>     at Module.require (module.js:354:17)
>>     at require (internal/module.js:12:17)
>>     at Object.<anonymous> (/Users/kumar/dev/addons-linter/dist/webpack:/external "babel-polyfill":1:1)
>>     at _webpack_require_ (/Users/kumar/dev/addons-linter/dist/webpack:/webpack/bootstrap 020e2f8f0c96331e7689:19:1)
>>     at Object.module.exports.Object.defineProperty.value (/Users/kumar/dev/addons-linter/dist/webpack:/src/main.js:5:1)
>>     at _webpack_require_ (/Users/kumar/dev/addons-linter/dist/webpack:/webpack/bootstrap 020e2f8f0c96331e7689:19:1)
>>     at module.exports.Object.defineProperty.value (/Users/kumar/dev/addons-linter/dist/webpack:/webpack/bootstrap 020e2f8f0c96331e7689:39:1)
>>     at Object.<anonymous> (/Users/kumar/dev/addons-linter/dist/webpack:/webpack/bootstrap 020e2f8f0c96331e7689:39:1)
>>     at Module._compile (module.js:410:26)
>>     at Object.Module._extensions..js (module.js:417:10)
>>     at Module.load (/Users/kumar/dev/web-ext/node_modules/coffee-script/lib/coffee-script/register.js:45:36)
>>     at Function.Module._load (module.js:301:12)
>>     at Module.require (module.js:354:17)
>>     at require (internal/module.js:12:17)
>>     at Object.module.exports.Object.defineProperty.value (/Users/kumar/dev/web-ext/dist/webpack:/external "addons-linter":1:1)
[...]

PR #865, opened 2016-08-02
Do not trigger a babel polyfill more than once

kumar303:
Fixes https://github.com/mozilla/addons-linter/issues/864

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -2,7 +2,9 @@ import cli from 'cli';
 import Linter from 'linter';
 import log from 'logger';
 
-import 'babel-polyfill';
+if (!global._babelPolyfill) {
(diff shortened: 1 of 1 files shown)


tofumatt:
r+

kumar303:
@tofumatt thanks for the review. @muffinresearch do you know if this is some safe magic to play with? It's all I could come up with.

muffinresearch:
Yeah it looks ok to me. We could look at using the https://babeljs.io/docs/plugins/transform-runtime/ but then that wouldn't solve some things like Array.prototype.includes. We could take a look at when support for that hits node at some point.

Short term though, this looks like it will make the most sense.

PR #866, opened 2016-08-03
Only load yargs when used directly (fix #735)

### LINKED ISSUE #735: Duplicate output in --help
### Steps to reproduce:
```bash
✗ ./node_modules/.bin/addons-linter --help
Usage: ./node_modules/.bin/addons-linter [options] library-json-file

Mozilla Dispensary v0.5.0

Options:
  --log-level         The log-level to generate
  [string] [choices: "fatal", "error", "warn", "info", "debug", "trace", "fatal", "error", "warn", "info", "debug", "trace"]
                                                                                                          [default: "fatal"]
  --output, -o, -o    The type of output to generate    [string] [choices: "json", "text", "json", "text"] [default: "json"]
  --metadata          Output only metadata as JSON                                              [boolean] [default: "false"]
  --pretty            Prettify JSON output                                                        [boolean] [default: false]
  --stack             Show stacktraces when errors are thrown                                     [boolean] [default: false]
  --boring            Disables colorful shell output                                              [boolean] [default: false]
  --self-hosted       Disables messages related to hosting on addons.mozilla.org.                 [boolean] [default: false]
  --max               Maximum number of concurrent HTTP requests                                    [string] [default: "35"]
  --libraries         Custom library file                [string] [default: "node_modules/addons-linter/src/libraries.json"]
  -h, --help, --help  Show help                                                                                    [boolean]
```

### Actual results:

> `--log-level`         The log-level to generate [string]
> [choices: "fatal", "error", "warn", "info", "debug", "trace", "fatal", "error", "warn", "info", "debug", "trace"]
> [default: "fatal"]

> `--output`, `-o`, `-o`    The type of output to generate    [string]
> [choices: "json", "text", "json", "text"]
> [default: "json"]

1. The "choices" are duplicated for both `--log-level` and `--output`.
2. For some reason `--output` shows the `-o` shortcut twice: `--output, -o, -o`
3. Same with `--help` flag showing twice: `-h, --help, --help`

tofumatt:
Addresses the issue in `web-ext`, and anything else including the
linter as a dep.

Whoever can r? first 😄

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,7 +1,64 @@
 import argv from 'yargs';
 
+import log from 'logger';
+import { singleLineString } from 'utils';
 import { version } from 'json!../package';
 
+
+export function getConfig({useCLI=true} = {}) {
+  if (useCLI === false) {
+    log.fatal(singleLineString`Config requested from CLI, but not in CLI mode.
+      Please supply a config instead of relying on the getConfig() call.`);
+    throw new Error('Cannot request config from CLI in library mode');
+  }
+
+  return argv
+    .usage(`Usage: ./$0 [options] addon-package-or-dir \n\n
+      Add-ons Linter (JS Edition) v${version}`)
+    .option('log-level', {
+      describe: 'The log-level to generate',
+      type: 'string',
+      default: 'fatal',
+      choices: ['fatal', 'error', 'warn', 'info', 'debug', 'trace'],
+    })
+    .option('output', {
+      alias: 'o',
+      describe: 'The type of output to generate',
+      type: 'string',
+      default: 'text',
+      choices: ['json', 'text'],
+    })
+    .option('metadata', {
+      describe: 'Output only metadata as JSON',
+      type: 'boolean',
+      default: 'false',
+    })
+    .option('pretty', {
+      describe: 'Prettify JSON output',
+      type: 'boolean',
+      default: false,
+    })
+    .option('stack', {
+      describe: 'Show stacktraces when errors are thrown',
+      type: 'boolean',
+      default: false,
+    })
+    .option('boring', {
+      describe: 'Disables colorful shell output',
+      type: 'boolean',
+      default: false,
+    })
+    .option('self-hosted', {
+      describe: 'Disables messages related to hosting on addons.mozilla.org.',
+      type: 'boolean',
+      default: false,
+    })
+    // Require one non-option.
+    .demand(1)
+    .help('help')
+    .alias('h', 'help')
+    .wrap(terminalWidth());
+}
 
 export function terminalWidth(_process=process) {
   if (_process && _process.stdout && _process.stdout.columns > 0) {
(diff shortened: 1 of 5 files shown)


src/main.js
```diff
@@ -1,12 +1,21 @@
-import cli from 'cli';
 import Linter from 'linter';
 import log from 'logger';
 
 import 'babel-polyfill';
 
+// Only include the CLI module if this is being run as a command line script.
+// Otherwise we trample over other apps using yargs. See:
+// https://github.com/mozilla/addons-linter/issues/735
+var cli = {argv: {logLevel: 'fatal'}};
```
muffinresearch: Still using var? 😄

muffinresearch:
r+wc lgtm

src/main.js
pdehaan: FWIW, I tried adding [`"no-var": "warn"`](http://eslint.org/docs/rules/no-var) to the .eslintrc and then running `$ grunt eslint` and got:

> ✖ 916 problems (0 errors, 916 warnings)

Similarly, if I add [`"prefer-const": "warn"`](http://eslint.org/docs/rules/prefer-const), I get:

> ✖ 91 problems (0 errors, 91 warnings)

The good news is that both those rules seem auto-fixable by newer ESLint versions, so the change isn't that difficult to quickly make. The bad news is, it's an ugly diff and touches virtually every file (plus we'll have a few errors with the `max-len` rule, since changing from `var` to `const` adds 2 characters, which means you'll either need to re-break some lines to be shorter, or tweak the `max-len` rule to accept 82 char lines.

kumar303:
I think we should avoid dynamic requires whenever possible. I think you could fix it with a bit of refactoring. Here is my suggestion:

Make the `argv` more lazy. Add something like this to `cli.js`:

````javascript
export function getConfig() {
  return argv.usage('...').option('...');
}
````

Make the `createInstance()` default config more lazy in `main.js`:
````javascript
export function createInstance({config=cli.getConfig(), runAsBinary=false} = {}) {
  // ...
}
````

File: src/main.js (hunk 1 above)
Comment (tofumatt) [inline src/main.js]: Old habits die hard!

**Comment by tofumatt** [comment]
@kumar303's solution is much nicer, I was being lazy.

Okay, I'll refactor things that way. I'll also update dispensary with a similar fix as it's doing the same thing.

**Comment by tofumatt** [comment]
Okay, changed this around and refactored something like what @kumar303 suggested. Ready for another r?

File: tests/test.main.js [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+
+describe('Main module tests', function() {
+
+  it('should not use yargs when not called from CLI', () => {
+    assert.doesNotThrow(() => {
+      createInstance({
+        // Passing `null` as a second argument means we don't supply
+        // a valid `argv` module, so this will fail if the code tries to
+        // call `argv`.
+        config: getConfig(false, null).argv,
```
kumar303: it's my personal preference but I would much rather see `getConfig({useCLI=false, _argv: null})`. You have to type some extra characters but I think it makes things easier to read and easier to expand upon in the future.

src/main.js
```diff
@@ -1,12 +1,21 @@
...
 import 'babel-polyfill';
 
-export function createInstance({config=cli.argv, runAsBinary=false} = {}) {
+
+export function isRunFromCLI(_module=module) {
+  return require.main === _module;
+}
+
+export function createInstance({
+  config=getConfig(isRunFromCLI()).argv, runAsBinary=false,
```
kumar303: what was your reasoning behind adding `isRunFromCLI()`? For added convenience?

In the case of `web-ext` it's not necessary because we will make the call like this:

````
const linter = createInstance({config: ourLocalConfig});
````

This would mean `getConfig()` never gets called and yargs options are never added. If someone were to not pass in their own config then I just don't see how this is useful. They would be stuck with a dummy config that does nothing.

Comment (tofumatt) [inline src/main.js]: It's for testing largely; coverage reporters will complain if we don't check the logic we use to check that we are running as the top-level module.

**Comment by tofumatt** [comment]
I changed to named arguments and answered the question re: `isRunFromCLI()`, let me know if that makes sense and if there are other issues. Hopefully I'm going about this right. This kind of contextual module stuff already seems weird to me in node, doubly so in ES6.

File: src/cli.js [hunk 4]
```diff
@@ -3,6 +3,64 @@ import argv from 'yargs';
 import { version } from 'json!../package';
 
 
+export function getConfig({useCLI=true, _argv=argv} = {}) {
+  if (useCLI === false) {
```
kumar303: I get how this is only used for testing but I think it could lead to confusion for a caller that is embedding the linter as a library. It would be confusing because their config could be incomplete. It's not a big deal but maybe logging a warning here would help -- something like "a dummy config was generated, you may need to specify a real config." Ideally I'd like to see it throw an error but I guess that won't let you get test coverage. Anyway, just a suggestion.

kumar303:
r+wc

src/cli.js
tofumatt: Good point. For now I'll at least log the config message.

PR #880, opened 2016-08-05
Don't use babel-polyfill in tests (fix #879)

### LINKED ISSUE #879: Don't polyfill in the tests
This also fixes the fact that we're slightly off from 100% coverage now because of https://github.com/mozilla/addons-linter/blob/master/src/main.js#L5

tofumatt:
r?


tofumatt:
Actually, no, this isn't right.

PR #896, opened 2016-08-11
Scan all JSON files (close #779)

### LINKED ISSUE #779: Scan all JSON files, not just manifest.json
**Steps to reproduce:**
Submit the attached webextension:
[iCloud-Bookmarks_v1.4.20.zip](https://github.com/mozilla/addons-linter/files/331481/iCloud-Bookmarks_v1.4.20.zip)

**Expected results:**
A validation error appears explaining that a json file contains block comments and they  should be removed or replaced with “//”.

**Actual results:**
The webextension passes the validation without any error, but it cannot be installed because appears to be corrupted.

**Additional notes:**
- This issue reproduces on AMO-dev, AMO-stage and AMO-production using Firefox 50.0a1 (2016-06-23) under Windows 10 64-bit.
- After submitting and installing the webextension, several errors appear in Browser Console such as: 1466754498299	addons.webextension.<unknown>	ERROR	Loading extension 'null': Loading locale file _locales/nl/messages.json: SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data     Log.jsm:753

tofumatt:
This turned out to be a lot more work than I thought it'd be, because of how we were scanning manifest.json files.

Now we scan all JSON files, so anything that would throw Firefox for a loop should be caught by the linter.

### DIFF
--- a/src/parsers/json.js
+++ b/src/parsers/json.js
@@ -1,0 +1,110 @@
+import esprima from 'esprima';
+import RJSON from 'relaxed-json';
+
+import * as messages from 'messages';
+
+export default class JSONParser {
+
+  constructor(jsonString, collector, {filename=null}={}) {
+    // Add the JSON string to the object; we'll use this for testing.
+    this._jsonString = jsonString;
+
+    // Provides ability to directly add messages to
+    // the collector.
+    this.collector = collector;
+
+    // Set the filename for this file
+    this.filename = filename;
+
+    // This marks whether a JSON file is valid; in the case of the base JSON
+    // parser, that's just whether it can be parsed and has duplicate keys.
+    this.isValid = null;
+  }
+
+  parse(RelaxedJSON=RJSON) {
+    try {
+      this.parsedJSON = JSON.parse(this._jsonString);
+    } catch (originalError) {
+      // First we'll try to remove comments with esprima;
+      // WebExtension manifests can contain comments, so we'll strip
+      // them out and see if we can parse the JSON.
+      // If not it's just garbage JSON and we error.
+      //
+      // Originally from https://github.com/abarreir/crx2ff/blob/d2b882056f902d751ad05e329efda7eddcb9d268/libs/ext-converter.js#L19-L37
+      var manifestString = `var o = ${this._jsonString}`;
+      try {
+        // This converts the JSON into a real JS object, and removes any
+        // comments from the JS code.
+        // This has some drawbacks because JSON and JS are not _100%_
+        // compatible. This is largely to do with Unicode characters we
+        // wouldn't expect to see in manifests anyway, and it should simply be
+        // a JSON parse error anyway.
+        // See:
+        // http://stackoverflow.com/questions/23752156/are-all-json-objects-also-valid-javascript-objects/23753148#23753148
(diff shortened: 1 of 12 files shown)


src/parsers/manifestjson.js
```diff
@@ -185,14 +125,12 @@ export default class ManifestJSONParser {
...
       this.collector.addError(messages.MANIFEST_UPDATE_URL);
-      isValid = false;
+      this.isValid = false;
     }
 
     if (isToolkitVersionString(this.parsedJSON.version)) {
       this.collector.addNotice(messages.PROP_VERSION_TOOLKIT_ONLY);
     }
-
-    return isValid;
```
muffinresearch: Presumably nothing was using this return value?

src/parsers/json.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // Provides ability to directly add messages to
+    // the collector.
+    this.collector = collector;
+
+    // Set the filename for this file
+    this.filename = filename;
+
+    // Any time parsing or validation fails this is set to `false`, so we start
+    // with the assumption it's `true`.
```
muffinresearch: But it's null :)

tests/parsers/test.json.js
```diff
@@ -0,0 +1,257 @@
...
+
+  it("doesn't evaluate JS code even though esprima is used", () => {
+    var addonLinter = new Linter({_: ['bar']});
+    var json = [
+      '{',
+      '// Required',
+      '"manifest_version": 2,',
+      '"name": "My Extension",',
+      '// Make the hell sure to use semvar.org if increasing this',
+      '"version": eval("alert(\'uh-oh\')")',
```
muffinresearch: I see this is a moved test and not new, but if the alert was fired, would it fail the test?

src/parsers/manifestjson.js
```diff
@@ -1,103 +1,43 @@
...
 import { isToolkitVersionString } from 'schema/formats';
 import { singleLineString } from 'utils';
 
-export default class ManifestJSONParser {
+
+export default class ManifestJSONParser extends JSONParser {
 
   constructor(jsonString, collector, {
-    rJSON=RJSON, selfHosted=getConfig().argv.selfHosted,
+    filename=MANIFEST_JSON, rJSON=RJSON, selfHosted=getConfig().argv.selfHosted,
```
muffinresearch: I wonder if this should be `relaxedJSON` instead of `rJSON` to make it clearer what it represents.

muffinresearch:
r+wc

src/parsers/manifestjson.js
tofumatt: Nope. Or not any more; it was a bit convoluted, where `ManifestJSONScanner.isValid` would be set but then also set using this... I just tidied it up.

src/parsers/json.js
tofumatt: Hah, thanks. I improved it but forgot to fix the comment 😄

tests/parsers/test.json.js
tofumatt: That's a good question... that should be a separate issue, and I think it's fine, but we could investigate further.

PR #924, opened 2016-09-01
prevent asterix in min_version

### LINKED ISSUE #914: Prevent use of * in strict_min_version
In https://bugzilla.mozilla.org/show_bug.cgi?id=1297752 its pointed out that usage of * in strict_min_version is confusing lots of people (48.* is not compatible with 48.0) and we should do our best to prevent this from happening until we come up with a better plan.

andymckay:
* fixes #914
* doesn't allow * for min version, but does for max version
* ...also @rpl

### DIFF
--- a/src/schema/manifest-schema.json
+++ b/src/schema/manifest-schema.json
@@ -17,12 +17,14 @@
                         "strict_min_version": {
                             "default": "42a1",
                             "description": "Minimum version of Gecko to support. Defaults to '42a1'. (Requires Gecko 45)",
-                            "type": "string"
+                            "type": "string",
+                            "pattern": "^[0-9]{1,3}(\\.[a-z0-9]+)+$"
                         },
                         "strict_max_version": {
                             "default": "*",
                             "description": "Maximum version of Gecko to support. Defaults to '*'. (Requires Gecko 45)",
-                            "type": "string"
+                            "type": "string",
+                            "pattern": "^[0-9]{1,3}(\\.[a-z0-9*]+)+$"
                         },
                         "update_url": {
                             "description": "Link to an add-on update manifest. (Requires Gecko 45)",
--- a/tests/schema/test.applications.js
+++ b/tests/schema/test.applications.js
@@ -62,6 +62,39 @@ describe('/applications/gecko/*', () => {
       '/applications/gecko/strict_max_version');
   });
 
+  // For the following tests I copied versions from:
+  // https://addons.mozilla.org/en-US/firefox/pages/appversions/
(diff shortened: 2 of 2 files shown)


tofumatt:
r+

rpl:
:+1:

src/schema/manifest-schema.json
```diff
@@ -17,12 +17,14 @@
...
-                            "type": "string"
+                            "type": "string",
+                            "pattern": "^[0-9]{1,3}(.[a-z0-9]+)+$"
                         },
                         "strict_max_version": {
                             "default": "*",
                             "description": "Maximum version of Gecko to support. Defaults to '*'. (Requires Gecko 45)",
-                            "type": "string"
+                            "type": "string",
+                            "pattern": "^[0-9]{1,3}(.[a-z0-9*]+)+$"
```
muffinresearch: Is the `.` correct or should that be escaped? This would allow `9#foo` for example.

tofumatt: Oh shoot, good catch. It should be escaped and there should be a test for it.

rpl: yeah, @muffinresearch is definitely right, the `.` is supposed to be escaped (and an another "invalid version number" test case should be added to the test, e.g. something similar to the above invalid example `40#0a2`)

src/schema/manifest-schema.json
```diff
@@ -17,12 +17,14 @@
                         "strict_min_version": {
                             "default": "42a1",
                             "description": "Minimum version of Gecko to support. Defaults to '42a1'. (Requires Gecko 45)",
-                            "type": "string"
+                            "type": "string",
+                            "pattern": "^[0-9]{1,3}(.[a-z0-9]+)+$"
```
muffinresearch: Same thing here with the `.`

src/schema/manifest-schema.json
```diff
@@ -18,13 +18,13 @@
                             "default": "42a1",
                             "description": "Minimum version of Gecko to support. Defaults to '42a1'. (Requires Gecko 45)",
                             "type": "string",
-                            "pattern": "^[0-9]{1,3}(.[a-z0-9]+)+$"
+                            "pattern": "^[0-9]{1,3}(\\.[a-z0-9]+)+$"
```
tofumatt: I think it should just be the one `\`, not two.

andymckay: It's in JSON where \ is a valid escape character, so you have to do \\ in the regex, which is annoying.

tofumatt: Ah okay. Figured it might've been the case but just wanted to make sure.

tofumatt:
r+

PR #925, opened 2016-09-01
allow homepage_url to be localised

### LINKED ISSUE #836: [WebExtensions] Localized `homepage_url` in manifest is wrongly considered invalid
I have localized my `homepage_url` manifest parameter by using `__MSG_addon_url__` as value and providing its locale-specific value via the message with the `addon_url` key in files `en/messages.json` and `ru/messages.json` in the `_locales` folder of my [extension](https://addons.mozilla.org/en-us/firefox/addon/consistent-https/) written using the WebExtensions API.

But I was **not** able to publish the new version just because the validation script of Addons.Mozilla.org wrongly treats such localized URL as invalid since the validator have apparently no clue that it’s actually not the URL itself, but a _reference_ to the corresponding _localized string_.

So I’ve been forced to replace the `homepage_url`-parameter value back to the real URL (hard-coded english one instead of own URL for each locale) for now. Thanks.

andymckay:
* fixes #836 
* until we can get the linter to understand localisations, allow to be a URI a localised string as per the docs https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Internationalizing_manifest.json

### DIFF
--- a/src/schema/manifest-schema.json
+++ b/src/schema/manifest-schema.json
@@ -42,7 +42,16 @@
         "homepage_url": {
             "description": "The URL of the homepage for this add-on. The add-on management page will contain a link to this URL.",
             "type": "string",
-            "format": "uri"
+            "oneOf": [
+                {
+                    "type": "string",
+                    "format": "uri"
+                },
+                {
+                    "type": "string",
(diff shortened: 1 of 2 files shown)


tofumatt:
r+

PR #976, opened 2016-10-05
Don't swallow the detailed error description on invalid JSON.

EnTeQuAk:
The problem here is that `error` is an actual `SyntaxError` instance
and the json export expects things to be strings which leads to
swallowing the actual message.

Fixes mozilla/addons-server#3385


tofumatt:
Huzzah, you fixed the test failure before I could finish my comment. r+ to the change too 👍

PR #1005, opened 2016-10-21
Added new manifest permissions

wbamberg:
Added "clipboardWrite" and "history" permissions.

Without this change, the linter generates a warning for "history-deleter":

    Unknown permissions "history"

With this change, history-deleter passes with 0 errors and 0 warnings.


tofumatt:
Fixes #1006.

Issue #1014, opened 2016-10-26
Add config option to treat warnings as errors


josteink:
### Describe the problem and steps to reproduce it:

First create a web-extension `manifest.json` with invalid permissions (like `"tbas"`). Then run `web-ext lint`.
### What happened?

`web-ext lint` produced a warning, but still exited with return code `0`. 
### What did you expect to happen?

Have ability to add config-flag to treat warnings like errors.

This would basically mean rewriting [the following code](https://github.com/mozilla/addons-linter/blob/master/src/linter.js#L428-L430):

``` javascript
if (this.config.runAsBinary === true) {
  process.exit(this.output.errors.length > 0 ? 1 : 0);
}
```

Into:

``` javascript
if (this.config.runAsBinary === true) {
  var exitcode = this.output.errors.length > 0 ? 1 : 0;
  if (exitcode === 0 && this.config.warningAsError === true) {
    exitcode = this.output.warnings.length > 0 ? 1 : 0;
  }
  process.exit(exitcode);
}
```

(Probably associated documentation needs to be updated as well, but that's the core code changes required.)

PR #1016, opened 2016-10-26
Add support for warningsAsErrors config-option.

josteink:
This partially resolves https://github.com/mozilla/addons-linter/issues/1014.


src/linter.js
```diff
@@ -426,7 +426,11 @@ export default class Linter {
         // test runs against un-instrumented code.
         /* istanbul ignore if  */
         if (this.config.runAsBinary === true) {
-          process.exit(this.output.errors.length > 0 ? 1 : 0);
+          var exitCode = this.output.errors.length > 0 ? 1 : 0;
```
tofumatt: A nitpick, but we should use `let` here 😄

src/linter.js
```diff
@@ -426,7 +426,11 @@ export default class Linter {
...
+          if (exitCode === 0 && this.config.warningsAsErrors === true) {
+            exitCode = this.output.warnings.length > 0 ? 1 : 0;
```
tofumatt: I think this code might not be checked by our code coverage tool, but could you add a CLI test (I think there are existing ones for guidance) to make sure this works as it should? (Though the code looks fine.)

src/linter.js
josteink: Agreed. My bad.

josteink: Agreed. My bad.

src/linter.js
josteink: Sounds reasonable. I'll see what I can get done.

josteink:
And there I think I should have addressed all concerns raised so far. Anything else you think needs fixing?

josteink:
@tofumatt Any news on this one?

tofumatt:
It's on my radar but I'm travelling/attending a conference and doing some AMO work right now, so I'll need to get to it after I clear some of that stuff up. Might not be able to review until Tuesday, but I'll try to find an extra moment! 😄

josteink:
Gotcha. And understood.

That said, I have a PR for `web-ext` which depends on this getting merged (which I need for a AMO-related production pipeline I'm building at work), so any time you can find is greatly appreciated :)

tofumatt:
Thanks for your patience, all merged; I'll do a release now as well so you can pull the newest version from npm.

josteink:
Awesome. Thanks for taking in this patch :)

atsay:
Thanks for the patch, Jostein! We've added this contribution to our [recognition page](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#October_2016).

Issue #1023, opened 2016-10-30
web-ext lint error: Too many non-option arguments: got 1, maximum of 0


kumar303:
### Describe the problem and steps to reproduce it:

To reproduce:
- upgrade the `addons-linter` dependency of `web-ext` to `0.15.7` (or higher)
- run `npm install` and `npm run build` to get a new `web-ext`
- cd into any add-on
- type `web-ext lint`
### What happened?

It shows a usage message with this error:

```
Too many non-option arguments: got 1, maximum of 0
```
### What did you expect to happen?

The command should show a lint report
### Anything else we should know?

`web-ext` simulates passing a directory of source as the first command line argument to `addons-linter` [like this](https://github.com/mozilla/web-ext/blob/master/src/cmd/lint.js#L75). Maybe something changed? I don't see anything obvious at a glance.

The mystery saga of https://github.com/mozilla/addons-linter/issues/969 continues! It was definitely fixed but now it's broken again. See https://github.com/mozilla/web-ext/pull/606 and https://github.com/mozilla/web-ext/pull/616

muffinresearch:
Looks like Yargs was updated 6 days ago: https://github.com/mozilla/addons-linter/pull/1003 and that was included in the 0.15.7 release.

josteink:
I'm having this error too. It's a bit of a bummer too, because with this, I finally planned to land the warnings-as-errors support in web-ext linting.

muffinresearch:
@kumar303 Running web-ext lint with a version of web-ext where the tests are failing, `web-ext lint --source-dir [SOURCE]` runs correctly so I think this might be an issue with how the cli tests in web-ext are being run and less a bug in addons-linter.

kumar303:
@muffinresearch Did you rebuild `web-ext` after upgrading its linter dep? I can still reproduce it. This is how I do it:
````
cd web-ext
npm install
npm run build
web-ext lint -s ~/dev/webextensions-examples/beastify/
# No errors
npm install addons-linter@0.15.9
npm run build
web-ext lint -s ~/dev/webextensions-examples/beastify/
# Too many non-option arguments: got 1, maximum of 0
````

muffinresearch:
@kumar303 It's working for me but the test still fails. What node / npm versions are you using?

josteink:
I can reproduce this error too and I'm using v4.6.1 (default on Fedora 24) and v4.2.6 (Ubuntu 16.04).

[...]

PR #1037, opened 2016-11-07
Prevent the linter from parsing yargs options

kumar303:
Fixes https://github.com/mozilla/addons-linter/issues/1023

The manifest parser needs to always pass the `selfHosted` flag from its own config. This will prevent the manifest parser class from re-parsing the command line and thus colliding with web-ext's yarg parser.


muffinresearch:
r+ if travis is green

josteink:
While this is great we also need to bump the version and publish a new package. 

Do any of you guys have the capability to do that?

muffinresearch:
> While this is great we also need to bump the version and publish a new package.

> Do any of you guys have the capability to do that?

0.15.10 is on its way to npm

PR #1042, opened 2016-11-15
Drop node 5 build

wagnerand:


wagnerand:
Heh, I can add it, but it *will* break tests...

muffinresearch:
We should discuss this further - I'm not convinced we should be running tests and agreeing to support non LTS node builds. If contributors want to run this code in newer versions and something breaks and we can fix it without breaking support for LTS then we could accept a PR but I don't think we should actively pursue support for non-LTS on an ongoing basis.

tofumatt:
Fair enough. I figured users who would run this locally might be on newer
versions of node, and if supporting them is easy I’m for it.

But I suppose I don’t want to put much time into it right now, so I
shouldn’t have asked for it ;-)

muffinresearch:
> But I suppose I don’t want to put much time into it right now, so I shouldn’t have asked for it ;-)

Yep - since node 6 is pretty new I think we can stick to just LTS for now and talk about it a bit more at the all-hands if not before. Plus of course, whatever we do now we can always revisit and change later if it makes sense to increase support.

wagnerand:
...revert **all** the PRs....

PR #1060, opened 2016-12-10
fix #1059: add WebExtensions Experiment type

### LINKED ISSUE #1059: Add WebExtension Experiment type to ADDON_TYPE_MAP
1. Add `<em:type>256</em:type>` to `install.rdf`
2. Run `addons-linter` on the resulting `.xpi` file

### What happened?
```
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code               Message                   Description                                                                            File          Line   Column
RDF_TYPE_INVALID   Invalid <em:type> value   The only valid values for <em:type> are 2, 4, 8, and 32. Any other values are either   install.rdf
                                             invalid or deprecated.
```

No errors

MDN Doc: https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#type

groovecoder:


andymckay:
Hmm, currently the linter does WebExtension add-ons only though right? Anything that's not a WebExtension should fail because it goes through the addons-validator? The original idea was lint everything in the linter, just wondering if it makes sense to double down on the linter only being for WebExtensions and even barf on having an `install.rdf` at all.

groovecoder:
That's a good point @andymckay. I originally ran this `addons-linter` against my `.xpi` because I was getting errors from it. Even on this patch, the errors remained. Then I realized I was trying to run my `.xpi` on Beta version and it required Aurora. So this particular issue was a red herring in my case.

That being said, this addons-linter is included as a requirement in the shield addon template, which can be SDK add-ons too.

But it's your call - I'm happy to close this and/or the corresponding issue.

EnTeQuAk:
> That being said, this addons-linter is included as a requirement in the shield addon template, which can be SDK add-ons too.

Hmm, the problem is that we simply do not have any proper support for sdk add-ons in the linter compared to what we have in amo-validator. There are some very basic rules afair but nothing compared to what the validator has. It'd be best to recommend the use of the validator for non-webextensions and the linter for webextensions.

Given the general direction of WebExtensions I'd be +1 on doubling down on using the linter for WebExtensions and maybe even error out with a clear "This is a legacy-style add-on. Please use the amo-validator" message or so including a link with more directions.

tofumatt:
Indeed, we should do that and eventually remove any code in the linter that
deals with legacy add-ons; I've meant to do it for some time but it hasn't
been that high priority. Now that it's confusing others seems a bit more
important. (I think there's even an issue open for it but I'm writing this
via email and will have to check later 😅)

Starting with the "legacy add-on; bailing" is a great idea though 👍🏻

- tofumatt

groovecoder:
Sounds good to me. Closing.

PR #1068, opened 2016-12-13
error on missing locale

### LINKED ISSUE #989: Show a warning for "default_locale" with no correspondent in _locales folder
Steps to reproduce:
1. Submit a web-extension containing a `default_value` set in the manifest.json that has no correspondent in **_locales** folder

Expected results:
A warning is displayed at validation.

Actual results:
No validation errors. User can make it to step 3 of submission process, where **is blocked** because it has no name field and cannot go forward because that field is mandatory.

Notes/Issues:
Verified on FF49(Win 7). Issue is reproducing all around AMO servers.
Screenshots for this issue: 
ok
submit
required

andymckay:
* fixes #989

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -235,7 +235,7 @@ export default class Linter {
               var manifestParser = new ManifestJSONParser(
                 json,
                 this.collector,
-                {selfHosted: this.config.selfHosted},
+                {selfHosted: this.config.selfHosted, io: this.io},
               );
               return manifestParser.getMetadata();
             });
@@ -250,7 +250,7 @@ export default class Linter {
       .then((addonMetadata) => {
         this.addonMetadata = addonMetadata;
 
-        // The type must be explcitly defined. This behaviour differs the
+        // The type must be explicitly defined. This behaviour differs the
         // historical approach by the amo-validator.
         // See mozilla/addons-linter#411.
         // In due course metadata checking code may surpass this error
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -107,3 +107,13 @@ export function manifestPropMissing(property) {
 
 export const PROP_NAME_MISSING = manifestPropMissing('name');
 export const PROP_VERSION_MISSING = manifestPropMissing('version');
+
+export const NO_MESSAGES_FILE = {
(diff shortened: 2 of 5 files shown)


src/messages/manifestjson.js
```diff
@@ -107,3 +107,13 @@ export function manifestPropMissing(property) {
 
 export const PROP_NAME_MISSING = manifestPropMissing('name');
 export const PROP_VERSION_MISSING = manifestPropMissing('version');
+
+export const NO_MESSAGES_FILE = {
+  code: 'NO_MESSAGES_FILE',
+  legacyCode: null,
+  message: _('The "default_locale" is missing localisations.'),
```
tofumatt: I think the Mozilla style guide says to use American English, at least as the default locale.

src/linter.js
```diff
@@ -235,7 +235,7 @@ export default class Linter {
               var manifestParser = new ManifestJSONParser(
                 json,
                 this.collector,
-                {selfHosted: this.config.selfHosted},
+                {selfHosted: this.config.selfHosted, files:this.io},
```
tofumatt: `files` seems like the wrong variable name here; I'd expect an array of files, but this is an `IO` object.

There really isn't a great variable name I can think of, so I'd say rename this to at least `io` so it's easier to trace through.

src/linter.js
```diff
@@ -250,7 +250,7 @@ export default class Linter {
       .then((addonMetadata) => {
         this.addonMetadata = addonMetadata;
 
-        // The type must be explcitly defined. This behaviour differs the
+        // The type must be explicitly defined. This behaviour differs the
```
tofumatt: Thanks! 😄

src/linter.js
tofumatt: Oh and very minor nitpick but we've been doing `files: this.io` (note the space after colon).

We should have stylelint rules for it but we seem to lack them. I've filed #1069 for that.

src/messages/manifestjson.js
andymckay: Grr 😄

src/linter.js
andymckay: Fair enough.

PR #1078, opened 2016-12-15
complain if _locales specified but no default_locale

### LINKED ISSUE #701: Validate `default_locale` is being set if _locales/ folder exists
Supports https://github.com/mozilla/addons-server/issues/1681

From https://developer.chrome.com/extensions/i18n:

> Important: If an extension has a _locales directory, the manifest must define "default_locale"

andymckay:
* fixes #701 which is kind of the inverse of the last one.
* not sure how to do this without the loop, but it should be pretty damn fast on anything but the biggest add-ons

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -117,3 +117,13 @@ export const NO_MESSAGES_FILE = {
     exists. See: https://mzl.la/2hjcaEE`),
   file: MANIFEST_JSON,
 };
+
+export const NO_DEFAULT_LOCALE = {
+  code: 'NO_DEFAULT_LOCALE',
+  legacyCode: null,
+  message: _('The "default_locale" is missing but "_locales" exist.'),
+  description: _(singleLineString`The "default_locale" value is not specifed in
+    the manifest, but a "_locales" directory exists.
+    See: https://mzl.la/2hjcaEE`),
+  file: MANIFEST_JSON,
+};
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -140,6 +140,17 @@ export default class ManifestJSONParser extends JSONParser {
         this.isValid = false;
       }
     }
+
+    if (!this.parsedJSON.default_locale && this.io) {
+      let match_re = /^_locales\/.*?\/messages.json$/;
+      for (let filePath in this.io.files) {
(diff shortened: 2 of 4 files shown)


src/parsers/manifestjson.js
```diff
@@ -140,6 +140,16 @@ export default class ManifestJSONParser extends JSONParser {
         this.isValid = false;
       }
     }
+
+    if (!this.parsedJSON.default_locale && this.io) {
+      for (let key in this.io.files) {
```
tofumatt: `filePath` would be a more appropriate variable name.

src/parsers/manifestjson.js
```diff
@@ -140,6 +140,16 @@ export default class ManifestJSONParser extends JSONParser {
...
+        if (key.startsWith('_locales')) {
```
tofumatt: Like you said I'm not a huge fan of the loop but our IO class doesn't yet have a usable `checkIfPathExists` type method. I've filed #1079 for it, though it's obviously low-priority.

andymckay:
I'm not sure where AMO will have a problem, but @EnTeQuAk might no more. Let's assume that having a `messages.json` inside `_locales` is the problem, but not an empty `_locales`. I've updated it to be a regex and look for `messages.json`. Because its a regex added more tests. How does that look?

src/messages/manifestjson.js
```diff
@@ -117,3 +117,13 @@ export const NO_MESSAGES_FILE = {
     exists. See: https://mzl.la/2hjcaEE`),
   file: MANIFEST_JSON,
 };
+
+export const NO_DEFAULT_LOCALE = {
+  code: 'NO_DEFAULT_LOCALE',
+  legacyCode: null,
+  message: _('The "default_locale" is missing but "_locales" exist.'),
+  description: _(singleLineString`The "default_locale" value is not specifed in
+    the manifest, but a "_locales" directory exists.
```
tofumatt: I would say `a "locales" directory with "messages.json" files`, just so they know exactly what to look for.

andymckay:
It wont catch `_locales/messages.json` because there has to be two slashes in there so `_locales//messages.json` would fail, I added a test for that. But lets face it, if we've got files like that, things are messed up somewhere.

tofumatt:
Agreed.

Cool, looks good to me! :-)

PR #1088, opened 2017-01-03
fix: Check for parent

tofumatt:
fix https://github.com/mozilla/addons/issues/303


tofumatt:
@netquik We tag every Tuesday and push every Thursday, so normally it would be live next Thursday as we missed this week's tag. Is that okay?

netquik:
@tofumatt Yes Sure many thanks... looking forward for it.. thanks again

Issue #1091, opened 2017-01-06
Logs error if async/await is used in JavaScript


asamuzaK:
### Describe the problem and steps to reproduce it:
Initially filed in [web-ext lint logs error if async/await is used · Issue #740 · mozilla/web-ext](https://github.com/mozilla/web-ext/issues/740).

When runs `web-ext lint`, in Validation Summary, JS SYNTAX ERROR is logged.

1. save attached file
2. run `web-ext lint`
3. in Validation Summary, JS SYNTAX ERROR is logged.

### What happened?
JS SYNTAX ERROR is logged if async/await is used in JavaScript file.

### What did you expect to happen?
No validation error.

### Anything else we should know?
[webExtTest.zip](https://github.com/mozilla/addons-linter/files/690380/webExtTest.zip)

kumar303:
In step 2, you could just do: `addons-linter .`

kumar303:
async/await landed Oct 2016: https://blog.nightly.mozilla.org/2016/11/01/async-await-support-in-firefox/

muffinresearch:
Looks like additional options are required to enable later ES features in eslint https://github.com/eslint/eslint/issues/1657#issuecomment-252500711

asamuzaK:
If this option is added to .eslintrc of addons-linter, will it be effective even on web-ext?
Or should this option be added to web-ext .eslintrc?
Or both?

tofumatt:
web-ext lint is using addons-linter so we just need to tell the linter what
to do. :-)

- tofumatt

asamuzaK:
So, add
```
  "parserOptions": {
    "ecmaVersion": 2017
  },
```
to [web-ext/.eslintrc at master · mozilla/web-ext](https://github.com/mozilla/web-ext/blob/master/.eslintrc) will fix the issue?

muffinresearch:
@asamuzaK I think we need a modification here: https://github.com/mozilla/addons-linter/blob/b301dda5e0c45799d6dc13e08de9227e50d65cfb/src/scanners/javascript.js#L44 since we're using the eslint api directly. The .eslintrc in our projects is intended to configure linting the code in the repository.

asamuzaK:
Oh, thanks.
Maybe insert this line below `env: {}` will fix?
```
parserOptions: { ecmaVersion: 2017 },
```
If so, I'll file a PR later.

muffinresearch:
@asamuzaK Yep, should do it.

PR #1095, opened 2017-01-13
Add parserOptions in eslint options

asamuzaK:
Fix [#1091](https://github.com/mozilla/addons-linter/issues/1091)

### DIFF
--- a/tests/scanners/test.javascript.js
+++ b/tests/scanners/test.javascript.js
@@ -30,7 +30,7 @@ describe('JavaScript Scanner', function() {
   });
 
   it('should not throw when async/await is used', () => {
-    var code = singleLineString`var foo = async a => a;`;
+    var code = 'var foo = async a => a;';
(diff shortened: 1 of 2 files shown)


tests/scanners/test.javascript.js
```diff
@@ -29,6 +29,16 @@ describe('JavaScript Scanner', function() {
     assert.equal(jsScannerWithOptions.options.foo, 'bar');
   });
 
+  it('should not throw when async/await is used', () => {
+    var code = singleLineString`var foo = async a => a;`;
```
tofumatt: Because the code is not over 80 lines, so you don't need to use `singleLineString` here, and regular single quotes (`'`) should be fine. 😄

asamuzaK: Fixed

asamuzaK:
Thanks!

Issue #571, opened 2016-03-17
Make it possible to run a single test from the test-suite


muffinresearch:
Either via an option passed along with grunt test or a watch-based test runner (which is often a good solution since changes would require a re-build via webpack.)

EnTeQuAk:
Let's close this in favor of #413

Issue #242, opened 2015-11-12
Add support for validating a single file


muffinresearch:
This would enable things like validation of a single file via a text editor.

EnTeQuAk:
This is a duplicate of https://github.com/mozilla/addons-linter/issues/413, let's keep discussion there.

PR #1100, opened 2017-01-16
feat: Added a --scan-file option to scan a single file

### LINKED ISSUE #413: Add option to validate single file
Main use case for this is allowing a linting tool (eg. Atom, Sublime, etc.) to validate a single file based on file type.

So users could do: `addons-linter --file foo/bar.js --type js`. Or even take `STDIN`.

rpl:
This PR defines a new `--scan-file` (or the shorter `-f`) cli option to optionally scan a single file.

This should fix  #413 (and it should help to achieve mozilla/web-ext#539, by creating a custom eslint plugin that is going to add the results of the addons-linter to the regular eslint warnings/errors).


tofumatt:
Whoa, neato!

src/cli.js
```diff
@@ -58,6 +58,19 @@ export function getConfig({useCLI=true} = {}) {
       type: 'boolean',
       default: false,
     })
+    .option('scan-file', {
+      alias: ['f'],
+      describe: 'Include only a single files in the scan',
```
tofumatt: Should be "single file".

Actually I'd change this to:

`describe: 'Scan a single file',`

src/cli.js
```diff
@@ -58,6 +58,19 @@ export function getConfig({useCLI=true} = {}) {
...
+      type: 'string',
+      requiresArg: true,
+      coerce: (value) => {
```
tofumatt: Huh, why is this here? Any harm in supplying it multiple times? (It's weird but harmless I'd think; I don't think we do checks like this elsewhere. If we should start, we should do this for anything that should only be supplied once.)

src/linter.js
```diff
@@ -415,12 +427,42 @@ export default class Linter {
       });
   }
 
+  get shouldScanFile() {
+    if (this.config.scanFile) {
+      const manifestFileNames = [
+        'install.rdf', 'manifest.json', 'package.json',
+      ];
+
+      return (fileOrDirName, isDir=false) => {
```
tofumatt: Can we make this a private method in the file somewhere? I just get surprised when I see inline functions returned as first-class objects as we don't do it elsewhere in the code often (if ever).

tests/test.collector.js
```diff
@@ -65,4 +65,31 @@ describe('Collector', function() {
     assert.equal(collection.notices.length, 0);
   });
 
+  it('should filter message by filename if config.scanFile is defined', () => {
```
tofumatt: Can you just test the actual code of the errors in the collection so we know the right ones are getting through? Thanks ^_^

rpl:
@tofumatt thanks for the super-fast review!

It should now be ready for a new round of review (I applied some changes, based of the previous review comment, and enabled it to work with multiple selected files).

rpl:
(and also added some additional test ;-))

rpl:
Hi @tofumatt 
I just pushed a new update on this PR:
it looks like that if the yargs coerce function is defined, yargs do not raise the expected error message when the option is used without the expected arguments, but normalizing `config.scanFile` (e.g. to ensure that it is always an Array and to filter out any undefined/empty strings items from it) in every place where we are going to use it... it is something that do not let me sleep well :-)

In this updated version I've added to the Linter class a new setter/getter for the config property, and so the scanFile normalization is now centralized in a single place (it happens once when we assign the property for the first time in the constructor). 

Let me know if this approach looks ok for you too (or, it it doesn't, we can have a brief chat about it and decide where we want to do it).

Thanks again for you help with this PR.

tests/test.collector.js
```diff
@@ -65,4 +65,40 @@ describe('Collector', function() {
...
+      scanFile: ['test.js', 'no-match-file.js'],
+    });
+
+    assert.equal(collection.length, 0);
+    collection.addError({
+      ...fakeMessageData,
+      // not file name
+    });
+    assert.equal(collection.length, 1);
+    assert.equal(collection.errors[0].type, 'error');
```
tofumatt: You shouldn't need to test the type here as things in the `errors` array are always `'error'` type (and the collection class tests that).

tests/test.collector.js
```diff
@@ -65,4 +65,40 @@ describe('Collector', function() {
...
+    });
+    assert.equal(collection.length, 1);
+
+    assert.equal(collection.length, 1);
+    collection.addError({
+      ...fakeMessageData,
+      file: 'test.js',
+    });
+    assert.equal(collection.length, 2);
+    assert.equal(collection.errors[1].type, 'error');
```
tofumatt: You might want to check the error `code` property matches the error you expect.

EnTeQuAk:
Does this also fix https://github.com/mozilla/addons-linter/pull/1100? Looks like it's a duplicate or at least related or the same to some degree. If so, let's close this one too once the PR is merged.

tofumatt:
> Does this also fix #1100? Looks like it's a duplicate or at least related or the same to some degree. If so, let's close this one too once the PR is merged.

This is #1100. Did you typo?

EnTeQuAk:
eh, yeah, this pull request is #1100, what I meant was https://github.com/mozilla/addons-linter/issues/242 https://github.com/mozilla/addons-linter/issues/571 which are duplicates of https://github.com/mozilla/addons-linter/issues/413 and closed now. Not to worry about it :)

EnTeQuAk:
@rpl do you have the time to follow-up on this pull request and get this merged some time?

rpl:
@EnTeQuAk sure, I'm going to update this asap (somehow I missed the last review comments notifications on this pullrequest from github)

rpl:
@tofumatt Thanks for the reviews, I've just tweaked the test case to check the error code as suggested.

@tofumatt @EnTeQuAk this should now be ready to go 
(but let me know if there is anything missing, or wrong, in the last updates on this pull request)

src/collector.js
```diff
@@ -22,6 +24,12 @@ export default class Collector {
   }
 
   _addMessage(type, opts, _Message=Message) {
+    if (this.config.scanFile && opts.file) {
+      if (!this.config.scanFile.some(v => v === opts.file)) {
+        return;
+      }
+    }
```
EnTeQuAk: I might be missing some context but I don't get this code, can we add a comment to why we skip here?

EnTeQuAk: Ah, reading more code... this is because files can be missing. This clearly needs a comment imho.

EnTeQuAk: Eh, actually, this is simply filtering messages by selected file-name. Still needs comments 😆

EnTeQuAk:
I'll leave the final review to @tofumatt but it looks good to me so far.

src/collector.js
rpl: :+1: agree, a new comment is coming ;-)

src/collector.js
```diff
@@ -22,6 +24,14 @@ export default class Collector {
   }
 
   _addMessage(type, opts, _Message=Message) {
+    // Filter the messages reported by file when the Linter has been configured
```
rpl: @EnTeQuAk how about this inline comment? does it help to grasp the role of this check without too much thinking?

tofumatt:
Oops, sorry I didn't reply sooner but r+ from me! @rpl you should have commit access so feel free to merge!

src/collector.js
EnTeQuAk: Looks good to me, thanks!

PR #1112, opened 2017-01-31
add in a naive and silly risk profile

andymckay:
try and guess a risk profile for an add-on, example:

https://gist.github.com/andymckay/242d478629790085c47627330c11d3db

this is naive and silly

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -201,13 +201,6 @@ export default class Linter {
     return output;
   }
 
-  getAddonRisk({
-    _log=log,
-    ManifestJSONParser=DefaultManifestJSONParser,
-  } = {}) {
-    return Pro
-  }
-
   getAddonMetadata({
     _log=log,
     ManifestJSONParser=DefaultManifestJSONParser,
--- a/src/linter.js
+++ b/src/linter.js
@@ -191,6 +191,7 @@ export default class Linter {
       count: this.collector.length,
       summary: {},
       metadata: this.addonMetadata,
+      risk: this.addonRisk,
     };
     for (let type of constants.MESSAGE_TYPES) {
       var messageType = `${type}s`;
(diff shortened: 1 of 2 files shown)


andymckay:
Do not merge this :)

tofumatt:
Neat idea though! 😄

pauljt:
One initial thought about this: should preserve the attributes that you use to calculate the risk? The numerical value won't be very useful to a reviewer, but if we preserve all the attributes it will make the system easier to audit and build upon? Ie maybe risk is an object like:

```
addon.risk =  {
 value: 0.4,
 vectors: {
     nativeMessaging: true,
     webRequest: false,
     webRequestBlocking: true,
     webNavigation: false,
     unlimitedStorage: true,
     ...
 }
```

Maybe something less cumbersome, but my gut tells me indexing this data might be useful for investigations.

muffinresearch:
I wonder if it would be useful to work out what's left to formalize the first cut of this patch, then expose it to the reviewers showing a message like this:  ***"If auto approvals were enabled this add-on would have qualified to be automatically approved"*** with a way to view the full data leading to the decision (I really like @pauljt suggestion that the criteria met is part of the object). 

We could also solicit feedback for when the risk metric seems wrong by setting up an email link with the full risk object being part of the body so that reviewers tell us when it something important is missed.

muffinresearch:
> Maybe something less cumbersome, but my gut tells me indexing this data might be useful for investigations.

It could also list the value given to each metric at the time since that could change as this gets tuned. 

I was considering leaving out the explicit false vectors but again since that could change over time too leaving them in for historical purposes would make sense - I'm assuming we would store this data.

Here's an example of that building on @pauljt's suggestion.

```
addon.risk =  {
 value: 0.7,
 flaggedVectors: {
     nativeMessaging: 0.5
     webRequest: false,
     webRequestBlocking: false,
     webNavigation: 0.2,
     unlimitedStorage: false,
 }
```

cr:
I think Paul is pushing in the right direction here: We should keep raw data and scan results where ever possible, – and add time stamps and scanner versions, too – to form an audit trail. Our algorithms will evolve over time and it would allow us to re-evaluate risk scoring in a dynamic fashion. The real work behind this will be in the weighting vector that turns the raw metrics into a singular score.

A risk score is great for allocating internal resources, but it may be just half of what we will need in the longer run. A trust score might be equally important. It should, for example, be much less of a problem to run high-risk, high-trust code than it should be to run high-risk, low-trust code.

Issue #275, opened 2015-11-16
rule: disallowed_file_type


tofumatt:
From https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L49

Looks for native files (JAR, etc.) and warns against their usage: https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L49

EnTeQuAk:
This is still relevant, we need a proper disallowed-files system to avoid people packaging up crap inside their zips.

andymckay:
Hmm I got a lot of this here: https://github.com/mozilla/addons-linter/blob/master/src/const.js#L123-L142 which is from #638

EnTeQuAk:
heh, that's true, sorry I didn't see this. I'm guessing this catches quite a lot of stuff already, it doesn't catch .jar files - not sure if we should though.

andymckay:
If I read that code in addons-validator it looks to see if there are more than 10 files that have the .class extension here: https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L96-L105 and then calls it a jar. It doesn't detect for jar files through its magic numbers (because its just a zip).

The linter just flags any add-on with a .class file anywhere in it. 

A simple addition would be to add in .jar as well into the flagged files.

andymckay:
We've got .jar, should we close this now @EnTeQuAk?

PR #1114, opened 2017-02-02
add in .jar

andymckay:
* supports #275


src/const.js
```diff
@@ -125,7 +125,7 @@ export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|\~$/i;
 export const ALREADY_SIGNED_REGEX = /^META\-INF\/manifest\.mf/;
 
 export const FLAGGED_FILE_EXTENSION_REGEX =
-  /\.exe$|\.dll$|\.dylib$|\.so$|\.sh$|\.class$|\.swf$/i;
+  /\.exe$|\.dll$|\.dylib$|\.so$|\.sh$|\.class$|\.swf$|\.jar$/i;
```
tofumatt: We should generate this regex from an array, which would be nicer for devs. Filed #1115. Obviously no need to do it in this PR.

PR #1136, opened 2017-02-20
Port eslint rules over to new eslint rule format.

### LINKED ISSUE #1135: Make it easier to include 3rd party eslint rules
This is a multi-step issue and part of https://github.com/mozilla/addons/issues/282 to simplify adding new rules potentially from 3rd party repositories, e.g security relevant rules from https://github.com/mozfreddyb/

Main task here for now: [ ] Port to new eslint rule format (http://eslint.org/blog/2016/07/eslint-new-rule-format) which is used by many rules and modern eslint configs.

EnTeQuAk:
See http://eslint.org/blog/2016/07/eslint-new-rule-format for a few more
details.

Fixes #1135

This opens a few more vectors for code quality and optimization and cleanup here and there but I'd like to make this as contained as possible so there will be follow-ups (e.g imho `ignorePrivateFunctions` isn't useful anymore).

Also, I tried integrating at least one 3rd party rule but there might be issues hiding here and there so they'll be resolved with every new rule that gets added.

Theoretically this allows us now to provide `fixer` functions when possible e.g for deprecated APIs etc. That could be a very useful feature to have in mind later.

Also, yes, I kept old rules for now since I wanted to port first to the new format and keep rules and tests in place to make sure I have any unittests I can rely on for the port :) Most of the rules will be removed later in the process.


src/rules/javascript/index.js
```diff
@@ -1,15 +1,36 @@
...
+import only_prefs_in_defaults from './only-prefs-in-defaults';
+import opendialog_nonlit_uri from './opendialog-nonlit-uri';
+import opendialog_remote_uri from './opendialog-remote-uri';
+import shallow_wrapper from './shallow-wrapper';
+import webextension_api from './webextension-api';
+import widget_module from './widget-module';
+
+
+export default {
+  rules: {
```
muffinresearch: I'm curious what was the rationale for making this an object instead of continuing to re-export? It makes adding rules more involved if you have to import and re-export every time.

src/scanners/javascript.js
```diff
@@ -34,11 +33,10 @@ export default class JavaScriptScanner {
       // ESLint is synchronous and doesn't accept streams, so we need to
       // pass it the entire source file as a string.
       var eslint = _ESLint.linter;
-      var rules = ignorePrivateFunctions(_rules);
 
-      for (let name in rules) {
+      for (let name in _rules) {
```
muffinresearch: As an aside: this `let`should probably be `const` see http://exploringjs.com/es6/ch_for-of.html#_iteration-variables-const-declarations-versus-var-declarations for the nuances.

muffinresearch:
Just a thought but see the new rule format - I wonder if the messages for a rule should now live on the meta data object for each eslint rule? That might make understanding the scope of a rule easier to understand without cross referencing?

src/rules/javascript/index.js
EnTeQuAk: The (probably naive) idea was that `rules/javascript` is more like a eslint-plugin (http://eslint.org/docs/developer-guide/working-with-plugins), eventually I'd like to move our json scanners and stuff to eslint processors but that's a step by step thing. Getting inspired by other eslint-plugins, e.g https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/src/index.js and I eventually wanted to get rid of our `const.ESLINT_RULE_MAPPING` and move that here to be more contained.

EnTeQuAk:
> Just a thought but see the new rule format - I wonder if the messages for a rule should now live on the meta data object for each eslint rule? That might make understanding the scope of a rule easier to understand without cross referencing?

That would be the ultimate goal, have messages and potential documentation and fixers live in the rule-scope like regular eslint rules. And eventually adapt that style to other scanners / rules too.

src/scanners/javascript.js
EnTeQuAk: Thanks for the hint, fixed in 9fd780d

src/rules/javascript/index.js
muffinresearch: So it might be better to make that config separately and leave the index file exports as is. That way it's easier to work with for now and the creation of the object to support the plugin structure can happen elsewhere. If you import rules collectively later for the plugin case your could probably then just spread them into the rules object.

EnTeQuAk: Sounds like a plan, let me try and change that :)

tests/scanners/test.javascript.js
```diff
@@ -4,8 +4,8 @@ import { DEPRECATED_APIS, ESLINT_ERROR, ESLINT_RULE_MAPPING, TEMPORARY_APIS,
          VALIDATION_ERROR, VALIDATION_WARNING } from 'const';
 import JavaScriptScanner from 'scanners/javascript';
 import * as messages from 'messages';
-import * as rules from 'rules/javascript';
-import { apiToMessage, ignorePrivateFunctions, singleLineString } from 'utils';
+import ruleConfig from 'rules/javascript';
```
muffinresearch: I think you could do `import { rules } from 'rules/javascript';` and then you could just do `Object.keys(rules)`

src/index.js
```diff
@@ -0,0 +1 @@
+export * from './rules/javascript';
```
muffinresearch: Is this used anywhere?

muffinresearch:
r+wc

src/index.js
EnTeQuAk: oh, that should never have been submitted. No, it isn't and it was part of me experimenting around with a real eslint-plugin. I'll get rid of it.

Issue #1113, opened 2017-02-01
Revisit whether there's a better way to consume eslint rules


muffinresearch:
This is to summarize a conversation on vidyo with @EnTeQuAk 

When the linter was originally built out we took the decision to use eslint internal APIs because that made it easier to run rules and define our own error messages. 

It would be a good idea to revisit this and check if there's any better options for being able to run eslint rules via the linter in standard form so that we can more easily consume existing eslint rules and so that any custom rules here are more readily usable via other eslint clients in editors directly.

Once all the legacy non-webext rules are ripped out there will be minimal eslint rules so that might provide a good time to do this if it's already been investigated and found to be viable.

As an example this json reporter might be one avenue worth exploring: https://github.com/sindresorhus/eslint-json

EnTeQuAk:
There is and this just got implemented with https://github.com/mozilla/addons-linter/issues/1140 and other PRs.

PR #1140, opened 2017-02-23
Add warning about use of unsafe innerHTML

### LINKED ISSUE #811: Linter doesn't warn about the use of innerHTML
If an add-on has code using innerHTML it should be flagged. 

See https://github.com/mozilla/amo-validator/blob/d04348a173f66fc6286aa38ce70dd1cf5c8e938c/validator/testcases/javascript/instanceproperties.py for some references to the amo-validator's take on it.

Originally filed here: https://github.com/mozilla/addons-server/issues/3041

EnTeQuAk:
Fixes #811 and references (but doesn't fix it yet imho) https://github.com/mozilla/addons-linter/issues/1113

This PR improves 3rd party rule support where needed along the way. Also adapts other rules a bit, e.g to correctly resolve the current working directory. See the actual commits for more details.

This now uses `CLIEngine.executeOnText` instead of `eslint.verify` to make sure plugins and processors are properly loaded. We could have done that ourselves but it felt more appropriate to just use that API. This messes a bit with the current working directory (now absolute paths are present in `context.getFilename`) but that's reasonable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "dispensary": "0.10.3",
     "es6-promisify": "5.0.0",
     "eslint": "3.16.0",
+    "eslint-plugin-no-unsafe-innerhtml": "1.0.15",
     "esprima": "3.1.3",
     "first-chunk-stream": "2.0.0",
     "postcss": "5.2.15",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "dispensary": "0.10.3",
     "es6-promisify": "5.0.0",
     "eslint": "3.16.0",
-    "eslint-plugin-no-unsafe-innerhtml": "1.0.15",
+    "eslint-plugin-no-unsafe-innerhtml": "EnTeQuAk/eslint-plugin-no-unsafe-innerhtml#4b6b606d50",
     "esprima": "3.1.3",
     "first-chunk-stream": "2.0.0",
     "postcss": "5.2.15",
--- a/src/rules/javascript/only-prefs-in-defaults.js
+++ b/src/rules/javascript/only-prefs-in-defaults.js
@@ -1,12 +1,15 @@
+import path from 'path';
+
 import { ONLY_PREFS_IN_DEFAULTS } from 'messages/javascript';
 import { getRootExpression } from 'utils';
 
+
 export default {
   create(context) {
-    var filename = context.getFilename();
+    var relPath = path.relative(process.cwd(), context.getFilename());
 
     // This rule only applies to files in defaults/preferences
-    if (filename.indexOf('defaults/preferences/') === 0) {
+    if (path.dirname(relPath).startsWith('defaults/preferences')) {
       return {
         CallExpression: function(node) {
           var root = getRootExpression(node);
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -32,52 +32,66 @@ export default class JavaScriptScanner {
     return new Promise((resolve) => {
       // ESLint is synchronous and doesn't accept streams, so we need to
       // pass it the entire source file as a string.
-      var eslint = _ESLint.linter;
-
-      for (const name in _rules) {
-        this._rulesProcessed++;
-        eslint.defineRule(name, _rules[name]);
-      }
-
-      var report = eslint.verify(this.code, {
+      var cli = new _ESLint.CLIEngine({
         env: { es6: true },
         parserOptions: { ecmaVersion: 2017 },
         ignore: false,
+        plugins: ['no-unsafe-innerhtml'],
         rules: _ruleMapping,
         settings: {
           addonMetadata: this.options.addonMetadata,
         },
-      }, {
         allowInlineConfig: false,
         filename: this.filename,
+        // Avoid loading the addons-linter .eslintrc file
+        useEslintrc: false,
       });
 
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -72,6 +72,7 @@
     "dispensary": "0.10.3",
     "es6-promisify": "5.0.0",
     "eslint": "3.16.0",
+    "eslint-plugin-no-unsafe-innerhtml": "EnTeQuAk/eslint-plugin-no-unsafe-innerhtml#4b6b606d50",
```
EnTeQuAk: There is an upstream PR that needs merging first (https://github.com/mozfreddyb/eslint-plugin-no-unsafe-innerhtml/pull/16) so this will be updated once that happens.

src/rules/javascript/only-prefs-in-defaults.js
```diff
@@ -1,12 +1,15 @@
+import path from 'path';
+
 import { ONLY_PREFS_IN_DEFAULTS } from 'messages/javascript';
 import { getRootExpression } from 'utils';
 
+
 export default {
   create(context) {
-    var filename = context.getFilename();
+    var relPath = path.relative(process.cwd(), context.getFilename());
```
EnTeQuAk: This might need a bit more work and testing. I'd love to forward the actual package path to the rules eventually but for now this seems to work just fine. (also this rule will be removed anyway but it's a good example)

src/scanners/javascript.js
```diff
@@ -30,54 +30,70 @@ export default class JavaScriptScanner {
...
         rules: _ruleMapping,
-        settings: {
-          addonMetadata: this.options.addonMetadata,
-        },
-      }, {
+        plugins: ['no-unsafe-innerhtml'],
         allowInlineConfig: false,
         filename: this.filename,
+        // Avoid loading the addons-linter .eslintrc file
+        useEslintrc: false,
```
muffinresearch: Nice!

muffinresearch: Presumably this change also means that eslintrc's nested in a webext can't override anything. It would be good to add an explicit test for something like that.

EnTeQuAk: I added https://github.com/mozilla/addons-linter/issues/1141 for this

src/rules/javascript/only-prefs-in-defaults.js
```diff
@@ -1,12 +1,15 @@
...
```
muffinresearch: I could see this could potentially break with using the linter via web-ext.

EnTeQuAk: Yeah, I want to make this configurable and maybe forward the actual package path, not sure yet, see https://github.com/mozilla/addons-linter/pull/1140#discussion_r102708372 on that.

Generally, this is how eslint does things anyway so I don't expect there to be any major problems.

src/scanners/javascript.js
muffinresearch: Ok, that's fine we should try and avoid a release until that's checked - a quick local check would suffice though.

EnTeQuAk: I checked locally and couldn't find an obvious way to overwrite default checks (checked with innerHTML)

EnTeQuAk: That doesn't work without that option too though as far as I see.

muffinresearch:
Looks pretty good. I like how the rule doesn't worry about static values.

I notice it doesn't catch bracket syntax:

```js
foo['innerHTML'] = "<a href='"+url+"'>About</a>";
```

But it's a good start in any case.

r+wc

muffinresearch:
One other thing - I see the rule allows for specific use of a escaping functions that would also in our case mean that if you wrap something in the right name you could bypass the warning. 

It might be nice to be able to disable that for our case. Or alternatively we could look to making some policy where we utiise that feature and define a list of acceptable sanitization function names. Either way that probably needs to be discussed with Andreas etc.

EnTeQuAk:
> I notice it doesn't catch bracket syntax

I'd say, let's file this at https://github.com/mozfreddyb/eslint-plugin-no-unsafe-innerhtml to get this in upstream. 

> One other thing - I see the rule allows for specific use of a escaping functions that would also in our case mean that if you wrap something in a function with the right name you would bypass the warning.

> It might be nice to be able to disable that for our case (probably an upstream patch). Or alternatively we could look to making some policy where we utiise that feature and define a list of acceptable sanitization function names. Either way that probably needs to be discussed with Andreas etc.

Can you elaborate on this a little bit more / maybe file appropriate tickets? I feel like I don't get what you mean right now.

EnTeQuAk:
I merged it, let's improve this incrementally from now on. I feel like it catches a lot and all rules should be subject to incremental improvement and re-evaluation.

kewisch:
> It might be nice to be able to disable that for our case (probably an upstream patch). Or alternatively we could look to making some policy where we utiise that feature and define a list of acceptable sanitization function names. Either way that probably needs to be discussed with Andreas etc.

Personally I think we can catch this with policy. Yes, the escapeHTML function would allow circumventing the warning, but escapeHTML is indeed the most common function name for functions that escape html. If wanted we could add another rule that makes sure the escapeHTML function does what it is supposed to do (by requiring the one we have in MDN). @wagnerand ?

Issue #1130, opened 2017-02-16
Add missing metadata to mozilla-central manifest schemas


mstriemer:
The manifest schema is being imported from mozilla-central in #1129. mozilla-central doesn't have all of the data that we need though, so we will need to add some of our metadata to the extracted schemas.

Things that are missing are some descriptions and validations such as the `strict_min_version` which mozilla-central just marks as optional but we have a regex to validate. For the `strict_min_version` case there is likely some validation being done elsewhere in Firefox. We should see about matching Firefox's validation and moving that into the schema, even if only as a `format`.

Depends on #1129.

mstriemer:
Be sure to look at the regexes that are coming from Firefox, some of them have case-insensitive flags at the beginning (like `(?i)^[a-f0-9]{4}...$`) that are currently being stripped out. We'll want to replace these too.

PR #1142, opened 2017-02-24
Import schemas from Firefox

### LINKED ISSUE #1129: Import initial manifest.json from mozilla-central
mozilla-central has some JSON schema-esque files to manage webextensions. We should import this data from there instead of maintaining a similar JSON schema file for the manifest.

There is some metadata that we need from mozilla-central that is missing in addons-linter. So the first step is to get the schema from mozilla-central then to add our metadata—which will be tracked separately.

mstriemer:
This imports and converts the schemas in Firefox to JSON schema that `ajv` supports. I think this would be JSON schema v4 (which is essentially v3?).

This doesn't actually use this code anywhere but this was the first step. I have another branch that uses these files in `schema/validator.js` and a bunch of tests fail for overly specific tests or the Firefox schemas not having enough data. Fixing that will come next.

You likely only need to look at `src/schema/firefox-schemas.js`, `src/schema/firefox-schemas-import.js` and `tests/schema/test.firefox-schemas.js ` (sorry the test file is big but partly it's data).

A spot check of some files wouldn't be bad but they're a bit hard to read so I was thinking I'd just let the tests sort that out when I start using them.

Fixes #1129.


src/schema/firefox-schemas-import.js
```diff
@@ -0,0 +1,57 @@
...
+    fs.readFileSync(`${path}/${file}`, 'utf-8'),
+    null, // reviver
+    true, // remove_comments
+  );
+}
+
+function writeSchema(path, file, schema) {
+  fs.writeFile(
+    `${path}/${file}`,
+    `${JSON.stringify(schema, undefined, 2)}\n`);
```
mstriemer: The newline isn't strictly necessary but we put newlines at the end of our other files and they show up as missing a new line in git if it isn't there.

src/schema/firefox-schemas-import.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function importSchemas() {
+  const path = process.argv[2];
+  const rawSchemas = loadSchemasFromFile(path);
+  const processedSchemas = processSchemas(rawSchemas);
+  writeSchemasToFile(path, processedSchemas);
+}
+
+importSchemas();
```
mstriemer: Yeah, I'd rather it not happen at import time, that's mostly why the file management stuff got pulled out into its own file.

I've just been running this file but it probably makes more sense to have something like `bin/firefox-import` that calls this rather than having it here.

src/schema/firefox-schemas.js
```diff
@@ -0,0 +1,189 @@
+const FLAG_PATTERN_REGEX = /^\(\?[im]*\)(.*)/;
+
+// Reference some functions on inner so they can be stubbed in tests.
+export const inner = {};
```
mstriemer: This is definitely a hack. I don't think you can stub the functions with sinon if they aren't attached to an object though, unfortunately.

Since these functions are internal to this module they need to be attached to something to get stubbed.

I think having them on the `inner` object shows that they're private, but I'm not sure it matters much. If it weren't for the sinon issue I would have exported them regularly.

src/schema/firefox-schemas.js
```diff
@@ -0,0 +1,189 @@
...
+
+function stripFlagsFromPattern(value) {
+  // TODO: Fix these patterns and remove this code.
```
mstriemer: I added a comment about this to #1130, we'll want to replace them there until they get updated in Firefox.

src/schema/firefox-schemas-import.js
EnTeQuAk: `bin/firefox-import` sounds like a great plan to me

src/schema/firefox-schemas.js
EnTeQuAk: I see, I'm fine with leaving it like that then

src/schema/firefox-schemas.js
EnTeQuAk: Perfect, thanks!

EnTeQuAk: Perfect, thanks!

src/schema/firefox-schemas-import.js
```diff
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import { join as joinPath } from 'path';
```
EnTeQuAk: hmm, what's wrong with `import path from 'path';` and use `path.join(...)` instead?

EnTeQuAk: or is that something you don't do in js-world?

mstriemer: Just that `path` was already being used. I can do that and rename those variables.

PR #1149, opened 2017-03-01
Add no-eval rule linting.

### LINKED ISSUE #1047: eval() not caught
`eval()` is missed here:
https://addons.mozilla.org/en-US/firefox/files/browse/544938/file/js/background.js#top

ref: transferred from here
https://github.com/mozilla/amo-validator/issues/370

EnTeQuAk:
This essentially mirrors what amo-validators does, only caring about
global eval() calls so we set allowIndirect to true. This can easily
be changed in case we want it though (which is actually the default
of the eslint rule).

Fixes #1047

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -31,8 +31,8 @@ export default class JavaScriptScanner {
   }={}) {
     return new Promise((resolve) => {
       var cli = new _ESLint.CLIEngine({
-        env: { es6: true },
         baseConfig: {
+          env: { es6: true, webextension: true, browser: true },
           parserOptions: { ecmaVersion: 2017 },
           settings: {
             addonMetadata: this.options.addonMetadata,
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -33,11 +33,11 @@ export default class JavaScriptScanner {
       var cli = new _ESLint.CLIEngine({
         baseConfig: {
           env: { es6: true, webextension: true, browser: true },
-          parserOptions: { ecmaVersion: 2017 },
           settings: {
             addonMetadata: this.options.addonMetadata,
           },
         },
+        parserOptions: { ecmaVersion: 2017 },
         ignore: false,
         rules: _ruleMapping,
         plugins: ['no-unsafe-innerhtml'],
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -32,12 +32,18 @@ export default class JavaScriptScanner {
     return new Promise((resolve) => {
       var cli = new _ESLint.CLIEngine({
         baseConfig: {
-          env: { es6: true, webextension: true, browser: true },
+          env: {
+            es6: true,
+            webextension: true,
+            browser: true,
+          },
           settings: {
             addonMetadata: this.options.addonMetadata,
           },
         },
-        parserOptions: { ecmaVersion: 2017 },
+        parserOptions: {
+          ecmaVersion: 2017,
+        },
         ignore: false,
         rules: _ruleMapping,
         plugins: ['no-unsafe-innerhtml'],
--- a/tests/rules/javascript/test.no_eval.js
+++ b/tests/rules/javascript/test.no_eval.js
@@ -1,0 +1,101 @@
+import { VALIDATION_WARNING } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+
+// These rules were mostly copied and adapted from
+// https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-eval.js
+// Please make sure to keep them up-to-date and report upstream errors.
+describe('no_eval', () => {
+  var validCodes = [
+    'Eval(foo)',
+
+    // User-defined eval methods.
+    'window.eval("foo")',
+    'window.eval("foo")',
+    'window.noeval("foo")',
+    'function foo() { var eval = "foo"; window[eval]("foo") }',
+    'global.eval("foo")',
+    'global.eval("foo")',
+    'global.noeval("foo")',
+    'function foo() { var eval = "foo"; global[eval]("foo") }',
+    'this.noeval("foo");',
+    'function foo() { "use strict"; this.eval("foo"); }',
+    'function foo() { this.eval("foo"); }',
+    'function foo() { this.eval("foo"); }',
+    'var obj = {foo: function() { this.eval("foo"); }}',
+    'var obj = {}; obj.foo = function() { this.eval("foo"); }',
+    'class A { foo() { this.eval(); } }',
+    'class A { static foo() { this.eval(); } }',
+
+    // Allows indirect eval
+    '(0, eval)("foo")',
+    '(0, window.eval)("foo")',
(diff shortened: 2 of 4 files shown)


wagnerand:
If I read http://eslint.org/docs/rules/no-eval correctly, we'd want `allowIndirect": false` in oder to catch `window.eval(...)`. What do you think?

tests/rules/javascript/test.no_eval.js
```diff
@@ -0,0 +1,101 @@
...
+  var validCodes = [
+    'Eval(foo)',
+
+    // User-defined eval methods.
+    'window.eval("foo")',
+    'window.eval("foo")',
+    'window.noeval("foo")',
+    'function foo() { var eval = "foo"; window[eval]("foo") }',
+    'global.eval("foo")',
+    'global.eval("foo")',
```
wagnerand: Why are some of these duplicated?

EnTeQuAk:
> If I read http://eslint.org/docs/rules/no-eval correctly, we'd want allowIndirect": false in oder to catch window.eval(...). What do you think?

I wasn't sure, amo-validator certainly didn't so I naively went the way of allowing it for now. I don't see any issues of disallowing indirect eval calls though so if that's what we want I can enable those too.

wagnerand:
Well, otherwise it would be trivial to work-around the rule just by prepending `window.`.

tests/rules/javascript/test.no_eval.js
EnTeQuAk: that was me failing adding the original tests since they were splitted between environments and I cleaned them up in a automated fashion and kinda failed 😆 I'll fix this.

EnTeQuAk:
> Well, otherwise it would be trivial to work-around the rule just by prepending window..

Sounds reasonable, I'll enable it.

EnTeQuAk:
Added it in https://github.com/mozilla/addons-linter/pull/1149/commits/5dfb90a440cb45f00eb53c39f082f65f15bb8b3e and improved tests, overall js-scanner configuration to find more globals and traverse them (see http://eslint.org/docs/user-guide/configuring#specifying-environments for more information)

tests/rules/javascript/test.no_eval.js
```diff
@@ -0,0 +1,114 @@
...
+describe('no_eval', () => {
+  var validCodes = [
+    'Eval(foo)',
+
+    // User-defined eval methods.
+    'window.noeval("foo")',
+    'function foo() { var eval = "foo"; window[eval]("foo") }',
+    'global.eval("foo")',
+    'global.noeval("foo")',
+    'function foo() { var eval = "foo"; global[eval]("foo") }',
```
wagnerand: Dupe of line 13?

tests/rules/javascript/test.no_eval.js
```diff
@@ -0,0 +1,114 @@
...
+    'this.noeval("foo");',
+    'function foo() { "use strict"; this.eval("foo"); }',
+    'var obj = {foo: function() { this.eval("foo"); }}',
+    'var obj = {}; obj.foo = function() { this.eval("foo"); }',
+    'class A { foo() { this.eval(); } }',
+    'class A { static foo() { this.eval(); } }',
```
wagnerand: Maybe we can sort this array (manually) so we can spot duplicates more easily?

src/scanners/javascript.js
```diff
@@ -31,13 +31,13 @@ export default class JavaScriptScanner {
...
       var cli = new _ESLint.CLIEngine({
-        env: { es6: true },
         baseConfig: {
-          parserOptions: { ecmaVersion: 2017 },
+          env: { es6: true, webextension: true, browser: true },
           settings: {
             addonMetadata: this.options.addonMetadata,
           },
         },
+        parserOptions: { ecmaVersion: 2017 },
```
muffinresearch: As above

muffinresearch:
This rule is purely using 'eval', it would be good to look at adding "no-implied-eval" to cover setInterval, setTimeout and friends - it's almost certainly better than our own rule.

muffinresearch:
Also http://eslint.org/docs/rules/no-new-func for new Function which is also related.

tests/rules/javascript/test.no_eval.js
```diff
@@ -0,0 +1,114 @@
+import { VALIDATION_WARNING } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+
+// These rules were mostly copied and adapted from
+// https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-eval.js
+// Please make sure to keep them up-to-date and report upstream errors.
+describe('no_eval', () => {
+  var validCodes = [
+    'Eval(foo)',
```
muffinresearch: This misses the cases in the src test file that this shouldn't raise an issue for setTimeout etc or is that a problem in this case due to the other rules we have?

tests/scanners/test.javascript.js
```diff
@@ -290,7 +290,7 @@ describe('JavaScript Scanner', function() {
   it('should export all rules in rules/javascript', () => {
     // We skip the "run" check here for now as that's handled by ESLint.
     var ruleFiles = getRuleFiles('javascript');
-    var externalRules = 1;
+    var externalRules = 2;
```
muffinresearch: At some point it would be good to look at updating the eslintrc of the codebase so that `const` and `let` are used (preferring const) and then all the var usage can be removed.

tests/rules/javascript/test.no_eval.js
EnTeQuAk: yeah, I removed those cases because we do have rules for it. Since our tests always test the whole test-suite at the moment (and I like to keep it that way tbh). We could separate this a bit but generally I like that we always look at the great picture for our rule tests.

tests/scanners/test.javascript.js
EnTeQuAk: heh, I'll put it on the list :)

EnTeQuAk:
> This rule is purely using 'eval', it would be good to look at adding "no-implied-eval" [1] to cover setInterval, setTimeout and friends - it's almost certainly better than anything we've currently got in that department.

yeah, I wanted to tackle that and replace what we currently have for setTimeout and others with that then. Need to figure out a few other things first though (e.g message overwriting)

tests/rules/javascript/test.no_eval.js
EnTeQuAk: Nope, `window` vs `global` namespace :)

tests/rules/javascript/test.no_eval.js
EnTeQuAk: Fixed in 62be224

tests/rules/javascript/test.no_eval.js
wagnerand: I swear I looked at this three times and still couldn't see it. Sorry.

EnTeQuAk: hehe, it actually took me a few minutes to spot this after sublime was sure it's not identical 🤣

PR #1160, opened 2017-03-04
docs: Added notes on API integration

### LINKED ISSUE #1156: Add documentation about how to use this package programmatically

I can't seem to find any documentation about how to use this package programmatically instead via cli.  It would be nice if there was a node api which one could use instead of having to use subprocesses to call the cli, then parse the stdout.

### What happened?

I can't seem to find any documentation about how to use this package programmatically.

There should be some docs that explain how to use this package via node/commonjs api instead of merely via cli.

rpl:
Just a small addition to the README.md with some notes related to "how to use the Linter programmatically"

should fix #1156

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,61 @@ you can enable/disable for the command-line app, use the `--help` option:
 addons-linter --help
 ```
 
+### Import Linter API into another NodeJS application
+
+```js
+import linter from 'addons-linter';
+
+const sourceDir = process.cwd();
+
+const linter = linter.createInstance({
+  config: {
+    // This mimics the first command line argument from yargs,
+    // which should be the directory to the extension.
+    _: [sourceDir],
+    logLevel: process.env.VERBOSE ? 'debug' : 'fatal',
+    stack: Boolean(process.env.VERBOSE),
+    pretty: false,
+    warningsAsErrors: false,
+    metadata: false,
+    output: 'none',
+    boring: false,
+    selfHosted: false,
+    // Lint only the selected files
+    //   scanFile: ['path/...', ...]
+    //
+    // Exclude files:
+    shouldScanFile: (fileName) => true,
+  },
+  // This prevent the linter to exit the nodejs application
+  runAsBinary: false,
+});
+
+linter.run()
+  .then((linterResults) => ...)
+  .catch((err) => console.error("addons-linter failure: ", err));
+```
+
+`linter.output` is composed by the following properties (the same of the 'json' report type):
+
+```js
+{
+  metadata: {...},
+  summary: {
+    error, notice, warning,
+  },
+  scanFile,
+  count,
+  error: [{
+    type: "error",
+    code, message, description,
+    column, file, line
+  }, ...],
+  warning: [...],
+  notice: [...]
+}
+```
+
 ## Development
 
 If you'd like to help us develop the addons-linter, that's great! It's
--- a/src/linter.js
+++ b/src/linter.js
@@ -91,6 +91,9 @@ export default class Linter {
   }
 
   print(_console=console) {
+    if (this.config.output === 'none') {
+      return;
+    }
     if (this.config.output === 'json') {
       _console.log(this.toJSON(this.config.pretty));
     } else {
@@ -392,7 +395,7 @@ export default class Linter {
           }
           this.collector._addMessage(message.type, message);
         }
-        return;
+        return this.output;
       });
   }
 
--- a/tests/test.linter.js
+++ b/tests/test.linter.js
@@ -397,6 +397,19 @@ describe('Linter.print()', function() {
     assert.ok(fakeConsole.log.called);
   });
 
+  it('should not print anything if config.output is none', () => {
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -45,6 +45,67 @@ you can enable/disable for the command-line app, use the `--help` option:
 addons-linter --help
 ```
 
+### Import Linter API into another NodeJS application
+
+```js
+import {createInstance as createLinter} from 'addons-linter';
```
tofumatt: If `createInstance` isn't a good name we should change it inside the linter rather than have our docs advocate importing it as a different name. I think this looks odd 🤷‍♀️

README.md
```diff
@@ -45,6 +45,67 @@ you can enable/disable for the command-line app, use the `--help` option:
...
+
+async lint(sourceDir) {
```
tofumatt: Using the `async` syntax (which is quite new and a lot of developers won't know) isn't very friendly, especially to novice developers. I don't know the `async` syntax well enough to understand this example as much as Promises-based one.

I think async devs will know how to translate it from Promises, but the inverse isn't true. We should provide Promise-based examples (either first or only).

README.md
```diff
@@ -45,6 +45,67 @@ you can enable/disable for the command-line app, use the `--help` option:
...
+      //   scanFile: ['path/...', ...]
+      //
+      // Exclude files:
+      shouldScanFile: (fileName) => true,
+    },
+    // This prevent the linter to exit the nodejs application
+    runAsBinary: false,
+  });
+
+  linter.print = () => {}; // Prevents the linter from printing the report
```
tofumatt: Unrelated but that's a bad API; we should make this a config option 😄

README.md
rpl: hehe, how about this?

```
import linter from 'addons-linter'

const linter = linter.createInstance({config: {...}});
...
```

README.md
rpl: :+1: that's true, snippet example rewritten into the Promises-based one.

README.md
rpl: yeah, I agree: I tweaked both the example and the Linter class so that a config with 'type: "none"' does not produce any linter output.

rpl:
Thanks @tofumatt for taking a look at it, I totally agree with the review comments, and I was even tempted to change the API accordingly from the start, but I though that by writing the example with the API like it is now and by creating a pull request, to briefly discuss how we want the API to look like and then apply the changes that we agreed on.

I applied some small change to the API (and the rewritten the example snippet accordingly) to:

- do not print any output if type is 'none'
- linter.run() resolved to the promise results

I didn't changed the exported function from `createInstance` to `createLinter` yet, mostly because `linter.createInstance` doesn't seem to fit too bad in the new snippet (but I'm completely open to rename it if you think that the example would like nicer).

EnTeQuAk:
I may be missing something but is there still something needed to land this?

rpl:
@EnTeQuAk just one last review, to be sure that the applied changes are covering the previous review comments. 

(in the last update on this PR I applied some minor change to the exposed API, rewritten the example to be Promise-based instead of based on the new async/await syntax, but I didn't renamed `createInstance` into `createLinter` as suggested because `linter.createInstance` can be probably good enough)

EnTeQuAk:
@tofumatt this PR looks great to me with the changes, do you have a moment to give this a second look to get this merged?

README.md
```diff
@@ -45,6 +45,61 @@ you can enable/disable for the command-line app, use the `--help` option:
...
+    //
+    // Exclude files:
+    shouldScanFile: (fileName) => true,
+  },
+  // This prevent the linter to exit the nodejs application
+  runAsBinary: false,
+});
+
+linter.run()
+      .then((linterResults) => ...)
```
tofumatt: This indentation seems off, are they tabs or something? Most JS indentation is 2 spaces and the above is but these are massive.

README.md
```diff
@@ -45,6 +45,61 @@ you can enable/disable for the command-line app, use the `--help` option:
...
+      .catch((err) => console.error("addons-linter failure: ", err));
+```
+
+`linter.output` is composed by the following properties (the same of the 'json' report type):
+
+```js
+     {
```
tofumatt: I'd use 2 space indents here too.

src/linter.js
```diff
@@ -91,6 +91,9 @@ export default class Linter {
   }
 
   print(_console=console) {
+    if (this.config.output === 'none') {
```
tofumatt: 👍  Thanks we needed this 😆

rpl:
@EnTeQuAk I pushed an update that fixes the indentation, this should be ready now! \o/

@tofumatt no worries, I've nothing but good to say about working with you and @EnTeQuAk! On the contrary, thank you both for your help!

rpl:
@EnTeQuAk Great! feel free to merge it at any time (I don't have the permissions needed to merge it on this repo)

(ping me if there will be any conflict to resolve upfront, in case something which conflicts with it has been landed in the meantime, I would be happy to handle it immediately)

PR #1162, opened 2017-03-05
Fix spelling of MULTIPLE_MANIFESTS

jwilk:
Found using [mwic](http://jwilk.net/software/mwic).


tofumatt:
@EnTeQuAk does the validation on the server pick this code out specifically for anything? If so we'd need to update addons-server, though I guess the tests would fail on a version update were that the case.

PR #1169, opened 2017-03-07
Add metadata to firefox schemas

### LINKED ISSUE #1130: Add missing metadata to mozilla-central manifest schemas

mstriemer:
Update the imported schemas so that they validate everything that we had in our unit tests. I'm still going to go through and test this some more manually to ensure that the output is about the same. I know some things now give multiple errors instead of just one so I'm not sure how that will format.

This also updates the validator to support reading the schema data from the firefox schemas or the linter, this is controlled by the `USE_FIREFOX_SCHEMAS` environment variable. The tests pass under both versions and they will both be run on Travis.

Fixes #1130.


package.json
```diff
@@ -51,6 +52,7 @@
     "json-loader": "0.5.4",
     "load-grunt-configs": "1.0.0",
     "load-grunt-tasks": "3.5.2",
+    "lodash": "^4.17.4",
```
pdehaan: Do you need all of **lodash**, or just [**lodash.ismatchwith**](https://www.npmjs.com/package/lodash.ismatchwith)?

mstriemer: That should work. I wasn't too concerned about it since this is just for the tests but may as well be specific about it.

tofumatt:
This is a big diff so I'm happy to review but there is one more mobile v1 bug I'd like to get out first so I won't be able to look at this until the weekend/Monday. Hope that's okay! 😄

mstriemer:
Lots of changes but I think most of them are the schema files and test.manifestjson.js where I move all the top-level describes into a main describe block with the shared prefix they all had. Might want to check that file with [whitespace ignored](https://github.com/mozilla/addons-linter/pull/1169/files?w=1).

tofumatt:
Yeah fair enough. Also OMG that is a cool trick ^_^

mstriemer:
Yeah, I was going to assign it to you when I first submitted this but figured you were probably busy 😛. No rush!

mstriemer:
@cgrebs took a look at #1175 which includes these changes, I made some updates here for his comments. Merging based on that review.

PR #1174, opened 2017-03-10
Detect unknown browser API usage

### LINKED ISSUE #1168: Warn on using an API that Firefox doesn't support
Now we've got the schema in place, can we tell if an API is not available (or not supported) in Firefox? Some examples:

chrome.gcm.*
chrome.identity.getAccounts

mstriemer:
This is branched off of the metadata stuff since it has all the data we need. Not ready to be used but it's working so I thought I'd throw it up here.

I added these lines to an XPI I downloaded off AMO and this is the output:

```js
browser.notathing.get('foo');
chrome.storage.google('wat');
browser.app.getDetails('deprecated');
```

```
Validation Summary:

errors          0              
notices         0              
warnings        4              

WARNINGS:

Code              Message                             Description                                                              File                   Line   Column
UNSUPPORTED_API   notathing.get is not supported      This API has not been implemented by Firefox.                            archive.js             34     13    
UNSUPPORTED_API   storage.google is not supported     This API has not been implemented by Firefox.                            archive.js             35     13    
APP_GETDETAILS    "app.getDetails" is deprecated or   This API has been deprecated by Chrome and has not been implemented by   archive.js             36     13    
                  unimplemented                       Firefox.                                                                                                     
ALREADY_SIGNED    Package already signed              Add-ons which are already signed will be re-signed when published on     META-INF/manifest.mf                
                                                      AMO. This will replace any existing signatures on the add-on.                                                
```

Fixes #1168.

### DIFF
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -1,4 +1,4 @@
-import { apiToMessage, gettext as _, singleLineString } from 'utils';
+emport { apiToMessage, gettext as _, singleLineString } from 'utils';
 
 export const JS_SYNTAX_ERROR = {
   code: 'JS_SYNTAX_ERROR',
@@ -251,9 +251,10 @@ export const UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT = {
     serious performance degradation.`),
 };
 
-export const UNKNOWN_API = {
-  code: 'UNKNOWN_API',
-  message: _('{{api}} is not support by Firefox'),
+export const UNSUPPORTED_API = {
+  code: 'UNSUPPORTED_API',
+  message: null,
+  messageFormat: _('{{api}} is not supported'),
   description: _('This API has not been implemented by Firefox.'),
   legacyCode: null,
 };
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -251,6 +251,12 @@ export const UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT = {
     serious performance degradation.`),
 };
 
+export const UNKNOWN_API = {
+  code: 'UNKNOWN_API',
+  message: _('{{api}} is not support by Firefox'),
+  description: _('This API has not been implemented by Firefox.'),
+};
+
 function deprecatedAPI(api) {
   return {
     code: apiToMessage(api),
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -1,4 +1,4 @@
-emport { apiToMessage, gettext as _, singleLineString } from 'utils';
+import { apiToMessage, gettext as _, singleLineString } from 'utils';
 
 export const JS_SYNTAX_ERROR = {
   code: 'JS_SYNTAX_ERROR',
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -255,6 +255,7 @@ export const UNKNOWN_API = {
   code: 'UNKNOWN_API',
   message: _('{{api}} is not support by Firefox'),
   description: _('This API has not been implemented by Firefox.'),
+  legacyCode: null,
 };
 
 function deprecatedAPI(api) {
--- a/src/rules/javascript/webextension-unsupported-api.js
+++ b/src/rules/javascript/webextension-unsupported-api.js
@@ -1,0 +1,21 @@
+import { UNSUPPORTED_API } from 'messages/javascript';
+import { hasBrowserApi } from 'schema/browser-apis';
+
+export default {
+  create(context) {
+    return {
+      MemberExpression: function(node) {
+        if (node.object.object &&
+            ['chrome', 'browser'].includes(node.object.object.name)) {
+          let namespace = node.object.property.name;
+          let property = node.property.name;
+          let api = `${namespace}.${property}`;
+
+          if (!hasBrowserApi(namespace, property)) {
+            context.report(node, UNSUPPORTED_API.messageFormat, { api });
+          }
+        }
+      },
+    };
+  },
+};
--- a/src/schema/browser-apis.js
+++ b/src/schema/browser-apis.js
@@ -1,4 +1,5 @@
-import schemaList from './firefox-schemas';
+import { DEPRECATED_APIS, TEMPORARY_APIS } from 'const';
+import schemaList from 'schema/firefox-schemas';
 
 const schemaArrayNames = ['functions', 'events'];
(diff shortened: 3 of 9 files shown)


src/schema/browser-apis.js
```diff
@@ -0,0 +1,67 @@
...
+  require('json!schema/imported/sessions'),
+  require('json!schema/imported/sidebar_action'),
+  require('json!schema/imported/storage'),
+  require('json!schema/imported/tabs'),
+  require('json!schema/imported/test'),
+  require('json!schema/imported/theme'),
+  require('json!schema/imported/top_sites'),
+  require('json!schema/imported/web_navigation'),
+  require('json!schema/imported/web_request'),
+  require('json!schema/imported/windows'),
```
andymckay: argh its pain to import those manually, could we loop over a directory?

src/schema/browser-apis.js
```diff
@@ -0,0 +1,67 @@
...
+          if (Array.isArray(namespaceProperties)) {
+            return namespaceProperties.some(
+              (schemaItem) => schemaItem.name === property);
+          }
+          // eslint-disable-next-line no-console
+          console.log(
+            `${namespaceProperties} is not Array `, namespaceProperties);
+        }
+        return false;
+      }));
```
andymckay: curious, how fast is all this? any concerns?

src/schema/browser-apis.js
mstriemer: Yeah, I'm going to add support for that but this was already written out in the firefox-schemas.js so I just stuck with it.

src/schema/browser-apis.js
mstriemer: I'm not sure how long it takes but it could definitely be optimised by converting the arrays into objects at startup. Would simplify the code here a bit too.

src/messages/javascript.js
```diff
@@ -240,6 +240,12 @@ export const KNOWN_LIBRARY = {
   ],
 };
 
+export const UNKNOWN_API = {
+  code: 'UNKNOWN_API',
+  message: _('{{api}} is not support by Firefox'),
```
andymckay: Yeah, i had to do this by creating const in messages for each combination. This isn't going to work. If you could figure this out, it could make some of the messages a lot simpler.

mstriemer:
This is likely blocked on a solution to #1183 since I think if we solve that we can solve the problem of this not having a code (or maybe that's not an issue?).

New output:

```
Validation Summary:

errors          0              
notices         0              
warnings        4              

WARNINGS:

Code                              Message                             Description                                                                           File                   Line   Column
notathing.get is not supported    notathing.get is not supported                                                                                            archive.js             34     13    
storage.google is not supported   storage.google is not supported                                                                                           archive.js             35     13    
APP_GETDETAILS                    "app.getDetails" is deprecated or   This API has been deprecated by Chrome and has not been implemented by Firefox.       archive.js             36     13    
                                  unimplemented                                                                                                                                                 
ALREADY_SIGNED                    Package already signed              Add-ons which are already signed will be re-signed when published on AMO. This will   META-INF/manifest.mf                
                                                                      replace any existing signatures on the add-on.

**Comment by mstriemer** [comment]
This depends on #1189 so the tests won't be passing but it should pass once that lands (and impending conflicts are resolved).

I also had this diff applied so either the checks for code should be removed and we standardise on rule-id or that will need to be added to #1189.

```diff
diff --git a/src/scanners/javascript.js b/src/scanners/javascript.js
index 7f05a2d..3914eaf 100644
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -90,6 +90,7 @@ export default class JavaScriptScanner {
             var overwrites = ESLINT_OVERWRITE_MESSAGE[message.ruleId];
             var shortDescription = overwrites.message || message.message;
             var description = overwrites.description || message.description;
+            code = overwrites.code || message.code;
           } else {
             var shortDescription = code;
             var description = null;

**Comment by mstriemer** [comment]
@EnTeQuAk this should be ready for review now that #1189 has landed.

**Comment by mstriemer** [comment]
@muffinresearch if you've got time to review this that'd be great since Matt and Chris are both out.

File: src/schema/browser-apis.js [hunk 4]
```diff
@@ -0,0 +1,39 @@
...
+  if (isDeprecatedApi(namespace, property)
+    || isTemporaryApi(namespace, property)) {
+    return true;
+  }
+  if (!schema) {
+    return false;
+  }
+  return schemaObjectNames.some((schemaProperty) => {
+    return schema[schemaProperty] && property in schema[schemaProperty];
+  }) || schemaArrayNames.some((schemaProperty) => {
```
muffinresearch: It might be more readable to separate these out into separate blocks that return.

src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -0,0 +1,21 @@
+import { UNSUPPORTED_API } from 'messages/javascript';
+import { hasBrowserApi } from 'schema/browser-apis';
+
+export default {
+  create(context) {
+    return {
+      MemberExpression: function(node) {
+        if (node.object.object &&
```
muffinresearch: Does this if need to be more defensive?

src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -0,0 +1,21 @@
...
+            ['chrome', 'browser'].includes(node.object.object.name)) {
+          let namespace = node.object.property.name;
```
muffinresearch: As above do you know these props are always available?

src/rules/javascript/webextension-unsupported-api.js
mstriemer: Looking at some eslint examples it seems like `node.object` is always available (at least in a `MemberExpression`) so this should be fine.

src/rules/javascript/webextension-unsupported-api.js
mstriemer: We had this code in the [webextension-api.js](https://github.com/mozilla/addons-linter/blob/e3c310e8efdb40dc50d96617f5c6708cbef1d786/src/rules/javascript/webextension-api.js#L10) before and it was fine.

Since this is a MemberExpression I think this should always be available.

mstriemer:
I would guess that this doesn't deal with assigning the API to a variable, but I don't think that's something we're aiming to detect here.

PR #1175, opened 2017-03-10
Test commands with Firefox schemas

### LINKED ISSUE #766: Validate commands
Copied across from mozilla/web-extension-manifest-schema#115 as filed by @andymckay 

Let's validate commands, since this is a pile of regexes, this could get messy.

mstriemer:
Includes changes from #1169.

I copied the schema from commands.json into manifest-schema.json so the tests pass with both versions.

Depends on #1169.
Fixes #766.

### DIFF
--- a/src/schema/manifest-schema.json
+++ b/src/schema/manifest-schema.json
@@ -150,6 +150,47 @@
         "author": {
             "description": "Legacy field from Chrome. Use \"developer\" instead.",
             "type": "string"
+        },
+        "commands": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
+                    "suggested_key": {
+                        "type": "object",
+                        "properties": {
+                            "default": {
+                                "$ref": "#/definitions/KeyName"
+                            },
+                            "mac": {
+                                "$ref": "#/definitions/KeyName"
+                            },
+                            "linux": {
+                                "$ref": "#/definitions/KeyName"
+                            },
+                            "windows": {
+                                "$ref": "#/definitions/KeyName"
+                            },
+                            "chromeos": {
+                                "type": "string"
+                            },
+                            "android": {
+                                "type": "string"
+                            },
+                            "ios": {
+                                "type": "string"
+                            },
+                            "additionalProperties": {
+                                "type": "string",
+                                "deprecated": "Unknown platform name"
+                            }
+                        }
+                    },
+                    "description": {
+                        "type": "string"
+                    }
+                }
+            }
         }
     },
     "required": [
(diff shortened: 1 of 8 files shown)


src/schema/manifest-schema.json
```diff
@@ -150,6 +150,47 @@
...
+                            "chromeos": {
+                                "type": "string"
+                            },
+                            "android": {
+                                "type": "string"
+                            },
+                            "ios": {
+                                "type": "string"
+                            },
+                            "additionalProperties": {
```
mstriemer: This is how the Firefox schema is set up but I think it's actually wrong, unfortunately. I don't think it's hurting anything but `additionalProperties` should be at the same level as `properties`.

src/schema/imported/manifest.json
```diff
@@ -72,20 +54,31 @@
...
-              "format": "url",
-              "preprocess": "localize"
+              "format": null,
+              "preprocess": "localize",
+              "oneOf": [
+                {
+                  "format": "url"
+                },
+                {
+                  "pattern": "^__MSG_.*?__$"
```
EnTeQuAk: this pattern feels wrong to me but it's being imported right? Afair the string only has to start with ``__MSG_`` if translated and that doesn't have to be the case all the time.

.travis.yml
```diff
@@ -27,3 +27,6 @@ env:
...
   - secure: R5a9OOHyki7lihwtbWuWaIBv7nwKY8lCGRmYaSiZ5rKZ5ZSkaoC55eq3JsFqW1sQTxsiri6UJKBXoxJJiY1veV5DVuER6ZRoYd3Vr55BjKMBBFSp5uTFrzX7Mtbfncuf5/tVbwoDxDgc/AwahbWvR/gV9yUDGrbMrgU1/jE0Otjw+LXEqc0+nINoMVyfYFT9Elt1zcl6iKIKS+LDO0S9B5jQpxyIudC2bh/k8uaSKYRaBG9srqMuPJKXOdcMxmtOLkjcr7pKT8+LL7Y6U9i/tpsYnXHv/hz6m084Wz/Eu634wPjEfOTExoXgH1frxGF24ZgmdSfo6xm3fjI+1FPyyB7kp+y4r/pyySjHtqWJgIIUR1no22ISA0z6vy/mtQ0IXPgqoXoF96F2U2hyqVp9Mu5IXASiJXGvafW/bq4cstwRNibC8llmT5QJwJFZLs47qwCroLnyapr8Zbx6+4vzqTX6jQDURy8b95ydewtMhmwLeTeOhX+3qWnYqkG/+thhAkk1bjnuZkFgLMx9ae7GaZJSXelxhLR3pSpr9yGQ/WGnWjeumk3bN/gM8c94VCv2WDFBqb4bOjYsHZk+gGkh4mr1MznrjHcf6K7QJf9eVOXGwq4F/AMAqNkYcx3qj02b2PRrZ1WvF4MPGOw2PHREBLUzIiJmYl/iopT76XBo+f8=
+  matrix:
+  - ''
```
mstriemer: It is to run the tests using the linter's old schema.

EnTeQuAk: ahhhhhh, okay, that definitely needs a comment or even better a `USE_FIREFOX_SCHEMAS=` (empty) or so.

src/schema/firefox-validator.js
```diff
@@ -0,0 +1,75 @@
...
+  require('json!schema/imported/sidebar_action'),
+  require('json!schema/imported/storage'),
+  require('json!schema/imported/tabs'),
+  require('json!schema/imported/test'),
+  require('json!schema/imported/theme'),
+  require('json!schema/imported/top_sites'),
+  require('json!schema/imported/web_navigation'),
+  require('json!schema/imported/web_request'),
+  require('json!schema/imported/windows'),
+];
```
mstriemer: Yeah, I just did this to get it working right away since it was easy. I'll update it.

src/schema/imported/manifest.json
mstriemer: I tried to figure out where Firefox handles this because I had no idea what it was. This is copied from our old pattern though, since Firefox uses a non-standard way of converting this (I don't see it checking the format but I think the localisation process could fail and cause an error at load time).

PR #1189, opened 2017-03-20
WIP: Initial implementation that allows message/description overwrite.

EnTeQuAk:


src/scanners/javascript.js
```diff
@@ -80,10 +81,15 @@ export default class JavaScriptScanner {
...
-          // message structure.
+          // message structure and allow us to optionally overwrite
+          // their `message` and `description`
           if (_messages.hasOwnProperty(code)) {
             var shortDescription = _messages[code].message;
             var description = _messages[code].description;
+          } else if (ESLINT_OVERWRITE_MESSAGE.hasOwnProperty(message.ruleId)) {
+            var overwrites = ESLINT_OVERWRITE_MESSAGE[message.ruleId];
+            var shortDescription = overwrites.message || message.message;
+            var description = overwrites.description || message.description;
```
mstriemer: I don't see the code getting set in my branch that uses this. Might want to add `code = overwrites.code || code`.

After discussion on IRC it looks like this is only used in the command line version so maybe not needed.

tests/rules/javascript/test.no_unsafe_innerhtml.js
```diff
@@ -1,5 +1,9 @@
 import { VALIDATION_WARNING } from 'const';
 import JavaScriptScanner from 'scanners/javascript';
+import {
+  UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT,
+  NO_DOCUMENT_WRITE } from 'messages';
```
mstriemer: This format looks odd to me but doesn't bother me much. This is what I'm used to but if this is consistent with this project it's cool.

```js

import {
  UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT,
  NO_DOCUMENT_WRITE,
} from 'messages'; 
// or
import {
  UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT, NO_DOCUMENT_WRITE,
} from 'messages';

File: src/scanners/javascript.js [hunk 3]
```diff
@@ -1,6 +1,7 @@
 import ESLint from 'eslint';
 
-import { ESLINT_RULE_MAPPING, ESLINT_TYPES } from 'const';
+import { ESLINT_RULE_MAPPING, ESLINT_TYPES,
+         ESLINT_OVERWRITE_MESSAGE } from 'const';
```
mstriemer: Same import formatting comment here, too.

src/messages/javascript.js
```diff
@@ -240,6 +240,17 @@ export const KNOWN_LIBRARY = {
   ],
 };
 
+export const UNSAFE_DYNAMIC_VARIABLE_ASSIGNMENT = {
+  code: 'UNSAFE_VAR_ASSIGNMENT',
```
mstriemer: If the code isn't used then we probably don't need to set it.

docs/rules.md
```diff
@@ -387,3 +387,4 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | :white_check_mark: | warning | Web extension | Temporary IDs can cause issues with identity.getRedirectURL | | | null | IDENTITY_GETREDIRECTURL |
 | :white_check_mark: | warning | Web extension | When default_locale is specified a matching messages.json must exist | | | null | NO_MESSAGES_FILE |
 | :white_check_mark: | warning | Web extension | When _locales directory exists, default_locale must exist | | | null | NO_DEFAULT_LOCALE |
+| :white_check_mark: | warning | Web extension | desc | | | | UNSAFE_VAR_ASSIGNMENT |
```
mstriemer: "desc" doesn't seem very descriptive. This code also isn't being set right now so maybe this isn't needed?

EnTeQuAk: 😆 "desc" was me counting columns haha. I'll fix that.

tests/rules/javascript/test.no_unsafe_innerhtml.js
mstriemer: Looks like there are a bunch of imports like this in the code base so these can be ignored.

src/scanners/javascript.js
```diff
@@ -80,10 +83,15 @@ export default class JavaScriptScanner {
           var code = message.message;
 
           // Support 3rd party eslint rules that don't have our internal
-          // message structure.
+          // message structure and allow us to optionally overwrite
+          // their `message` and `description`
```
mstriemer: nit: this should end with a period.

src/scanners/javascript.js
mstriemer: Should the code also be getting set?

src/scanners/javascript.js
```diff
@@ -1,6 +1,9 @@
 import ESLint from 'eslint';
 
-import { ESLINT_RULE_MAPPING, ESLINT_TYPES } from 'const';
+import {
+  ESLINT_RULE_MAPPING, ESLINT_TYPES,
```
mstriemer: I think these should all be on the same line or have 1 per line.

src/scanners/javascript.js
EnTeQuAk: No, the code will go away eventually.

src/messages/javascript.js
EnTeQuAk: There is currently a test that requires it and since it won't matter I just added it for now until we get rid of codes in favor of rule-ids.

PR #1195, opened 2017-03-22
Show all error messages and hide duplicates

### LINKED ISSUE #1194: Show all unique error messages
We currently only show the first error message for a specific JSON pointer. This helps reduce the number of errors and likely improves the readability of the errors but some errors are more useful than others.

For example:

> "/background/scripts" is not a valid key or has invalid extra properties

This error message can be useful, but it is not as useful as the one it is hiding:

> "/background/scripts" should be array

We should only show `additionalProperties` errors if there are no other errors for a JSON pointer.

Supports #1131 and #765.

mstriemer:
With `"background": { "scripts": "foo.js" }` in the schema, an invalid permission and an invalid extension id...

Error messages using Firefox schemas:

```
errors          7              
notices         0              
warnings        3              

ERRORS:

Code                      Message                                Description                                                   File            Line   Column
JSON_INVALID              "/applications/gecko/id" should        Your JSON file could not be parsed.                                                        
                          match pattern                                                                                                                     
                          "^\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f…                                                                                              
                          ]{4}-[0-9a-f]{4}-[0-9a-f]{12}\}$"                                                                                                 
JSON_INVALID              "/applications/gecko/id" should        Your JSON file could not be parsed.                                                        
                          match pattern                                                                                                                     
                          "^[a-z0-9-._]*@[a-z0-9-._]+$"                                                                                                     
JSON_INVALID              "/applications/gecko/id" should        Your JSON file could not be parsed.                                                        
                          match some schema in anyOf                                                                                                        
JSON_INVALID              "/background/scripts" is not a valid   Your JSON file could not be parsed.                                                        
                          key or has invalid extra properties                                                                                               
MANIFEST_FIELD_REQUIRED   "/background/page" is a required       See https://mzl.la/1ZOhoEN (MDN Docs) for more information.   manifest.json                
                          property                                                                                                                          
MANIFEST_FIELD_INVALID    "/background/scripts" should be        See https://mzl.la/1ZOhoEN (MDN Docs) for more information.   manifest.json                
                          array                                                                                                                             
JSON_INVALID              "/background" should match some        Your JSON file could not be parsed.                                                        
                          schema in anyOf                                                                                                                   
WARNINGS:

Code                   Message                             Description                                                              File                   Line   Column
MANIFEST_PERMISSIONS   /permissions: Unknown permissions   See https://mzl.la/1R1n1t0 (MDN Docs) for more information.              manifest.json                       
                       "wat" at 5.                                                                                                                                      
APP_GETDETAILS         "app.getDetails" is deprecated or   This API has been deprecated by Chrome and has not been implemented by   archive.js             36     13    
                       unimplemented                       Firefox.                                                                                                     
ALREADY_SIGNED         Package already signed              Add-ons which are already signed will be re-signed when published on     META-INF/manifest.mf                
                                                           AMO. This will replace any existing signatures on the add-on.                                                

```

Fixes #1194.

### DIFF
--- a/tests/test.collector.js
+++ b/tests/test.collector.js
@@ -1,7 +1,7 @@
 import { default as Collector } from 'collector';
 import { fakeMessageData } from './helpers';
 
-describe.only('Collector', function() {
+describe('Collector', function() {
 
   it('should be thrown an error if Message is created without a type', () => {
     assert.throws(() => {
--- a/tests/test.collector.js
+++ b/tests/test.collector.js
@@ -65,6 +65,69 @@ describe('Collector', function() {
     assert.equal(collection.notices.length, 0);
   });
 
+  it('should not add a duplicate dataPath', () => {
+    var collection = new Collector();
+    collection.addWarning({ ...fakeMessageData, dataPath: '/foo' });
+    collection.addWarning({ ...fakeMessageData, dataPath: '/foo' });
+    assert.equal(collection.warnings.length, 1);
+    assert.equal(collection.warnings[0].type, 'warning');
+    assert.equal(collection.errors.length, 0);
+    assert.equal(collection.notices.length, 0);
+  });
+
+  it('should overwrite an old message about additionalProperties', () => {
+    var collection = new Collector();
+    collection.addWarning({
+      ...fakeMessageData,
+      keyword: 'additionalProperties',
+      dataPath: '/foo',
+    });
+    collection.addWarning({
+      ...fakeMessageData,
+      keyword: 'type',
+      dataPath: '/foo',
+    });
+    assert.equal(collection.warnings.length, 1);
+    assert.equal(collection.warnings[0].keyword, 'type');
+    assert.equal(collection.errors.length, 0);
+    assert.equal(collection.notices.length, 0);
+  });
+
+  it('should allow the same dataPath in different message types', () => {
+    var collection = new Collector();
+    collection.addWarning({
+      ...fakeMessageData,
+      dataPath: '/foo',
+    });
+    collection.addError({
+      ...fakeMessageData,
+      dataPath: '/foo',
+    });
+    assert.equal(collection.warnings.length, 1);
+    assert.equal(collection.errors.length, 1);
+    assert.equal(collection.notices.length, 0);
+  });
+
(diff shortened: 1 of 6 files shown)


mstriemer:
This seems like an improvement but there's probably more we can do. I just realised there's a similar issue with `anyOf` and the extension id.

If you have `"id": "foo@"` then with the Firefox schemas you'll just get `"/applications/gecko/id" should match pattern "^\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\}$"` which shows that it can be a guid, but we're hiding `"/applications/gecko/id" should match pattern "^[a-z0-9-._]*@[a-z0-9-._]+$"` and `"/applications/gecko/id" should match some schema in anyOf`.

The last one isn't very useful so we could probably hide `anyOf` and `additionalProperties` errors if there are other errors at that JSON pointer.

Actually, I'm not sure how overly verbose this is. Maybe we just shouldn't hide any errors unless we find specific cases where it's confusing?

mstriemer:
I changed this so that it just hides duplicate messages. This makes the old tests pass so you no longer get 38 warnings for an invalid id (which seems like a good thing).

Multiple messages on the same JSON pointer are maintained. It might be worth hiding the additionalProperties and anyOf messages if there are other messages at the same JSON selector but we can handle that later.

tests/test.collector.js
```diff
@@ -65,6 +65,30 @@ describe('Collector', function() {
...
+  });
+
+  it('should add a message that differs on one prop', () => {
+    var collection = new Collector();
+    collection.addWarning({ ...fakeMessageData, dataPath: '/foo' });
+    collection.addWarning({
+      ...fakeMessageData,
+      dataPath: '/foo',
+      message: 'Foo message',
+    });
```
EnTeQuAk: This explicitly only tests `message` but for regular messages, say, some kind of warning that happens like 10x in a file mostly only the `line` is different afaik (I could be completely wrong though). So can we maybe add a test and make sure that we don't de-duplicate messages where only the `line` (or other properties) are because we should keep them.

PR #1201, opened 2017-03-27
Configure flagged file extensions in an Array.

### LINKED ISSUE #1115: Generate flagged files Regexp from an array
Right now our flagged files are stored directly as a regex in our constants file, but as mentioned it https://github.com/mozilla/addons-linter/pull/1114/files#diff-c4c3d3aa75ed8ae6e651ade4d628296aR128 we could just make that regex out of an array which would be a nicer API for developers. We should do that.

EnTeQuAk:
Fixes #1115 

r? @tofumatt

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -134,8 +134,9 @@ export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|\~$/i;
 export const ALREADY_SIGNED_REGEX = /^META\-INF\/manifest\.mf/;
 
-export const FLAGGED_FILE_EXTENSION_REGEX =
-  /\.exe$|\.dll$|\.dylib$|\.so$|\.sh$|\.class$|\.swf$|\.jar$/i;
+export const FLAGGED_FILE_EXTENSIONS = [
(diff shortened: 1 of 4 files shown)


src/const.js
```diff
@@ -134,8 +134,9 @@ export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|\~$/i;
 export const ALREADY_SIGNED_REGEX = /^META\-INF\/manifest\.mf/;
 
-export const FLAGGED_FILE_EXTENSION_REGEX =
-  /\.exe$|\.dll$|\.dylib$|\.so$|\.sh$|\.class$|\.swf$|\.jar$/i;
+export const FLAGGED_FILE_EXTENSIONS = [
+  '.exe', '.dll', '.dylib', '.so', '.sh', '.class', '.swf', '.jar',
```
tofumatt: Nitpick, I'd probably but one item per line but whatevs! 😆

src/scanners/filename.js
```diff
@@ -7,6 +9,8 @@ export default class FilenameScanner extends BaseScanner {
 
   scan() {
     return new Promise((resolve) => {
+      var ext = extname(this.filename);
```
tofumatt: I'd use `const` here instead of `var` (I know some of the linter code was using `var`, but might as well use `const/let` going forward).

Also I'd call it `extension`, we have the space!

Issue #1209, opened 2017-04-04
Handle Firefox schemas with multiple namespaces


mstriemer:
There are some schemas in Firefox that have multiple namespaces defined. Specifically privacy which defined "privacy" (empty), "privacy.network" and "privacy.websites"; and omnibox which defines "omnibox" and "omnibox_internal".

We either need to handle this by creating multiple files somehow or updating these schemas in Firefox if that makes sense.

PR #1210, opened 2017-04-04
Script to import schemas from Firefox

mstriemer:
This adds support for pulling the schema files from Firefox and importing them. It removes src/schemas/firefox folder since it was only used during the import and that folder it now created and removed automatically by the script.

I noticed that the new privacy schema actually defined two schemas that are needed. Looking at the old schemas it looks like omnibox is as well. This is a bit of an issue for us unfortunately since we are only keeping one namespace from each file. I filed #1209 to handle this.

First commit is the changes to the script, and the next two bring us up to date with the latest aurora. I could update this to run off of beta now or we can update it once it meets its demise. Using the latest non-nightly build made sense to me though.


mstriemer:
There was an issue with the update for firefox 54 so I added an extra commit. This should be good for review now.

src/schema/firefox-schemas-import.js
```diff
@@ -324,9 +347,35 @@ function loadSchemasFromFile(basePath) {
...
+export function importSchemas(firefoxPath, ourPath, importedPath) {
   const rawSchemas = loadSchemasFromFile(firefoxPath);
   const ourSchemas = readSchema(ourPath, 'manifest.json');
   const processedSchemas = processSchemas(rawSchemas, ourSchemas);
-  writeSchemasToFile(firefoxPath, processedSchemas);
+  writeSchemasToFile(firefoxPath, importedPath, processedSchemas);
+}
+
+function downloadUrl(version) {
+  return `https://hg.mozilla.org/mozilla-central/archive/FIREFOX_AURORA_${version}_BASE.tar.gz`;
```
mstriemer: I only see tags for aurora on [mozilla-central](https://hg.mozilla.org/mozilla-central/tags), but it's possible I'm looking in the wrong spot. I'm guessing those will start being FIREFOX_BETA_<version>_BASE.tar.gz at some point.

We still need to run this script manually for the time being so when the URL changes we can update it and run it.

mstriemer: As in, I don't expect it to change other than that one time so updating the source is probably the easiest and most convenient.

PR #1215, opened 2017-04-07
package up json files in final build

EnTeQuAk:
cc @mstriemer on this, I'm not too experienced with webpack and I'm running into errors building with this…

```
grunt build
Running "webpack:build" (webpack) task
Version: webpack 1.14.0     
               Asset    Size  Chunks             Chunk Names
    addons-linter.js  269 kB       0  [emitted]  main
addons-linter.js.map  267 kB       0  [emitted]  main

ERROR in ./~/json-loader!./src/schema/manifest-schema.json
Module build failed: SyntaxError: Unexpected token m in JSON at position 0
  at JSON.parse (<anonymous>:null:null)
  at Object.module.exports (/home/chris/Projects/addons-linter/node_modules/json-loader/index.js:7:48)
 @ ./src/schema/linter-validator.js 16:42-80

ERROR in ./~/json-loader!./src/schema/imported/manifest.json
Module build failed: SyntaxError: Unexpected token m in JSON at position 0
  at JSON.parse (<anonymous>:null:null)
  at Object.module.exports (/home/chris/Projects/addons-linter/node_modules/json-loader/index.js:7:48)
 @ ./src/schema/firefox-validator.js 23:19-59
```

EDIT:

* Upgrades to WebPack 2
* Fixes various json import problems, explicitly requires them in the bundle
* unfortunately increases bundle size to 6xxkB but I feel like that's alright since it's not being used in the browser 
* Finally works!


EnTeQuAk:
Also cc @tofumatt for more magic ideas :)

mstriemer:
This error is just because there are some imports like `const schemaObject = require('json!schema/imported/manifest');` They can switch to `import schemaObject from 'schema/imported/manifest.json';` now.

This won't get the schemas in the bundle though, since they're being read with `fs.readFile`. It looks like you can use something like `require.context` but it appears as though webpack won't do anything with that so maybe still won't end up in the bundle.

Why do we have the bundle, exactly?

tofumatt:
I tried that though it caused the coverage tests to fail, oddly. :-/

tofumatt:
We use `fs.readFile` if I recall correctly because we need to be able to use commentJSON. Actually, I wonder if the issue is some of manifests have comments and we should be reading all JSON use `cjson`?

mstriemer:
We only need to deal with comments when they get imported from Firefox in the import script. Using the JSON loader should work fine.

Using `require.context` appears to import the schemas into the bundle and makes the code read much better but it just seems to move the error to runtime from build time.

EnTeQuAk:
I am still failing to fix this properly or to figure out what's wrong here. If anyone has an idea or a few minutes to spare to try things out I'd hugely appreciate any help!

mstriemer:
Why are you trying to include the JSON files in the bundle? They don't need to be transpiled so I don't think there's a need to include them?

EnTeQuAk:
Because they aren't packaged up in our npm package so that you get an
error when starting the addons linter. (feel free to try and install the
latest version and start it, you should get an error)




On Mon, Apr 10, 2017, at 11:48 PM, Mark Striemer wrote:

> Why are you trying to include the JSON files in the bundle? They don't
> need to be transpiled so I don't think there's a need to include them?
> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].
> 


Links:

  1. https://github.com/mozilla/addons-linter/pull/1215#issuecomment-293089103

EnTeQuAk:
@mstriemer see https://travis-ci.org/mozilla/addons-server/jobs/219572423 for what I'm talking about

EnTeQuAk:
I think I finally managed to figure this out, @tofumatt or @mstriemer r? Would be nice to get this in in today's tag if possible :)

The schema objects are now in the final build and tests pass so it's one step closer to work as expected :)

EnTeQuAk:
Interesting enough, `grunt test-no-coverage` works but regular `npm test` doesn't…

EnTeQuAk:
Upgraded to webpack2 and now this appears to be working more or less as expected. At least the tests work now, including coverage (let's see if travis agrees).

Now it's only fixing `firefox-schemas.js` and we should be good to go.

tasks/webpack.js
```diff
@@ -37,12 +37,16 @@ module.exports = {
...
-        exclude: /(node_modules|bower_components|test)/,
-        loader: 'babel-istanbul',
-      }],
-    }),
+    module: {
+      rules: [
+        {
+          use: 'babel-istanbul-loader',
+          // babel options are in .babelrc
+          exclude: /(node_modules|bower_components)/,
```
mstriemer: This probably needs to ignore tests too.

EnTeQuAk:
@tofumatt mind giving https://github.com/mozilla/addons-linter/pull/1215 a final r? It's finally green and appears to be working! :)

PR #1229, opened 2017-04-13
Handle namespace collisions in Firefox schemas

### LINKED ISSUE #1209: Handle Firefox schemas with multiple namespaces

mstriemer:
This adds support for handling namespace collisions in different schema files and multiple schemas that need importing in the same file.

I also added support for reading the tarball from the filesystem since I had to run this a few times.

Fixes #1209.


bin/firefox-schema-import
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+// vim: set ft=javascript:
```
EnTeQuAk: Hmm, shouldn't vim recognize this via the shebang already? I'm -1 settings like that tbh.

src/schema/firefox-schemas-import.js
```diff
@@ -26,6 +26,9 @@ export const refMap = {
 // Reference some functions on inner so they can be stubbed in tests.
 export const inner = {};
 
+// Consider moving this to a Set if you add more schema namespaces.
+export const ignoredSchemas = ['omnibox_internal'];
```
EnTeQuAk: Why are some ignored if I may ask? Maybe worth noting in the comment.

src/schema/firefox-schemas-import.js
```diff
@@ -250,24 +253,108 @@ export function rewriteExtend(schemas, schemaId) {
...
+    return schemas;
+  }
+
+  // There is folding to do, join the matching schemas.
+  const foldedSchemas = [];
+
+  // The order of the schemas will be maintained since they were inserted in
+  // the order of schemas.
+  Object.keys(schemasByPrefix).forEach((namespace) => {
+    const { __base = {}, ...nestedSchemas } = schemasByPrefix[namespace];
```
EnTeQuAk: Not a fan of actually calling this `__base` instead of ``baseNamespace`` but I don't have too strong opinions about it tbh… just noticed it :)

src/schema/firefox-schemas-import.js
```diff
@@ -290,18 +377,38 @@ inner.loadSchema = (schema, file) => {
...
+  return schemas;
+};
+
+export function processSchemas(schemas) {
   const loadedSchemas = {};
   schemas.forEach(({ file, schema }) => {
     // Convert the Firefox schema to more standard JSON schema.
     const loadedSchema = inner.loadSchema(schema, file);
-    loadedSchemas[loadedSchema.id] = { file, schema: loadedSchema };
+    const { id } = loadedSchema;
```
EnTeQuAk: It actually took me a while to notice the difference between loadedSchema and loaderSchema**s**, can we maybe call the ``id`` ``loadedSchemaId`` or at least ``loadedId`` or even work directly with ``loadedSchema.id`` instead?

bin/firefox-schema-import
mstriemer: It should, I added a plugin to set it in my vim and it still isn't working though :(

I'll remove this and figure that out on my own though.

src/schema/firefox-schemas-import.js
mstriemer: This schema namespace isn't actually exposed to add-ons, or they shouldn't be using it if it is. I'll update the comment.

src/schema/firefox-schemas-import.js
mstriemer: I figured it would be best to call it something that wouldn't be used by a schema.

It could be `baseNamespace`, that does seem pretty unlikely.

Issue #1152, opened 2017-03-02
Incorrect error message for duplicated permission


vasilicamihasca:
**Steps to reproduce:**
Submit the attached webextension on AMO-dev: 
[duplicatedpermission.zip](https://github.com/mozilla/addons-linter/files/813534/duplicatedpermission.zip)

**Expected results:**
The proper error message about the duplicated permission is successfully displayed.

**Actual results:**
The following error appears in Validation Report: "Permissions must be strings. If a permission is an object, it's likely from a Chrome App Extension and will not work in Firefox." 

**Additional notes:**
- This issue reproduces on AMO-dev using Firefox 54.0a1 (2017-03-01) under Windows 10 64-bit.
- Please see the attached screenshots for this issue:
![2017-03-02_1248](https://cloud.githubusercontent.com/assets/18114210/23504120/dd952818-ff46-11e6-9036-82f0f4065135.png)
![2017-03-02_1249](https://cloud.githubusercontent.com/assets/18114210/23504127/e1e3e922-ff46-11e6-996f-ca349d0c0d98.png)

wagnerand:
@vasilicamihasca Could you please see if this is still an issue?

vasilicamihasca:
Retested this issue on all 3 AMO branches (dev,stage and production) on Firefox 57.0a1 (2017-08-10) under Windows 10 64-bit and seems that the initial error is no longer reproducible and the following message is displayed in its place:     Permissions should NOT have duplicate items (items ## 0 and 1 are identical). 

@wagnerand, I have some doubts related to 2 things:
- Is it ok how the “##” are displayed? Is this a standard format?
- Having more than 2 identical permissions, they will not be all specified in the same error. They are grouped by 2. Is this intended? Or all duplicated items should be mentioned in one error? (Tested with the attached webextension, which has 5 identical permissions [duplicatedpermission1.zip](https://github.com/mozilla/addons-linter/files/1217384/duplicatedpermission1.zip))

![2017-08-11_1217](https://user-images.githubusercontent.com/18114210/29207490-0076888c-7e8f-11e7-9442-6c29768b5b00.png)

wagnerand:
It seems like we're using a third party library (*ajv*) to detect and output duplicates. The notation seems ok to me, I am not sure how much we can do about finding multiple duplicate pairs. This might be an upstream bug, but I am not that familiar with *ajv*.

Summoning @mstriemer ...

andymckay:
[...]

PR #1232, opened 2017-04-14
improve message on a duplicated permission

andymckay:
it now says:

screenshot 2017-04-13 17 08 21

When you try and do a non-string, its gets quite mean:

screenshot 2017-04-13 17 09 25

But that's the JSON schema code, we should probably try and compact that down. Anyway combining the error the JSON schema returns and simplifying the description down to "go read the docs" seems ok to me.

Supports #1152


mstriemer:
We might want to revise the changes from #1195 to also hide the duplicates here. I Would have imagined we'd only see the "should be a string" message once here but the output is the same so we have no need to show it.

I filed #1240 to remove the duplicate messages.

Just sending people to the docs sounds good to me.

PR #1237, opened 2017-04-20
Add error to reject static theme uploads. Fixes #1213

### LINKED ISSUE #1213: Auto-reject static themes
AMO isn't ready to support static (WebExtension) themes yet, so we need the linter to automatically reject them. We need to detect the `theme` key in the manifest, [like here](https://github.com/andymckay/webextensions-examples/blob/static-theme/static-theme/manifest.json#L9-L17), and throw an error with the following message: `WebExtension theme uploads are currently not supported.`

EnTeQuAk:
There's one open question in https://github.com/mozilla/addons-linter/issues/1213#issuecomment-295657175 that needs to be answered before this can be merged.

But since I already did the work I thought I can open the PR to get initial feedback.


EnTeQuAk:
Closing in favor of https://github.com/mozilla/addons-server/issues/5210

PR #1238, opened 2017-04-20
Use Firefox schemas for manifest validation

### LINKED ISSUE #1131: Replace existing manifest-schema.json with mozilla-central schema
Once we have schemas imported in #1129 and updated with our extra metadata in #1130 we should be able to start using the schema data that was imported from Firefox.

### LINKED ISSUE #1196: Fix output from ajv with bin/addons-linter
When you run addons-linter with the Firefox schemas there is some output about the format property of some fields. We should define these formats so the output is supressed.

mstriemer:
1. Update the manifest validation to use the Firefox schemas version.
2. Clean up the noise that was generated about missing formats when running the validator with the Firefox schemas version.
3. Rewrite the case-insensitive regexes from the Firefox version.
4. Fix up TODOs related to the Firefox schemas.

Fixes #1131.
Fixes #1196.


src/schema/firefox-schemas-import.js
```diff
@@ -8,6 +8,12 @@ import request from 'request';
 import tar from 'tar';
 
 const FLAG_PATTERN_REGEX = /^\(\?[im]*\)(.*)/;
+export const FLAG_PATTERN_REWRITES = {
+  '(?i)^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$':
+    // eslint-disable-next-line max-len
```
EnTeQuAk: Can we maybe add that to lines 11-16 to avoid having the comment in the midst of the regex?

src/schema/firefox-schemas-import.js
```diff
@@ -8,6 +8,12 @@ import request from 'request';
...
+    '^\\{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\\}$',
+  '(?i)^[a-z0-9-._]*@[a-z0-9-._]+$': '^[a-zA-Z0-9-._]*@[a-zA-Z0-9-._]+$',
+};
```
EnTeQuAk: Can you add a example of what the regex matches as a comment or maybe put the separate regexes into constants to make the object easier to read? Not sure what I prefer tbh… I think I am generally missing some details of why the rewrite is even needed. Maybe a comment solves my problem too :)

mstriemer:
I added some comments about why the object exists and to what each line is supposed to match. I think that's probably clearer than breaking them out into named variables.

PR #1243, opened 2017-04-22
warn on inline script tags in HTML

### LINKED ISSUE #68: Deal with inline JS and script blocks when parsing the HTML
We're going to need to handle any JS we come across whilst processing the HTML files.

andymckay:
* the prefwindow test wasn't very useful because that was for XUL add-ons, so lets remove that
* re-use that code into a checking for script missing a src or nonce
* nonce can be used to allow inline with CSP from what I gather https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script, but perhaps we should just not even bother about that
* this should fix #68


andymckay:
Oh wait, the nonce only works if you change the CSP declaration, so again you have to change the CSP. Sorry I misunderstood the MDN page. I'll remove nonce from this and make it a teeny bit shorter.

PR #1249, opened 2017-04-24
Handle allOf with optional

### LINKED ISSUE #1245: run_at declared as mandatory while not mandatory in the documentation
Running addons linter on an app with a manifest.json that does not contain a "run_at" property in the "content_scripts" section throws an error, while the documentation clearly states that the only mandatory field is "matches":

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/content_scripts

The error object looks like this:

```javascript
{ keyword: 'required',
  dataPath: '/content_scripts/0/run_at',
  schemaPath: '#/required',
  params: { missingProperty: 'run_at' },
  message: 'is a required property',
  schema:
   { matches: { type: 'array', minItems: 1, items: [Object] },
     exclude_matches: { type: 'array', minItems: 1, items: [Object] },
     include_globs: { type: 'array', items: [Object] },
     exclude_globs: { type: 'array', items: [Object] },
     css:
      { type: 'array',
        description: 'The list of CSS files to inject',
        items: [Object] },
     js:
      { type: 'array',
        description: 'The list of CSS files to inject',
        items: [Object] },
     all_frames:
      { type: 'boolean',
        description: 'If allFrames is <code>true</code>, implies that the JavaScript or CSS should be injected into all frames of current page. By default, it\'s <code>false</code> and is only injected into the top frame.' },
     match_about_blank:
      { type: 'boolean',
        description: 'If matchAboutBlank is true, then the code is also injected in about:blank and about:srcdoc frames if your extension has access to its parent document. Code cannot be inserted in top-level
about:-frames. By default it is <code>false</code>.' },
     run_at: { allOf: [Object] } },
  parentSchema:
   { type: 'object',
     description: 'Details of the script or CSS to inject. Either the code or the file property must be set, but both may not be set at the same time. Based on InjectDetails, but using underscore rather than camel case naming conventions.',
     properties:
      { matches: [Object],
        exclude_matches: [Object],
        include_globs: [Object],
        exclude_globs: [Object],
        css: [Object],
        js: [Object],
        all_frames: [Object],
        match_about_blank: [Object],
        run_at: [Object] },
     required: [ 'matches', 'run_at' ] },
  data:
   { matches: [ '*://*/*' ],
     css:
      [ 'data/css/lib/cssreset-context-min.css',
        'data/css/lib/opensans.css',
        'data/css/contentscript.css' ],
     js:
      [ 'data/js/lib/tether.min.js',
        'data/js/lib/drop.min.js',
        'data/js/lib/jquery.min.js',
        'data/js/formfill-browser-client.js',
        'data/js/formfill.js' ] } }
```

(Especially this part is interesting: **required: [ 'matches', 'run_at' ]** )

### What happened?

Error gets thrown with the following details:

Code: MANIFEST_FIELD_REQUIRED
Message: "/content_scripts/0/run_at" is a required property
Description: See https://mzl.la/1ZOhoEN (MDN Docs) for more information.
manifest.json

That no error gets thrown as this field is not a mandatory field

mstriemer:
This handles parsing schemas with an `allOf` that has an optional schema in it. This was causing an error on `content_scripts` without a `run_at` property.

Fixes #1245.

### DIFF
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -68,14 +68,17 @@ export function rewriteOptionalToRequired(schema) {
     const value = schema[key];
     if (!Array.isArray(value) && typeof value === 'object') {
       const { optional, ...rest } = value;
+      // If there is an allOf, check the inner schemas for optional.
       if (Array.isArray(rest.allOf)) {
         let someOptional = false;
+        // Update the inner schemas to remove the optional property and record
+        // if we found any optional schemas. See issue #1245.
         rest.allOf = rest.allOf.map((inner) => {
-          someOptional = someOptional || inner.optional;
-          const innerCopy = { ...inner };
-          delete innerCopy.optional;
-          return innerCopy;
+          const { optional: innerOptional, ...innerRest } = inner;
+          someOptional = someOptional || innerOptional;
+          return innerRest;
         });
+        // If none of the inner schemas are optional then this property is required.
         if (!someOptional) {
           required.push(key);
         }
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -68,7 +68,18 @@ export function rewriteOptionalToRequired(schema) {
     const value = schema[key];
     if (!Array.isArray(value) && typeof value === 'object') {
       const { optional, ...rest } = value;
-      if (!optional) {
+      if (Array.isArray(rest.allOf)) {
+        let someOptional = false;
+        rest.allOf = rest.allOf.map((inner) => {
+          someOptional = someOptional || inner.optional;
+          const innerCopy = { ...inner };
+          delete innerCopy.optional;
+          return innerCopy;
+        });
+        if (!someOptional) {
+          required.push(key);
+        }
+      } else if (!optional) {
         required.push(key);
       }
       return { ...obj, [key]: rest };
--- a/tests/schema/test.firefox-schemas-import.js
+++ b/tests/schema/test.firefox-schemas-import.js
@@ -83,7 +83,7 @@ describe('firefox schema import', () => {
       });
     });
 
-    it('handles an allOf with one being optional', () => {
+    it('handles an allOf with one nested schema being optional', () => {
       const obj = {
         foo: { type: 'string', optional: true },
         bar: {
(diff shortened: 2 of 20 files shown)


tests/schema/test.firefox-schemas-import.js
```diff
@@ -82,6 +82,50 @@ describe('firefox schema import', () => {
...
+          allOf: [
+            { $ref: '#/types/Whatever' },
+            { type: 'string' },
+          ],
+        },
+        baz: { type: 'boolean' },
+      };
+      assert.deepEqual(rewriteOptionalToRequired(obj), {
+        foo: { type: 'string' },
+        bar: { allOf: [
```
tofumatt: Nitpick but it would be nice if this followed the layout the same as above to visually compare them.

tests/schema/test.firefox-schemas-import.js
```diff
@@ -82,6 +82,50 @@ describe('firefox schema import', () => {
         required: [],
       });
     });
+
+    it('handles an allOf with one being optional', () => {
```
tofumatt: I think you should say "with one property being optional" or something like that because this isn't really clear.

tests/schema/test.firefox-schemas-import.js
```diff
@@ -82,6 +82,50 @@ describe('firefox schema import', () => {
...
+        bar: { allOf: [
+          { $ref: '#/types/Whatever' },
+          { description: 'a thing' },
+        ]},
+        baz: { type: 'boolean' },
+        required: ['baz'],
+      });
+    });
+
+    it('handles an allOf with none being optional', () => {
```
tofumatt: Same here; this should be "with no property flagged as optional" or something like that.

src/schema/firefox-schemas-import.js
```diff
@@ -68,7 +68,18 @@ export function rewriteOptionalToRequired(schema) {
     const value = schema[key];
     if (!Array.isArray(value) && typeof value === 'object') {
       const { optional, ...rest } = value;
-      if (!optional) {
+      if (Array.isArray(rest.allOf)) {
```
tofumatt: This is a bit of an odd to read chunk of code. I get it after looking at the issue, but I wouldn't mind a link back to the issue just to explain the reasoning behind it.

mstriemer:
I had added a test to reproduce the bug earlier and forgot to add it to git. It's added now. I gave it a once over and it looks okay but let me know if something is amiss and I can update it.

I also refactored the `allOf.map` loop to use destructuring a bit better and avoid deleting the property explicitly. I think it makes it a bit simpler. Added some comments too so it should be easier to understand now.

tofumatt:
Looks good to me 👍

PR #1272, opened 2017-05-08
Drop tests on node4

wagnerand:
```
<TheOne> muffinresearch: have you folks talked about dropping tests on node4, now that node6 has been on LTS for a while? Tests on node4 take significantly longer
<@muffinresearch> yeah we already have elsewhere
<@muffinresearch> So feel free to do that anywhere that needs it.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-- '4'
 - '6'
 script: npm test
 notifications:


muffinresearch:
@wagnerand before we land this would be good to check that web-ext isn't relying on node4.

muffinresearch:
See https://github.com/mozilla/web-ext#installation-from-source

wagnerand:
Hm, both `dispensary` and `web-ext` tests pass on node 6. Is that enough?

muffinresearch:
> Hm, both dispensary and web-ext tests pass on node 6. Is that enough?

If web-ext says it supports node 4 and then the linter doesn't we could break them. So I think before going ahead, it would be good to look at getting web-ext updated to support a similar policy of node 6+ support.

wagnerand:
Ah, got it.

Issue #1252, opened 2017-04-26
fetchSchemas() does not report stream errors


kumar303:
In passing, I noticed that the promise returned by `fetchSchemas()` of `src/schema/firefox-schemas-import.js` is not rejected when there is a stream error. It is [resolved here](https://github.com/mozilla/addons-linter/pull/1210/files#diff-44cfe69567cbcc640cec70b9fde0adebR378) but it should probably reject the promise like this:

````js
request.get(downloadUrl(version))
  .pipe(...)
  .on('error', (error) => {
    reject(error);
  })
  .on('end', () => {
    resolve();
  });
````

...otherwise any errors may go unreported which would be confusing.

muffinresearch:
This one should be fixed as part of #1278

caitmuenster:
@muffinresearch Does the pull request cover this issue? Not sure if this is duped or not.

PR #1278, opened 2017-05-10
Mozilla greenkeeper/tar 3.0.0

mstriemer:
So things changed a bit with the rewrite. Not too bad except it now seems to include null bytes in filenames when using the `Parse` class which breaks `fs.createWriteStream` and took me some time to figure out. Also the parsing won't chain off the stream from `request` anymore so it now writes to file and uses that.

Closes #1277.

### DIFF
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -1,6 +1,5 @@
 import fs from 'fs';
 import path from 'path';
-import zlib from 'zlib';
 
 import commentJson from 'comment-json';
 import merge from 'deepmerge';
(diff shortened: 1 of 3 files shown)


src/schema/firefox-schemas-import.js
```diff
@@ -504,26 +503,59 @@ inner.isBrowserSchema = (path) => {
...
+            if (inner.isBrowserSchema(entryPath)) {
+              const filePath = path.join(outputPath, path.basename(entryPath));
+              entry.pipe(fs.createWriteStream(filePath));
+            } else {
+              entry.resume();
+            }
+          })
+          .on('end', () => {
+            resolve();
+          });
```
muffinresearch: Would be nice to take the opportunity to listen to error events and reject there too.

muffinresearch: That would fix https://github.com/mozilla/addons-linter/issues/1252

muffinresearch:
Version 3.0.1 is released now too.

muffinresearch:
Looks like there's a test fail here in any case.

mstriemer:
This passes locally for me =\. I should have time later this week to look at it again though.

EnTeQuAk:
@mstriemer did you have time to look at this again? I think it'd be great to land this update sometime.

PR #1282, opened 2017-05-13
handle duplicate permissions error messages

### LINKED ISSUE #1240: Handle duplicate permissions error messages
1. Set a permission to a non-string (like `24`).
2. Validate the extension.

### What happened?
You probably got a multiple errors stating that the permission must be a string. This is a bit annoying.

One message stating it should be a string.

There was some work done in #1195 to hide duplicate messages. That should have some leads on where to look to fix this.

ccarruitero:
Fixes #1240 

I did a first attemp, but I have a doubt about how to continue.

I saw that the [validator for the manifest relies on ajv](https://github.com/mozilla/addons-linter/blob/master/src/schema/validator.js#L21). Sadly I can't find a way to get the line where the error occurs from ajv's response. But looking to the response I think that `dataPath` is the most near reference to where the error occurs.

I added a check for `dataPath` for the manifest in order to add just one error to the collector.
But testing a new build with the patch, the error continues because the `dataPath` is [removed](https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L65) when come from permissions in the parser.

So I'm not sure why is the `dataPath` removed in this case? This is the desired functionality?

If we can't get the `dataPath`, I think the other alternative to solve this could be get the line where the error occurs, then we just need give the line to the collector since the collector already [handle that case](https://github.com/mozilla/addons-linter/blob/master/tests/test.collector.js#L78). But should do that patch in [`ajv`](https://github.com/epoberezkin/ajv), right?

Do you have any advice on this?

### DIFF
--- a/src/collector.js
+++ b/src/collector.js
@@ -69,10 +69,14 @@ export default class Collector {
 
   isDuplicateMessage(message) {
     if (message.dataPath) {
-      var previousMessages = this.messagesAtDataPath(message.dataPath);
-      return previousMessages.some((prevMessage) => {
-        return prevMessage.matches(message);
-      });
+      let previousMessages = this.messagesAtDataPath(message.dataPath);
+      if (message.file === 'manifest.json') {
+        return previousMessages.length > 0;
+      } else {
+        return previousMessages.some((prevMessage) => {
+          return prevMessage.matches(message);
+        });
+      }
     }
     return false;
   }
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -62,7 +62,7 @@ export default class ManifestJSONParser extends JSONParser {
                typeof error.data !== 'undefined' &&
                typeof error.data !== 'string') {
       baseObject = messages.MANIFEST_BAD_PERMISSION;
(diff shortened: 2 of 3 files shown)


wagnerand:
@EnTeQuAk Could you help out here?

EnTeQuAk:
oh damn, I must have missed that PR, sorry about that! I'll take a look at this shortly.

EnTeQuAk:
> So I'm not sure why is the dataPath removed in this case? This is the desired functionality?

This got introduced in https://github.com/mozilla/addons-linter/commit/b01faa74b8c2d5e8d4d5084506c1a3532f77098f

As far as I can see it you should be perfectly able to add `dataPath` to the `overrides` and use that later on. I think that idea is great.

Does that help?

EnTeQuAk:
Hey @ccarruitero, did you have time to look at this patch again, can I help with anything else to push this forward?

ccarruitero:
@EnTeQuAk sorry, I missed this.

Thanks for you comments. I will continue with this.

ccarruitero:
@EnTeQuAk  I think the PR is ready for review. Take a look when you have a chance.

PR #1296, opened 2017-05-22
Overwrite eslint default ignore patterns.

### LINKED ISSUE #1288: Do not ignore node-modules/bower_components folder in add-ons
If an add-on includes a `node-modules` folder, each JS file in that folder raises the following warning:
`Warning: File ignored by default. Use "--ignore-pattern '!node_modules/*'" to override`.

We should not ignore contents of that folder.

EnTeQuAk:
This scans node_modules and bower_components. Fixes #1288


src/scanners/javascript.js
```diff
@@ -48,10 +48,15 @@ export default class JavaScriptScanner {
         parserOptions: {
           ecmaVersion: 2017,
         },
-        ignore: false,
         rules: _ruleMapping,
         plugins: ['no-unsafe-innerhtml'],
         allowInlineConfig: false,
+
+        // Disable ignore-mode but overwrite eslint default ignore patterns
+        // since these are matched nonetheless. Fixes #1288
```
tofumatt: Can you link to the actual issue URL here instead of the number? It will make it easier for newbies, etc. to get to the actual issue.

tests/test.linter.js
```diff
@@ -183,6 +183,21 @@ describe('Linter', function() {
...
+      'tests/fixtures/webextension_node_modules_bower']});
+
+    // Stub print to prevent output.
+    addonLinter.print = sinon.stub();
+
+    assert.equal(addonLinter.collector.warnings.length, 0);
+
+    return addonLinter.scan()
+      .then(() => {
+        assert.equal(addonLinter.collector.warnings.length, 0);
```
tofumatt: I would think we would want an error in the example add-on we scan here to verify it's being scanned and not ignored. I'm not sure what this test is proving.

src/scanners/javascript.js
EnTeQuAk: Fixed in 03f931a

tests/test.linter.js
EnTeQuAk: I added some comments about what it's proving in 78e6c41

If I remove the new `patterns` option the test fails with exactly 2 warnings.

EnTeQuAk:
@tofumatt fixed up all the things, mind a re-r?

tests/test.linter.js
```diff
@@ -183,6 +183,38 @@ describe('Linter', function() {
...
+  it('Eslint ignore patterns and .eslintignorerc should be ignored', () => {
+    // Verify https://github.com/mozilla/addons-linter/issues/1288 is fixed
+    var addonLinter = new Linter({_: [
+      'tests/fixtures/webextension_node_modules_bower']});
+
+    // Stub print to prevent output.
+    addonLinter.print = sinon.stub();
+
+    assert.equal(addonLinter.collector.warnings.length, 0);
+    assert.equal(addonLinter.collector.warnings.length, 0);
```
tofumatt: Duplicate line here. Should this be removed or something different?

tests/fixtures/webextension_node_modules_bower/index.js
```diff
@@ -0,0 +1,5 @@
+function dummy(text, callback) {
+  callback(text;
```
tofumatt: I skimmed this and didn't immediately see the syntax error, can you just add a comment explaining it? Thanks!

tests/test.linter.js
```diff
@@ -183,6 +183,38 @@ describe('Linter', function() {
...
+
+    var getFileSpy = sinon.spy(addonLinter, 'scanFile');
+
+    return addonLinter.scan()
+      .then(() => {
+        // Ensure no warnings are raised. If we would not overwrite eslint
```
tofumatt: This wording is a bit awkward, maybe try:

```js
// Ensure no warnings are raised. We override eslint's `ignore_patterns`
// so eslint scans `bower_components` and `node_modules`, which eslint
// doesn't scan by default.
```

tests/test.linter.js
```diff
@@ -183,6 +183,38 @@ describe('Linter', function() {
...
+        // ignore patterns eslint ignores `node_modules` and
+        // `bower_components` by default. We want them to be scanned though.
+        assert.equal(addonLinter.collector.warnings.length, 0);
```
tofumatt: It just isn't clear to me why `warnings.length` still being `0` proves those folders were scanned. Nothing has changed. Wouldn't you want a file with a warning inside the `node_modules` folder so we can ensure it's scanned and raises an error?

tests/test.linter.js
```diff
@@ -183,6 +183,38 @@ describe('Linter', function() {
...
+
+        // We have an expected SyntaxError in index.js, this verifies
+        // that .eslintignore is also ignored.
+        assert.equal(addonLinter.collector.errors.length, 1);
+
+        // Make sure we also scan them ourselves.
+        assert.ok(getFileSpy.calledWith('index.js'));
+        assert.ok(getFileSpy.calledWith('manifest.json'));
+        assert.ok(getFileSpy.calledWith('node_modules/foo.js'));
```
tofumatt: Ah, but this itself seems pretty good, we shouldn't need to check warnings length again (or at all?) if we're sure the file is passed to the file spy here.

tofumatt: But I still think it would be nice to have a file that causes a warning or error inside bower and node folders.

tests/test.linter.js
EnTeQuAk: Should actually be `errors`, doh!

tests/test.linter.js
EnTeQuAk: I'll see if I can get sinon mocking up to spy into eslint directly but the idea here is, without the new configuration eslint actually throws a warning that the folder is ignored and we assume in the test, if that warning isn't thrown that the folder is scanned. It's a weird way of testing that, I agree.

I think I'll go with the error-approach, raising a error in the node_modules and bower_components folders but keep the test that there are no warnings being raised (thus the folders by default being ignored)

tests/fixtures/webextension_node_modules_bower/index.js
EnTeQuAk: heh, absolutely :)

EnTeQuAk:
@tofumatt this should be ready for another review. Unfortunately turned into a bigger refactor but I hope it's still clear what it's doing. I'm now exposing the scanned filenames from our scanners now so that we can test that more easily, that's the part though that lead to touching many other tests and I started to unify naming things and `scan` vs `getContents` along the way whenever I touched it (and actually had to because of test failures)

src/linter.js
```diff
@@ -383,13 +386,17 @@ export default class Linter {
...
-        for (let message of messages) {
+      .then(({linterMessages, scannedFiles}) => {
+        for (let message of linterMessages) {
           if (typeof message.type === 'undefined') {
             throw new Error('message.type must be defined');
           }
           this.collector._addMessage(message.type, message);
         }
+
+        for (let filename of scannedFiles) {
```
muffinresearch: Should be const

EnTeQuAk: fixed b7c10d1

muffinresearch: Should deal with this everywhere and add the necessary lint rules to prefer const at some point since using const here is a little counter intuitive. See also http://www.ecma-international.org/ecma-262/6.0/index.html#sec-for-in-and-for-of-statements-static-semantics-boundnames

src/collector.js
```diff
@@ -76,6 +77,10 @@ export default class Collector {
     return false;
   }
 
+  recordScannedFile(filename) {
+    this.scannedFiles.push(filename);
```
muffinresearch: I wonder if this should be default behaviour?

src/linter.js
muffinresearch: I think the commit above actually changed a different line.

src/collector.js
muffinresearch: Presumably with some of the scanner running over everything and some runnning over a subset this will have dupes?

src/linter.js
EnTeQuAk: I actually reverted the commit and I think we should do that in a separate pull request if that works for you?

src/collector.js
muffinresearch: I also wonder if the lists need to be categorized by scan type too?

src/linter.js
tofumatt: Changing it everywhere else later is good but I think it's worth changing this one to `const` here. 😄

EnTeQuAk: changed it :)

src/collector.js
tofumatt: Categorising this list by scan type would definitely make it a lot more useful for debugging, yeah.

src/scanners/javascript.js
```diff
@@ -67,6 +76,12 @@ export default class JavaScriptScanner {
       var report = cli.executeOnText(this.code, this.filename, true);
 
       for (const result of report.results) {
+        // eslint prepends the filename with the current workdir, strip
```
tofumatt: Could you write `working directory` instead of `workdir` here? Thanks!

src/scanners/javascript.js
```diff
@@ -48,10 +51,16 @@ export default class JavaScriptScanner {
         parserOptions: {
           ecmaVersion: 2017,
         },
-        ignore: false,
         rules: _ruleMapping,
         plugins: ['no-unsafe-innerhtml'],
         allowInlineConfig: false,
+
+        // Disable ignore-mode but overwrite eslint default ignore patterns
+        // since these are matched nonetheless. Fixes
```
tofumatt: This is still a bit awkwardly phrased. Could it be?

```js
// Disable ignore-mode and overwrite eslint default ignore patterns
// so an add-on's bower and node module folders are included in
// the scan. See: https://github.com/mozilla/addons-linter/issues/1288
```

src/scanners/javascript.js
```diff
@@ -67,6 +76,12 @@ export default class JavaScriptScanner {
       var report = cli.executeOnText(this.code, this.filename, true);
 
       for (const result of report.results) {
+        // eslint prepends the filename with the current workdir, strip
+        // that out.
+        var relPath = path.relative(process.cwd(), result.filePath);
```
tofumatt: Can this be `relativePath` instead? We have the bytes! 😉

tests/rules/test.rdf.js
```diff
@@ -46,36 +38,29 @@ describe('RDF: mustNotExist', () => {
       });
   });
 
-  it('should not allow certain tags contextually (eg. when listed)', () => {
+  it('should not allow certain tags when listed', () => {
```
tofumatt: 👍

tests/test.linter.js
```diff
@@ -183,6 +183,25 @@ describe('Linter', function() {
...
+    // Verify https://github.com/mozilla/addons-linter/issues/1288 is fixed
+    var addonLinter = new Linter({_: [
+      'tests/fixtures/webextension_node_modules_bower']});
+
+    // Stub print to prevent output.
+    addonLinter.print = sinon.stub();
+
+    return addonLinter.scan()
+      .then(() => {
+        assert.deepEqual(addonLinter.collector.scannedFiles, [
```
tofumatt: This looks good but as mentioned above (https://github.com/mozilla/addons-linter/pull/1296/files#r118443364) this is only showing it was scanned at all and not explicitly by eslint from what I can tell. Might be worth putting a JS error back in the `bar.js` and `foo.js` just to make sure they were scanned by eslint.

That and/or categorise the files scanned, but if you wanted to do that in a separate issue and just check for eslint errors for now to get this patch through that works for me.

src/collector.js
muffinresearch: Thinking aloud, maybe an object with filename keys and value of a list containing all scanners used could work and keep it reasonably compact. `Object.keys(obj)` would then get the full file list easily too.

Very crude example:

```js
{ 
  'foo/bar.js': ['js', 'filetype'],
  'bar.css': ['css', 'filetype'],
  'someRandomFile': ['filetype'],
}
```

tofumatt: Yeah, that works for me.

EnTeQuAk: I agree, I'll change it to record a structure similar to that. That'll also resolve the dupes problem and make the method more worthwile to keep around.

EnTeQuAk:
@tofumatt @muffinresearch I hope I fixed all remaining comments, mind giving this a final r?

EnTeQuAk:
hrm, I have no idea why coverage still thinks `scannerName` is untested, gotta check that more in depth or try to find something else to test to get it back up to 100% :-/

src/scanners/base.js
```diff
@@ -18,11 +18,16 @@ export default class BaseScanner {
     return 'string';
   }
 
+  static get scannerName() {
+    throw new Error('scannerName is not implemented');
```
tofumatt: This seems like something that should be checked during class initialisation, not during access.

tests/scanners/test.base.js
```diff
@@ -13,6 +13,16 @@ class BaseScannerWithContents extends BaseScanner {
 
 describe('Base Scanner Class', function() {
 
+  it('scannerName should throw an error', () => {
+    try {
+      var scanner = new BaseScanner('', 'filename.txt');
+      scanner.scannerName;
```
tofumatt: This seems wrong; accessing a property shouldn't throw like this. It seems very much like an unintended side-effect. I think it would be better if during initialisation of this class it checked/validated `this.scannerName` and threw during class creation so:

```js
var scanner = new BaseScanner('', 'filename.txt');
```
would `throw` without having to access the property.

Also you can use `const` here :-)

tests/scanners/test.base.js
```diff
@@ -13,6 +13,16 @@ class BaseScannerWithContents extends BaseScanner {
...
+    } catch (err) {
+      assert.instanceOf(err, Error);
+      assert.equal(err.message, 'scannername is not implemented');
```
tofumatt: It's weird that this works as the actual error is `scannerName is not implemented` (note the camelCase), and it should definitely be camelCase everywhere so it's easier to scan/grok. Maybe `strictEqual` would do the trick?

tests/scanners/test.binary.js
```diff
@@ -32,4 +34,7 @@ describe('Binary', function() {
     assert.equal(BinaryScanner.fileStreamType, 'chunk');
   });
 
+  it('should report a proper scanner name', () => {
+    assert.equal(BinaryScanner.scannerName, 'binary');
```
tofumatt: Should these be constants somewhere? They seem a bit like magic constants if we're using them elsewhere, but if it's just for reporting I guess that's okay. But if they're in more than one file we should probably make them constants.

src/collector.js
```diff
@@ -76,6 +77,18 @@ export default class Collector {
     return false;
   }
 
+  recordScannedFile(filename, scanner) {
+    // TODO: Add some code that verifies and normalizes `filename`
```
tofumatt: Can you add an issue for this and add it to the code? That way we don't forget 😄

src/collector.js
```diff
@@ -76,6 +77,18 @@ export default class Collector {
...
+    // to better avoid duplicates.
+    if (filename in this.scannedFiles) {
+      if (!this.scannedFiles[filename].includes(scanner)) {
+        this.scannedFiles[filename].push(scanner);
```
tofumatt: I think this should be inverted; it's possible to run multiple scanners over the same file, as I understand it (and I think we do already).

I would make the key the `scannerName` and have it contain an array of `filename`s. That way you can see all the files scanned by each scanner. That makes more sense to me.

src/scanners/base.js
EnTeQuAk: This is modelled after how `_getContents` works, also since this is a static method I can't check this properly in the constructor since that'd already be too late.

tests/scanners/test.base.js
EnTeQuAk: This is modelled after how _getContents works, also since this is a static method I can't check this properly in the constructor since that'd already be too late.

I actually had the test the wrong way around so it didn't make any sense and accidentally suceeded.

src/collector.js
EnTeQuAk: The primary focus and usecase of this property is to log the files that were scanned though, the array of scanners is there for debugging. We will have the need to access this list more often in the future (e.g also for new checks in autoapprove stuff heading to AMO) and always reversing it to just get a list of filenames feels wrong to me.

EnTeQuAk: Also… https://github.com/mozilla/addons-linter/pull/1296#discussion_r118454823

src/collector.js
EnTeQuAk: Done :)

tests/scanners/test.binary.js
EnTeQuAk: Yeah, I think they could be constants. Let's do that though once we actually use them somewhere else.

src/scanners/base.js
tofumatt: Ah actually okay seeing the class itself that's better. Thanks!

src/collector.js
```diff
@@ -76,6 +77,19 @@ export default class Collector {
     return false;
   }
 
+  recordScannedFile(filename, scanner) {
+    // TODO: Add some code that verifies and normalizes `filename`
+    // to better avoid duplicates.
+    // See https://github.com/mozilla/addons-linter/issues/1310
+    if (filename in this.scannedFiles) {
```
muffinresearch: This will check the prototype chain too whereas `this.scannedFiles.hasOwnProperty(filename)` won't

tofumatt: Oh shoot, right, I missed that. Should fix it in a future patch.

@EnTeQuAk ^^

EnTeQuAk: ah, I think I read about this. It doesn't seem to be a performance issue though mostly a potential for false-positives. But given that our filenames always have a `.css` or other extensions I doubt that we'll run into that.

Anything else I'm missing?

Issue #1063, opened 2016-12-13
Drop all support for non-WebExtensions add-ons


groovecoder:
### Describe the problem and steps to reproduce it:
I ran the `addons-linter` against my old-school add-on `.xpi` file.
I saw old-school rule violations & error messages.

### What happened?
I worked on the violations & error messages

### What did you expect to happen?
I should see an error message to use amo-validator on old-school add-ons.

### Anything else we should know?
https://github.com/mozilla/addons-linter/pull/1060

muffinresearch:
@andymckay do we have a plan on the Firefox side of things re: dictionaries and langpacks migrating to a web-ext wrapper?

andymckay:
None in the works yet.

muffinresearch:
> None in the works yet.

@andymckay Ok - so if we go ahead and remove all non-web-ext related features from the linter it would mean we'll continue to use the amo-validator for dictionaries and lang-packs until such point that we do something different with them in the future. 

I think that makes sense - do you think that sounds ok? It would be nice to not block on making the linter web-ext only.

andymckay:
Ditching the old validator (eventually) would be nice. I can imagine though prioritizing support for moving them to the linter won't come until we are getting ready to drop the validator as we drop support for legacy add-ons.

wagnerand:
@EnTeQuAk is your branch https://github.com/mozilla/addons-linter/tree/feature/remove-legacy-addons doing this? What keeps us from merging it?

EnTeQuAk:
actually my branch is old and should be deleted. @andymckay did most of the work and I think we're in a pretty good shape now, there's a few more things like `legacyCode` that should go away but maybe we should file separate issues for them?

wagnerand:
Yeah I think it makes sense to file issues and remove code as we come across it. Can't be that much left, can it? 😆

ValentinaPC:
Please add some STR for us to check this or the label "qa:not needed". Thanks!

PR #1308, opened 2017-05-25
remove the rdf scanner and old extension support

### LINKED ISSUE #424: Detect when a package is in a nested directory.
It's easy to package a dir which might result in things like manifest.json being under a directory in the package rather than being at the top level. We should flag that.

### LINKED ISSUE #464: Switch to a proper rdf parser
We only need to parse install.rdf file for dictionaries and langpacks but it would be a good idea to use a better parser.

andymckay:
* supports #1063, but I think there's a bit more clean up to do
* if there's no manifest.json it just throws that as an error
* removes the RDF parser, I don't think its used anywhere else

screenshot 2017-05-25 16 51 03

.. maybe also fixes #424 and we can close #464


EnTeQuAk:
I don't want to be a party pooper but afaik we eventually wanted to move language-pack and theme support from the validator over to the linter given that their support might last quite a bit longer? So we'd still need a rdf parser I think.

tofumatt:
Yeah, it's likely worth keeping RDF support intact, though we probably want to trim down our usage of it for WebExtensions. Maybe we can use some of the code from this patch for that?

andymckay:
Maybe: https://bugzilla.mozilla.org/show_bug.cgi?id=1365709 is about converting to a new webextension format language pack. I admit I don't know about timelines though.

Themes are just going to be webextensions too, heavy weight themes are dead in 57.

jvillalobos:
There are also dictionaries, which also use `install.rdf`.

andymckay:
So you want me to leave the RDF parser there, but just make linter.js just fail and say clearly no manifest.json rather than trying to guess what to do if one isn't present?

tofumatt:
> So you want me to leave the RDF parser there, but just make linter.js just fail and say clearly no manifest.json rather than trying to guess what to do if one isn't present?

I think that makes the most sense for now. When we add support for other add-on types we can still use the RDF parser that way...

andymckay:
Rightio, will do.

muffinresearch:
fwiw (and feel free to correct me if I'm missing something) but I don't see the need to worry too much about removing the RDF parser if it's not being used in the linter today. If we want it back it's in history. It might benefit from re-implementing when we have solid use-case to use it for too.

andymckay:
Going to step back from this back until the path is clearer.

hameleonka:
I'll take this

PR #1327, opened 2017-06-07
Require icons defined in the manifest to exist

### LINKED ISSUE #1026: Validate icons
We should check that images referenced in the manifest are present, we should also check that they are an allowed type (Not sure exactly what this list is, chrome web exts allow any image format supported by webkit [1] MDN docs don't specify allowed formats [2])

We could also go further and check files are the filetype the extension suggests they are.

[1] https://developer.chrome.com/apps/manifest/icons
[2] https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/icons

mstriemer:
Icons weren't being validated at all before, so if you defined an icon but it didn't exist Firefox would try to use it. This makes browser action buttons buggy and shows a broken image next to the add-on name.

We should likely validate the size of the icons and possibly check if they're valid, too. I'll file some more issues.

Fixes #1026.

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -1,4 +1,4 @@
-import { gettext as _, singleLineString } from 'utils';
+import { gettext as _, singleLineString, sprintf } from 'utils';
 import { MANIFEST_JSON } from 'const';
 
 
@@ -102,6 +102,20 @@ export function manifestPropMissing(property) {
     legacyCode: null,
     message: _(`No "${property}" property found in manifest.json`),
     description: _(`"${property}" is required`),
+    file: MANIFEST_JSON,
+  };
+}
+
+export const MANIFEST_ICON_NOT_FOUND = 'MANIFEST_ICON_NOT_FOUND';
+export function manifestIconMissing(path) {
+  return {
+    code: MANIFEST_ICON_NOT_FOUND,
+    legacyCode: null,
+    message: _(
+      'An icon defined in the manifest could not be found in the package.'),
+    description: sprintf(
+      _('Icon could not be found at "%(path)s".'),
+      {path}),
     file: MANIFEST_JSON,
   };
 }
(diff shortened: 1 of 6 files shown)


src/messages/manifestjson.js
```diff
@@ -106,6 +106,20 @@ export function manifestPropMissing(property) {
...
+export const MANIFEST_ICON_NOT_FOUND = 'MANIFEST_ICON_NOT_FOUND';
+export function manifestIconMissing(path) {
+  return {
+    code: MANIFEST_ICON_NOT_FOUND,
+    legacyCode: null,
+    message: _(
+      'An icon defined in the manifest could not be found in the package.'),
+    description: sprintf(
+      _('Icon could not be found at "%(path)s".'),
+      {path}),
```
EnTeQuAk: I like this I think. We should apply this to how `UNSUPPORTED_API` is defined too.

src/parsers/manifestjson.js
```diff
@@ -146,6 +150,17 @@ export default class ManifestJSONParser extends JSONParser {
...
+  validateIcons() {
+    const { icons } = this.parsedJSON;
+    Object.keys(icons).forEach((size) => {
+      const path = icons[size];
+      if (!this.io.files.hasOwnProperty(path)) {
+        this.collector.addError(messages.manifestIconMissing(path));
+        this.isValid = false;
+      }
+    });
+  }
```
EnTeQuAk: I actually wonder if we should verify that they're not empty and match their anticipated file-type (e.g a `.png` file is an actual PNG).

mstriemer: I filed some follow up issues to do that. #1328 to check that they're valid images and #1329 to ensure they're reasonable sizes.

As long as Firefox can render the image the file type might not matter, but that seems reasonable.

src/messages/manifestjson.js
mstriemer: It looks like the unsupported API code is using `context.report()` which can use eslint's built-in string formatting. I think that works fine.

I was hoping to use it here but since the manifest validation updates the collector directly I don't think we have support for it.

PR #1331, opened 2017-06-08
Error on small icons and warn about smallish icons

### LINKED ISSUE #1329: Validate the size of icons
Firefox expects icons to be at least 16 pixels in size. We should require that there be at least one icon that is at least 16 pixels in size. This should check the size defined in the manifest but it would also be nice to check the size of the actual image building off of #1328.

mstriemer:
For now just validating the keys in the icons object. The UI gets wonky if an icon is less than 16 pixels so we need at least one icon of that size, if icons are provided.

The docs recommend an icon of 48x48 so add a warning if there is an icon specified less than that.

Fixes #1329.


mstriemer:
I have a rough draft of checking the image's for sizes using [sharp](https://www.npmjs.com/package/sharp). Unfortunately reading the data from the image is async and we don't expect `ManifestJSONParser` to be async. So that will take some refactoring.

I'm going to close this for now since I think it will all change.

PR #1332, opened 2017-06-08
Don't validate computed properties for WE APIs

### LINKED ISSUE #1309: Misleading warning when using dynamic property names to use API
Use an API with a dynamic property accessor, like the following:
```js
const AREA = "local";
browser.storage[AREA].set({
  migrated: true
});
```

### What happened?
The warning printed by the linter is "storage.AREA is not supported".

A warning along the lines of "Could not determine if API is implemented for storage[AREA]"

Source code with which I have experienced this: https://github.com/freaktechnik/justintv-stream-notifications/blob/master/src/background/preferences.js
Travis CI build with the warnings: https://travis-ci.org/freaktechnik/justintv-stream-notifications#L1971-L1976

mstriemer:
We won't be able to validate that the member expression is correct since it is computed, so don't check it rather than treating the name of the variable as the property.

Fixes #1309.

### DIFF
--- a/src/rules/javascript/webextension-unsupported-api.js
+++ b/src/rules/javascript/webextension-unsupported-api.js
@@ -5,7 +5,8 @@ export default {
   create(context) {
     return {
       MemberExpression: function(node) {
-        if (node.object.object &&
+        if (!node.computed &&
+            node.object.object &&
             ['chrome', 'browser'].includes(node.object.object.name)) {
           let namespace = node.object.property.name;
           let property = node.property.name;
--- a/src/rules/javascript/webextension-unsupported-api.js
+++ b/src/rules/javascript/webextension-unsupported-api.js
@@ -1,5 +1,6 @@
 import { UNSUPPORTED_API } from 'messages/javascript';
 import { hasBrowserApi } from 'schema/browser-apis';
+import { isBrowserNamespace } from 'utils';
 
 export default {
   create(context) {
@@ -7,7 +8,7 @@ export default {
       MemberExpression: function(node) {
         if (!node.computed &&
             node.object.object &&
-            ['chrome', 'browser'].includes(node.object.object.name)) {
+            isBrowserNamespace(node.object.object.name)) {
           let namespace = node.object.property.name;
           let property = node.property.name;
           let api = `${namespace}.${property}`;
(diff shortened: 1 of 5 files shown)


src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -5,7 +5,8 @@ export default {
   create(context) {
     return {
       MemberExpression: function(node) {
-        if (node.object.object &&
+        if (!node.computed &&
+            node.object.object &&
             ['chrome', 'browser'].includes(node.object.object.name)) {
```
muffinresearch: This isn't new code but wondering if it should be a constant?

src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -5,7 +5,8 @@ export default {
   create(context) {
     return {
       MemberExpression: function(node) {
-        if (node.object.object &&
+        if (!node.computed &&
```
muffinresearch: Presumably actually determining the value would only work if the value defined in the same file?

muffinresearch: What is the potential impact on schema validation for all APIS. Presumably this would mean we'd skip everything that's dynamically referenced. Are reviewers ok with that vs the flipside of more false positives?

/cc @wagnerand

wagnerand: Since we rely more and more on auto-detection, I'm not sure it's a good idea to just skip them.

Do you think this would lead to a lot of real false positives?

mstriemer: This test is mostly here as a convenience to developers from what I understand. If they want to try and use `chrome.gcm.send()` they can try it but we don't support it.

I'm not sure this is something that would need to extra review. If there are certain APIs that need extra validation then we can check those but that likely belongs somewhere else.

mstriemer: As far as actually determining the value that would likely involve running the code.

In the case that prompted this we would be able to do it (a const assigned at top of file and used throughout) but that's a special case that could be handled with static analysis (in theory).

If it was instead pulling the value from a user-managed variable or generating a random string for some reason we'd probably have no possible way of knowing what value or values it would be.

mstriemer:
I think if we want to flag something for follow up review it should likely be for specific APIs.

This should be good for a final review now.

PR #1334, opened 2017-06-09
Switch our test-suite to use Jest

EnTeQuAk:
* [x] Get tests to run and be green as a first step
* [x] Go through code and fix up broken tests that weren't properly modified by codemods
* [x] Fix eslint errors
* [x] 👯 

Notes:

* We used `shelljs` to test the command line client but I ran into massive problems running it via jest on travis so I implemented a workaround helper for now.
* It's lots of commits but I plan to rebase them right before merging, they are mostly self-containing except for a few travis testing commits.
* **jest has to be run with `-i` since for some reason it breaks down on processes without** (any suggestions, happily welcome)

Regarding the error that I get when running without `-i`

```
 PASS  tests/test.utils.js
 FAIL  tests/test.main.js
  ● Test suite failed to run

    ProcessTerminatedError: cancel after 2 retries!

      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:81:25)
      at Array.forEach (native)
      at Farm.<anonymous> (node_modules/worker-farm/lib/farm.js:75:36)
      at ontimeout (timers.js:386:14)
      at tryOnTimeout (timers.js:250:5)
      at Timer.listOnTimeout (timers.js:214:5)

A worker process has quit unexpectedly! Most likely this is an initialization error.
```

and it stops executing all other workers.


EnTeQuAk:
Perfect, first tests are passing, that's nice. Now let's see if shelljs really messes things up…

EnTeQuAk:
Failing tests because of shelljs may have been running into https://github.com/shelljs/shelljs/pull/524 but I'm not yet sure so let's see how my hacky helper performs...

tests/parsers/test.manifestjson.js
```diff
@@ -11,16 +11,16 @@ describe('ManifestJSONParser', function() {
...
-    assert.include(errors[0].message, 'Your JSON is not valid.');
+    expect(errors.length).toEqual(1);
+    expect(errors[0].code).toEqual(messages.JSON_INVALID.code);
+    expect(errors[0].message).toContain('Your JSON is not valid.');
 
     var metadata = manifestJSONParser.getMetadata();
-    assert.equal(metadata.manifestVersion, null);
-    assert.equal(metadata.name, null);
-    assert.equal(metadata.version, null);
+    expect(metadata.manifestVersion).toEqual(null);
```
EnTeQuAk: Funny part is, this actually caught a bug that I had to fix… don't ask me why this didn't fail before though. The problem I'm talking about is:

 
https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L28 defines `mainfestVersion` while https://github.com/mozilla/addons-linter/blob/master/src/parsers/manifestjson.js#L177 uses `manifest_version` since that's what's supplied if a correct manifest is given. So the default value in `parsedJSON` was kinda wrong 🤔

EnTeQuAk:
Apparently there was really something missing, https://github.com/mozilla/addons-linter/pull/1334/commits/aaa5b80e5b49075636366a11f5969e64cfd37c5e was necessary. So now coverage should be green again. 🕺

tests/io/test.base.js
```diff
@@ -8,11 +10,11 @@ describe('io.IOBase()', function() {
...
   it('should init class props as expected', () => {
     var io = new IOBase('foo/bar');
-    assert.equal(io.path, 'foo/bar');
-    assert.equal(io.entries.length, 0);
-    assert.equal(Object.keys(io.files).length, 0);
-    assert.equal(typeof io.files, 'object');
-    assert.equal(io.maxSizeBytes, 104857600);
+    expect(io.path).toEqual('foo/bar');
+    expect(io.entries.length).toEqual(0);
+    expect(Object.keys(io.files).length).toEqual(0);
```
muffinresearch: There's an open thread here: https://github.com/mozilla/addons-frontend/pull/2540/files/9ecf8b9664f359e27c0269cfa29988304f781814#diff-2eb5d63af99d473bf0c208ce9c699740 on whether to use `toBe` for things like this - might be worth having a look at this because we'll probably want to come to a shared consensus.

tests/io/test.crx.js
```diff
@@ -102,8 +104,8 @@ describe('crx.getFiles()', function() {
     };
     return myCrx.getFiles()
       .then((files) => {
-        assert.deepEqual(files, myCrx.files);
-        assert.notOk(this.fromBufferStub.called);
+        expect(files).toEqual(myCrx.files);
+        expect(this.fromBufferStub.called).toBeFalsy();
```
muffinresearch: Can follow up on this - I'm just mentioning it in passing, but this is the type of thing that would get better messages either using sinon.assert directly (no expect) or perhaps better still, getting some jasmine matchers working for sinon (the existing package for that is incompatible) to you can do expect syntax similar to what jest.fn has.

tests/io/test.directory.js
```diff
@@ -175,7 +177,7 @@ describe('Directory.getFileAsString()', function() {
         return myDirectory.getFileAsString('dir3/foo.txt');
       })
       .then((content) => {
-        assert.notOk(content.charCodeAt(0) === 0xFEFF);
+        expect(content.charCodeAt(0) === 0xFEFF).toBeFalsy();
```
muffinresearch: This could use `toBe()` with the value instead of this style.

tests/io/test.xpi.js
```diff
@@ -610,14 +610,14 @@ describe('Xpi.getFilesByExt()', function() {
...
-        assert.equal(htmlFiles[2], 'third.html');
+        expect(htmlFiles.length).toEqual(3);
+        expect(htmlFiles[0]).toEqual('index.html');
+        expect(htmlFiles[1]).toEqual('second.htm');
+        expect(htmlFiles[2]).toEqual('third.html');
 
         for (let i = 0; i < htmlFiles.length; i++) {
-          assert.ok(htmlFiles[i].endsWith('.html') ||
-                    htmlFiles[i].endsWith('.htm'));
+          expect(htmlFiles[i].endsWith('.html') ||
```
muffinresearch: This would be better to use https://facebook.github.io/jest/docs/expect.html#tomatchregexporstring

tests/rules/javascript/test.eslintRulesObject.js
```diff
@@ -3,25 +3,24 @@ import { existsSync, readdirSync } from 'fs';
...
   it('should have files that match the keys', () => {
     var files = readdirSync('src/rules/javascript')
       .filter((fileName) => fileName !== 'index.js');
     for (let fileName of files) {
-      assert.ok(ESLINT_RULE_MAPPING.hasOwnProperty(path.parse(fileName).name),
-                singleLineString`fileName "${fileName}" does not have a
-                matching key in the eslint rules object.`);
+      expect(
+        ESLINT_RULE_MAPPING.hasOwnProperty(path.parse(fileName).name)
+      ).toBeTruthy();
```
muffinresearch: I think it's worth look to use `.toBe(true)` more in cases like this and to avoid truthy/falsey comparisons where possible (I know this is a result of the code-mod) but it's good to not have potentially problematic loose comparisions.

tests/scanners/test.filename.js
```diff
@@ -152,9 +149,9 @@ describe('Hidden and Flagged File Regexes', function() {
 
   for (const filePath of nonMatchingFlaggedFileExtensions) {
     it(`should not match ${filePath} as a flagged file extension`, () => {
-      assert.isNotOk(
-        constants.FLAGGED_FILE_EXTENSIONS.includes(extname(filePath)),
-        `${filePath} should not match flagged file extension regex`);
+      expect(
+        constants.FLAGGED_FILE_EXTENSIONS.includes(extname(filePath))
```
muffinresearch: If this used https://facebook.github.io/jest/docs/expect.html#tocontainitem it would give better errors.

tests/scanners/test.filename.js
```diff
@@ -139,9 +136,9 @@ describe('Hidden and Flagged File Regexes', function() {
 
   for (const filePath of matchingFlaggedFileExtensions) {
     it(`should match ${filePath} as a flagged file extensions`, () => {
-      assert.isOk(
-        constants.FLAGGED_FILE_EXTENSIONS.includes(extname(filePath)),
-        `${filePath} should not match flagged file extension regex`);
+      expect(
+        constants.FLAGGED_FILE_EXTENSIONS.includes(extname(filePath))
```
muffinresearch: If this used https://facebook.github.io/jest/docs/expect.html#tocontainitem it would give better errors.

.travis.yml
```diff
@@ -2,14 +2,17 @@ language: node_js
...
-script: npm test
+script: npm run test-ci
 notifications:
   irc:
     channels:
     - irc.mozilla.org#amo-bots
     on_success: change
     on_failure: always
 after_script: npm run publish-coverage
+cache:
```
EnTeQuAk: already prepared a branch for that, should be coming up soon.

.eslintrc
```diff
@@ -22,6 +23,10 @@
   "globals": {
     "assert": false,
```
EnTeQuAk: oh, good idea. I'll do that.

tests/io/test.base.js
```diff
@@ -1,3 +1,5 @@
+import sinon from 'sinon';
```
muffinresearch: This could be a global sandbox instance created in the global test setup that gets automatically torn down too. See https://github.com/mozilla/addons-frontend/blob/master/tests/setup.js#L62 for an example - although as mentioned I've filed an issue re: calling this sinon (see https://github.com/mozilla/addons-frontend/issues/2546)

If you follow the example above you probably should remove the line re: format since sinon.format has been deprecated anyway.

.eslintrc
muffinresearch: Actually if you use:

```
"env": {
  "jest": true,
} 
```

That will cover all of that :)

EnTeQuAk: fixed in 7022714

jest.config.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  coveragePathIgnorePatterns: [
```
EnTeQuAk: fixed in 17c83ef and showed that 029ef6d was needed, nice!

mstriemer:
The tests are so fast after this! Thanks for doing this! 🎉🎈🎉

PR #1336, opened 2017-06-09
[WIP] Validate icon files

mstriemer:
I need to fix how the async works. I hacked it in the `getMetadata()` since that worked well with `async`/`await` but then that stopped working after an npm/yarn install.

I think we'll want the API for `ManifestJSONParser` to change to something like:

```js
const parser = new ManifestJSONParser(
  json,
  this.collector,
  { io: this.io, ... });
await parser.validate();
parser.getMetadata();
```

Alternatively the icon validation could go into a separate class like `ManifestIconValidator` and that can be async. That might be easier.

The tests are scoped to the new stuff so they pass but this should break a bunch of stuff.


src/const.js
```diff
@@ -163,3 +163,7 @@ export const TEMPORARY_APIS = [
   'storage.local',
   'storage.sync',
 ];
+
+// Minimum sizes for icons in pixels.
+export const MIN_ICON_SIZE = 16;
+export const RECOMMENDED_ICON_SIZE = 48;
```
EnTeQuAk: is there a max size?

EnTeQuAk:
Sorry btw for breaking all the things in that branch with my jest refactor 😇

mstriemer:
> I think I like the idea, especially if we're starting to use async and await in the linter

If we can get it working I think it's a good call. I thought it was a bit confusing before but I've started using it more now and it is wonderful.

> I do think that the icon validation fits nicely into ManifestJSONParser though.

I agree, I won't have time to get back to this until next week most likely though. I might make a separate async class for the time being and handle making `ManifestJSONParser` async in another patch. I'll likely do whatever is easiest to get this in then follow it up with a refactor.

> Sorry btw for breaking all the things in that branch with my jest refactor 😇

Totally worth it!

src/const.js
mstriemer: I'm not sure. I don't think so but we obviously don't need a 4K image here. It might make sense to limit it to 256 just cause we can.

EnTeQuAk:
> I'll likely do whatever is easiest to get this in then follow it up with a refactor.

That sounds like a perfect plan to me.

EnTeQuAk:
@mstriemer are you still working on this?

mstriemer:
I would like to get this landed but it grew a bit bigger than expected. I will try and split this up when I have time but for now I don't think this PR will get landed.

EnTeQuAk:
Cool, thanks! Let me know if I can be of any help working on this.

PR #1344, opened 2017-06-12
Import schemas 55+ and nightly

mstriemer:
I wanted to verify that the schemas would import once the 55 cut over happens so I ran the import against nightly. This seems like it could be worth keeping around.

Just looked into it some more and importing 55 didn't work because the tag is now called BETA instead of AURORA (of course). So this now supports 55+ and nightly.


mstriemer:
Updated with support for 55+ imports. Currently looking through the import to see if it's good to land. That will be a separate PR.

PR #1348, opened 2017-06-14
Import Firefox 55 schemas

### LINKED ISSUE #1345: Update schemas to Firefox 55
Firefox 55 is now on Beta. We should import the schemas so that we stay up to date.

Once we've cut a release web-ext should also be updated to use the new version.

mstriemer:
This includes an update to how URL formats are handled. Firefox has an interesting naming convention for these and when checking the new `chrome_settings_override` options I noticed our checks weren't quite the same.

The `url-parse` library will always parse whatever you give it as a valid URL even if it is obviously not valid (like `['lol', 10, 'you', 'CANNOT', 9.912, 'beserious']`), so I switched us to `whatwg-url` since Firefox is using the `URL` class that it polyfills.

Fixes #1345.

### DIFF
--- a/src/schema/formats.js
+++ b/src/schema/formats.js
@@ -1,4 +1,4 @@
-import parse from 'url-parse';
+import { URL } from 'whatwg-url';
 
 const VALIDNUMRX = /^[0-9]{1,5}$/;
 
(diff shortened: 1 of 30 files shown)


src/schema/formats.js
```diff
@@ -43,12 +43,52 @@ export function isValidVersionString(version) {
...
+  const codes = new Array(length);
+  for (let i = 0; i < length; i++) {
+    codes[i] = Math.floor(Math.random() * 26) + offset;
+  }
+  const protocol = String.fromCharCode(...codes);
+  return `${protocol}:`;
+}
+
+function isRelativeUrl(value) {
+  const protocol = getRandomProtocol();
```
EnTeQuAk: Why does this have to be random? Can't we just use something predictable that get's re-used every time?

tests/schema/test.formats.js
```diff
@@ -1,73 +1,134 @@
...
+      '1.000000',
+      '2.99999',
+      '3.65536',
+      '1.0.0-beta2',
+    ];
+
+
+    for (let validVersionString of validVersionStrings) {
+      it(`should find ${validVersionString} to be valid`, () => {
+        expect(isValidVersionString(validVersionString)).toBeTruthy();
```
EnTeQuAk: from recent discussions with @muffinresearch this should probably be `expect(...).toEqual(true)` to unify our asserts a bit and to avoid testing for anything else but the expected `true` value.

tests/schema/test.formats.js
```diff
@@ -1,73 +1,134 @@
...
+      });
+    }
+
+    for (let invalidVersionString of invalidVersionStrings) {
+      it(`should find ${invalidVersionString} to be invalid`, () => {
+        expect(isValidVersionString(invalidVersionString)).toBeFalsy();
```
EnTeQuAk: Same here

tests/schema/test.formats.js
```diff
@@ -1,73 +1,134 @@
...
+
+  describe('URL formats', () => {
+    it('domain', () => {
+      const value = 'https://example.com';
+
+      expect(isAnyUrl(value)).toBeTruthy();
+      expect(isAbsoluteUrl(value)).toBeTruthy();
+      expect(isSecureUrl(value)).toBeTruthy();
+
+      expect(isStrictRelativeUrl(value)).toBeFalsy();
```
EnTeQuAk: And all of these

src/schema/formats.js
EnTeQuAk: actually, reading a bit more… it's because it needs a protocol and we want to avoid people to workaround our validation, I think that could work.

mstriemer: Yeah, it feels a little silly. There's no advantage to using the magic protocol that we use that I can tell so it isn't a big deal if it happens but we don't want it to. Another option would be to define a magic constant then check if the URL is valid on its own and uses that protocol.

tests/schema/test.formats.js
mstriemer: I was thinking the same thing when I started writing this but opted to stick with the same format that was being used before. I'll switch them all over though.

mstriemer:
Should be good now.

Issue #1350, opened 2017-06-15
Noisy test output


mstriemer:
Bunyan is causing some noisy test output. This has something to do with dtrace-provider which is optional but doesn't work on mac or not with node 6+ or something. Either way we aren't using dtrace and this is annoying.

Pino is a bunyan replacement that is supposed to be faster and doesn't depend on dtrace at all. In theory it will fix the test output.

```
 PASS  tests/test.logger.js
  ● Console

    console.error node_modules/dtrace-provider/dtrace-provider.js:26
      { Error: Cannot find module './build/Release/DTraceProviderBindings' from 'dtrace-provider.js'
          at Resolver.resolveModule (/Users/markstriemer/work/addons-linter/node_modules/jest-resolve/build/index.js:179:17)
          at Resolver._getVirtualMockPath (/Users/markstriemer/work/addons-linter/node_modules/jest-resolve/build/index.js:296:23)
          at Resolver._getAbsolutPath (/Users/markstriemer/work/addons-linter/node_modules/jest-resolve/build/index.js:279:10)
          at Resolver.getModuleID (/Users/markstriemer/work/addons-linter/node_modules/jest-resolve/build/index.js:252:31)
          at Runtime._shouldMock (/Users/markstriemer/work/addons-linter/node_modules/jest-runtime/build/index.js:574:37)
          at Runtime.requireModuleOrMock (/Users/markstriemer/work/addons-linter/node_modules/jest-runtime/build/index.js:402:14)
          at Object.<anonymous> (/Users/markstriemer/work/addons-linter/node_modules/dtrace-provider/dtrace-provider.js:17:23)
          at Runtime._execModule (/Users/markstriemer/work/addons-linter/node_modules/jest-runtime/build/index.js:513:13)
          at Runtime.requireModule (/Users/markstriemer/work/addons-linter/node_modules/jest-runtime/build/index.js:329:14)
          at Runtime.requireModuleOrMock (/Users/markstriemer/work/addons-linter/node_modules/jest-runtime/build/index.js:405:19) code: 'MODULE_NOT_FOUND' }
    console.error node_modules/dtrace-provider/dtrace-provider.js:26
      { Error: Cannot find module './build/default/DTraceProviderBindings' from 'dtrace-provider.js'
          at Resolver.resolveModule (/Users/markstriemer/work/addons-linter/node_modules/jest-resolve/build/index.js:179:17)
[...]

PR #1351, opened 2017-06-15
Switch to pino from bunyan

mstriemer:
I can't figure out how to get dispensary installed from my git branch so this just updates addons-linter. I did a `yarn link dispensary` locally and it fixed the annoying test output.

Supports #1350.


mstriemer:
I think we could update dispensary in a separate patch but I can put it in here too.

mstriemer:
This now includes the dispensary update and fixes the test output for me locally. We are also defining a default for `logLevel` so we're good there.

Issue #1287, opened 2017-05-16
`code` and `message` is the same in JSON output 


potch:
### Describe the problem and steps to reproduce it:
Get results from programmatic usage of the linter (as an API)

### What happened?
In results, the `code` field of warnings is the same thing as the `message` field. example:
```js
{
  _type: 'warning',
  code: 'The Function constructor is eval.',
  message: 'The Function constructor is eval.',
  ...
}
```

### What did you expect to happen?
The `code` field that would be something like "UNSUPPORTED_API", as listed in the messages definition: https://github.com/mozilla/addons-linter/blob/286ebbc91025baa19fcb8eace30d28144193f966/src/messages/javascript.js#L266

### Anything else we should know?
Nah.

mstriemer:
Since `code` is something we added and isn't supported when adding an error through eslint I think we're stuck with this for now.

@EnTeQuAk and I discussed moving to the eslint standard of using the rule name to replace `code` and we agreed that was the best path forward but I don't see an issue for that.

EnTeQuAk:
Correct, it's a bit of a tricky problem. We do have `code` and `message` and at the moment use both and for our internal messages `code` is set properly but for messages coming from eslint it's set to `message` which isn't a problem at the moment since more or less `code` was unused.

We do have the ability to overwrite `code` for messages coming from eslint, we should probably do and require that. I do agree that having a `code` is useful longerm since it allows ignoring or filtering messages on the receiving client (AMO, web-ext etc).

potch:
I'm looking to filter for specific codes to filter under programmatic usage. Can I pass a flag that's like "this isn't eslint, give me generalized codes"?

EnTeQuAk:
I marked this as fixed with merging https://github.com/mozilla/addons-linter/pull/1356. We are now able to overwrite the `code` for a 3rd party rule as well, we are currently overwriting all 3rd party rules so as far as I can tell this issue should be fixed now.

Mind testing if this helps with what you wanted to do @potch?

vcarciu:
Please add some STRs to this bug or mark it as "qa: not needed"

PR #1356, opened 2017-06-19
Overwrite 'code' on overwritten 3rd party messages too.

### LINKED ISSUE #1355: Overwrite `code` for 3rd party rule overwrites
This should ref https://github.com/mozilla/addons-linter/issues/1287 and helps using that code for our post-review work in addons-linter.

EnTeQuAk:
Fixes #1355 
Refs and might even fix https://github.com/mozilla/addons-linter/issues/1287

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -112,6 +112,7 @@ export default class JavaScriptScanner {
             var overwrites = ESLINT_OVERWRITE_MESSAGE[message.ruleId];
             var shortDescription = overwrites.message || message.message;
             var description = overwrites.description || message.description;
+            code = overwrites.code || message.message;
           } else {
             var shortDescription = code;
             var description = null;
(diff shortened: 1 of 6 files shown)


src/scanners/javascript.js
```diff
@@ -112,6 +112,7 @@ export default class JavaScriptScanner {
             var overwrites = ESLINT_OVERWRITE_MESSAGE[message.ruleId];
             var shortDescription = overwrites.message || message.message;
             var description = overwrites.description || message.description;
+            code = overwrites.code || message.message;
```
mstriemer: `code` is already set to `message.message` so I think this could reuse `code` instead, or only set it in the case that it will change.

```js
code = overwrites.code || code;
// or
if (overwrites.code) {
  code = overwrites.code;
}

EnTeQuAk: Fixed in ecde23d

EnTeQuAk:
> I see that we test the length of the code if it exists but maybe we should check that exported non-function objects always have a code?

Great idea, there's quite a few tests failing in case I set some code to `null` but none that explicitly tests for it and helps debugging why the others are failing. Added in 997d9cf.

PR #1365, opened 2017-06-20
re-add build step before deployment

EnTeQuAk:


.travis.yml
```diff
@@ -13,6 +13,8 @@ after_script: npm run publish-coverage
 cache:
   directories:
   - node_modules
+before_deploy:
+  - npm run build
```
muffinresearch: Building should be part of the test run in CI so that if the build fails, the test run fails.

EnTeQuAk: Should be fixed in 570c550 - let's see if travis agrees.

PR #1388, opened 2017-07-04
fix(package): update chalk to version 2.0.1

wagnerand:
Closes #1383


muffinresearch:
The fix for chalk was to remove the failing test. Since the output of chalk has changed there's no longer a good key for checking the colours.

PR #1429, opened 2017-07-27
Warn when detecting a <script> tag with a remote URL.

### LINKED ISSUE #1321: Warn when detecting a <script> tag with a remote URL
For post-review, we want to increase priority of add-ons using a `<script>` with a remote URL. The linter should detect this and add a warning with a specific code that addons-server can look for.

EnTeQuAk:
Fixes #1321 

r? @muffinresearch for linter sanity, r? @wagnerand for messaging and cc @diox that this is happening.

### DIFF
--- a/src/messages/html.js
+++ b/src/messages/html.js
@@ -12,7 +12,7 @@ export const INLINE_SCRIPT = {
 export const REMOTE_SCRIPT = {
   code: 'REMOTE_SCRIPT',
   legacyCode: null,
-  message: _('Remote scripts are blocked by default'),
-  description: _(singleLineString`Default CSP rules prevent remote JavaScript
-    from being loaded (https://mzl.la/2pn32nd).`),
+  message: _('Remote scripts are not allowed as per the Add-on Policies.'),
+  description: _(singleLineString`Please include all scripts in the add-on.
+    For more information, refer to https://mzl.la/2uEOkYp.`),
 };
--- a/src/messages/html.js
+++ b/src/messages/html.js
@@ -7,3 +7,12 @@ export const INLINE_SCRIPT = {
   description: _(singleLineString`Default CSP rules prevent inline JavaScript
     from running (https://mzl.la/2pn32nd).`),
 };
+
+
+export const REMOTE_SCRIPT = {
+  code: 'REMOTE_SCRIPT',
+  legacyCode: null,
+  message: _('Remote scripts are blocked by default'),
+  description: _(singleLineString`Default CSP rules prevent remote JavaScript
+    from being loaded (https://mzl.la/2pn32nd).`),
+};
--- a/src/rules/html/warn-on-remote-script.js
+++ b/src/rules/html/warn-on-remote-script.js
@@ -1,0 +1,25 @@
+import { VALIDATION_WARNING } from 'const';
+import { isStrictRelativeUrl } from 'schema/formats';
+import * as messages from 'messages';
+
+
+export function warnOnRemoteScript($, filename) {
+  return new Promise((resolve) => {
+    var linterMessages = [];
+
+    $('script').each((i, element) => {
+      const src = $(element).attr('src');
+
+      if (src !== undefined && !isStrictRelativeUrl(src)) {
+        linterMessages.push(
+          Object.assign({}, messages.REMOTE_SCRIPT, {
+            type: VALIDATION_WARNING,
+            file: filename
+          })
+        );
+      }
+    });
+
+    resolve(linterMessages);
+  })
+}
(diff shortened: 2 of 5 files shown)


src/rules/html/warn-on-remote-script.js
```diff
@@ -0,0 +1,25 @@
...
+
+
+export function warnOnRemoteScript($, filename) {
+  return new Promise((resolve) => {
+    var linterMessages = [];
+
+    $('script').each((i, element) => {
+      const src = $(element).attr('src');
+
+      if (src !== undefined && !isStrictRelativeUrl(src)) {
```
wagnerand: This should also respect `src` that starts with `moz-extension://`. Please also add this to the test.

src/messages/html.js
```diff
@@ -7,3 +7,12 @@ export const INLINE_SCRIPT = {
   description: _(singleLineString`Default CSP rules prevent inline JavaScript
     from running (https://mzl.la/2pn32nd).`),
 };
+
+
+export const REMOTE_SCRIPT = {
+  code: 'REMOTE_SCRIPT',
+  legacyCode: null,
+  message: _('Remote scripts are blocked by default'),
```
wagnerand: `Remote scripts are not allowed as per the Add-on Policies.`

src/messages/html.js
```diff
@@ -7,3 +7,12 @@ export const INLINE_SCRIPT = {
...
+  description: _(singleLineString`Default CSP rules prevent remote JavaScript
+    from being loaded (https://mzl.la/2pn32nd).`),
```
wagnerand: `Please include all scripts in the add-on. For more information, refer to https://mzl.la/2uEOkYp .`

src/rules/html/warn-on-remote-script.js
```diff
@@ -0,0 +1,25 @@
...
+        linterMessages.push(
+          Object.assign({}, messages.REMOTE_SCRIPT, {
+            type: VALIDATION_WARNING,
```
wagnerand: This should be an error, since remote scripts are not allowed.

diox: https://docs.google.com/spreadsheets/d/1CZEECHHqEmK87fNiWNslAsFQBL7zxYF-c63-zZVZuqw/edit#gid=0 mentions remote scripts as one of the post-review prioritization weights, which seems to contradict this (can't be a hard error and affect post-review priority at the same time...)

wagnerand: Since this is a hard policy rule, I don't see what sense it makes to let this pass in the linter and then rely on a human to spot it and take action.

EnTeQuAk: summoning @jvillalobos for a final statement here.

I do agree with @wagnerand, this should be an ERROR since that's what our policies say. On the other hand, that'd mean that inremove scripts can be removed as a risk-factor for post-review.

src/rules/html/warn-on-remote-script.js
EnTeQuAk: Fixed in 9a4d596

src/messages/html.js
EnTeQuAk: Fixed in e8a272a

src/messages/html.js
EnTeQuAk: Fixed in e8a272a

src/rules/html/warn-on-remote-script.js
jvillalobos: There are certainly cases where we have allowed remote scripts in the past, so it can't be a hard error. Otherwise, making an exception means an admin will have to upload every update in order to be accepted.

wagnerand: We haven't made any exceptions in far more than a year, and I'd like to keep it that way, as I don't think we should make any exceptions at all. It nullifies the purpose of signing. I believe @kewisch agrees.

jvillalobos: That's a policy change that will need to be discussed separately from this issue. The risk factors and behavior of the linter were agreed upon months ago when we covered this in Toronto. I'd rather not make last-minute changes without discussion.

tests/scanners/test.html.js
```diff
@@ -57,6 +57,41 @@ describe('HTML', function() {
       });
   });
 
+  it('should warn on remote <script> tag src attribute', () => {
+    var badHTML = validHTML(singleLineString`
+      <script src="http://foo.bar/my.js"></script>
+      <script src="https://foo.bar/my.js"></script>
+      <script src="file://foo.bar/my.js"></script>
+      <script src="ftp://foo.bar/my.js"></script>
+      <script src="moz-extension://foo/my.js"></script>
```
wagnerand: Can we please add the following test as well?

`<script src="//foo.bar/my.js"></script>`

EnTeQuAk: Added in f17a4e0

EnTeQuAk: Added in f17a4e0

PR #1432, opened 2017-07-31
Support more icon paths

### LINKED ISSUE #1413: MANIFEST_ICON_NOT_FOUND error even if there is an icon
addons-linter reports MANIFEST_ICON_NOT_FOUND error even if there is an icon.

### What happened?
run web-ext lint
lint reports MANIFEST_ICON_NOT_FOUND error.

No error if icon exists.

Uploading add-on to AMO is rejected due to this error.

mstriemer:
Adds support for icons with URL fragments, .. relative URLs that resolve
within the package and domains for some reason cause Firefox supports
that.

Fixes #1413.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -2,6 +2,7 @@ import path from 'path';
 
 import RJSON from 'relaxed-json';
 import validate from 'schema/validator';
+import { URL } from 'whatwg-url';
 
 import { getConfig } from 'cli';
 import { MANIFEST_JSON, PACKAGE_EXTENSION } from 'const';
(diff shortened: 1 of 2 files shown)


src/parsers/manifestjson.js
```diff
@@ -12,12 +13,8 @@ import { isToolkitVersionString } from 'schema/formats';
 import { singleLineString } from 'utils';
 
 function normalizePath(iconPath) {
-  if (iconPath.startsWith('/')) {
-    return iconPath.slice(1);
-  } else if (iconPath.startsWith('./')) {
-    return iconPath.slice(2);
-  }
-  return iconPath;
+  const { pathname } = new URL(iconPath, 'https://example.com/');
```
muffinresearch: Could use a comment to explain why the example.com origin part is needed.

mstriemer:
Thanks, should have checked the PTO calendar before I asked for the review.

I think we should consider picking this into Thursday's release.

tofumatt:
I always forget to check it, no worries. I didn't know it was urgent enough to cherry pick, my bad!

Matthew Riley MacPherson (sent from mobile)

PR #1434, opened 2017-08-02
Parse and validate CSP rules more properly.

### LINKED ISSUE #1298: Only flag CSP that is actually less strict than the default
We currently flag any definition of CSP in the manifest.
This causes quite some false positives, and for post-review, we need it to be more reliable.

We are mostly interested in remote scripts, other remote resources should be fine.
So, I *think* what we need to check for is whether `http` or `*` is in `script-src` or `default-src`.

@jonathanKingston @marumari could you confirm what I wrote above is correct and complete? Thanks!

EnTeQuAk:
Fixes #1298 

cc @wagnerand for potential messaging updates.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src#Sources mentions 

```
         

<scheme-source>
    A schema such as 'http:' or 'https:'. The colon is required, single quotes shouldn't be used. You can also specify data schemas (not recommended).

        data: Allows data: URIs to be used as a content source. This is insecure; an attacker can also inject arbitrary data: URIs. Use this sparingly and definitely not for scripts.
        mediastream: Allows mediastream: URIs to be used as a content source.
        blob: Allows blob: URIs to be used as a content source.
        filesystem: Allows filesystem: URIs to be used as a content source.
```

Do we want to disallow `data:` URIs too?

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -169,7 +169,7 @@ export const TEMPORARY_APIS = [
 // See https://mzl.la/2vwqbGU for more details and allowed options.
 export const CSP_KEYWORD_RE = new RegExp([
   '(self|none|strict-dynamic|',
-  'unsafe-inline|unsafe-eval|unsafe-hashed-attributes)',
+  'unsafe-inline|unsafe-hashed-attributes)',
   // Only match these keywords, anything else is forbidden
   '(?!.)',
   '|(sha(256|384|512)-|nonce-)',
--- a/src/const.js
+++ b/src/const.js
@@ -168,7 +168,8 @@ export const TEMPORARY_APIS = [
 // `script-src`. Used in manifest.json parser for validation.
 // See https://mzl.la/2vwqbGU for more details and allowed options.
 export const CSP_KEYWORD_RE = new RegExp([
-  '(self|none|strict-dynamic|unsafe-hashed-attributes)',
+  '(self|none|unsafe-inline|unsafe-eval|strict-dynamic|',
+  'unsafe-hashed-attributes)',
   // Only match these keywords, anything else is forbidden
   '(?!.)',
   '|(sha(256|384|512)-|nonce-)',
--- a/src/const.js
+++ b/src/const.js
@@ -163,3 +163,14 @@ export const TEMPORARY_APIS = [
   'storage.local',
   'storage.sync',
 ];
+
+// All valid CSP keywords that are options to keys like `default-src` and
+// `script-src`. Used in manifest.json parser for validation.
+// See https://mzl.la/2vwqbGU for more details and allowed options.
+export const CSP_KEYWORD_RE = new RegExp([
+  '(self|none|strict-dynamic|',
+  'unsafe-inline|unsafe-eval|unsafe-hashed-attributes)',
+  // Only match these keywords, anything else is forbidden
+  '(?!.)',
+  '|(sha(256|384|512)-|nonce-)',
+].join(''));
--- a/src/const.js
+++ b/src/const.js
@@ -168,8 +168,7 @@ export const TEMPORARY_APIS = [
 // `script-src`. Used in manifest.json parser for validation.
 // See https://mzl.la/2vwqbGU for more details and allowed options.
 export const CSP_KEYWORD_RE = new RegExp([
-  '(self|none|strict-dynamic|',
-  'unsafe-inline|unsafe-hashed-attributes)',
+  '(self|none|strict-dynamic|unsafe-hashed-attributes)',
   // Only match these keywords, anything else is forbidden
   '(?!.)',
   '|(sha(256|384|512)-|nonce-)',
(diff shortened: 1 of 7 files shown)


EnTeQuAk:
cc @jonathanKingston @april to verify tests are correct and we're catching sensible stuff here.

src/parsers/manifestjson.js
```diff
@@ -170,6 +170,49 @@ export default class ManifestJSONParser extends JSONParser {
...
+
+            if (url.host) {
+              // Let's warn for everything trying to allow hosts as
+              // protocol prefixes are optional a user might try enabling
+              // `web.example.com:80` which is perfectly valid.
+              this.collector.addWarning(messages.MANIFEST_CSP);
+            }
+
+            // set value to protocol to validate it later.
+            value = url.protocol;
```
muffinresearch: Does this need to redefine value? Couldn't this be a new const?

src/utils.js
```diff
@@ -235,3 +235,30 @@ export function apiToMessage(string) {
 export function isBrowserNamespace(string) {
   return ['browser', 'chrome'].includes(string);
 }
+
+
+export function parseCspPolicy(policy) {
```
muffinresearch: Are there no good libs out there for this?

EnTeQuAk: There's https://www.npmjs.com/package/csp-parse and a few others which are all essentially the same or similar to this. I didn't want to pull in an external library for such an easy part.

src/parsers/manifestjson.js
EnTeQuAk: heh, I already refactored that code so that it is clearer.

EnTeQuAk:
This should now be ready for review, sorry for the spam, I changed it's implementation a few times back and forth and I think I'm happy now.

src/utils.js
muffinresearch: Yeah the ones I saw seems quite basic. The only benefit would be that they might benefit from updates - since the spec changes quite a bit that might be worth something.

tests/parsers/test.manifestjson.js
```diff
@@ -332,17 +332,96 @@ describe('ManifestJSONParser', function() {
 
   describe('content security policy', function() {
 
-    it('should warn that csp will mean more review', () => {
+    it('dont warn on CSP not including default-src or script-src)', () => {
```
muffinresearch: The `it` part should read as a sentence along with the string.

tests/parsers/test.manifestjson.js
```diff
@@ -332,17 +332,96 @@ describe('ManifestJSONParser', function() {
...
+        'script-src ws:',
+        'script-src wss:',
+        'script-src http:',
+        'script-src https:',
+        'script-src ftp:',
+        'script-src http://cdn.example.com/my.js',
+        'script-src https://cdn.example.com/my.js',
+
+        // Properly match mixed with other directives
+        'script-src https: \'unsafe-eval\'; object-src \'self\'',
```
muffinresearch: If you use a template string here you can avoid the escapes.

EnTeQuAk: Unfortunately not allowed by our eslint rules currently but a double-quote works too.

src/utils.js
EnTeQuAk: Well, none of them have been touched in recent years so I wouldn't be surprised if they're even maintained. If you're okay with it I feel more confident having that code in our code-base since we have to validate those rules on a relatively low level anyway.

EnTeQuAk:
>  I will have another look at this once the changes about 'script-src web.example.com' lands, as discussed in IRC.

That just landed. I'm still verifying wether or not we should match `worker-src` and @april did mention something so there might be a few more tests being added.

src/const.js
```diff
@@ -163,3 +163,13 @@ export const TEMPORARY_APIS = [
   'storage.local',
   'storage.sync',
 ];
+
+// All valid CSP keywords that are options to keys like `default-src` and
+// `script-src`. Used in manifest.json parser for validation.
+// See https://mzl.la/2vwqbGU for more details and allowed options.
+export const CSP_KEYWORD_RE = new RegExp([
+  '(self|none|strict-dynamic|unsafe-hashed-attributes)',
```
wagnerand: what about `unsafe-eval` and `unsafe-inline`?

tests/parsers/test.manifestjson.js
```diff
@@ -332,17 +332,111 @@ describe('ManifestJSONParser', function() {
...
+        'worker-src web.example.com',
+        'worker-src web.example.com:80',
+        'worker-src web.example.com:443',
+
+        // Properly match mixed with other directives
+        "script-src https: 'unsafe-eval'; object-src 'self'",
+
+        // unsafe-eval and unsafe-inline are forbidden too.
+        "script-src 'self' 'unsafe-eval';",
+        "script-src 'self' 'unsafe-inline';",
```
wagnerand: I don't think we are ready yet to forbid `unsafe-eval` and `unsafe-inline`.

src/const.js
EnTeQuAk: fixed in dd8f893

tests/parsers/test.manifestjson.js
EnTeQuAk: Fixed in dd8f893 - as discussed in IRC.

PR #1440, opened 2017-08-04
Fix path handling

### LINKED ISSUE #1431: validateIcons raises MANIFEST_ICON_NOT_FOUND on Windows even if icon exists

`MANIFEST_ICON_NOT_FOUND` is raised when run `web-ext lint`.

1. Set icon path with files in sub-directory.
```json
  "icons": {
    "48": "icons/icon.png",
    "96": "icons/icon@2x.png"
  },

```
2. Run `web-ext lint` in working directory on Windows.
3. MANIFEST_ICON_NOT_FOUND

### What happened?

`MANIFEST_ICON_NOT_FOUND` even if icons exists.

No error on Windows too.

It's because of path separator like #1426.

https://github.com/mozilla/addons-linter/blob/517b98b21cdff4fd850a5f713153506c4cdf92cd/src/parsers/manifestjson.js#L166 :
```javascript
if (!this.io.files.hasOwnProperty(path)) {
```

`this.io.files` has path like `icons\icon.png` not `icons/icon.png` when running on Windows. It causes `MANIFEST_ICON_NOT_FOUND`.

I tried fixing it and sending PR but it's not small to fix alone. `io.files`(it means both return value of `IOBase.getFiles()` and variables named `files` in some files) is used widely without any wrapping method. It seems to need design decision. In my opinion, paths should be stored as something other instead of plain string joinned with `path.join()`.

asamuzaK:
Fixes #1431

Convert URL `pathname` to native file path.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -17,7 +17,9 @@ function normalizePath(iconPath) {
   // . and .. automatically. We need an absolute URL to use as a base so we're
   // using https://example.com/.
   const { pathname } = new URL(iconPath, 'https://example.com/');
-  return path.join(pathname.split('/').filter(i => i));
+  return pathname.split("/").reduce((p, c) =>
+    p && path.join(p, decodeURIComponent(c)) || decodeURIComponent(c)
+  );
 }
 
 export default class ManifestJSONParser extends JSONParser {
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -17,7 +17,7 @@ function normalizePath(iconPath) {
   // . and .. automatically. We need an absolute URL to use as a base so we're
   // using https://example.com/.
   const { pathname } = new URL(iconPath, 'https://example.com/');
-  return pathname.slice(1);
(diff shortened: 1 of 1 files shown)


EnTeQuAk:
Thanks a lot for the patch @asamuzaK! Unfortunately I think it's best to think about a more generic fix for this and not handle this only specifically for the icon validation. I'm sure there are a few more places that handle this wrong and we should think about handling this directly in the `io` system, I don't have a plan yet but wanted to look at this on Monday.

If you come up with another idea, I'm all ears and looking forward to patches!

In any case, this needs at least a test to prove it's working - which will also help fixing it in another way.

asamuzaK:
Okay.
Should I close this?

EnTeQuAk:
I just did, sorry about that and thanks again for your work on that patch!

Please feel free to re-open this if you have a proper test or maybe another implementation :)

asamuzaK:
@EnTeQuAk
If it takes time to fix #1431, how about applying this PR temporarily?
I will add necessary updates and tests.

asamuzaK:
I mean if it take time to fix `io` system.

PR #1446, opened 2017-08-08
Flag Mozilla conditions of use violation.

### LINKED ISSUE #1285: Flag violations of "Mozilla conditions of use"
This will be based on a list of forbidden terms.

EnTeQuAk:
Fixes #1285 

This is currently a `notice` to not be too intrusive, is that okay? Imho this is primarily used for post-review so a notice upfront sounds fair to me and a warning overkill.

Also, this currently matches all files which I think is good but that would mean any other add-on with a stopword list like this will be flagged like hell.


src/linter.js
```diff
@@ -629,4 +634,25 @@ export default class Linter {
       });
   }
 
+  _markBadwordUsage(filename, fileData) {
+    if (fileData) {
+      const badwordsRe = new RegExp(badwords.en.join('|'), 'gi');
```
EnTeQuAk: This is prepared to match other languages but not really implemented yet unless we have a use-case for it.

wagnerand:
A `notice` sounds ok to me. We can always adjust if necessary.

src/linter.js
```diff
@@ -629,4 +634,25 @@ export default class Linter {
...
+
+      const words = fileData.split(' ');
+
+      for (let word of words) {
+        // Filter out none a-Z characters to simplify matching with our
+        // badword list.
+        word = word.toLowerCase().replace(/[^a-zA-Z]/, '');
+
+        if (word.match(badwordsRe)) {
```
wagnerand: I am a bit worried about the performance impacts of these broad regular expressions. Especially line 639 could be quite slow. Have we measured this with some sample set or real add-ons?

EnTeQuAk: I haven't, let me do that and get back to you on that. I'm not too worried though tbh, we are iterating through the AST quite a few times during the linting process which imho should be slower even than regexes.

Especially line 639 shouldn't be too slow imho. I'll bechmark that though with a few bigger add-ons and see if it has performance implications.

wagnerand: Thank you. Let me know if you need help finding big add-ons.

EnTeQuAk:
@wagnerand I moved the badwords regular expression construction to a constant which should improve performance. This indeed speeds things up a notch, apparently the regex in `word = word.toLowerCase().replace(/[^a-zA-Z]/, '');` get's properly cached or is simple enough that moving that to a constant didn't have any performance impact.

r?

src/linter.js
```diff
@@ -629,4 +633,23 @@ export default class Linter {
...
   }
 
+  _markBadwordUsage(filename, fileData) {
+    if (fileData) {
+      const words = fileData.split(' ');
+
+      for (let word of words) {
+        // Filter out none a-Z characters to simplify matching with our
+        // badword list.
+        word = word.toLowerCase().replace(/[^a-zA-Z]/, '');
```
wagnerand: As far as I can see, this only filters the first non `a-Z` character. Don't we want to replace all occurrences? (Note that this might have a bigger impact on performance).

src/linter.js
```diff
@@ -629,4 +633,23 @@ export default class Linter {
       });
   }
 
+  _markBadwordUsage(filename, fileData) {
+    if (fileData) {
+      const words = fileData.split(' ');
+
+      for (let word of words) {
+        // Filter out none a-Z characters to simplify matching with our
```
wagnerand: Nit: `none` => `non`

src/linter.js
EnTeQuAk: This only normalizes words from the code-file so that we can easier match them with the badwords regular expression. All words in the badwords json file are `a-Z` only for now. I want to avoid doing needless matches for terms like `{` and stuff. I just pushed 24b0f48 that implements that.

wagnerand: Quick update: We discussed this in IRC and the regexp needs an update to match *all* non a-z characters. The current one only matches the very first one.

src/linter.js
```diff
@@ -634,21 +635,15 @@ export default class Linter {
...
-
-        if (word && word.match(constants.BADWORDS_RE.en)) {
-          this.collector.addNotice(
-            Object.assign({}, messages.MOZILLA_COND_OF_USE, {
-              file: filename,
-            })
-          );
-        }
+    if (fileData && fileData.trim()) {
+      const sanitizedFileData = fileData.replace(/[^a-z]/g, '').trim();
```
wagnerand: I think the `.trim()` is unnecessary since the `replace` should already have taken care of whitespaces?

EnTeQuAk: fair point, I'l skip that then.

wagnerand: I'd also remove the trim in line 638, because the regexp is doing it anyway.

PR #1449, opened 2017-08-08
Import Firefox 56 schemas

### LINKED ISSUE #1346: Update schemas to Firefox 56
Once Firefox 56 is on Beta we should import the schemas, cut a new release and update web-ext to use it for linting.

If there are interesting things landing in Firefox 56 please add a comment so it's a bit easier to track what's incoming.

mstriemer:
Looked through it and it looks okay. The first update to manifest.json doesn't look quite right. Going to open the PR and see if it has any side effects.

Fixes #1346.

### DIFF
--- a/src/schema/imported/alarms.json
+++ b/src/schema/imported/alarms.json
@@ -58,7 +58,8 @@
                   "$ref": "#/types/Alarm"
                 },
                 {
-                  "name": "alarm"
+                  "name": "alarm",
+                  "optional": true
                 }
               ]
             }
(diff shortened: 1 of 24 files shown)


mstriemer:
Some quick testing shows that permissions are validated fine. I'm not sure what's causing it to format like that though.

muffinresearch:
@mstriemer see the lint config PR there is a broken test relating to schema permissions. https://github.com/mozilla/addons-linter/pull/1451#discussion_r131971852

mstriemer:
I filed #1459 but I don't think that's blocking this. It fails the same way on this branch and on master.

mstriemer:
These changes are up for review again in #1461.

PR #1451, opened 2017-08-08
Configure amo eslint config

### LINKED ISSUE #1391: yarn test should run watch and not coverage by default
Would be nice to have the default test command be more useful in development. 

By default running coverage slows things down too so that could be skipped by default.

Moving what the current test command does to a CI specific command would mean that the right things continue to be run for CI.

### LINKED ISSUE #1450: Use the amo eslint config
This will sync this repo up to use the same eslint config as the addons-frontend project.

muffinresearch:
Fixes #1450
Fixes #1391 (it made testing easier)

It found a couple of real issues:

* There was a broken regex which would match `browser/components/extensions/schemas/lkfjghdsfhskdjhfjson` and not just `browser/components/extensions/schemas/lkfjghdsfhskdjhf.json`
* There's a test that had a bad loop I've commented out the data that doesn't validate - it's a schema test and I will call it out in the diff where possible. This one will need a bug filed to fix up in due course.

A lot of the noise is spaces around `{` etc. 
The bigger changes were changes to a lot of the loops to move away from `for..in` and `for..of`. 

There's a lot of changes here so it'll be worth trying to be diligent with review incase something got broken in the process.


tests/schema/test.permissions.js
```diff
@@ -24,30 +24,31 @@ describe('/permissions', () => {
...
   });
 
-  var matchingPatterns = [
+  const matchingPatterns = [
     '*://developer.mozilla.org/*',
     'http://developer.mozilla.org/*',
-    'https://foo.com',
-    'ftp://do.people.still.use.this',
-    'app://wat',
+    // FIXME: the loop in this test was previously broken.
```
muffinresearch: This is the bad loop that needs a bug filing to fix. The `matchPattern` seen in the test was always the last value in the list.

src/io/xpi.js
```diff
@@ -97,7 +94,7 @@ export class Xpi extends IOBase {
   }
 
   checkPath(path) {
-    if (!this.files.hasOwnProperty(path)) {
+    if (!Object.prototype.hasOwnProperty.call(this.files, path)) {
```
EnTeQuAk: what's the difference here?

src/scanners/binary.js
```diff
@@ -13,7 +13,9 @@ export default class BinaryScanner extends BaseScanner {
   }
 
   check(buffer, values) {
-    for (let v in values) {
+    const valuesKeys = Object.keys(values);
+    for (let i = 0; i < valuesKeys.length; i++) {
+      const v = valuesKeys[i];
```
EnTeQuAk: why no forEach here?

src/schema/firefox-schemas-import.js
```diff
@@ -28,8 +30,8 @@ const UNRECOGNIZED_PROPERTY_REFS = [
 ];
 
 const schemaRegexes = [
-  new RegExp('browser/components/extensions/schemas/.*\.json'),
-  new RegExp('toolkit/components/extensions/schemas/.*\.json'),
+  new RegExp('browser/components/extensions/schemas/.*\\.json'),
+  new RegExp('toolkit/components/extensions/schemas/.*\\.json'),
```
EnTeQuAk: great catch!

src/schema/formats.js
```diff
@@ -22,11 +18,13 @@ export function isValidVersionString(version) {
...
   }
-  var parts = version.split('.');
+  const parts = version.split('.');
   if (parts.length > 4) {
     return false;
   }
-  for (var part of parts) {
+
+  for (let i = 0; i < parts.length; i++) {
+    let part = parts[i];
```
EnTeQuAk: also no forEach here?

tests/io/test.xpi.js
```diff
@@ -542,19 +532,17 @@ describe('Xpi.getFileAsStream()', function() {
         expect(err.message).toContain('¡hola!');
       });
   });
-
 });
 
-describe('Xpi.getFilesByExt()', function() {
-
+describe('Xpi.getFilesByExt()', function getFilesByExtCallback() {
```
EnTeQuAk: Why are they named now? Can't we just use lambdas here as well, afair we do that quite often in other tests?

src/io/xpi.js
muffinresearch: See http://eslint.org/docs/rules/no-prototype-builtins

package.json
```diff
@@ -10,9 +10,9 @@
     "start": "grunt start",
     "build": "grunt build",
     "eslint": "eslint .",
-    "test": "jest tests/ --runInBand --coverage",
-    "test-no-coverage": "jest tests/ --runInBand",
-    "test-ci": "npm run test && cat ./coverage/lcov.info | coveralls",
+    "test": "jest tests/ --runInBand --watch",
```
muffinresearch: Would be good to get to the bottom of why that is. Not using watch defeats one of the best features of Jest.

src/scanners/binary.js
muffinresearch: Because there's no way to break out of a `forEach` without throwing. 

Thinking about it this is probably something where we could be using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

It could probably be re-written as:

```
if (Object.keys(values).some((v) => values[v] !== buffer[v])) {
  return;
}
```

src/schema/formats.js
muffinresearch: Same reason as before.

tests/io/test.xpi.js
muffinresearch: One of the lint rules prevents unnamed functions. In our tests we have a few places where we rely on `this`, and the semantics of a fat arrow breaks that code from working . We should probably look at refactoring those tests and set `const`s instead of relying on `this`.

pdehaan: FWIW, you could just create a nested ./tests/.eslintrc which disables that one unnamed function rule, and all the other rules from the root .eslintrc file would still be inherited.

package.json
muffinresearch: I suppose one of the other goals I was aiming for was to try and make the test command setup more consistent between addons-frontend and addons-linter since that's another difference we can eliminate. 

If I added the `test-once` command would that provide what you need? See https://github.com/mozilla/addons-frontend/#development-commands for more details.

tests/io/test.xpi.js
muffinresearch: @pdehaan If we end up concluding we don't need the rule I'd rather disable it centrally in the shared config since we're trying to make our config more consistent across repos. Of course there might be some cases where we _do_ need overrides but we'd probably want to keep that to an absolute minimum.

src/scanners/binary.js
muffinresearch: I'm going to switch this to use `Array.prototype.some`. It's purpose built for this use-case.

package.json
muffinresearch: I've added `yarn test-once` and `test-coverage-once`.

muffinresearch:
@EnTeQuAk if you're happy with this and want to land it feel free to go ahead and squash/merge it. Otherwise we can fix up any additional bits next week.

package.json
EnTeQuAk: Perfect, thanks a lot! I think that works for me.

EnTeQuAk:
Thank you so much for the explanations and work that went into this! This is looking great :)

So I just merged this, if there's anything else that needs to be done I'm sure we can easily do it in a separate PR.

Issue #572, opened 2016-03-17
Point user to README or run build if they try to run the binary without building it first


tofumatt:
If the user doesn't RTFM, they will get an error trying to run `./bin/addons-linter` and may think the linter is busted. We should catch that error and handle it; this is silly but polite.

eviljeff:
This would be a good first bug for someone who knows Javascript.

Getting started for any contributors: https://wiki.mozilla.org/Add-ons/Contribute/AMO/Code

Revanth47:
@eviljeff Can I take up this issue?

eviljeff:
@Revanth47 sure!  It should just be a matter of changing https://github.com/mozilla/addons-linter/blob/master/bin/addons-linter to check for the presence of the executable or folder and giving an appropriate message to the user if it's not built yet. @tofumatt should be able to provide more detail.

Revanth47:
So I just check if the "dist" folder exists and if it doesn't I'll just display this link '[https://github.com/mozilla/addons-linter#readme](https://github.com/mozilla/addons-linter#readme)'

wagnerand:
@tofumatt @eviljeff could you reply to @Revanth47 please?

tofumatt:
> So I just check if the "dist" folder exists and if it doesn't I'll just display this link 'https://github.com/mozilla/addons-linter#readme'? @tofumatt @eviljeff

Sorry I missed this!

Basically: yes. You specifically want to check for the [built file in `dist/` that `/bin/addons-linter` tries to `require`](https://github.com/mozilla/addons-linter/blob/master/bin/addons-linter#L8): that file is `dist/addons-linter.js`.

wagnerand:
@Revanth47 are you still working on this?

Divya063:
can I take up this issue?

tofumatt:
@Divya063 Yes, feel free to work on it, thanks 😀

Divya063:
@tofumatt please review

wagnerand:
This got fixed in https://github.com/mozilla/addons-linter/pull/1492

PR #1485, opened 2017-08-21
Points the user to README or run build if they try to run the binary without building it first

Divya063:
Fixes [mozilla/addons#572](https://github.com/mozilla/addons-linter/issues/572)

### DIFF
--- a/bin/addons-linter
+++ b/bin/addons-linter
@@ -10,6 +10,6 @@ try {
   .run();
 } catch (err) {
   if (err.code == 'MODULE_NOT_FOUND') {
-    console.log('You did not build addons-linter yet. Please run 'yarn' and 'yarn build' or see README.md for more information');
+    console.log("You did not build addons-linter yet. Please run 'yarn' and 'yarn build' or see README.md for more information");
   }
 }
--- a/bin/addons-linter
+++ b/bin/addons-linter
@@ -4,14 +4,13 @@ var path = require('path');
 
 var absoluteAppRoot = path.resolve(path.join(__dirname, '..'));
 global.appRoot = path.relative(process.cwd(), absoluteAppRoot);
-
-try {
+try{
     require('../dist/addons-linter')
-       .createInstance({runAsBinary: true})
-        .run();
-} catch (e) {
+    .createInstance({runAsBinary: true})
+    .run();
+}
+catch (e) {
     if (e.code == 'MODULE_NOT_FOUND') {
-        console.log("Please refer https://github.com/mozilla/addons-linter#readme");
+       console.log('You did not build addons-linter yet. Please run \'npm install\' and \'yarn build\' or see README.md for more information');
     }
-   
 }
--- a/bin/addons-linter
+++ b/bin/addons-linter
@@ -4,13 +4,12 @@ var path = require('path');
 
 var absoluteAppRoot = path.resolve(path.join(__dirname, '..'));
 global.appRoot = path.relative(process.cwd(), absoluteAppRoot);
-try{
-    require('../dist/addons-linter')
-    .createInstance({runAsBinary: true})
-    .run();
-}
-catch (e) {
-    if (e.code == 'MODULE_NOT_FOUND') {
-       console.log('You did not build addons-linter yet. Please run \'npm install\' and \'yarn build\' or see README.md for more information');
-    }
+try {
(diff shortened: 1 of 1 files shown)


bin/addons-linter
```diff
@@ -5,6 +5,13 @@ var path = require('path');
...
-require('../dist/addons-linter')
-  .createInstance({runAsBinary: true})
-  .run();
+try {
+    require('../dist/addons-linter')
+       .createInstance({runAsBinary: true})
+        .run();
+} catch (e) {
+    if (e.code == 'MODULE_NOT_FOUND') {
+        console.log("Please refer https://github.com/mozilla/addons-linter#readme");
```
EnTeQuAk: I would add a bit more information to it. Let's do maybe something like `You did not build addons-linter yet. Please run 'npm install' and 'yarn build' or 'npm build' or see https://github.com/mozilla/addons-linter#readme for more information` cc @tofumatt for wording.

Let's also use single-quotes as everywhere else in the module.

tofumatt: I think you'll need `npm run build` and I would stick with solely npm or yarn commands (don't mix them) but otherwise sounds good.

bin/addons-linter
```diff
@@ -5,6 +5,13 @@ var path = require('path');
 var absoluteAppRoot = path.resolve(path.join(__dirname, '..'));
 global.appRoot = path.relative(process.cwd(), absoluteAppRoot);
 
-require('../dist/addons-linter')
-  .createInstance({runAsBinary: true})
-  .run();
+try {
+    require('../dist/addons-linter')
```
tofumatt: This indentation is off; please make sure you run `npm test` before committing to make sure you don't have any style errors.

bin/addons-linter
tofumatt: I think you could simply say "see README.md" for more information.

Divya063:
@tofumatt @EnTeQuAk Thanks for the review, I applied the requested changes.

bin/addons-linter
```diff
@@ -4,7 +4,13 @@ var path = require('path');
 
 var absoluteAppRoot = path.resolve(path.join(__dirname, '..'));
 global.appRoot = path.relative(process.cwd(), absoluteAppRoot);
-
-require('../dist/addons-linter')
-  .createInstance({runAsBinary: true})
-  .run();
+try{
+    require('../dist/addons-linter')
```
tofumatt: It's odd the linter isn't catching this but these indents are off compared to our usual style. would you mind changing the layout to:

```js
try {
  require('../dist/addons-linter')
  .createInstance({ runAsBinary: true })
  .run();
} catch (err) {
  if (e.code == 'MODULE_NOT_FOUND') {
    console.log('You did not build addons-linter yet. Please run \'npm install\' and \'yarn build\' or see README.md for more information');
  }
}
```

tofumatt: I'm guessing, actually, that we don't lint this file. We should do that! I've filed #1487.

bin/addons-linter
```diff
@@ -4,7 +4,13 @@ var path = require('path');
...
+    .createInstance({runAsBinary: true})
+    .run();
+}
+catch (e) {
+    if (e.code == 'MODULE_NOT_FOUND') {
+       console.log('You did not build addons-linter yet. Please run \'npm install\' and \'yarn build\' or see README.md for more information');
```
tofumatt: As mentioned please use just one command for both things. Looks like the README leans toward `yarn`, so this should read `"Please run `yarn` and `yarn build` or see README.md for more information"`.

Note also the use of double-quotes; they're fine to use here because it saves you having the escape the `'` character.

bin/addons-linter
Divya063: @tofumatt  yes I used the command `npm test` to check the style errors and I didn't receive any error message regarding this.

tofumatt: Yeah, sorry about that. Looks like this file isn't checked for style errors 😢

Divya063: @tofumatt no problem :smile:

Divya063:
@tofumatt applied the changes

tofumatt:
Looks good, I've made a few tweaks in another PR but I think this is basically good to merge. Thanks for your contribution! 😄

Issue #1172, opened 2017-03-09
Let SemVer for WebExtensions add-on versioning


asamuzaK:
### Describe the problem and steps to reproduce it:
(Please include as many details as possible.)

1. Create some WebExtensions add-on
2. Fill in `"version": "1.0.0-a.1"` for example in manifest.json
3. Upload to AMO

### What happened?

Upload is rejected due to version string.

### What did you expect to happen?

Accept version control in SemVer.

### Anything else we should know?
(Please include screenshots and any relevant files.)

Initially filed at [1345685 - WebExtensions: Let SemVer for versioning](https://bugzilla.mozilla.org/show_bug.cgi?id=1345685)

gneatgeek:
+1

Albeit that the version number is a tad weird, but our generator came up with "10.1704.14.024" as a version number and the linter threw an error for it. Worked fine once the 0 was removed "10.1704.14.24".

EnTeQuAk:
Before this can be tackled, addons-server needs to understand these versions properly too: https://github.com/mozilla/addons-server/blob/master/src/olympia/versions/compare.py - this is needed so that we can understand and compare versions across the system.

julienw:
Here is my conversion function in JS:
```javascript
function semverToWebExtVersion(version) {
  return version.replace(/-(\w+)\.(\d+)$/, '$1$2');
}
```

EnTeQuAk:
So this got closed because semver isn't supported by Firefox thus the linter can't support it either. @asamuzaK added lots of tests that verify a more correct behavior of our version format matching.

PR #1494, opened 2017-08-23
Support semver

### LINKED ISSUE #1439: Error message for versions like 1.4.0-pre.0
Error message for a bad version format is too cryptic.

### What happened?
Used a version like 1.4.0-pre.0 for a beta version of my addon.
Got the error: "/version" should match format "versionString"

Either no error, or an error explaining the format

asamuzaK:
Fixes  #1439, #1172

### DIFF
--- a/src/schema/formats.js
+++ b/src/schema/formats.js
@@ -1,48 +1,29 @@
 import { URL } from 'whatwg-url';
 
-const VALIDNUMRX = /^[0-9]{1,5}$/;
-
 // Firefox's version format is laxer than Chrome's, it accepts:
 // https://developer.mozilla.org/en-US/docs/Toolkit_version_format
 // We choose a slightly restricted version of that format (but still more
 // permissive than Chrome) to allow Beta addons, per:
 // https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance
-const TOOLKIT_VERSION_REGEX = /^(\d+\.?){1,3}\.(\d+([A-z]+(-?[\dA-z]+)?))$/;
+// We also support Semantic Versioning:
+// http://semver.org/
+const CHROME_PART = '(?:0|[1-9]\\d{0,3}|[1-5]\\d{4}|6(?:[0-4]\\d{3}|5(?:[0-4]\\d{2}|5(?:[0-2]\\d|3[0-5]))))';
+const CHROME_VERSION = `${CHROME_PART}(?:\\.${CHROME_PART}){0,3}`;
+const ADDON_BETA_PART = '(?:a(?:lpha)?|b(?:eta)|pre|rc)\\d*';
+const ADDON_VERSION = `${CHROME_VERSION}(?:${ADDON_BETA_PART})?`;
+const SEMVER = '(?:0|[1-9]\\d*)(?:\\.(?:0|[1-9]\\d*)){2}(?:-[A-Za-z\\d\\-]+(?:\\.[A-Za-z\\d\\-]+)*)?(?:\\+[A-Za-z\\d\\-]+(?:\\.[A-Za-z\\d\\-]+)*)?';
+const WEBEXT_VERSION_REGEXP = new RegExp(`^(?:${ADDON_VERSION}|${SEMVER})$`);
 
 export function isValidVersionString(version) {
   // We should be starting with a string.
-  if (typeof version !== 'string') {
+  if (!(typeof version === 'string' || version instanceof String)) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


asamuzaK:
Notes:
* `'0.1.12dev-cb31c51'`,
* `'4.1.1dev-abcdef1'`,
moved to invalid string, because `dev` is not allowed for beta string.

* `'1.0.0-beta2'`
moved to valid string, because it is valid semver.

asamuzaK:
https://github.com/mozilla/addons-linter/issues/1439#issuecomment-324538056

I'll rewrite PR when how to treat '0.1.12dev-cb31c51', '4.1.1dev-abcdef1' should be is decided.

Another question from me.
Currently, isToolkitVersionString() === isValidVersionString() (Some kind of an alias).
But, when we support semver, valid version string is not always equal to toolkit version string.

Should we keep isToolkitVersionString an alias? (current implement)

Or should we treat those function to match there name, like below? (may break some functions)

```
const TOOLKIT_VERSION_REGEX = ...;
function isValidVersionString(version) {
  if (TOOLKIT_VERSION_REGEX.test(version)) {
    return true;
  }
  const semverReg = ...;
  return semverReg.test(version);
}
function isToolkitVersionString(version) {
  return TOOLKIT_VERSION_REGEX.test(version);
}
```

EnTeQuAk:
`isToolkitVersionString` and `isValidVersionString` interestingly call each other which I think feels wrong. Imho `isValidVersionString` should call `isToolkitVersionString` for an early return and then go on parse the rest. The usage of `isToolkitVersionString` in `src/parsers/manifestjson.js` should be replaced by `isValidVersionString`.

src/schema/formats.js
```diff
@@ -1,48 +1,27 @@
...
+  // permissive than Chrome) to allow Beta addons, per:
+  // https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance
+  // We also support Semantic Versioning:
+  // http://semver.org/
+  const chromePart = '(?:0|[1-9]\\d{0,3}|[1-5]\\d{4}|6(?:[0-4]\\d{3}|5(?:[0-4]\\d{2}|5(?:[0-2]\\d|3[0-5]))))';
+  const chromeVer = `${chromePart}(?:\\.${chromePart}){0,3}`;
+  const betaPart = '(?:a(?:lpha)?|b(?:eta)?|pre|rc)\\d*';
+  const addonVer = `${chromeVer}(?:${betaPart})?`;
+  const semVer = '(?:0|[1-9]\\d*)(?:\\.(?:0|[1-9]\\d*)){2}(?:-[A-Za-z\\d\\-]+(?:\\.[A-Za-z\\d\\-]+)*)?(?:\\+[A-Za-z\\d\\-]+(?:\\.[A-Za-z\\d\\-]+)*)?';
+  return (new RegExp(`^(?:${addonVer}|${semVer})$`)).test(version);
```
wagnerand: I think those regular expressions should use what Firefox accepts, not more and not less.

asamuzaK: I am not a native English speaker, so I can not understand a bit of the meaning of what you are pointing out.
What does "use what Firefox accepts" mean?

wagnerand: We should use the format that Firefox supports.

asamuzaK: You mean only Toolkit?

asamuzaK: Chrome version format always matches Toolkit. (&lt;number-a&gt;)
SemVer should match Toolkit too.
For example, `1.0.0-a1b1`.
According to [Toolkit version format | MDN](https://developer.mozilla.org/en-US/docs/Toolkit_version_format "Toolkit version format | MDN"), this version looks like it doesn't match Toolkit version format.
In 3rd version part (`0-a1b1`), &lt;number-a&gt; is `0` &lt;string-b&gt; is `-a` &lt;number-c&gt; is `1` &lt;string-d&gt; is `b`, last `1` does not match.
But according to [nsIVersionComparator - Mozilla | MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIVersionComparator "nsIVersionComparator - Mozilla | MDN"), it says
&gt; &lt;number-a&gt;&lt;string-b&gt;&lt;number-c&gt;&lt;string-d (everything else)&gt;
If so, then &lt;string-d&gt; becomes `b1`, and it matches Toolkit?

asamuzaK:
Closing this PR. See https://github.com/mozilla/addons-linter/issues/1439#issuecomment-326158984

PR #1507, opened 2017-09-03
Fix version format regexp

asamuzaK:
The current TOOLKIT_REGEX regular expression has a bug.
* For example, 1.00 is invalid but 1.00a1 is marked as valid.
* It treats other string than a|alpha|b|beta|pre|rc plus optional integer as valid.

This PR fix them.

NOTE: '0.1.12dev-cb31c51', '4.1.1dev-abcdef1' will be invalid in this PR.
Please don't merge unless decision is made that those strings should be invalid.
See [#1439 (comment)](https://github.com/mozilla/addons-linter/issues/1439#issuecomment-324538056 "Error message for versions like 1.4.0-pre.0") and [#1439 (comment)](https://github.com/mozilla/addons-linter/issues/1439#issuecomment-324712461 "Error message for versions like 1.4.0-pre.0")


EnTeQuAk:
This looks great so far, I'll dig a bit deeper into the regexes later today and check them against a few more live add-ons

> NOTE: '0.1.12dev-cb31c51', '4.1.1dev-abcdef1' will be invalid in this PR.
Please don't merge unless decision is made that those strings should be invalid.

That decision was made in the comments you linked. Firefox won't be able to parse them properly - or rather might allow them but doesn't properly update - so we should forbid them.

asamuzaK:
Added tests for isToolkitVersionString()

src/schema/formats.js
```diff
@@ -1,48 +1,32 @@
...
-    return false;
-  }
-
-  for (let i = 0; i < parts.length; i++) {
-    let part = parts[i];
-    // Leading or multiple zeros not allowed.
-    if (part.startsWith('0') && part.length > 1) {
-      return false;
-    }
-    // Disallow things like 123e5 which parseInt will convert.
```
EnTeQuAk: This warrants a test, can you add that?

asamuzaK: Done.
And also added more test patterns.

EnTeQuAk: Ah ` '123e5'` in the invalid patterns, somehow I missed that.

Nice, thanks!

EnTeQuAk:
I just merged this, thank you so much for taking care of this!

PR #1514, opened 2017-09-05
Limit version string length.

EnTeQuAk:
Fixes bug 1393644.

There is more work going on on this at https://github.com/mozilla/addons-linter/pull/1507 but it'd like to get this merged first.

### DIFF
--- a/src/schema/formats.js
+++ b/src/schema/formats.js
@@ -14,7 +14,7 @@ const TOOLKIT_REGEXP =
   new RegExp(`^${VERSION_PART}(?:\\.${VERSION_PART}){0,3}${BETA_PART}$`);
 
 export function isValidVersionString(version) {
-  // We should be starting with a string.
+  // We should be starting with a string. Limit length, see bug 1393644
   if (typeof version !== 'string' || version.length > 100) {
     return false;
   }
@@ -22,8 +22,8 @@ export function isValidVersionString(version) {
 }
 
 export function isToolkitVersionString(version) {
-  // We should be starting with a string.
-  if (typeof version !== 'string') {
+  // We should be starting with a string. Limit length, see bug 1393644
+  if (typeof version !== 'string' || version.length > 100) {
     return false;
   }
   return TOOLKIT_REGEXP.test(version);
--- a/src/schema/formats.js
+++ b/src/schema/formats.js
@@ -11,7 +11,7 @@ const TOOLKIT_VERSION_REGEX = /^(\d+\.?){1,3}\.(\d+([A-z]+(-?[\dA-z]+)?))$/;
 
 export function isValidVersionString(version) {
   // We should be starting with a string.
-  if (typeof version !== 'string') {
+  if (typeof version !== 'string' || version.length > 100) {
     return false;
   }
   // If valid toolkit version string, return true early
--- a/tests/schema/test.formats.js
+++ b/tests/schema/test.formats.js
@@ -103,6 +103,7 @@ describe('formats', () => {
       '1.0.0.0.0a',
       '1.0.0a-1',
       '1.0.0a1.1',
+      `1.${'9'.repeat(100)}`,
     ];
 
     validToolkitVersionStrings.forEach((validToolkitVersionString) => {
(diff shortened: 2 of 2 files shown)


tests/schema/test.formats.js
```diff
@@ -41,6 +41,8 @@ describe('formats', () => {
       '1.0.0-rc1.0+001',
       '0.1.12dev-cb31c51',
       '4.1.1dev-abcdef1',
+      // eslint-disable-next-line prefer-template
+      '1.9' + '9'.repeat(99),
```
muffinresearch: You can have expressions in a template string so this could be:

```javascript
`1.${'9'.repeat(100)}`
```
and save on the eslint override.

src/schema/formats.js
```diff
@@ -15,7 +15,7 @@ const TOOLKIT_REGEXP =
 
 export function isValidVersionString(version) {
   // We should be starting with a string.
-  if (typeof version !== 'string') {
+  if (typeof version !== 'string' || version.length > 100) {
```
muffinresearch: Are we happy that there's no version strings that would hit this limit currently? It's unlikely but it would be good to check.

EnTeQuAk: Yes we are, I think. The following are the 15 longest version strings we have in our database.

| Length | Version string                                   | add-on slug                        |
| ------ | ------------------------------------------------ | ---------------------------------- |
| 47     | 1.9.2.20110409044302.1-signed.1-signed.1-signed  |    jeteye                          |
| 47     | 1.9.3.20120129113830.1-signed.1-signed.1-signed  |    jeteye                          |
| 46     | 0.4b2.2013051811.1-signed.1-let-fixed.1-signed   |    autoproxy                       |
| 46     | baidu_after_right-click.rev3.1-signed.1-signed   |    baidurightclick                 |
| 46     | 2.4.1+0.1-signed.1-signed.1-signed.1-let-fixed   |    direct-currency-converter-2-fo  |
| 44     | Release_Candidate_1.rev115.1-signed.1-signed     |    symfony2-debug-switch           |
| 44     | Virus_total_scan_url.rev11.1-signed.1-signed     |    virus-total-scan-url            |
| 44     | SimSimi_for_Firefox.rev227.1-signed.1-signed     |    simsimi-for-firefox             |
| 44     | 0.772.1-signed.1-signed.1-let-fixed.1-signed     |    vlcwatcha                       |
| 43     | initial-production.rev267.1-signed.1-signed      |    panther-front-end               |
| 43     | 0.4b2.2015031118.1-signed.1-signed.1-signed      |    autoproxymod                    | 
| 42     | 0.4.3.1Build201012161000.1-signed.1-signed       |    customize-your-web              |
| 42     | 0.5.3.1Build201101051500.1-signed.1-signed       |    mouseless-browsing              |
| 42     | 0.4.3.2Build201102090000.1-signed.1-signed       |    customize-your-web              |
| 42     | 0.4.3.3Build201201122030.1-signed.1-signed       |    customize-your-web              |

tofumatt: I think a TODO reminding us to explain why we have made this change is reasonable. 😅

asamuzaK: Those version strings are not WebExtensions compatible, so `false` will be returned anyway without having to determine the length of the string.
I thought the legacy add-ons are responsible for amo-validator, and addons-linter only need to consider WebExtensions? [Remove support for webextensions in amo-validator · mozilla/amo-validator@cce447e](https://github.com/mozilla/amo-validator/commit/cce447ebe7db90f32cae732f1e5c704ca248b1bc "Remove support for webextensions in amo-validator · mozilla/amo-validator@cce447e")

EnTeQuAk: That's true, I just ran the query through all add-ons no matter what to get an idea of how long they get. So 100 is a fairly good limit here.

> I think a TODO reminding us to explain why we have made this change is reasonable. 😅

True, I'll reference the bugzilla bug

asamuzaK: Indeed, problems can arise if the length of the string exceeds the range of Number.MAX_SAFE_INTEGER, so I can understand if it was `version.length > Number.MAX_SAFE_INTEGER`.
But I do not see much basis for `100`?

asamuzaK:
> `1.${'9'.repeat(100)}`
If you want to exactly check length, it should be `1.${'9'.repeat(99)}`?
You also should add beta string test too, I suppose.
```
`1.${'9'.repeat(99)}`,
`1a${'9'.repeat(99)}`,
```

EnTeQuAk:
mhmm, that really should not matter as they won't get splitted before and our length limit is super hard over the whole string.

PR #1515, opened 2017-09-05
fix(package): update eslint to version 4.6.1

EnTeQuAk:
Closes #1484

### DIFF
--- a/tests/helpers.js
+++ b/tests/helpers.js
@@ -17,7 +17,7 @@ export const fakeMessageData = {
 export const EMPTY_PNG = Buffer
   .from(oneLine`iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACklEQVR4nGMA
                 AQAABQABDQottAAAAABJRU5ErkJggg==`,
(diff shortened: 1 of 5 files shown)


muffinresearch:
Once eslint 4.x is in place the eslint-config-amo can be updated to 0.2.0.

wagnerand:
@EnTeQuAk r? (can't formally set you as the reviewer since you created the PR ¯\_(ツ)_/¯)

tests/helpers.js
```diff
@@ -17,7 +17,7 @@ export const fakeMessageData = {
 export const EMPTY_PNG = Buffer
   .from(oneLine`iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACklEQVR4nGMA
                 AQAABQABDQottAAAAABJRU5ErkJggg==`,
-        'base64')
```
EnTeQuAk: uhm, why? This is an argument to `from` so it should be indented like that? Oh, because two whitespaces? Maybe an idea to indent the base64 string too, mhmm, I don't care enough I guess

PR #1520, opened 2017-09-05
Partially revert "Fix version format regexp (#1507)"

EnTeQuAk:
This reverts commit 7dc81ce726457a5f5ad7fd9bbdbb642619452551

This adds the same fix to invalidate '0.1.12dev-cb31c51' and '4.1.1dev-abcdef1' though.

Might need another follow-up to (partially) fix mentioned issues from #1439 or #1172.

### DIFF
--- a/src/schema/formats.js
+++ b/src/schema/formats.js
@@ -1,24 +1,44 @@
 import { URL } from 'whatwg-url';
 
+const VALIDNUMRX = /^[0-9]{1,5}$/;
+
 // Firefox's version format is laxer than Chrome's, it accepts:
 // https://developer.mozilla.org/en-US/docs/Toolkit_version_format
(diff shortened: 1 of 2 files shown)


src/schema/formats.js
```diff
@@ -1,32 +1,52 @@
...
+    }
+    part = parseInt(part, 10);
+    if (Number.isNaN(part) || part < 0 || part > 65535) {
+      return false;
+    }
+  }
+  return true;
 }
 
 export function isToolkitVersionString(version) {
```
muffinresearch: Doesn't this essentially duplicate `isValidVersionString` but with duplication since it's running `isValidVersionString` anyway?

muffinresearch: Probably making it a simple alias is the way to go then.

Issue #1323, opened 2017-06-07
 Warn when detecting libraries and other unreadable code


diox:
For post-review, we want to increase priority of add-ons using unknown libraries we haven't specifically allowed (or denied) or other unreadable code. The linter should detect this and add a warning with a specific code that addons-server can look for.

This is vague, but I think a basic obfuscation detection is what we want here. @jvillalobos and @wagnerand for more details.

wagnerand:
Yes, obfuscation is what comes into my mind here. I *think* that should cover it pretty much, but let's hear from @jvillalobos to see whether there is something beyond that.

jvillalobos:
Yes, we want to detect code that is either minified or obfuscated. @kewisch pointed out that whitespace density could be a good indicator of this.

kewisch:
Also, complete lack of code comments (although that should not be a factor alone)

jvillalobos:
Somewhat related, in a meeting today @cr brought up retire.js (https://retirejs.github.io/retire.js/) as a tool that can detect libraries and flag outdated ones. I don't know if we're considered this for the linter before, but it looks like an interesting option.

EnTeQuAk:
There's already an issue for that, see https://github.com/mozilla/addons-linter/issues/699 for a few more details.

EnTeQuAk:
@diox @wagnerand I wonder if this should be a real warning for other users too or just a flag in the metadata so that addons-server knows about this given that we allow uploading of sources and stuff on AMO?

Maybe a warning that says "please remember to upload sources if you submit minified or otherwise unreadable code"?

wagnerand:
> Maybe a warning that says "please remember to upload sources if you submit minified or otherwise unreadable code"?

We are doing that already with mixed results.

I think the main idea here is to have a risk factor that detects minified code.

EnTeQuAk:
> I think the main idea here is to have a risk factor that detects minified code.

which means it'll just be added to the metadata similar to how jsLibs or other flagged files are. That sounds like a plan.

EnTeQuAk:
Implementation / Experimentation path:

**Easier, approach matching everything potentially minified:**
 * Does the file end with ".min.js"? (adds to a `could_be_minified` score)

From firefox devtools (also adds to a `could_be_minified` score):

```javascript 
const SAMPLE_SIZE = 30; // no of lines
const INDENT_COUNT_THRESHOLD = 20; // percentage

/**
 * Determines is the source text is minified.
[...]

PR #1525, opened 2017-09-06
First version of detecting minified code.

EnTeQuAk:
Refs #1323

cc @muffinresearch @wagnerand for ideas on what tests I could add to avoid false positives. On my list so far:

 * a file that only has two lines > 500 characters should not match
 * a file with only 8 lines of code but properly indented should not match
 * potentially clone one of our more fully-featured webextension zips and put a modified library in there to avoid any detections of other files in there

Any more ideas?

One more thing I noticed while grepping  through files: We can catch files that have a source map definition in them too, they're very certainly minified. E.g `//# sourceMappingURL=jquery.mobile-1.4.2.min.map` and stuff. ~~I'm working on adding this, that adds less guessing, more knowing to the mix.~~ -> done


src/utils.js
```diff
@@ -238,3 +237,58 @@ export function parseCspPolicy(policy) {
...
+  // Threshold in percent of indented lines to mark a file as not
+  // minified.
+  const indentCountThreshold = 20; // percentage
+
+  // Lenght of a line that looks suspicious of being minified
+  const hugeLinesLength = 500;
+
+  // Number of huge lines to also mark a file as potentially minified
+  // Hint: Minified AngularJS has 12 lines, jQuery 4
+  const hugeLinesThreshold = 4;
```
muffinresearch: I would consider reducing this to 1? I think it's quite unlikely to have a sane un-minified src file with over 500chars on a single line - or do you have a reason to think that would lead to false-positives?

src/utils.js
```diff
@@ -238,3 +237,58 @@ export function parseCspPolicy(policy) {
...
+  let lineStartIndex = 0;
+  let lines = 1;
+  let indentCount = 0;
+  let hugeLinesCount = 0;
+
+  // Strip comments.
+  const normalizedCode = code.replace(/\/\*[\S\s]*?\*\/|\/\/.+/g, '');
+
+  while (lines < sampleSize) {
+    lineEndIndex = normalizedCode.indexOf('\n', lineStartIndex);
```
muffinresearch: What if a file has no `\n`?

src/utils.js
```diff
@@ -238,3 +237,58 @@ export function parseCspPolicy(policy) {
...
+ */
+export function couldBeMinifiedCode(code) {
+  // Number of lines to look at, taken from the head of the code.
+  const sampleSize = 30;
+
+  // Threshold in percent of indented lines to mark a file as not
+  // minified.
+  const indentCountThreshold = 20; // percentage
+
+  // Lenght of a line that looks suspicious of being minified
```
muffinresearch: Typo

src/utils.js
EnTeQuAk: I could think of base64 encoded files, people do that for absolutely no reason 🤷‍♂️

EnTeQuAk: And, the indentation detection would detect files that are minified but have <4 lines that are super long so I'm not too worried.

src/utils.js
EnTeQuAk: Then it's still matched and could very well be a minified code. Do you mean, if there are absolutely no `\n` we should watch out for longer lines? Or are you referring to hacks like https://github.com/mozilla/addons-server/blob/master/static/js/lib/semicolon.js to fix some compiled code?

wagnerand:
> a file that only has two lines > 500 characters should not match

Hm, can you elaborate why this should not match?

EnTeQuAk:
> Hm, can you elaborate why this should not match?

Well, it might if these are the only two lines and there isn't any other properly indented code but if it's only these two but otherwise indented code it could be just a regular JavaScript file inho.

src/utils.js
```diff
@@ -238,3 +240,69 @@ export function parseCspPolicy(policy) {
...
+    if (lineEndIndex >= hugeLinesLength) {
+      hugeLinesCount++;
+    }
+
+    lineStartIndex = lineEndIndex + 1;
+    lines++;
+  }
+
+  return (
+    (indentCount * 100) < indentCountThreshold ||
```
muffinresearch: Unless I'm missing something, this doesn't seem to be checking a percentage of checked lines.

src/utils.js
muffinresearch: You're going to break out of the loop if no `\n` is found in the entire file. Whilst the check will return true currently you probably want a test to cover this eventually in-case the return logic changes in some way.

EnTeQuAk: I'll add a test

src/utils.js
EnTeQuAk: indeed, that should be `((indentCount / lines) * 100) < indentCountThreshold`, I blame the keyboard 🙈

wagnerand:
> Well, it might if these are the only two lines and there isn't any other properly indented code...

But shouldn't this mean the linter should match (ie recognize a minified file)?

src/utils.js
EnTeQuAk: Fixed in e05bb91de236bab06eb62021fb4ed4ea9e700853

src/utils.js
EnTeQuAk: Fixed in e3ea083

EnTeQuAk:
> But shouldn't this mean the linter should match (ie recognize a minified file)?

So… let's be a bit more detailed: If a file has exactly two lines that aren't properly indented but <500 chars long it will be detected by our indentation checker since 100% of the lines in the sample size of lines (30) are not indented.

Does this answer your question?

This should be ready for another review now @wagnerand @muffinresearch, I added a few more tests and fixed your comments.

EnTeQuAk:
I'm going to merge this so we'll get more testing and throw more code against the detection. if there is anything we can fix them one at a time.

PR #1528, opened 2017-09-07
Added a check for type of script in warn-on-js

### LINKED ISSUE #1516: Inline script warning should only appear when scripts are actually executed
Visit https://addons.mozilla.org/en-US/firefox/files/compare/704052...577665/file/templates/plids.extensions.popup.main.search.htm#top

image

### What happened?
A validator message appeared saying `Warning: Inline scripts blocked by default`, but the "script" is actually just a HTML template: `<script type="text/html" id="my-html-template-used-in-knockout">`

The validator should just warn for scripts that actually execute, i.e. `type="something with javascript"` and without a type attribute. Possibly ask security if there are other cases.

harry-7:
Fixes #1516 
before raising a warning check for type of javascript and only rise the warning if the script is of type javascript.

### DIFF
--- a/tests/scanners/test.html.js
+++ b/tests/scanners/test.html.js
@@ -35,7 +35,7 @@ describe('HTML', () => {
       });
   });
 
-  it('should require <script> tag to have a src attribute', () => {
+  it('should warn on <script> tag without a src attribute and without type attribute', () => {
     const badHTML = validHTML('<script>alert()</script>');
     const htmlScanner = new HTMLScanner(badHTML, 'index.html');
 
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
Thanks a lot for the patch! This needs an update to the tests though, mind doing that too? You should find everything you need in `tests/scanners/test.html` in the test `should require <script> tag to have a src attribute`

harry-7:
@EnTeQuAk  Added the tests . PTAL

EnTeQuAk:
Uhm, did you push your changes?

harry-7:
@EnTeQuAk Aah Sorry. I pushed the changes now. PTAL

tests/scanners/test.html.js
```diff
@@ -47,6 +47,32 @@ describe('HTML', () => {
       });
   });
 
+
```
EnTeQuAk: let's just use one linebreak here

tests/scanners/test.html.js
```diff
@@ -47,6 +47,32 @@ describe('HTML', () => {
...
+    return htmlScanner.scan()
+      .then(({ linterMessages }) => {
+        expect(linterMessages.length).toEqual(1);
+        expect(linterMessages[0].code).toEqual(messages.INLINE_SCRIPT.code);
+        expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
+      });
+  });
+
+
```
EnTeQuAk: same here, just one line break

harry-7:
Will check for lint errors before pushing changes from the next time :-). Made changes. PTAL

PR #1533, opened 2017-09-10
Added bin folder to eslint scans 

### LINKED ISSUE #1487: Run eslint on bin/addons-linter
JS linting/style errors aren't being caught in `bin/addons-linter`; I'm guessing we don't include it in our files to lint.

We should! 😄

This should just be a matter of adding it to the files `eslint` scans.

Divya063:
Fixes #1487 
bin folder is being included while using command `npm eun eslint`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,7 @@
-{
+{ 
+  "rules":{
+    "no-console":0
+  },
   "env": {
     "node": true,
     "browser": true,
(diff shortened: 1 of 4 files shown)


muffinresearch:
@Divya063 thanks for the patch! To get CI passing, the next step is to fix the lint violations in the relevant files - see the output for details (here's a screenshot - click on it to view larger):

job__4038_2_-_mozilla_addons-linter_-_travis_ci

Divya063:
@muffinresearch Thanks for the response, I fixed those lint violations.

PR #1555, opened 2017-09-20
Validates the size of icons

### LINKED ISSUE #1329: Validate the size of icons

apoorvasingh17:
Fixes #1329


EnTeQuAk:
Thanks a lot @apoorvasingh17 for that patch!

Unfortunately I think there are a few issues with that patch. Checking the icon size based on a filename is mostly trusting our developers to name things correctly which I think we should never do. Like, if I have an icon with the name `my-icon-64` but in reality it's just a 32x32 icon.

So eventually we'll have to start opening the files and read the metadata for those files which for example is what https://github.com/mstriemer/addons-linter/commit/5cbdaaeed8edabb43efff20ca256a6199b003cd1 started to do. Maybe you can draw some inspiration from that?

EnTeQuAk:
Worth noting that someone else is working on checking validity of icon files in https://github.com/mozilla/addons-linter/pull/1558 so it's best to try to have your work as separated as possible.

EnTeQuAk:
Hey @apoorvasingh17 how is it going with this patch? #1558 just got merged so you should make sure to rebase with the current master.

Let me know if you run into any problems or have any questions.

EnTeQuAk:
I'm closing this PR due to inactivity. Please feel free to re-open anytime once you have time to follow-up on this patch.

PR #1558, opened 2017-09-22
Added check to make sure icon has a valid extension

### LINKED ISSUE #1328: Check that icons are valid image files
We should try to validate that the image is valid, to avoid getting broken images in Firefox.

ravneette:
Fixes #1328

      PRs open for the same issue.
* [x] Add `Fixes #ISSUENUM` at the top of your PR.
* [x] Add a description of the the changes introduced in this PR.

Added a check to make sure that the icon has a valid extension. Gives a warning in case the extension of the image doesn't match a valid image format.
The valid extensions included for now are :   jpg, jpeg, webp, giff, png, svg

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -151,7 +151,6 @@ export const NO_DEFAULT_LOCALE = {
 
 export const WRONG_ICON_EXTENSION = {
   code: 'WRONG_ICON_EXTENSION',
-  legacyCode: null,
   message: _('Unsupported image extension'),
   description: _('Icons should be one of JPG, JPEG, WebP, GIFF, PNG or SVG.'),
   file: MANIFEST_JSON,
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -153,6 +153,6 @@ export const WRONG_ICON_EXTENSION = {
   code: 'WRONG_ICON_EXTENSION',
   legacyCode: null,
   message: _('Consider Adding another icon'),
-  description: sprintf('Icons Should be of type jpg, jpeg, webp, giff, png or svg.'),
+  description: _('Icons should be one of JPG, JPEG, WebP, GIFF, PNG or SVG.'),
   file: MANIFEST_JSON,
 };
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -152,7 +152,7 @@ export const NO_DEFAULT_LOCALE = {
 export const WRONG_ICON_EXTENSION = {
   code: 'WRONG_ICON_EXTENSION',
   legacyCode: null,
-  message: _('Consider Adding another icon'),
+  message: _('Unsupported image extension'),
   description: _('Icons should be one of JPG, JPEG, WebP, GIFF, PNG or SVG.'),
   file: MANIFEST_JSON,
 };
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -148,3 +148,11 @@ export const NO_DEFAULT_LOCALE = {
     See: https://mzl.la/2hjcaEE`),
   file: MANIFEST_JSON,
 };
+
+export const WRONG_ICON_EXTENSION = {
+  code: 'WRONG_ICON_EXTENSION',
+  legacyCode: null,
+  message: _('Consider Adding another icon'),
+  description: sprintf('Icons Should be of type jpg, png or svg'),
+  file: MANIFEST_JSON,
+};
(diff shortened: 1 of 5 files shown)


EnTeQuAk:
Thanks @ravneette for that patch!

For some reason I thought I answered to https://github.com/mozilla/addons-linter/issues/1328#issuecomment-331130082 but apparently I didn't so here is what I wanted to comment on that issue:

I do think eventually we should verify that the image can be opened and check for potential image sizes. I think looking at https://github.com/mstriemer/addons-linter/commit/5cbdaaeed8edabb43efff20ca256a6199b003cd1#diff-baa432d9354e3de899d28a2638d1499eR31 and the patch in general for inspiration is a great idea.

Otherwise this is a great start and we can improve on that in a separate patch so let's start with a short review for this patch:

 * You added `jpg`, `png` and `svg` to the list of allowed files. Let's maybe add at least `jpeg` (as an often used alias for `jpg`), `webp` and `giff` to the list. I don't know yet what Firefox supports but we can figure that out somewhere to make sure we don't allow anything Firefox doesn't support (e.g tiff images are where I'm not so sure)
 * It's best to compare a lower cased file extension to the list to make sure we match all file names, lower or upper cased
 * This obviously needs tests but I think you're already working on them?

Otherwise, keep on going!

EnTeQuAk:
Worth noting that someone else is working on checking validity of icon files in https://github.com/mozilla/addons-linter/pull/1555 so it's best to try to have your work as separated as possible.

ravneette:
Thanks for the review @EnTeQuAk 
Will make sure to keep my work disjoint from #1555 .
Added the other extensions and shifted to making the comparison among lowercase strings.
Also, do you think I should add this as a warning or as an error?
Working on writing tests now.

EnTeQuAk:
> Will make sure to keep my work disjoint from #1555 .

Cool, thanks!

> Added the other extensions and shifted to making the comparison among lowercase strings. Also, do you think I should add this as a warning or as an error?

I don't think it's worth an error yet. If we are going to check for validity of images, e.g check that they can be opened and that they represent what they should, we can start raising errors if that check fails.

But for now we are only checking file-extensions so that should not be too agressive.

ravneette:
Great!
I restricted the code to throw a warning instead of an error.

src/messages/manifestjson.js
```diff
@@ -165,3 +165,11 @@ export const NO_DEFAULT_LOCALE = {
     See: https://mzl.la/2hjcaEE`),
   file: MANIFEST_JSON,
 };
+
+export const WRONG_ICON_EXTENSION = {
+  code: 'WRONG_ICON_EXTENSION',
+  legacyCode: null,
+  message: _('Consider Adding another icon'),
+  description: sprintf('Icons Should be of type jpg, jpeg, webp, giff, png or svg.'),
```
EnTeQuAk: This should not use `sprintf` since you're not using format-strings. You can just use `_('')` to mark the string as translatable.

Also, I'd write something like `Icons should be one of JPG, WebP, GIFF, PNG or SVG` which is what we want eventually.

tests/parsers/test.manifestjson.js
```diff
@@ -739,5 +739,26 @@ describe('ManifestJSONParser', () => {
         description: 'Icon could not be found at "icons/icon-64.png".',
       });
     });
+
+    it('adds an error if the icon does not have a valid extension', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        icons: {
+          32: 'icons/icon-32.txt',
+          64: 'icons/icon-64.html',
```
EnTeQuAk: Let's add one or two valid icons so that we can verify we don't match types we're not supposed to match.

ravneette:
Hey @EnTeQuAk 
Modified the test and removed sprintf

src/messages/manifestjson.js
```diff
@@ -165,3 +165,11 @@ export const NO_DEFAULT_LOCALE = {
     See: https://mzl.la/2hjcaEE`),
   file: MANIFEST_JSON,
 };
+
+export const WRONG_ICON_EXTENSION = {
+  code: 'WRONG_ICON_EXTENSION',
+  legacyCode: null,
+  message: _('Consider Adding another icon'),
```
EnTeQuAk: Sorry, I didn't notice this earlier. This message is prominently shown to developers so it should be more explicit.

Something along the lines of `Unsupported image extension` and then the `description` goes into the details.

Mind changing that?

ravneette:
Ya sorry my bad on that.
Should've been more explicit.
Changed it to "Unsupported image extension"

ravneette:
Hey @EnTeQuAk , thanks for the guidance. 

Since you mentioned the follow up, I've been brainstorming on the same. Drawing inspiration from mstriemer's work (mstriemer/addons-linter@5cbdaae#diff-baa432d9354e3de899d28a2638d1499eR31), I'm using the concept of ImageMetadata errors (Here's some progress on the same https://github.com/ravneette/addons-linter/commits/detect-corrupt-icons). Do you think this will work?

PR #1562, opened 2017-09-22
check that files in the background script actually exist

### LINKED ISSUE #1561: Check all background scripts and pages exist
Apparently people include things which don't exist. We'll harden WebExtensions against this, but let's let the developer know. Example: https://bugzilla.mozilla.org/show_bug.cgi?id=1402205#c2

andymckay:
fixes #1561

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -143,6 +143,35 @@ export function manifestIconMissing(path) {
   };
 }
 
+export const MANIFEST_BACKGROUND_SCRIPT_NOT_FOUND = 'MANIFEST_BACKGROUND_SCRIPT_NOT_FOUND';
+export const MANIFEST_BACKGROUND_PAGE_NOT_FOUND = 'MANIFEST_BACKGROUND_PAGE_NOT_FOUND';
+export function manifestBackgroundMissing(path, script_or_page) {
+  if (script_or_page === 'script') {
+    return {
+      code: MANIFEST_BACKGROUND_SCRIPT_NOT_FOUND,
+      legacyCode: null,
+      message: _(
+        'A background script defined in the manifest could not be found.'),
+      description: sprintf(
+        _('Background script could not be found at "%(path)s".'),
+        { path }),
+      file: MANIFEST_JSON,
+    };
+  }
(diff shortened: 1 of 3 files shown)


src/messages/manifestjson.js
```diff
@@ -143,6 +143,33 @@ export function manifestIconMissing(path) {
...
+    legacyCode: null,
+    message: _(
+      'A background page defined in the manifest could not be found.'),
+    description: sprintf(
+      _('Background page could not be found at "%(path)s".'),
+      { path }),
+    file: MANIFEST_JSON,
+  };
+}
+
```
EnTeQuAk: Both return values are mostly identical, I'm sure you can unify that to one constant and just replace the `code`, `message` and `description`. Maybe, if not I'm fine with using it like that too

src/parsers/manifestjson.js
```diff
@@ -129,6 +129,17 @@ export default class ManifestJSONParser extends JSONParser {
       this.validateIcons();
     }
 
+    if (this.parsedJSON.background) {
+      if (this.parsedJSON.background.scripts) {
+        for (let i = 0; i < this.parsedJSON.background.scripts.length; i++) {
+          this.validateExists(this.parsedJSON.background.scripts[i], 'script');
+        }
```
EnTeQuAk: We recently started using more `forEach` etc so this could be something like

```javascript
this.parsedJSON.background.scripts.forEach(() => {
    this.validateExists(...)
});
```

src/parsers/manifestjson.js
```diff
@@ -180,6 +191,15 @@ export default class ManifestJSONParser extends JSONParser {
     });
   }
 
+  validateExists(filePath, type) {
```
EnTeQuAk: I'd call this `validateFileExistsInPackage` or just `validateFileExists` since that's what it does.

src/parsers/manifestjson.js
andymckay: wfm

src/parsers/manifestjson.js
andymckay: okey doke

PR #1580, opened 2017-10-02
fixes #351, replaces none critical throws with collector.addError

### LINKED ISSUE #351: All errors unless totally fatal should add errors to the collector not throw.
This bug is to review that is the case and if not fix it.

Blocks #348

hameleonka:
I replaced some throws with collector.addError, just not sure if I got it right, If you(@EnTeQuAk) ok with this approach, I will finish this job. Working on this problem as an applicant for outreach program Dec 2017 -March 2018. Thanks.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -187,6 +187,12 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | :white_check_mark: | error | | | top level guid missing from install.rdf | install.rdf | | | RDF_ID_MISSING |
 | :white_check_mark: | error | | | top level version missing from install.rdf. | install.rdf | | | RDF_VERSION_MISSING |
 | :white_check_mark: | error | | | top level name missing from install.rdf | install.rdf | | | RDF_NAME_MISSING |
+| :white_check_mark: | error | | | RDF tag should have only one child  | install.rdf | | | RDF_MANY_CHILDREN |
+| :white_check_mark: | error | | | RDF tag not found | install.rdf | | | RDF_TAG_NOT_FOUND |
+| :white_check_mark: | error | | | RDF name should be presented only once | install.rdf | | | RDF_MULTIPLE_NAME_TAGS |
+| :white_check_mark: | error | | | RDF version should be presented only once | install.rdf | | | RDF_MULTIPLE_VERSION_TAGS |
+| :white_check_mark: | error | | | RDF guid should be presented only once | install.rdf | | | RDF_MULTIPLE_ID_TAGS |
+| :white_check_mark: | error | | | RDF type should be presented only once | install.rdf | | | RDF_MULTIPLE_TYPE_TAGS |
 
 
 ## Jetpack
--- a/src/io/xpi.js
+++ b/src/io/xpi.js
@@ -93,19 +93,21 @@ export class Xpi extends IOBase {
     });
   }
 
-  checkPath(path) {
+  _checkPath(path) {
     if (!Object.prototype.hasOwnProperty.call(this.files, path)) {
-      throw new Error(`Path "${path}" does not exist in this XPI`);
+      return new Error(`Path "${path}" does not exist in this XPI`);
     }
 
     if (this.files[path].uncompressedSize > this.maxSizeBytes) {
-      throw new Error(`File "${path}" is too large. Aborting.`);
+      return new Error(`File "${path}" is too large. Aborting.`);
     }
+    return null;
   }
 
   getFileAsStream(path) {
     return new Promise((resolve, reject) => {
-      this.checkPath(path);
+      const pathErr = this._checkPath(path);
+      if (pathErr) return reject(pathErr);
       return this.open()
         .then((zipfile) => {
           zipfile.openReadStream(this.files[path], (err, readStream) => {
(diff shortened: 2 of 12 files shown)


EnTeQuAk:
Hey @hameleonka, thank you so much for that pull request! It looks generally quite good and I don't see anything being majorly wrong so please feel free to go ahead and add a few tests if they're still needed. I'll have a detailed look and review later today, I'm so sorry for the late response.

hameleonka:
you are welcome, waiting for your review, I think this PR covers all the functionality described in the issue, there are some throws left, but in my opinion they are critical.

docs/rules.md
```diff
@@ -187,6 +187,12 @@ TODO: A lot of these are generated so this will need expanded with each unique c
 | :white_check_mark: | error | | | top level guid missing from install.rdf | install.rdf | | | RDF_ID_MISSING |
 | :white_check_mark: | error | | | top level version missing from install.rdf. | install.rdf | | | RDF_VERSION_MISSING |
 | :white_check_mark: | error | | | top level name missing from install.rdf | install.rdf | | | RDF_NAME_MISSING |
+| :white_check_mark: | error | | | RDF tag should have only one child  | install.rdf | | | RDF_MANY_CHILDREN |
```
EnTeQuAk: I would name this `RDF_TOO_MANY_CHILDREN` which is more explicit imho.

src/io/xpi.js
```diff
@@ -140,7 +142,8 @@ export class Xpi extends IOBase {
 
   getChunkAsBuffer(path, chunkLength) {
     return new Promise((resolve, reject) => {
-      this.checkPath(path);
+      const pathErr = this._checkPath(path);
+      if (pathErr) return reject(pathErr);
```
EnTeQuAk: I think from a semantic point of view I'd still like to have `checkPath` raise an exception but use `try/except` here to catch the exception and then `reject` it properly. I also don't think we should rename `checkPath` to `_checkPath` but that's nothing to worry about too much.

src/messages/rdf.js
```diff
@@ -71,6 +71,30 @@ export const RDF_TYPE_MISSING = {
   file: INSTALL_RDF,
 };
 
+export function rdfMultipleTags(tag) {
+  return {
+    code: `RDF_MULTIPLE_${tag.toUpperCase()}_TAGS`,
+    message: _(`Multiple <${tag}> elements found`),
+    description: _(oneLine`There should be only one tag <${tag}>`),
```
EnTeQuAk: You don't need the `oneLine` helper here. It's only used to strip newlines and whitespaces if a string spans across multiple lines.

src/messages/rdf.js
```diff
@@ -71,6 +71,30 @@ export const RDF_TYPE_MISSING = {
...
+    file: INSTALL_RDF,
+  };
+}
+
+export const RDF_TAG_NOT_FOUND = {
+  code: 'RDF_TAG_NOT_FOUND',
+  message: _(`RDF Node is not defined`),
+  description: _(oneLine`RDF tag should be defined`),
```
EnTeQuAk: Same here with `oneLine`

src/messages/rdf.js
```diff
@@ -71,6 +71,30 @@ export const RDF_TYPE_MISSING = {
...
+  file: INSTALL_RDF,
+};
+
+export const RDF_MANY_CHILDREN = {
+  code: 'RDF_MANY_CHILDREN',
+  message: _(`RDF node should only have a single descendant <Description>`),
+  description: _(oneLine`RDF tag should have only one child`),
```
EnTeQuAk: Here `oneLine` isn't really needed.

src/parsers/installrdf.js
```diff
@@ -15,12 +15,23 @@ export default class InstallRdfParser {
...
-      restartless: this._getIsBootstrapped(),
+    return new Promise((resolve, reject) => {
+      const guid = this._getGUID();
+      const name = this._getName();
+      const type = this._getAddonType();
+      const version = this._getVersion();
+      const restartless = this._getIsBootstrapped();
+      if (guid === null || name === null || version === null) {
+        return reject(new Error(oneLine`Cannot get metadata,
+                                  version, name, or id tag is missing`));
```
EnTeQuAk: You can put the whole string on the next line to make it easier to read, something like:

```javascript
return reject(new Error(oneLine`
    Cannot get metadata, version, name or id tag is missing`));
```

Also, I'd write the error message a bit differently: `Cannot get metadata, version, name or id tag.` The `is missing` part is a bit confusing imho.

EnTeQuAk:
Soooooo, we just had a discussion in IRC and I falsely marked #351 as a good first bug while we are actually pretty ready to remove all RDF related stuff from this code-base.

Languagepacks are moving into WebExtension direction right now, we never had dictionary validation in the linter and there isn't much else that needs RDF to work.

So, unfortunately I'm going to close this PR BUT as a compensation if you want, you can give https://github.com/mozilla/addons-linter/pull/1308 a look and maybe re-base or redo the whole removal part which should be pretty fun I hope.

Sorry for keeping you waiting so long and thanks a lot for all the work that you invested into this!

PR #1581, opened 2017-10-02
Added check to add warning in case of corrupt image file

ravneette:
In continuation of https://github.com/mozilla/addons-linter/pull/1558

* [x] Add a description of the changes introduced in this PR.

Adds a warning in case the promise from sharp library rejects on reading an image file for an icon.


ravneette:
Hey @EnTeQuAk 

I'm facing the following issue in the Sharp library suggested in https://github.com/mstriemer/addons-linter/commit/5cbdaaeed8edabb43efff20ca256a6199b003cd1#diff-baa432d9354e3de899d28a2638d1499eR31 

If I run the following code independently, neither the info nor the error event is triggered:
```js
var transformer = sharp()
  .resize(300)
  .on('info', function(info) {
    console.log('Image height is ' + info.height);
  })
  .on('error', function(err) {
    console.log('error' + err);  	
  })
var readableStream = fs.createReadStream('default.png');
readableStream.pipe(transformer)
```

However, the following code does throw an error incase of a corrupted file:
```js
sharp('./a.png')
  .toFile('output.jpg', function(err) {
  	console.log(err);
  });
```
The problem with this approach is it ends up creating an unnecessary output file. 

Which one of the above do you think is on the right lines?

ravneette:
Found a metadata function which works fine.
Using the following code to test : https://github.com/mozilla/addons-linter/pull/1581/files#diff-baa432d9354e3de899d28a2638d1499eR31

EnTeQuAk:
Sorry, I must have missed the notifications about your comments. Anything that is still unclear? Using the metadata function you linked sounds like a great idea and way forward.

Otherwise I'll give this pull request a quick look very soon.

ravneette:
Hey @EnTeQuAk 
This P.R has one issue to be resolved.

On using a corrupt image, the promise returned by the metadata function rejects, i.e., https://github.com/mozilla/addons-linter/pull/1581/files#diff-baa432d9354e3de899d28a2638d1499eR200 gets executed.

But it still does not appear as a warning on running tests. My guess is that this is happening because the 
```js
this.collector.addWarning(messages.CORRUPT_ICON_FILE);
``` 
is executed after some time (since the metadata function hasn't rejected the promise yet) and so the warning is not added.

I think we need to prevent the scanner from returning the warnings till the time this promise is resolved/rejected.

Is there a way to do that?

ravneette:
Fixed the merge conflicts.
Also, I added checks to validate the size of images, following up from https://github.com/mozilla/addons-linter/pull/1555 since it was closed.

tests/parsers/test.manifestjson.js
```diff
@@ -1,3 +1,4 @@
+
```
EnTeQuAk: That one line caught my eye, mind removing it again? 😆

src/messages/manifestjson.js
```diff
@@ -174,3 +199,11 @@ export const WRONG_ICON_EXTENSION = {
...
   file: MANIFEST_JSON,
 };
+
+
+export const CORRUPT_ICON_FILE = {
+  code: 'CORRUPT_ICON_FILE',
+  message: _('Corrupt image file'),
+  description: _('Icon file is not a valid Image file'),
+  file: MANIFEST_JSON,
+};
```
EnTeQuAk: I think it makes sense similar to how `iconIsNotSquare` works that we print the icon filename too.

ravneette:
Hi @EnTeQuAk 

I wrote the tests, but they don't seem to pass because of the issue I mentioned in https://github.com/mozilla/addons-linter/pull/1581#issuecomment-335411390

The getMetadataFunction returns a promise, and by the time the promise gets resolved/rejected, the test checks if this.collector.warnings has length == 1.

Thus we need to make sure that the tests checks only after the promise has resolved/rejected. Is there a way of doing that?

ravneette:
From what I can get, we need to add a promise somewhere here : 
https://github.com/mozilla/addons-linter/blob/master/src/linter.js#L264
to make sure we return only after the icon validation has occurred.

ravneette:
Made the `validateIcons` function return a promise @EnTeQuAk 
Not sure why the tests are failing, seem to be running fine locally. Could you please take a look?

EnTeQuAk:
>  Not sure why the tests are failing, seem to be running fine locally. Could you please take a look?

Absolutely, I'll take a look and see if they run fine for me locally too.

EnTeQuAk:
Okay, so the test in travis fails for me locally too. You don't see any issues locally?

EnTeQuAk:
Ahhh, @ravneette so this change is coming for a recent pull request that I merged into master - the value of `4` is perfectly fine there and expected. Would be great if you can update the PR, I don't see much else that stands in the way of merging it.

ravneette:
Hey @EnTeQuAk 
Thanks for pointing that out. Fixed the tests!

ravneette:
It was a lovely experience, learnt a lot while working on this issue. 

Thank you @EnTeQuAk for your guidance all along :D

Issue #1560, opened 2017-09-22
Cleanup rules.md to not include old amo-validator rules anymore.


EnTeQuAk:

ravneette:
Seems like a continuation of https://github.com/mozilla/addons-linter/issues/1559
Can I take this up if you haven't worked on it already @EnTeQuAk ?

EnTeQuAk:
hmm, it's less of a continuation of #1559 unfortunately since it's basically syncing with what `https://github.com/mozilla/amo-validator` implements, what we have in our `rules.md` and which of these rules still apply to WebExtensions (most don't) and give that document a new live / formatting.

That might also affect the `should have updated rules.md with new message codes` test in `tests/test.messages.js` since that's validating the rules.

I honestly think it's wasted time and energy for you to work on that and I really appreciate you already started working on https://github.com/mozilla/addons-linter/pull/1581 which adds way more value to the linter.

PR #1607, opened 2017-10-16
removes all rdf related code and tests

hameleonka:
repeats #1308 with additional cleanup

@EnTeQuAk not sure if I should also remove all mentions of rdf from rules.md, couple of them are still there line [175](https://github.com/mozilla/addons-linter/compare/master...hameleonka:code-cleanup/remove-rdf-scanner#diff-4c27e978f82559bae938fd10bd608486R175)


EnTeQuAk:
> @EnTeQuAk not sure if I should also remove all mentions of rdf from rules.md, couple of them are still there line 175

Everything that directly links to `install.rdf` - which includes the ones you mentioned - can be removed from the rules.md

There's https://github.com/mozilla/addons-linter/issues/1560 for a broader cleanup but getting rid of them here is highly appreciated.

docs/types.md
```diff
@@ -32,19 +32,17 @@ Docs: https://developer.mozilla.org/Add-ons/WebExtensions
...
 
 ## Lang Packs
 
-* Validate `install.rdf`
 * Validate `chrome.manifest`
 * Deal with any other langpack specific rules.
 
 Docs appear to be a bit thin on the ground for these. They need looking
-into further and we shouldm look at the existing rules.
+into further and we should look at the existing rules.
```
EnTeQuAk: Nice catch!

docs/types.md
```diff
@@ -32,19 +32,17 @@ Docs: https://developer.mozilla.org/Add-ons/WebExtensions
...
-* Deal with any other dictionary specific rules.
+* Deal with any dictionary specific rules.
 
 Docs: https://developer.mozilla.org/docs/Creating_a_spell_check_dictionary_add-on
 
 ## Lang Packs
 
-* Validate `install.rdf`
 * Validate `chrome.manifest`
 * Deal with any other langpack specific rules.
```
EnTeQuAk: Hmm, this page urgently needs a cleanup… we don't really lint dictionaries and don't have any support for `chrome.manifest` anymore (which was legacy as well). Doesn't have to be happening here, or rather, it shouldn't and we can open another cleanup issue for docs specifically.

tests/io/test.crx.js
```diff
@@ -16,16 +16,16 @@ const chromeManifestEntry = Object.assign({}, defaultData, {
...
+  fileName: 'file.entry',
 });
 
-const dupeInstallRdfEntry = Object.assign({}, defaultData, {
+const dupeInstallFileEntry = Object.assign({}, defaultData, {
   compressedSize: 416,
   uncompressedSize: 851,
-  fileName: 'install.rdf',
+  fileName: 'file.entry',
 });
```
EnTeQuAk: I think it's more real-life if we call these `manifest.json` instead since that's the correct replacement for `install.rdf` in the WebExtension world.

tests/io/test.directory.js
```diff
@@ -12,7 +12,7 @@ describe('Directory.getFiles()', () => {
       size: 1,
     };
     myDirectory.files = {
-      'install.rdf': fakeFileMeta,
+      'file.entry': fakeFileMeta,
       'chrome.manifest': fakeFileMeta,
```
EnTeQuAk: Same here, let's use `manifest.json` which is a bit more correct.

tests/io/test.xpi.js
```diff
@@ -356,7 +356,7 @@ describe('Xpi.getChunkAsBuffer()', function getChunkAsBufferCallback() {
   it('should resolve with a buffer', () => {
     const myXpi = new Xpi('foo/bar', this.fakeZipLib);
     myXpi.files = {
-      'install.rdf': installRdfEntry,
+      'file_to.read': installFileEntry,
     };
```
EnTeQuAk: Same here, let's use `manifest.json` to be more real-world like. I am aware that it doesn't matter here but having our tests to use real-life names makes it easier to read in the future

EnTeQuAk:
Perfect, thanks a lot for all the changes and for getting rid of that huge chunk of old code!

caitmuenster:
Thank you so much for the patch, @hameleonka! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

If you would like to set up a profile on [mozillians.org](https://mozillians.org), I would be happy to vouch for you. 

Welcome onboard! I look forward to seeing you around.

PR #1613, opened 2017-10-18
Add support for WebExtension Langpack

### LINKED ISSUE #1526: Support WebExtensions Language Packs in the linter
Part of https://github.com/mozilla/addons-server/issues/6297

[Bug 1365709](https://bugzilla.mozilla.org/show_bug.cgi?id=1365709) introduces language packs that use WebExtensions packaging. We need to be able to validate them like the old validator does.

- [ ] The main thing to check for is that language packs should only be allowed to use metadata manifest properties and language pack manifest properties. They must not include any others. The schema [can be found here](https://hg.mozilla.org/mozilla-central/rev/960c3df4ced4#l2.12).
- [ ] I don't know if the legacy validator checks for anything else. If so, we need to discuss if it's worth porting.

EnTeQuAk:
Fixes #1526

* Adds `.properties` and `.doctype` parsers for potential completeness tests or at least very basic validation if those files look fine
  * Once all language packs moved to FTL we can drop them
* Adds a new `validateLangPack` validator, rewrites a few schemas to now explicitly grab the correct validator type (webext/langpack)
* Adds a new command line option that explicitly switches the mode to langpack and verifies against `#/types/WebExtensionLangpackManifest` instead of `WebExtensionManifest`

it's not 100% complete, it adds basic validation against the schema but doesn't do anything with the actual l10n files yet. That'll come in another iteration

**Edit**

* Added support for FTL files, they are the first ones to throw proper warnings, yeay!

screenshot from 2017-10-20 12-54-24


src/parsers/doctype.js
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor(dtdString, collector, { filename = null } = {}) {
+    this._dtdString = dtdString;
+    this.collector = collector;
+    this.filename = filename;
+    this.isValid = null;
+  }
+
+  parse() {
+    const entities = this._dtdString.match(new RegExp(ENTITY_RE, 'g'));
```
wagnerand: Could it be `<!entity` as well? If so, we might want to add the `i` flag.

tests/parsers/test.properties.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it('should ignore invalid entities', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const propertiesParser = new PropertiesParser(`
+        this should be ignored.
+        foo=
+        bar
+        abc.def=x
+        yz`, addonLinter.collector);
```
wagnerand: Could we add another string that comprises of three lines? Just to make sure we don't break the implementation in the future.

src/parsers/doctype.js
EnTeQuAk: hmm, good point. I have not found anything that is lowercase and I'm not even sure about the spec here but I think it's better to be safe than sorry :)

I'll add that.

tests/parsers/test.properties.js
EnTeQuAk: Will do :)

EnTeQuAk:
Fixed the review comments and added support for FTL files.

@mstriemer, @wagnerand - this PR is now really ready for a review :)

src/scanners/langpack.js
```diff
@@ -0,0 +1,51 @@
...
+
+        if (ext === '.properties') {
+          ParserClass = PropertiesParser;
+        } else if (ext === '.ftl') {
+          ParserClass = FluentParser;
+        } else if (ext === '.dtd') {
+          ParserClass = DoctypeParser;
+        } else {
+          throw new Error('Unsupported file type');
+        }
```
wagnerand: Are we sure langpacks cannot contain any file types other than the ones listed?

tests/parsers/test.fluent.js
```diff
@@ -0,0 +1,173 @@
...
+        attrs: {
+          accesskey: 'R',
+          label: 'Reconnect to &syncBrand.shortName.label;…',
+        },
+        val: undefined,
+      },
+    });
+  });
+
+  it('supports placeholders', () => {
```
wagnerand: Nit: Use the official term `placeable`?

src/scanners/langpack.js
EnTeQuAk: The linter only forwards files that are either `.properties`, `.dtd` or `.ftl` and nothing else right now. Practically a langpack can also contain icons and xml because of overwriting search but we don't validate that for right now.

src/parsers/doctype.js
```diff
@@ -0,0 +1,45 @@
...
+    entities.forEach((entity) => {
+      const [, key, value] = entity.match(new RegExp(ENTITY_RE, 'i'));
+
+      // strip enclosing quotation marks
+      const normalizedValue = value.slice(1, -1);
+
+      this.parsedData[key] = normalizedValue;
+    });
+
+    this.isValid = true;
```
mstriemer: It looks like this always sets `this.isValid = true`. Would this throw on an invalid dtd file?

tests/parsers/test.properties.js
```diff
@@ -0,0 +1,53 @@
...
+        c
+        d`, addonLinter.collector);
+
+    propertiesParser.parse();
+
+    expect(propertiesParser.isValid).toEqual(true);
+    expect(propertiesParser.parsedData).toEqual({
+      foo: 'bar',
+      'abc.def': 'xyz',
+      'three.lines': 'abcd',
```
mstriemer: Do we care what the values are? I'm guessing not but I'd imagine you'd want a space between the lines.

src/schema/firefox-schemas-import.js
```diff
@@ -198,18 +198,16 @@ inner.mapExtendToRef = (schemas) => {
...
-      if (updatedType) {
-        updatedSchemas[namespace] = {
-          ...updatedSchemas[namespace],
-          schema: {
-            ...extendSchema,
-            types: {
-              ...extendSchema.types,
-              [type]: updatedType,
-            },
+      updatedSchemas[namespace] = {
```
mstriemer: Looks like `updatedType` is always set so I'm not sure why this if was here in the first place... Thanks for removing it.

src/schema/validator.js
```diff
@@ -37,4 +37,14 @@ validator.addFormat('secureUrl', isSecureUrl);
...
 
-export default validator.compile(schemaObject);
+export const validateAddon = validator.compile({
+  ...schemaObject,
+  id: 'manifest',
+  $ref: '#/types/WebExtensionManifest',
+});
+
+export const validateLangPack = validator.compile({
+  ...schemaObject,
```
mstriemer: Rather than use reuse the webextension manifest I think you could avoid updating the import code by creating a new schema for langpacks and just using that one.

```js
import langpackSchema from 'schema/langpack-manifest.json';
const langpackValidator = ajv({
  allErrors: true,
  errorDataPath: 'property',
  jsonPointers: true,
  verbose: true,
});
export const validateLangPack = validator.compile(langpackSchema);

tests/parsers/test.properties.js
EnTeQuAk: I don't think we do, at least not right now and I'm sure if we would we mostly care about the keys (e.g for completeness checks).

src/schema/validator.js
EnTeQuAk: Good idea, but not necessary right now afaik. I might rethink that though when adding support for static themes or webext dictionaries.

src/schema/firefox-schemas-import.js
EnTeQuAk: heh, yeah. That was a good hack to increase coverage :)

src/parsers/doctype.js
EnTeQuAk: If the file is completely invalid, yes. We parse dtd files very resilient to errors though so that we don't care right now. We mostly just care about whether or not we can open the file. More checking might come later.

PR #1623, opened 2017-10-23
fixes #1436

### LINKED ISSUE #1436: Add a switch to easily disable/enable existing rules
In order to mitigate the risk when having more and more complicated rules, it'd be good to have a system to dynamically enable and disable existing rules without having to wait for a push.

@EnTeQuAk said it should be fairly straightforward 😊

I assume this needs a counterpart on *addons-server* for a UI?

hameleonka:
introduces new cli param `disable-linter-rules` which accepts coma separated list of rules to be disabled

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -12,9 +12,9 @@ import * as messages from 'messages';
 import { rules } from 'rules/javascript';
 import { ensureFilenameExists } from 'utils';
 
-export function excludeRules(excludeFrom = {}, excludeWhat = {}) {
+export function excludeRules(excludeFrom = {}, excludeWhat = []) {
   return Object.keys(excludeFrom).reduce((result, ruleName) => {
-    if (excludeWhat[ruleName]) return result;
+    if (excludeWhat.includes(ruleName)) return result;
     return {
       ...result,
       [ruleName]: excludeFrom[ruleName],
@@ -34,13 +34,8 @@ export default class JavaScriptScanner {
     this.scannedFiles = [];
     this._rulesProcessed = 0;
     this.disabledRules = typeof options.disabledRules === 'string' ? options.disabledRules.split(',')
-      .reduce((result, i) => {
-        if (!i.trim()) return result;
-        return {
-          ...result,
-          [i.trim()]: true,
-        };
-      }, {}) : {};
+      .map((rule) => rule.trim())
+      .filter((notEmptyRule) => notEmptyRule) : [];
     ensureFilenameExists(this.filename);
   }
 
@@ -57,8 +52,6 @@ export default class JavaScriptScanner {
     _ruleMapping = ESLINT_RULE_MAPPING,
     _messages = messages,
   } = {}) {
-    const ruleMappingWithExclusion = excludeRules(_ruleMapping, this.disabledRules);
-    const rulesAfterExclusion = excludeRules(_rules, this.disabledRules);
     return new Promise((resolve) => {
       const cli = new _ESLint.CLIEngine({
         baseConfig: {
@@ -74,7 +67,7 @@ export default class JavaScriptScanner {
         parserOptions: {
           ecmaVersion: 2017,
         },
-        rules: ruleMappingWithExclusion,
+        rules: _ruleMapping,
         plugins: ['no-unsafe-innerhtml'],
         allowInlineConfig: false,
 
@@ -89,6 +82,7 @@ export default class JavaScriptScanner {
         useEslintrc: false,
       });
 
+      const rulesAfterExclusion = excludeRules(_rules, this.disabledRules);
       Object.keys(rulesAfterExclusion).forEach((name) => {
         this._rulesProcessed++;
         cli.linter.defineRule(name, rulesAfterExclusion[name]);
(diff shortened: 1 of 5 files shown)


src/scanners/javascript.js
```diff
@@ -23,7 +33,14 @@ export default class JavaScriptScanner {
...
     this._rulesProcessed = 0;
-
+    this.disabledRules = typeof options.disabledRules === 'string' ? options.disabledRules.split(',')
+      .reduce((result, i) => {
+        if (!i.trim()) return result;
+        return {
+          ...result,
+          [i.trim()]: true,
+        };
+      }, {}) : {};
```
EnTeQuAk: I wonder if a simple `map` (given that `disabledRules` can be an array instead of an object, see comment down below) would work easier here.

```javascript
this.disabledRules = typeof options.disabledRules === 'string' ? options.disabledRules.split(',')
    .map(rule => { rule.trim() })
    .filter(() => { return true; });
```

`map` cleans the strings and `filter` filters out empty elements by default so the result should be similar.

tests/scanners/test.javascript.js
```diff
@@ -47,6 +67,41 @@ describe('JavaScript Scanner', () => {
     expect(jsScannerWithOptions.options.foo).toEqual('bar');
   });
 
+  it('should be initialised with empty excluded rules object', () => {
+    const jsScanner = new JavaScriptScanner('', 'filename.txt');
+    expect(typeof jsScanner.disabledRules).toEqual('object');
+    // This test assures us the disabledRules can be accessed like an object.
+    expect(typeof jsScanner.disabledRules.someUndefinedProp).toEqual('undefined');
+  });
```
EnTeQuAk: I think that test should be more something like

```javascript
it('should have no rules disabled by default', () => {
    const jsScanner = new JavaScriptScanner('', 'filename.txt');
    expect(jsScanner.disabledRules).ToEqual({}); // or [] given my comments above
});
```

That way you can merge those two checks into one.

tests/scanners/test.javascript.js
```diff
@@ -47,6 +67,41 @@ describe('JavaScript Scanner', () => {
...
+
+  it('should be initialised with disabledRules from options', () => {
+    const jsScanner = new JavaScriptScanner('', 'filename.txt', {
+      disabledRules: 'no-eval, no-implied-eval,                 no-unsafe-innerhtml/no-unsafe-innerhtml',
+    });
+    expect(typeof jsScanner.disabledRules).toEqual('object');
+    // This test assures us the disabledRules built properly.
+    expect(jsScanner.disabledRules['no-unsafe-innerhtml/no-unsafe-innerhtml']).toBe(true);
+  });
```
EnTeQuAk: Same here, I would compare the actual expected `disabledRule` object.

tests/scanners/test.javascript.js
```diff
@@ -47,6 +67,41 @@ describe('JavaScript Scanner', () => {
...
+
+  it('should be initialised with empty excluded rules object, when there is no string', () => {
+    const jsScanner = new JavaScriptScanner('', 'filename.txt', {
+      disabledRules: true,
+    });
+    expect(typeof jsScanner.disabledRules).toEqual('object');
+    expect(jsScanner.disabledRules).toEqual({});
+  });
```
EnTeQuAk: Ah, here you did the check correctly. I don't think you need the `typeof` check though.

tests/scanners/test.javascript.js
```diff
@@ -47,6 +67,41 @@ describe('JavaScript Scanner', () => {
...
+    const jsScanner = new JavaScriptScanner('', 'filename.txt', {
+      disabledRules: 'no-eval, no-implied-eval,                 no-unsafe-innerhtml/no-unsafe-innerhtml,,,,,',
+    });
+    expect(jsScanner.disabledRules).toEqual({
+      'no-eval': true,
+      'no-implied-eval': true,
+      'no-unsafe-innerhtml/no-unsafe-innerhtml': true,
+    });
+  });
+
```
EnTeQuAk: That's precisely what I meant 👍

src/scanners/javascript.js
hameleonka: it's not the same in my case we have map
```
{ 
  rule0: true,
  rule1: true,
}
```
in your case we have array of rules. it just faster to look in the object if it has property, than lookup in the array if it has item, if you want I can change it to array

src/scanners/javascript.js
```diff
@@ -40,6 +57,8 @@ export default class JavaScriptScanner {
     _ruleMapping = ESLINT_RULE_MAPPING,
     _messages = messages,
   } = {}) {
+    const ruleMappingWithExclusion = excludeRules(_ruleMapping, this.disabledRules);
+    const rulesAfterExclusion = excludeRules(_rules, this.disabledRules);
```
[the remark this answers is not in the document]
hameleonka: this._disabledRules - is object not an array

src/scanners/javascript.js
```diff
@@ -55,7 +74,7 @@ export default class JavaScriptScanner {
         parserOptions: {
           ecmaVersion: 2017,
         },
-        rules: _ruleMapping,
+        rules: ruleMappingWithExclusion,
```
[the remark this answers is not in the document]
hameleonka: that's correct, my mistake, sorry

PR #1627, opened 2017-10-27
Fixes #1616: Fix zip decoding when hitting an encoding boundary between two chunks;

### LINKED ISSUE #1616: A syntax error prevents upload (false positive)
I tried to upload an extension on AMO. That was an update.

### What happened?
It says: “There is a JavaScript syntax error in your code; validation cannot continue on this file.”
But this code works.
There no syntax error here.

bnjbvr:
      PRs open for the same issue.
* [x] Add `Fixes #ISSUENUM` at the top of your PR.
* [ ] Add a description of the the changes introduced in this PR.
* [ ] Add tests to cover the changes added in this PR.

The issue is that we hit a boundary between the two bytes encoding an utf8 char: say we have a char that has encoding \a\b, a first zip chunk is read and ends with \a, the second one starts with \b, but the current implementation thinks these are two different chars.

Instead, we should store the data in a raw byte buffer and do the conversion to utf8 at the end.

The test is basically the whole grammalecte zip file, which needs this precise file layout (to match the internal zip chunk size), so not sure how useful it is to include it here.

@diox, can you have a look, please? Thanks!

### DIFF
--- a/src/io/xpi.js
+++ b/src/io/xpi.js
@@ -123,13 +123,14 @@ export class Xpi extends IOBase {
     return this.getFileAsStream(path)
       .then((fileStream) => {
         return new Promise((resolve, reject) => {
-          let fileString = '';
+          let buf = new Buffer('');
           fileStream.on('data', (chunk) => {
-            fileString += chunk;
+            buf = Buffer.concat([buf, chunk]);
           });
 
           // Once the file is assembled, resolve the promise.
           fileStream.on('end', () => {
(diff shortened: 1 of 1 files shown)


bnjbvr:
(just fixed the eslint issue)

EnTeQuAk:
Thanks a lot for the pull request! It looks great from my perspective.

It'd be awesome to have some tests for it but we do have general coverage for that part of the code so I'm relatively fine without.

PR #1652, opened 2017-11-14
Block specific content script match patterns.

### LINKED ISSUE #1650: Show error for certain manifest patterns
We need to filter certain add-ons, based on information in the manifest, for example matching a content script hostmask.

Fictional example:
```
"content_scripts": [
  {
    "matches": ["*://*.baddomain.com/*"],
    "js": ["borderify.js"]
  }
]
```

The linter should have a list of `matches` with their corresponding actions, e.g. `error`. It would be good if that list would be a separate file.

EnTeQuAk:
Fixes #1650


src/messages/manifestjson.js
```diff
@@ -166,6 +166,15 @@ export function manifestContentScriptFileMissing(path, type) {
   };
 }
 
+// https://github.com/mozilla/addons-linter/issues/1650
+// Potentially temporary
+export const MANIFEST_CSCRIPT_INVALID = {
+  code: 'MANIFEST_CSCRIPT_INVALID',
+  message: _('Invalid "matches" entry in "content_script".'),
```
wagnerand: Let's go with:

message: Forbidden content found in add-on.
description: This add-on contains forbidden content.

PR #1661, opened 2017-11-21
fixes #1640 adds missing content scripts file rule for static strings

### LINKED ISSUE #1640: linter doesn't throw an error when content_scripts referenced from code are missing
there are 2 ways to introduce content scripts to addon
1) declaration in manifest.json covered in #1630 
2) programmatically with tabs.executeScript()

### What happened?
linter doesn't trow an error in the second case if script is missing

I would expect error notification

wondering how to deal with this if there is dynamic file name i.e. instead of path there could be something like `/content_scripts/${conditionalFileName}`

hameleonka:
please take a careful look into this, I'm not sure if it is the best way doing this validation, I basically lazy looking for any static string which starts from `content_scripts` and ends on `.js`. Why I do it this way: because people can specify this file path in the variables like:
```
var filePath = '/content_scripts/file.js';
browser.tabs.executeScript({file: filePath})
```
but with this approach if user wants to write some weird sentence containing `/content_scripts/file.js` without having `file.js` in content_scripts folder, rule will throw an error.
In the same time I do not consider template strings or concatenations:
```
var filePath ='file.js'
browser.tabs.executeScript({file:  `/content_scripts/${filePath}`;})
```
```
browser.tabs.executeScript({file:  '/content_scripts/'+'file.js';})
```
both examples above will pass the validation even without having `file.js` in content_scripts

### DIFF
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -12,9 +12,9 @@ export default {
         }
         const namespace = node.object.property.name;
         const property = node.property.name;
-        // Namespace should be tabs function should be executeScript.
-        // I.E. browser.tabs.executeScript.
-        if (namespace !== 'tabs' || property !== 'executeScript') {
+        // Namespace should be tabs function should be executeScript and it should be a call.
+        // I.E. browser.tabs.executeScript().
+        if (namespace !== 'tabs' || property !== 'executeScript' || node.parent.type !== 'CallExpression') {
           return;
         }
         node.parent.arguments.forEach((arg) => {
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -5,21 +5,26 @@ export default {
   create(context) {
     const existingFiles = context.settings.existingFiles || {};
     return {
-      Program(node) {
-        const stringTokensIncludesContentScript = node.tokens.filter((token) => {
-          return token.type === 'String' && token.value.match(/content_scripts.*?\.js/);
-        });
+      CallExpression(node) {
+        if (node.callee.type === 'MemberExpression' &&
+          node.callee.property &&
+          node.callee.property.name === 'executeScript') {
+          node.arguments.forEach((arg) => {
+            if (arg.type !== 'ObjectExpression') return;
 
-        stringTokensIncludesContentScript.forEach((token) => {
-          const trimmedFileName = normalizePath(token.value.replace(/["']/g, ''));
-          if (!Object.prototype.hasOwnProperty.call(existingFiles, trimmedFileName)) {
-            return context.report({
-              loc: token.loc,
-              message: CONTENT_SCRIPT_NOT_FOUND.code,
-            });
-          }
-          return null;
-        });
+            const fileProperty = arg.properties.find((prop) => prop.key && prop.key.name === 'file');
+            const fileValue = fileProperty && fileProperty.value.type === 'Literal' ? fileProperty.value.value : '';
+            if (fileValue.match(/content_scripts.*\.js/)) {
+              const normalizedName = normalizePath(fileValue);
+              if (!Object.prototype.hasOwnProperty.call(existingFiles, normalizedName)) {
+                context.report({
+                  loc: fileProperty.value.loc,
+                  message: CONTENT_SCRIPT_NOT_FOUND.code,
+                });
+              }
+            }
+          });
+        }
       },
     };
   },
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -1,0 +1,26 @@
+import { normalizePath } from "utils";
+import { JS_CONTENT_SCRIPT_FILE_NOT_FOUND } from 'messages/javascript';
+
+export default {
+  create(context) {
+    const existingFiles = context.settings.existingFiles || {};
+    return {
+      Program(node) {
+        const stringTokensIncludesContentScript = node.tokens.filter((token) => {
+          return token.type === 'String' && token.value.match(/content_scripts.*?\.js/);
+        });
+
+        stringTokensIncludesContentScript.forEach((token) => {
+          const trimmedFileName = normalizePath(token.value.replace(/["']/g, ''));
+          if (!Object.prototype.hasOwnProperty.call(existingFiles, trimmedFileName)) {
+            return context.report({
+              loc: token.loc,
+              message: JS_CONTENT_SCRIPT_FILE_NOT_FOUND.code,
+            });
+          }
+          return null;
+        });
+      },
+    };
+  },
+};
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -11,32 +11,36 @@ export default {
           const property = node.property.name;
           // namespace should be tabs function should be executeScript
           // i.e. browser.tabs.executeScript
-          if (namespace !== 'tabs' || property !== 'executeScript') return;
-          node.parent.arguments.forEach((arg) => {
-            // skip argument is not an object
-            if (arg.type !== 'ObjectExpression') return;
-            const fileProperty = arg.properties.find((prop) => prop.key && prop.key.name === 'file');
-            const fileValue = fileProperty.value && fileProperty.value.value;
-            // skip argument if file property is not a static string
-            if (fileProperty.value.type !== 'Literal') return;
-            // if filename is empty, report an issue
-            if (fileValue === '') {
-              context.report({
-                loc: fileProperty.value.loc,
-                message: CONTENT_SCRIPT_EMPTY.code,
-              });
-              return;
+          if (namespace === 'tabs' && property === 'executeScript') {
+            /* eslint-disable no-restricted-syntax */
+            for (const arg of node.parent.arguments) {
+              // work with the argument if it is object
+              if (arg.type === 'ObjectExpression') {
+                const fileProperty = arg.properties.find((prop) => prop.key && prop.key.name === 'file');
+                const fileValue = fileProperty && fileProperty.value && fileProperty.value.value;
+                // continue validation if file property is not a static string
+                if (fileProperty && fileProperty.value.type === 'Literal' && typeof fileValue === 'string') {
+                  // if filename is empty, report an issue
+                  if (fileValue === '') {
+                    context.report({
+                      loc: fileProperty.value.loc,
+                      message: CONTENT_SCRIPT_EMPTY.code,
+                    });
+                    return;
+                  }
+                  const normalizedName = normalizePath(fileValue);
+                  // if file exists then we are good
+                  if (!(normalizedName in existingFiles)) {
+                    // file not exists report an issue
+                    context.report({
+                      loc: fileProperty.value.loc,
+                      message: CONTENT_SCRIPT_NOT_FOUND.code,
+                    });
+                  }
+                }
+              }
             }
-
-            const normalizedName = normalizePath(fileValue);
-            // if file exists then we are good
-            if (normalizedName in existingFiles) return;
-            // file not exists report an issue
-            context.report({
-              loc: fileProperty.value.loc,
-              message: CONTENT_SCRIPT_NOT_FOUND.code,
-            });
-          });
+          }
         }
       },
     };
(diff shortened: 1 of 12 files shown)


hameleonka:
Yep, thanks, will make necessary corrections

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -0,0 +1,31 @@
...
+      CallExpression(node) {
+        if (node.callee.type === 'MemberExpression' &&
+          node.callee.property &&
+          node.callee.property.name === 'executeScript') {
+          node.arguments.forEach((arg) => {
+            if (arg.type !== 'ObjectExpression') return;
+
+            const fileProperty = arg.properties.find((prop) => prop.key && prop.key.name === 'file');
+            const fileValue = fileProperty && fileProperty.value.type === 'Literal' ? fileProperty.value.value : '';
+            if (fileValue.match(/content_scripts.*\.js/)) {
```
rpl: I don't think that we should condition this validation on this regexp match, we can't assume that the filenames of the extension content scripts are going to contains "content_scripts" in their name.

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -0,0 +1,31 @@
...
+              const normalizedName = normalizePath(fileValue);
+              if (!Object.prototype.hasOwnProperty.call(existingFiles, normalizedName)) {
```
rpl: An alternative way to do this check could be:

```js
if (normalizedName in existingFiles) {
  // the content script normalized filename exists.
  return;
}

// Report the error for the non existent filename.
context.report({...});
```

src/rules/javascript/content-scripts-file-absent.js
rpl: on the contrary, we should probably check if the file literal string is an empty string and produce a different validation error message (something like "unexpected empty filename" or similar error message).

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -0,0 +1,31 @@
...
+  create(context) {
+    const existingFiles = context.settings.existingFiles || {};
+    return {
+      CallExpression(node) {
+        if (node.callee.type === 'MemberExpression' &&
+          node.callee.property &&
+          node.callee.property.name === 'executeScript') {
+          node.arguments.forEach((arg) => {
+            if (arg.type !== 'ObjectExpression') return;
+
```
rpl: To make false positive less likely to happen because of this validation rule, we should also check if this executeScript call expression is happening on a actual WebExtension API object, e.g. like we do for the existent validation rule for the deprecated APIs:
https://github.com/mozilla/addons-linter/blob/6109428a065ff175e97ef483d1ee9a29ef3d94ad/src/rules/javascript/webextension-api.js#L9

Looking at the AST received by this CallExpression hook, it should be something like:
```js
isBrowserNamespace(node.callee.object.object.name);
```

We should also add some test cases that ensures that this rule doesn't generate the false positive errors, e.g. on JS fragments like:

```js
executeScript(...);

myObject.executeScript(...);

myObject.tabs.executeScript(...);
```

hameleonka:
hey @rpl, could you please check it once again? Hope it's better now

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -0,0 +1,191 @@
...
+        'content_scripts/existingFile.js': '',
+      },
+      addonMetadata: {
+        id: 'test',
+      },
+    });
+
+    return jsScanner.scan()
+      .then(({ linterMessages }) => {
+        expect(linterMessages.length).toEqual(0);
```
rpl: On the assertion that checks that the validation messages are empty, it could be better to assert that the 
`linterMessages` value is equal to an empty array, so that when the assertion fails it prints the unexpected validation errors in the failure message, e.g. `expect(linterMessages).toEqual([]);` would raise a failure message like this:

```
 expect(received).toEqual(expected)
    
    Expected value to equal:
      []
    Received:
      [{"code": "CONTENT_SCRIPT_EMPTY", "column": 35, "description": "Content script file name should not be empty", "file": "file-with-code.js", "line": 1, "message": "Content script file name should not be empty.", "sourceCode": "browser.tabs.executeScript({file: '' });", "type": "error"}]
    
    Difference:
    
    - Expected
    + Received
    
    - Array []
    + Array [
    +   Object {
    +     "code": "CONTENT_SCRIPT_EMPTY",
    +     "column": 35,
    +     "description": "Content script file name should not be empty",
    +     "file": "file-with-code.js",
    +     "line": 1,
    +     "message": "Content script file name should not be empty.",
    +     "sourceCode": "... /* the validated source code */ ...",
    +     "type": "error",
    +   },
    + ]
```

And the above error would make it way easier to guess the reasons behind a failure.

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -0,0 +1,191 @@
+import { VALIDATION_ERROR } from 'const';
+import { CONTENT_SCRIPT_NOT_FOUND, CONTENT_SCRIPT_EMPTY } from 'messages/javascript';
+import JavaScriptScanner from 'scanners/javascript';
+
+describe('content_scripts_file_absent', () => {
+  it('should show an error when content script is missing', () => {
+    const code = `browser.tabs.executeScript({file: '/content_scripts/absentFile.js'});`;
```
rpl: Nit, it seems that all the tests in this test file are sharing some of the "test case setup steps"  (e.g. like the code between line 7 and line 16 of this test case), it could be reasonable to refactor it into a small shared helper function defined and used in this test file, something like:

```js
function createJsScanner(code, validatedFilename, existingFiles = {}) {
  return new JavascriptScanner(code, validatedFileName, {
    addonMetadata: { id: 'test' },
    existingFiles: {
       ...existingFiles,
       [validatedFilename]: ''
    },
  });
}
```

Then using the helper every test case could be reduce to something like:

```js
it('should ...', async () => {
  const code = `...`
  const jsScanner = createJsScanner(code, 'validated-file-name.js');
  // or createJsScanner(code, 'validated-file-name.js', {'another/file/name.js': ''});
  const { linterMessages } = await jsScanner.run();
  ... // assertions on the linterMessages
});
```

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -0,0 +1,191 @@
...
+        fileRequiresContentScript,
+      },
+      addonMetadata: {
+        id: 'test',
+      },
+    });
+
+    return jsScanner.scan()
+      .then(({ linterMessages }) => {
+        linterMessages.sort();
```
rpl: do we need to sort the array of the linter messages here?
(given that it seems that we expect an array on a single element based on the asserion at line 21).

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -0,0 +1,191 @@
...
+import { CONTENT_SCRIPT_NOT_FOUND, CONTENT_SCRIPT_EMPTY } from 'messages/javascript';
+import JavaScriptScanner from 'scanners/javascript';
+
+describe('content_scripts_file_absent', () => {
+  it('should show an error when content script is missing', () => {
+    const code = `browser.tabs.executeScript({file: '/content_scripts/absentFile.js'});`;
+    const fileRequiresContentScript = 'file-requires-content-script.js';
+    const jsScanner = new JavaScriptScanner(code, fileRequiresContentScript, {
+      existingFiles: {
+        fileRequiresContentScript,
```
rpl: I guess these existingFiles properties should be `{ [fileRequiresContentScript]: '' }` (using the [ES2015 computed property name syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015), which would be equivalent to `{ 'file-requires-content-script.js': '' }`), am I right?

hameleonka:
@rpl there is new code to review :)

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -0,0 +1,48 @@
...
+export default {
+  create(context) {
+    const existingFiles = context.settings.existingFiles || {};
+    return {
+      MemberExpression(node) {
+        if (node.object.object && isBrowserNamespace(node.object.object.name)) {
+          const namespace = node.object.property.name;
+          const property = node.property.name;
+          // namespace should be tabs function should be executeScript
```
rpl: Nit, as a convention all the inline comments should have the first letter capitalized and ends with a dot.

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -0,0 +1,48 @@
...
+          // i.e. browser.tabs.executeScript
+          if (namespace === 'tabs' && property === 'executeScript') {
+            /* eslint-disable no-restricted-syntax */
```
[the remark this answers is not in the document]
hameleonka: eslint doesn't like (`for...of`)[https://github.com/airbnb/javascript/issues/1271] and (`continue`)[https://eslint.org/docs/rules/no-continue]

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -0,0 +1,89 @@
...
+  it('should not break validator when file property missing', async () => {
+    const code = `browser.tabs.executeScript({code: 'console.log("lol")'});`;
+    const fileRequiresContentScript = 'file-requires-content-script.js';
+    const jsScanner = createJsScanner(code, fileRequiresContentScript);
+
+    try {
+      const { linterMessages } = await jsScanner.scan();
+      expect(linterMessages).toEqual([]);
+    } catch (e) {
+      expect(true).toBe(false);
```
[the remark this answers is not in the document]
hameleonka: I merged it to one test, but it could be not clear why test failed.

hameleonka:
@rpl new changes, now there is `forEach` with early returns instead of `for...of` with early `continue`
`for..of`: https://github.com/airbnb/javascript/issues/1271
`continue`: https://eslint.org/docs/rules/no-continue

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -6,42 +6,47 @@ export default {
...
-          }
+        if (!node.object.object || !isBrowserNamespace(node.object.object.name)) {
+          // Early return when it's not our case.
+          return;
+        }
+        const namespace = node.object.property.name;
+        const property = node.property.name;
+        // Namespace should be tabs function should be executeScript.
+        // I.E. browser.tabs.executeScript.
+        if (namespace !== 'tabs' || property !== 'executeScript') {
```
rpl: Just noticed that we should also add something like ` || node.parent.type !== 'CallExpression'` here, so that we can be 100% sure that `node.parent.arguments.forEach` at line 20 is going to work.

(we should also add `browser.tabs.executeScript;` to the rest of scenarios in the "no false positive" test case, and we expect it to fail because the above check is missing, and pass once we add it).

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -35,36 +35,30 @@ describe('content_scripts_file_absent', () => {
 
   it('should not report any errors when filename is not static string', async () => {
```
rpl: How about rephrasing this into something that also include the other scenarios from the code template string used here? e.g. something like `it('should not report errors for code ignored by this rule', async () => {})` or `it('should not report false positive errors', async () => {})`.

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -35,36 +35,30 @@ describe('content_scripts_file_absent', () => {
...
     const code = `
-      const fileName = 'script.js';
-      browser.tabs.executeScript({file: \`/content_scripts/\${fileName}\`});
-      browser.tabs.executeScript({file: '/content_scripts/'+ fileName });
-      browser.tabs.executeScript({file: '/content_scripts/' + 'absentFile.js' });
-      browser.tabs.executeScript(1,{file: '/content_scripts/' + 'absentFile.js' });
-      browser.tabs.executeScript(1,{file: 2 });
-      browser.tabs.executeScript(1,{file: null });
+      browser.tabs.executeScript({ file: \`/content_scripts/\${fileName}\` });
```
rpl: Nit, I would put some inline comments inside this template string to make it more clear which are the grouped the scenarios, e.g.:

```
// API calls using a non literal strings file attribute.
....

// API calls with a non string file attribute.
...

// API calls without a file attribute.
...

// non WebExtensions API function calls.
...
```

PR #1664, opened 2017-11-21
Mozilla greenkeeper/deepmerge 2.0.1

mstriemer:
This updates the import script to use the old array merge but still updates deepmerge to the latest version.


mstriemer:
I definitely don't love it. I wish they just included this function in deepmerge rather than putting it in the docs 🤷‍♂️

```js
import merge, { oldArrayMerge } from 'deepmerge';
merge(a, b, { arrayMerge: oldArrayMerge });  // Much nicer...

PR #1665, opened 2017-11-22
More precise CSP evaluation

### LINKED ISSUE #1660: CSP wrongly considers default-src+script-src unsafe
https://reviewers.addons.mozilla.org/en-US/firefox/files/browse/782622/#top

has default-src with various hosts, but script-src self. The validator things that this allows remote scripts.

This is actually ok since the script-src is reset to self. The validator should understand this.

wagnerand:
Fix #1660.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -285,13 +285,23 @@ export default class ManifestJSONParser extends JSONParser {
     // Not sure about FTP here but CSP spec treats ws/wss as
     // equivalent to http/https.
     const validProtocols = ['ftp:', 'http:', 'https:', 'ws:', 'wss:'];
-    const candidates = ['script-src', 'default-src', 'worker-src'];
+    // The order is important here, 'default-src' needs to be
+    // before 'script-src'
+    const candidates = ['default-src', 'script-src', 'worker-src'];
 
+    let insecureDefaultSrc = false;
+    let secureScriptSrc = false;
     for (let i = 0; i < candidates.length; i++) {
       /* eslint-disable no-continue */
       const candidate = candidates[i];
       if (Object.prototype.hasOwnProperty.call(directives, candidate)) {
         const values = directives[candidate];
+
+        // If the 'default-src' is insecure, check whether the 'script-src'
+        // makes it secure, ie 'script-src: self;'
+        if (insecureDefaultSrc && values.length === 1 && values[0] === '\'self\'') {
+          secureScriptSrc = true;
+        }
 
         for (let j = 0; j < values.length; j++) {
           let value = values[j].trim();
@@ -306,7 +316,13 @@ export default class ManifestJSONParser extends JSONParser {
             (validProtocols.some((x) => value.startsWith(x))));
 
           if (hasProtocol) {
-            this.collector.addWarning(messages.MANIFEST_CSP);
+            if (candidate === 'default-src') {
+              // Remember insecure 'default-src' to check whether a later
+              // 'script-src' makes it secure
+              insecureDefaultSrc = true;
+            } else {
+              this.collector.addWarning(messages.MANIFEST_CSP);
+            }
             continue;
           }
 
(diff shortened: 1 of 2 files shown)


src/parsers/manifestjson.js
```diff
@@ -285,14 +285,24 @@ export default class ManifestJSONParser extends JSONParser {
...
       /* eslint-disable no-continue */
       const candidate = candidates[i];
       if (Object.prototype.hasOwnProperty.call(directives, candidate)) {
         const values = directives[candidate];
 
+        // If the 'default-src' is insecure, check whether the 'script-src'
+        // makes it secure, ie 'script-src: self;'
+        if (insecureDefaultSrc && values.length === 1 && values[0] === '\'self\'') {
+          secureScriptSrc = true;
+        }
```
EnTeQuAk: I feel like this is missing a check for `candidate` since this doesn't work in case `worker-src: self;` is being set (for whatever reason)

src/parsers/manifestjson.js
```diff
@@ -285,14 +285,24 @@ export default class ManifestJSONParser extends JSONParser {
     // Not sure about FTP here but CSP spec treats ws/wss as
     // equivalent to http/https.
     const validProtocols = ['ftp:', 'http:', 'https:', 'ws:', 'wss:'];
-    const candidates = ['script-src', 'default-src', 'worker-src'];
+    // The order is important here, 'default-src' needs to be
+    // before 'script-src'
```
EnTeQuAk: add "to ensure it can overwrite `default-src` security policies" or similar

src/parsers/manifestjson.js
EnTeQuAk: I'd also not use two different variables but rather just use `script-src` to overwrite any insecurity that got set earlier. For example, define a `insecureSrcDirective` which get's set in case we discover an insecure `default-src` and unset in case `script-src` fixes it again.

wagnerand: Good catch, thank you!

wagnerand:
@EnTeQuAk Changes added. I also added a test to cover `default-src http:; worker-src 'self'`.

tests/parsers/test.manifestjson.js
```diff
@@ -431,6 +431,13 @@ describe('ManifestJSONParser', () => {
         // unsafe-inline is not supported by Firefox and won't be for the
         // forseeable future. See http://bit.ly/2wG6LP0 for more details-
         "script-src 'self' 'unsafe-inline';",
+
+        // 'default-src' is insecure, but the limiting 'script-src' prevents
+        // remote script injection
+        "default-src *; script-src 'self'",
```
muffinresearch: This CSP would also allow `object-src` to fallback to `*`.

muffinresearch: @wagnerand out of curiosity do you know what spec of CSP web-ext's support? Or is it tied to what firefox supports for web content by default?

wagnerand: Hm, yes, but we don't check for `object-src` either way.

wagnerand: @muffinresearch It mostly follows the web content spec, the specifics (and differences) are documented at https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/content_security_policy

wagnerand: @muffinresearch We discussed this with Security (I can't find the issue right now), and came to the conclusion that only `script-src` and `worker-src` are relevant to catch a CSP that allows remote script injection.

EnTeQuAk: This was implemented in https://github.com/mozilla/addons-linter/issues/1298 but I might have that `object-src` information from IRC, I can't find it unfortunately.

muffinresearch: > @muffinresearch We discussed this with Security (I can't find the issue right now), and came to the conclusion that only script-src and worker-src are relevant to catch a CSP that allows remote script injection. Do you have info that suggests otherwise?

There's presumably a reason the default web-ext CSP sets `object-src 'self'` to prevent remote objects by default?

wagnerand: This seems out of scope for this PR. Can we file a new issue, please?

muffinresearch: @wagnerand @EnTeQuAk also FWIW - not setting object-src is called out here as explicitly invalid: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/content_security_policy#Invalid_examples

PR #1692, opened 2017-12-07
Babel setup for async/await

hameleonka:
@rpl this PR adds babel support for async/await with transpilation to ES2015 generators for backward compatibility with node 6 + working example of function converted to async/await syntax in linter.js. Please check it out.


.eslintrc
```diff
@@ -2,7 +2,7 @@
   "rules":{
     "no-console": 0,
     // Inherited from airbnb via eslint-config-amo
-    'function-paren-newline': 0,
+    "function-paren-newline": 0
```
rpl: thanks!

package.json
```diff
@@ -35,15 +35,20 @@
     "babel-eslint": "8.0.2",
     "babel-jest": "21.2.0",
     "babel-loader": "7.1.2",
+    "babel-plugin-syntax-async-functions": "^6.13.0",
```
rpl: Every dependency is tha package.json file should be an exact version (without the ^), so that the version is fixed and greenkeeper can open a pull request for us when one of our dependency released a new version on npm.

src/linter.js
```diff
@@ -273,23 +273,20 @@ export default class Linter {
...
+      const stats = await _lstatPromise(filepath);
+      if (stats.isFile() === true || stats.isDirectory() === true) {
+        return stats;
+      }
+    } catch (err) {
+      if (err.code !== 'ENOENT') {
+        throw err;
+      }
+    }
+    throw invalidMessage;
```
rpl: 👍 it looks great (I would add an empty line at line 289, but that is just a small nit)

tests/test.linter.js
```diff
@@ -246,15 +246,15 @@ describe('Linter', () => {
     addonLinter.io = { files: { whatever: {} } };
     addonLinter.io.getFile = () => Promise.resolve();
     addonLinter.getScanner = sinon.stub();
-    class fakeScanner {
+    class FakeScanner {
```
rpl: Thanks! (it seems a convention that it could be also nice to enforce on all our sources using an eslint rules, but we can evaluate that in a follow up because I guess that an eslintrc rule could catch more of these and they would be changes unrelated to this pull request goal).

hameleonka:
@rpl I finished rewriting linter.js. Could you please check it out when you have a chance.

src/linter.js
```diff
@@ -465,78 +447,69 @@ export default class Linter {
...
+      const filesWithoutJSLibraries = Object.keys(files).filter((file) => {
+        return !(file in this.addonMetadata.jsLibs);
+      });
+
+      await this.scanFiles(filesWithoutJSLibraries);
+
+      this.print(deps._console);
+      // This is skipped in the code coverage because the
+      // test runs against un-instrumented code.
+      /* istanbul ignore if  */
```
rpl: Nit, do you mind to remove one space from the end of this comment? (it was already there, but given that we are touching this function... ;-), same at line 492)

src/linter.js
```diff
@@ -465,78 +447,69 @@ export default class Linter {
...
-        })
-        .catch((err) => {
-          log.debug(err);
-          this.handleError(err, deps._console);
-          throw err;
-        });
+      }
     }
-    return this.scan(deps).then(() => this.output);
+    await this.scan(deps);
```
rpl: Nit, do you mind to add one empty line at line 504 (only to make it a bit more visible).

src/linter.js
```diff
@@ -327,119 +313,115 @@ export default class Linter {
...
-        if (ScannerClass !== BinaryScanner && ScannerClass !== FilenameScanner) {
-          // Check for badwords across all code files
-          this._markBadwordUsage(filename, fileData);
+      scanResult = {
+        linterMessages: [filesizeError],
+        scannedFiles: [filename],
+      };
+    } else {
+      const fileData = await this.io.getFile(filename, ScannerClass.fileResultType);
+      if (ScannerClass !== BinaryScanner && ScannerClass !== FilenameScanner) {
```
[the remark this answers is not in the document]
hameleonka: @rpl I moved fileData from line 339 to line 320, but I don't understand how can I merge else on line 338 with if from line 340. It seems for me like they are both necessary.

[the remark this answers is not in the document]
rpl: @hameleonka yeah, don't worry about that, you are right.

Also I think that it is better that we don't change too much this function in this PR, while we are rewriting many of these functions into async functions (and it sound that making this function a bit more readable could be part of the issue mentioned at line 354).

hameleonka:
@rpl I've made all changes that you've mentioned above. Please check it again.

src/linter.js
```diff
@@ -465,78 +446,70 @@ export default class Linter {
...
-            const _files = this.config.scanFile.join(', ');
-            throw new Error(`Selected file(s) not found: ${_files}`);
-          }
+  async scan(deps = {}) {
+    try {
+      await this.extractMetadata(deps);
+      const files = await this.io.getFiles();
+
+      if (this.config.scanFile &&
+        !this.config.scanFile.some((f) => Object.keys(files).includes(f))) {
```
rpl: Nit, align `!this.config.scanFile...` with `this.config.scanFile` from the previous line.

PR #1693, opened 2017-12-07
Detect file type before read as stream

### LINKED ISSUE #1646: Binary images are incorrectly reported as corrupted

Icons linked in the manifest.json are reported as CORRUPT_ICON_FILE, if they are PNGs.

When validating a web extension from a directory, images are read using `encoding: 'utf8'`:

https://github.com/mozilla/addons-linter/blob/ff7ffa2ce46f27738d83416b66acc950da9b1fe2/src/io/directory.js#L62-L66

This makes binary images unreadable for sharp and it will return the error "Input buffer contains unsupported image format" here:

https://github.com/mozilla/addons-linter/blob/f7514790a7d5aaa24cf611620cc92e82eafccde2/src/parsers/manifestjson.js#L28-L31

The tests don't catch this, because they don't use the same method to create the readable stream:

https://github.com/mozilla/addons-linter/blob/f7514790a7d5aaa24cf611620cc92e82eafccde2/tests/parsers/test.manifestjson.js#L786

As soon as you change this to also use `encoding: 'utf8'` the test fails.

asamuzaK:
Fixes #1646

Detect file type from file extension.

The following extensions will be treated as ASCII file, other as BINARY.
[css, htm, html, js, json, mml, svg, txt, xml, xhtml]

I'm not sure that's enough or too much for linter.
Please tell me.

* [✓] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [✓] Add `Fixes #ISSUENUM` at the top of your PR.
* [✓] Add a description of the the changes introduced in this PR.
* [✓] The change has been successfully run locally.
* [✓] Add tests to cover the changes added in this PR.


asamuzaK:
Sorry, commit by mistake.
It's not related to this PR, reverted.

rpl:
Hey @asamuzaK Thank you so much for opening this PR and mention it in #1714, unfortunately I did not notice that you were already working on it when I was investigating the issue in web-ext (mozilla/web-ext#1164) and so I worked on it on a different PR.

I ended up to land my PR, mostly because I preferred a more "unit test" testing strategy (which in this PR is more like an integration test, which we should definitely add to this repo, but I would prefer to add them into a separate integration test suite, instead of inside the current unit test suite) and also because of the `isAscii` check inside of the `getFileAsStream` (which could be hard to keep in sync over the time), nevertheless I kept into account your proposals from this PR and I tweaked the #1714 accordingly before landing it.

And so, long story short, even if I'm closing this PR, I would say that there is still a bit of your PR in the final #1714 version!

Thanks again!

PR #1699, opened 2017-12-13
Async/await refactoring for files in scanners folder 

hameleonka:
@rpl in this PR I refactored files to async/await syntax in scanners folder. Please check it out, waiting for your review.

### DIFF
--- a/src/scanners/base.js
+++ b/src/scanners/base.js
@@ -41,56 +41,39 @@ export default class BaseScanner {
     ensureFilenameExists(this.filename);
   }
 
-  scan(_rules = this._defaultRules) {
-    return new Promise((resolve, reject) => {
-      this.getContents()
-        .then((contents) => {
-          const promises = [];
-          // Ignore private functions exported in rule files.
-          //
-          // (These are exported for testing purposes, but we don't want
-          // to include them in our linter's rules.)
-          const rules = ignorePrivateFunctions(_rules);
+  async scan(_rules = this._defaultRules) {
+    const contents = await this.getContents();
+    // Ignore private functions exported in rule files.
+    //
+    // (These are exported for testing purposes, but we don't want
+    // to include them in our linter's rules.)
+    const rules = ignorePrivateFunctions(_rules);
 
-          Object.keys(rules).forEach((rule) => {
-            this._rulesProcessed++;
-            promises.push(rules[rule](contents, this.filename, this.options));
-          });
+    const ruleResults = await Promise.all(Object.keys(rules).map((rule) => {
+      this._rulesProcessed++;
+      return rules[rule](contents, this.filename, this.options);
+    }));
 
-          return Promise.all(promises);
-        })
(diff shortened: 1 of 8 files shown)


src/scanners/binary.js
```diff
@@ -2,7 +2,6 @@ import BaseScanner from 'scanners/base';
 import * as messages from 'messages';
 import * as constants from 'const';
 
-
```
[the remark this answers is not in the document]
hameleonka: @rpl all changes have been made

src/scanners/base.js
```diff
@@ -41,56 +41,39 @@ export default class BaseScanner {
...
 
-          resolve(this._parsedContent);
-        })
-        .catch(reject);
-    });
+  async getContents() {
+    if (this._parsedContent !== null) {
+      return this._parsedContent;
+    }
+    const contents = await this._getContents();
```
rpl: it seems that we could remove the `const contents` here and just set the value returned from `await this._getContents();` to `this._parsedContent` and then return it.

src/scanners/css.js
```diff
@@ -52,56 +52,49 @@ export default class CSSScanner extends BaseScanner {
...
-          // absolute paths.
-          file: this.filename,
-        }));
-
-
-        // A syntax error has been encounted so it's game over.
-        return resolve(null);
+  async _getContents(_cssParser = postcss) {
+    try {
+      const rootNode = _cssParser.parse(this.contents, { from: this.filename });
```
rpl: `rootNode` doesn't seem to be used anywhere in this function, we could just return the value returned by `_cssParser.parse`.

src/scanners/base.js
hameleonka: @rpl I've made all changes

src/scanners/base.js
```diff
@@ -41,56 +41,38 @@ export default class BaseScanner {
...
 
-          resolve(this._parsedContent);
-        })
-        .catch(reject);
-    });
+  async getContents() {
+    if (this._parsedContent !== null) {
+      return this._parsedContent;
+    }
+    this._parsedContent = await this._getContents();
```
rpl: I would defer it for a follow up, because it is sightly different from what this function was doing before the async/await refactoring, but to prevent `this._getContents` to be called twice if `getContents` has been called twice while the this._getContents() is still not resolved, it could be better to store the promise and return it from the async function, e.g.:

```js
async getContents() {
  if (this._promiseParsedContent !== null)
    return this._promiseParsedContent;
  }

  this._promiseParsedContent = this._getContents();

  return this._promiseParsedContent
}
```

EnTeQuAk:
@rpl is this good to land?

rpl:
@EnTeQuAk yes, the behavior currently implemented is exactly the same as it was in the previous version, we can evaluate the improvement suggested in my last comment (https://github.com/mozilla/addons-linter/pull/1699#discussion_r160096877) in a follow up.

Issue #1535, opened 2017-09-11
[tracker] Localize the Linter


muffinresearch:
This will need additional bugs filing but this is an attempt to capture the steps needed to make the linter fully localized.

* [x] Setup locales - there's some potentially useful scripts in addons-frontend that will provide a guide for the typical mozilla way to do this.
* [x] Create a locale object to provide the gettext + sprintf etc. We use Jed for this in addons-frontend.
* [ ] Ensure all strings are wrapped in gettext calls.
* [x] Setup locale extraction - https://github.com/mozilla/babel-gettext-extractor/ might be a good solution for this we're using in mozilla/addons-frontend.
* [ ] Automate locale extraction.
* [ ] Setup localization via the CLI  - the locale passed from the terminal should set the correctly language.
* [ ] Setup localization via the submission flow in addons-server. This will warrant a separate bug to be filed - this will probably mean the CLI will have a locale option so the web-ui can pass this to the task so that errors are localized. 

/cc @EnTeQuAk for additional ideas + input.

drashti4:
I would like to work on this issue.

EnTeQuAk:
@drashti4 we currently have this issue set aside for consideration as part of outreachy. For now it will be preferable if you can find other issues to work on, however I will be sure to let you know if this issue is not going to be worked on as part of that program.

There are issues marked with [contrib: maybe good first bug](https://github.com/mozilla/addons-linter/labels/contrib%3A%20maybe%20good%20first%20bug) or [contrib: welcome](https://github.com/mozilla/addons-linter/issues?q=is%3Aissue+is%3Aopen+label%3A%22contrib%3A+welcome%22) that could be good starting points.

Ambalgi-Pankaja:
I would like to work on this project through outreachy. I applied for this project, hopefully I get a chance.

Ambalgi-Pankaja:
Can you please give me task to start working on this project?

EnTeQuAk:
@Ambalgi-Pankaja this issue is specifically meant for a outreachy mentee to work on once selected. Please have a look at other [issues](https://github.com/mozilla/addons-linter/issues) to select from (note the `contrib: welcome` label) but generally the outreachy application period just ended.

We're always happy to see new contributions happening though!

akve17:
Hi @EnTeQuAk , I have applied for outreachy and would like to work on bugs related to linter. Can you suggest me some good bugs i can take up?
Thanks!

[...]

PR #1702, opened 2017-12-14
Created create-locales script

ravneette:
In Reference to #1535

This PR adds the create-locales script in the bin folder to create po files from pot files. 
I have hardcoded a pot file for now, and would work on the script to create it in a separate PR.


ravneette:
Hey @EnTeQuAk . Yes I think that makes sense.
Removed the locale folder from the root directory for now.

PR #1704, opened 2017-12-14
Improve messaging for missing manifest.json

### LINKED ISSUE #424: Detect when a package is in a nested directory.

wagnerand:
Fix #424.

### DIFF
--- a/src/messages/layout.js
+++ b/src/messages/layout.js
@@ -22,7 +22,7 @@ export const TYPE_NO_MANIFEST_JSON = {
   code: 'TYPE_NO_MANIFEST_JSON',
   message: _('manifest.json was not found'),
   description: _(oneLine`A manifest.json in the root of the extension was not
-    found. Check that your code is not nested in a subdirectory.
+    found. Make sure that the manifest.json is in the top-level of the file.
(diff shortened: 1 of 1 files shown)


src/messages/layout.js
```diff
@@ -21,9 +21,9 @@ export const BAD_ZIPFILE = {
 export const TYPE_NO_MANIFEST_JSON = {
   code: 'TYPE_NO_MANIFEST_JSON',
   message: _('manifest.json was not found'),
-  description: _(oneLine`A manifest.json in the root of the
-     extension was not found. See: https://mzl.la/2r2McKv for more on
-     packaging`),
+  description: _(oneLine`A manifest.json in the root of the extension was not
+    found. Check that your add-on code is not nested in a subfolder.
+    See: https://mzl.la/2r2McKv for more on packaging.`),
```
EnTeQuAk: When we're updating it… we're first saying `extension` but then referring to an `add-on`, I think we should unify this a bit?

wagnerand: Updated messaging.

muffinresearch:
:shipit:

PR #1707, opened 2017-12-15
fix: prevent Object properties to be detected as existent files by tabs.executeScript rule

rpl:
This PR fix a small issue in the rule that checks if the files used in a tabs.executeScript API calls actually exist (even if it is unlikely that an extension would call one of its content script file "constructor", or like other Object properties, the rule is supposed to raise an error if the filename is not an actual property of the existingFiles object).


rpl:
@EnTeQuAk it could be nice to have an helper function to use where we check that a filename exists, so that it would be harder to get caught by the subtle differences between using `... in ...` or `Object.hasOwnProperties` to check for the existence of a property name in an object.

PR #1712, opened 2017-12-17
Async/await refactoring for schema folder

hameleonka:
@rpl in this PR I refactored to async/await syntax schema folder (I found only 1 file there that needed refactoring) . Please check it out, waiting for your review.

### DIFF
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -567,6 +567,7 @@ async function getTarballPath({ inputPath, version }) {
         });
     });
   }
+
   throw new Error('inputPath or version is required');
 }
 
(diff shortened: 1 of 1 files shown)


src/schema/firefox-schemas-import.js
```diff
@@ -565,42 +565,36 @@ function getTarballPath({ inputPath, version }) {
         .on('close', () => {
           resolve(tmpPath);
         });
-    } else {
-      reject(new Error('inputPath or version is required'));
-    }
-  });
+    });
+  }
+  throw new Error('inputPath or version is required');
```
rpl: Nit, add an empty line between the closed bracket and the `throw new Error(...);`

hameleonka: Done

PR #1713, opened 2017-12-17
Refactoring to async functions: utils.js and rules/html folder

hameleonka:
@rpl in this PR I refactored functions that return Promise to async functions (without await). There are 3 files that I've changed : utils.js and 2 files in rules/html folder. Please check it out, waiting for your review.

### DIFF
--- a/src/rules/html/warn-on-inline.js
+++ b/src/rules/html/warn-on-inline.js
@@ -6,8 +6,8 @@ export async function warnOnInline($, filename) {
   const linterMessages = [];
   $('script').each((i, element) => {
     if ($(element).attr('src') === undefined &&
-      ($(element).attr('type') === undefined ||
-      $(element).attr('type') === 'text/javascript')) {
+        ($(element).attr('type') === undefined ||
+         $(element).attr('type') === 'text/javascript')) {
       linterMessages.push(
         Object.assign({}, messages.INLINE_SCRIPT, {
           /* This could occur in any HTML file, so let's make it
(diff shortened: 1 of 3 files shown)


src/rules/html/warn-on-inline.js
```diff
@@ -2,23 +2,21 @@ import { VALIDATION_WARNING } from 'const';
...
-            file: filename,
-          }));
-      }
-    });
-    resolve(linterMessages);
+export async function warnOnInline($, filename) {
+  const linterMessages = [];
+  $('script').each((i, element) => {
+    if ($(element).attr('src') === undefined &&
+      ($(element).attr('type') === undefined ||
```
rpl: Nit, do you mind to align this line and the next line to the `$(...)` from the previous line?
e.g.
```
if ($(element)....
    ($(element)...
     $(element)...)) {
  ...
}
```

src/rules/html/warn-on-inline.js
```diff
@@ -2,23 +2,21 @@ import { VALIDATION_WARNING } from 'const';
...
+        Object.assign({}, messages.INLINE_SCRIPT, {
+          /* This could occur in any HTML file, so let's make it
+            * a warning in case they've included any other file.
+            */
+          type: VALIDATION_WARNING,
+          file: filename,
+        }));
+    }
   });
+  return linterMessages;
```
rpl: Nit (optional, but nice) add an empty line between the `});` and the `return linterMessage`.

hameleonka: @rpl all changes have been made

src/rules/html/warn-on-inline.js
hameleonka: Done

PR #1714, opened 2017-12-18
fix: addons-linter is reporting png file as corrupted when using io.directory module

rpl:
The last released version of the addons-linter is reporting icon png files as corrupted when linting an extension from a directory (using the "io/directory" module), while it is correctly recognizing and linting the icon size when linting the same extension from a zip file (which uses the "io/xpi" or "io/crx" module).

The reason for the different behavior is the `getFileAsStream` method provided by the "io/directory" module, which is currently always passing the `encoding: 'utf-8'` to `createReadStream`.

The tests are not currently able to catch this issue because the unit tests related to "parsers/manifestjson.js" do not use the underlaying "io/*" modules and the needed file stream are created as part of the shimmed environment provided by the test case.

I'm not thrilled about the workaround currently applied in this PR (because it adds a parameter which is only provided by the "io/directory" and not from "io/xpi" or "io/crx", which do not specify any particular encoding for the returned stream), but I had to look into it more deeply to be sure that we are not going to introduce other regressions by not including an encoding by default in the "io/directory" module as an alternative fix.

As an additional side note, the "io/directory" module is likely to be the one used by addon developers that are linting their addon locally (e.g. using "web-ext lint") and so this issue does not impact the validation happening as part of an addon upload on addons.mozilla.org.


asamuzaK:
See also #1693

asamuzaK:
@rpl should close #1646

PR #1721, opened 2017-12-20
refactor normalizePath handling

asamuzaK:
Previously the normalizePath() function was used inside `src/parsers/manifestjson.js`, but now it has been moved to `src/utils.js` and been exported.
However, unit test of normalizePath() is not prepared.

In this PR, together with addition of unit test, I refactored function to fix some issues.

Refactored things are:
* It did not decode after converting to URL, so for example, if the path contained spaces it was not processed properly.
  Decoding character is added.
* Linter throws file not found error on Windows.
  Converted slashes `/` to OS specific separators.

Issue #1649, opened 2017-11-13
NO_MESSAGES_FILE reported on Windows, but not on macOS


matatk:
### Describe the problem and steps to reproduce it:

When I lint an add-on on macOS, it reports no errors. When I lint the same add-on on Windows, I get a `NO_MESSAGES_FILE` error. 

### What happened?

A `NO_MESSAGES_FILE` error was reported.

### What did you expect to happen?

No errors to be reported.

### Anything else we should know?

[Repo for the add-on I'm working on.](https://github.com/matatk/landmarks)

matatk:
Sorry, to build the add-on, issue `npm run build:firefox` (or `npm run build:all`). I have also attached an add-on file to this message.

[landmarks-2.1.0-firefox.zip](https://github.com/mozilla/addons-linter/files/1492829/landmarks-2.1.0-firefox.zip)

wagnerand:
Hm, I tried to install the linter from npm on Windows, but getting weird dependency installation errors. @EnTeQuAk have you ever tried to run the linter on Windows?

matatk:
The errors are probably due to needing to compile stuff to install Sharp.

I got it to work by running [`npm install --global windows-build-tools`](https://www.npmjs.com/package/windows-build-tools) from an admin command prompt.

wagnerand:
Thanks, yes I just found that on the internets...running now...

matatk:
There's a [function in my build script that creates the ZIP file](https://github.com/matatk/landmarks/blob/bc1b09f28d41abbeee7282cb38f4e9c832795288/scripts/build.js#L209-L227).  The ZIP file passes all the tests on AMO (and on macOS).  Actually I am pretty sure I transferred the ZIP file generated on macOS to Windows and found that addons-linter reported the error on Windows, but not on macOS with the same ZIP file.

However I am very-much open to the possibility of there being a bug in my code :-).

EnTeQuAk:
This should have been fixed via https://github.com/mozilla/addons-linter/pull/1730 and the `sharp` dependency has been removed in another PR as well.

In case it hasn't please let us know.

vcarciu:
Tried on Windows 10x64 with latest linter, [webext](**https://github.com/mozilla/addons-linter/files/1492829/landmarks-2.1.0-firefox.zip**) from this issue and there are no errors  : 
[...]

PR #1730, opened 2018-01-03
Convert scanned and collected file paths to unix path separator

rpl:
This PR contains a proposed approach to fix a regression related to some of the newly added validation rules (in particular the ones related to the icon image validations and the ones related to the content script files existence) when running on windows with an unpacked extension as target.

This bug has been reported on web-ext as mozilla/web-ext#1134, and on the addons-linter as #1649 and #1431

There is also another PR, #1721, which aims to fix these issues, but when I've been testing it on windows it doesn't seem to be enough and so I took a deeper look and prepared this PR to make the addons-linter test suite to pass completely when running on windows (and I also tried to run the addons-linter tool on a bunch of webextensions-examples).

- [x] convert scanned and collected file paths to unix path separator
- [x] run all the test suite on windows
- [x] skip firefox schema import tests on windows
- [x] run addons-linter test suite on appveyor (windows CI service)
- [ ] re-look at #1721 and check if there are further changes useful to cover other corner cases


rpl:
@kumar303 @EnTeQuAk appveyor is not enabled on the repo and the test suite is passing on windows as well \o/ Thanks a lot for the help.

(I've currently disabled coveralls submission from appveyor, because it doesn't recognize it as a valid job, but we can re-evaluate to fix it and re-enable it in a follow up).

This should now be ready for a final sign-off and landing.

asamuzaK:
Filed a follow up #1737

PR #1733, opened 2018-01-05
Async/await refactoring for tests/io folder - test.directory.js 

hameleonka:
@rpl in this PR I've worked on test.directory.js in tests/io folder. Please check it out, waiting for your review.

### DIFF
--- a/tests/io/test.directory.js
+++ b/tests/io/test.directory.js
@@ -103,8 +103,9 @@ describe('Directory.getFileAsStream()', () => {
 
     const readStream = await myDirectory.getFileAsStream('dir2/dir3/file3.txt');
 
-    const content = await readStringFromStream(readStream);
-    expect(content).toEqual('123\n');
+    await expect(
+      readStringFromStream(readStream)
+    ).resolves.toBe('123\n');
   });
 
   it('should not enforce utf-8 when encoding = null', async () => {
@@ -160,8 +161,9 @@ describe('Directory.getFileAsString()', () => {
     const myDirectory = new Directory('tests/fixtures/io/');
 
     await myDirectory.getFiles();
-    const string = await myDirectory.getFileAsString('dir2/dir3/file3.txt');
-    expect(string).toEqual('123\n');
+    await expect(
+      myDirectory.getFileAsString('dir2/dir3/file3.txt')
+    ).resolves.toBe('123\n');
   });
 
   it('should reject if stream emits error', async () => {
--- a/tests/io/test.directory.js
+++ b/tests/io/test.directory.js
@@ -6,7 +6,7 @@ import { unexpectedSuccess } from '../helpers';
 
 
 describe('Directory.getFiles()', () => {
-  it('should return cached data when available', () => {
+  it('should return cached data when available', async () => {
     const myDirectory = new Directory('tests/fixtures/io/');
     const fakeFileMeta = {
       size: 1,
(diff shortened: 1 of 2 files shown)


tests/io/test.directory.js
```diff
@@ -61,85 +55,86 @@ describe('Directory.getFiles()', () => {
...
-            });
-        })
-          .then((content) => {
-            expect(content).toEqual('123\n');
-          });
+    await myDirectory.getFiles();
+
+    const readStream = await myDirectory.getFileAsStream('dir2/dir3/file3.txt');
+
+    const readStringFromStream = () => {
```
rpl: Nit, there is no need to declare it as a function if we are not going to reuse it,
we could just store the promise:

```js
const onceReadString = new Promise((resolve, reject) => {
  ...
});

const content = await onceReadString;
...
```

or we could move the function outside of the test function and pass the readStream as a parameter:

```js
function readStringFromStream(readStream) {
    ...
  });
}

describe('...', () => {
  it('...', async () => {
    ...
    const content = await readStringFromStream(readStream);
    ...
  });
});
```

hameleonka: @rpl I decided to go with second option because readStringFromStream(readStream) is used in 2 tests

EnTeQuAk:

tests/io/test.directory.js
```diff
@@ -61,85 +55,84 @@ describe('Directory.getFiles()', () => {
...
-            expect(err.message).toContain(
-              '"whatever" does not exist in this dir.'
-            );
-          });
-      });
+
+    await myDirectory.getFiles();
+    try {
+      await myDirectory.getPath('whatever');
+      unexpectedSuccess();
```
rpl: it looks that jest with a version >= 20 supports a `expect(...).rejects` assertion that we could use in out tests instead of the `try { ... } catch (...) { ... }`:

- https://facebook.github.io/jest/docs/en/expect.html#rejects

e.g. this one would become something like:

```js
    await expect(
      myDirectory.getPath('whatever')
    ).rejects.toThrow('"whatever" does not exist in this dir.');
```

and as a plus jest produce a very nice failure message when the rejected promise doesn't match the expected error message (or if it doesn't fail as expected).

There are a number of `try { ... } catch (...) { ... }` in this file that we can make shorter and cleaner applying this approach, and it could be worth given that we are already changing this test file.

tests/io/test.directory.js
```diff
@@ -61,85 +55,84 @@ describe('Directory.getFiles()', () => {
...
+    try {
+      await myDirectory.getPath('/file1.txt');
+      unexpectedSuccess();
+    } catch (err) {
+      expect(err.message).toContain('Path argument must be relative');
+    }
   });
 });
 
+function readStringFromStream(readStream, transform) {
```
rpl: This helpers seems to be useful enough to be something exported by the helpers.js module (from the tests/ directory), I would not be surprised if it could be helpful in the xpi and crx test files as well.

tests/io/test.directory.js
```diff
@@ -61,85 +55,56 @@ describe('Directory.getFiles()', () => {
...
-          .then((content) => {
-            expect(content).toEqual('123\n');
-          });
-      });
+    await myDirectory.getFiles();
+
+    const readStream = await myDirectory.getFileAsStream('dir2/dir3/file3.txt');
+
+    const content = await readStringFromStream(readStream);
+    expect(content).toEqual('123\n');
```
hameleonka: @rpl thanks for the very helpful Jest link. Now I have a couple of suggestions about using Jest methods. For example, maybe here we can replace lines 106, 107 with 
await expect(readStringFromStream(readStream).resolves.toBe('123\n');

tests/io/test.directory.js
```diff
@@ -194,39 +140,31 @@ describe('Directory.getFileAsStream()', () => {
...
-      .then(() => {
-        return myDirectory.getFileAsString('dir2/dir3/file3.txt');
-      })
-      .then((string) => {
-        expect(string).toEqual('123\n');
-      });
+
+    await myDirectory.getFiles();
+    const string = await myDirectory.getFileAsString('dir2/dir3/file3.txt');
+    expect(string).toEqual('123\n');
```
hameleonka: @rpl same here 
await expect(myDirectory.getFileAsString('dir2/dir3/file3.txt').resolves.toBe('123\n');

tests/io/test.directory.js
rpl: @hameleonka 👍 sure! go for it!

tests/io/test.directory.js
rpl: @hameleonka 👍

tests/io/test.directory.js
hameleonka: @rpl I've changed some Jest methods in this file

PR #1739, opened 2018-01-07
Added the webpack configuration file to extract text

ravneette:
In Reference to #1535 

This PR adds the webpack.l10n.config.babel.js to facilitate extraction of strings.


EnTeQuAk:

package.json
```diff
@@ -33,6 +33,7 @@
   "devDependencies": {
     "babel-core": "6.26.0",
     "babel-eslint": "8.0.2",
+    "babel-gettext-extractor": "git+https://github.com/muffinresearch/babel-gettext-extractor.git#0d39d3882bc846e7dcb6c9ff6463896c96920ce6",
```
EnTeQuAk: @muffinresearch can we at some point get whatever changed into upstream so that we don't have to rely on git-dependencies? I fear that's bad for packaging up the linter since users need to have `git` installed as well.

webpack-common.js
```diff
@@ -0,0 +1,72 @@
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+
+const postCssPlugins = [];
```
EnTeQuAk: The linter doesn't use postcss so I think that can be removed

webpack-common.js
```diff
@@ -0,0 +1,72 @@
...
+          loader: 'sass-loader',
+          options: {
+            outputStyle: 'expanded',
+            sourceMap: true,
+            sourceMapContents: true,
+          },
+        },
+      ],
+    }),
+  },
```
EnTeQuAk: Actually all references to css and scss or sass files can and should be removed since we don't make use of them. So the whole `styleRules` can be removed.

webpack-common.js
```diff
@@ -0,0 +1,72 @@
...
+      query: babelQuery,
+    },
+    ...styleRules,
+    {
+      test: /\.svg$/,
+      use: [{ loader: 'svg-url-loader', options: urlLoaderOptions }],
+    }, {
+      test: /\.(jpg|png|gif|webm|mp4|otf|woff|woff2)$/,
+      use: [{ loader: 'url-loader', options: urlLoaderOptions }],
+    },
```
EnTeQuAk: You might want to remove the `styleRules` reference, the `svg` and `url-loader` rules here as well. The linter doesn't make use of them.

EnTeQuAk: Plus, looking at this. The whole file doesn't add anything new that isn't already in our `webpack.config.js` file. Can we remove the dependency on `getRules` in `webpack.l10n.config.babel.js` and remove this file maybe?

webpack.l10n.config.babel.js
```diff
@@ -0,0 +1,62 @@
...
+];
+
+const BABEL_QUERY = Object.assign({}, babelrcObject, {
+  plugins: babelPlugins.concat(babelL10nPlugins),
+});
+
+module.exports = Object.assign({}, webpackConfig, {
+  module: {
+    rules: getRules({ babelQuery: BABEL_QUERY }),
+  },
```
EnTeQuAk: You should be able to replace the `getRules` call with something like `webpackConfig.module.rules` which contains all our defined rules for babel.

ravneette:
Yes just went through the files. 
Looks like a lot of unneeded stuff inspired from addons-frontend.
Working on this now.

ravneette:
Hey @EnTeQuAk 

I've made the changes you requested. I think the next step would be to build a yarn command to allow users to extract locales.

Something like 
```
yarn extract-locales
```

From what I could get I need to add a new file inside the tasks/ folder that would run the webpack script and make a few changes in the GruntFile

Does that sound right?

ravneette:
So I replaced the rules in the webpack.config.js file to 

https://github.com/mozilla/addons-linter/pull/1739/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR29

and this seems to have broken a few things.

EnTeQuAk:
> I've made the changes you requested. I think the next step would be to build a yarn command to allow > users to extract locales.

> Something like `yarn extract-locales`

Just forgot to answer that, yeah - that'd be super awesome! Let's maybe prioritize that a bit more?

LinusU:
This pull request added the webpack dependency a second time, and on the latest master this package now depends on both webpack 3.x and 4.x at the same time 😆 

Also, I guess that `extract-text-webpack-plugin` should be added to `devDependencies` and not dependencies? I'm now getting a warning when installing this package in my module since `extract-text-webpack-plugin` has a `peerDependency` on WebPack 3.x

Actually, I don't see `extract-text-webpack-plugin` being used at all, am I'm missing something?

EnTeQuAk:
> This pull request added the webpack dependency a second time, and on the latest master this package now depends on both webpack 3.x and 4.x at the same time laughing

Heh, good eyes! There's already a PR that fixes that that should be merged some time today.

> Also, I guess that extract-text-webpack-plugin should be added to devDependencies and not dependencies?

True again, I'll fix that.

> Actually, I don't see extract-text-webpack-plugin being used at all, am I'm missing something?

I could be wrong but as far as I know it's implicitly used by our webpack config but I'll take another look.

LinusU:
> I could be wrong but as far as I know it's implicitly used by our webpack config but I'll take another look.

I think that `babel-gettext-extractor` is the package that are doing the work, `extract-text-webpack-plugin` is for extracting CSS...

I've submitted a PR to fix all webpack related dependencies: #1889

PR #1744, opened 2018-01-10
Async/await refactoring manifestjson.js

hameleonka:
@rpl in this PR I've worked on manifestjson.js. I'm not sure that new version of validateIcons() is a good one. Please check it out, waiting for your review.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -254,6 +254,7 @@ export default class ManifestJSONParser extends JSONParser {
       this.collector.addWarning(messages.corruptIconFile({
         path: iconPath,
       }));
+      log.debug('Icon validation message', err);
     }
   }
 
@@ -268,9 +269,7 @@ export default class ManifestJSONParser extends JSONParser {
       } else if (!IMAGE_FILE_EXTENSIONS.includes(_path.split('.').pop().toLowerCase())) {
         this.collector.addWarning(messages.WRONG_ICON_EXTENSION);
       } else {
-        promises.push(
-          this.validateIcon(_path, size)
-        );
+        promises.push(this.validateIcon(_path, size));
       }
     });
     return Promise.all(promises);
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -251,10 +251,10 @@ export default class ManifestJSONParser extends JSONParser {
         }));
       }
     } catch (err) {
+      log.debug(`Unexpected error raised while validating icon "${iconPath}"`, err);
       this.collector.addWarning(messages.corruptIconFile({
         path: iconPath,
       }));
-      log.debug('Icon validation message', err);
     }
   }
 
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -17,7 +17,8 @@ import { isToolkitVersionString } from 'schema/formats';
 import { parseCspPolicy, normalizePath } from 'utils';
 import BLOCKED_CONTENT_SCRIPT_HOSTS from 'blocked_content_script_hosts.txt';
 
-function getImageMetadata(io, iconPath) {
+
+async function getImageMetadata(io, iconPath) {
   // Get a non-utf8 input stream by setting encoding to null.
   // (only needed for the 'io/directory' module which open the file using the utf-8
   // encoding by default).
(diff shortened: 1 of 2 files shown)


EnTeQuAk:

hameleonka:
@EnTeQuAk @rpl thanks! now all checks have passed!

src/parsers/manifestjson.js
```diff
@@ -234,6 +236,27 @@ export default class ManifestJSONParser extends JSONParser {
     }
   }
 
+  async validateIcon(_path, size) {
```
rpl: given that we are extracting this into a new method I would also remove the `_` from `_path` and rename `size` into `expectedSize`. 

I think that it makes sense to make it a method, @EnTeQuAk how that sounds to you?

(if we opt to extract it into this new `validateIcon` method, we should also add a new test case for it in the same pull request).

src/parsers/manifestjson.js
```diff
@@ -246,31 +269,14 @@ export default class ManifestJSONParser extends JSONParser {
...
-              }));
-            })
+          this.validateIcon(_path, size)
         );
       }
     });
     return Promise.all(promises);
   }
 
+
```
rpl: Nit, remove this empty line

EnTeQuAk: Weird that eslint isn't warning here, hmm

src/parsers/manifestjson.js
EnTeQuAk: > I think that it makes sense to make it a method, @EnTeQuAk how that sounds to you?

Agreed, `_path` is a weird argument name for a method that should have a nicer interface.

src/parsers/manifestjson.js
rpl: We have an explicit eslint rule in web-ext to allow it:

https://github.com/mozilla/web-ext/blob/7832938473bea398768c35a093fdc11395923369/.eslintrc#L95

it doesn't seem to be part of the addons-linter eslint config, and it doesn't seem to be part of the rules coming from the eslint-config-amo preset, but it is part of the config if I run "./node_modules/.bin/eslint --print-config .", and the following is its current value:

```
    "no-multiple-empty-lines": [
      "error",
      {
        "max": 2,
        "maxEOF": 1
      }
    ],
```

tests/parsers/test.manifestjson.js
```diff
@@ -808,125 +806,136 @@ describe('ManifestJSONParser', () => {
...
-      const manifestJSONParser = new ManifestJSONParser(
-        json, addonLinter.collector, { io: getStreamableIO(files) });
-      return manifestJSONParser.validateIcons()
-        .then(() => {
-          expect(manifestJSONParser.isValid).toBeTruthy();
-          const { warnings } = addonLinter.collector;
-          expect(warnings.length).toEqual(1);
+      it('adds a warning if the image size is not the same as mentioned', async () => {
+        const addonLinter = new Linter({ _: ['bar'] });
+        const icon32 = 'tests/fixtures/icon-33.png';
```
rpl: Nit, how about rename this to `icon33`? so that the size mismatch is also visible when looking at the identifiers names.

src/parsers/manifestjson.js
```diff
@@ -234,6 +236,27 @@ export default class ManifestJSONParser extends JSONParser {
...
+      } else if (info.mime !== 'image/svg+xml' &&
+                  parseInt(info.width, 10) !== parseInt(expectedSize, 10)) {
+        this.collector.addWarning(messages.iconSizeInvalid({
+          path: iconPath,
+          expected: parseInt(expectedSize, 10),
+          actual: parseInt(info.width, 10),
+        }));
+      }
+    } catch (err) {
+      this.collector.addWarning(messages.corruptIconFile({
```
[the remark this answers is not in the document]
hameleonka: @rpl log.debug is a good idea!I've made changes, please check it out.

src/parsers/manifestjson.js
```diff
@@ -234,6 +236,28 @@ export default class ManifestJSONParser extends JSONParser {
...
+        path: iconPath,
+      }));
+      log.debug('Icon validation message', err);
```
rpl: Nit, do you mind to move the `log.debug` call before the `this.collector.addWarning` call?

also, how about changing the message from 'Icon validation message' into something like `Unexpected error raised while validating icon "${iconPath}"`? 

(so that it makes it clear that we are logging the underlying error and not the related validation message, and we also ensure that the `iconPath` is part of the logged message, there are chances that the `iconPath` could be already part of the raised error message but we can't be 100% sure because it is an unexpected exception).

hameleonka: @rpl it's done. Please take a look.

PR #1749, opened 2018-01-11
Fixup dependency mess

EnTeQuAk:
* Upgrade wbepack-dev-server to a fixed version
* Downgrade jest and jest-cli to a seemingly working version
* Make sure we actually install any package updates on test-runs

Refs https://github.com/webpack/webpack-dev-server/issues/1259#issuecomment-356243604
Refs https://github.com/facebook/jest/issues/3793


package.json
```diff
@@ -62,6 +62,7 @@
     "is-mergeable-object": "1.1.0",
     "isparta": "4.0.0",
     "jest": "22.0.4",
+    "jest-cli": "22.0.4",
```
rpl: :+1: it looks that is this explicit dependency on jest-cli that is fixing the issue

I think that we can remove the "jest" dependency above it, so that we don't need to keep them in sync when we upgrade jest (I tried locally and it worked fine).

EnTeQuAk: Hmm, that's actually true but given that `jest` only has `jest-cli` as a dependency we don't have to pin anything right now. Once they pin `jest-cli` we can get rid of the explicit dependency.

rpl: sounds good

PR #1763, opened 2018-01-17
Initial attempt to warn about coin miners.

### LINKED ISSUE #1643: Show warning if add-on uses 'coinhive'
Including coin miners in add-ons is strictly forbidden, and [Coinhive](https://coinhive.com/) is the method most often used by developers for this. We need to look for the string 'coinhive' in add-on code and show an error if present. It can lead to some false positives, but they should be rare and easy to work around.

Error message: `Firefox add-ons are not allowed to run coin miners.`

EnTeQuAk:
This does not attempt to detect all of them but primarily focuses on
detecting CoinHive for now.

There's a whole bunch to explore and block when looking at NoCoin but
I'd rather see how many false positives we get in the real world with
this code first.

Fixes #1643 but might need a few follow up issues to add more miners.


src/miner_blocklist.js
```diff
@@ -0,0 +1,16 @@
+export const MINER_BLOCKLIST = {
+  code: [
+    // CoinHive
+    /window.CoinHive=window.CoinHive/,
+    /CoinHive\.CONFIG/,
+    /coin-hive\.com/,
+    /coinhive\.com/,
+    /cnhv\.co/,
```
wagnerand: Hm, using domains is too broad. There are lots of anti-coin mining add-ons and adblockers that include blocklists containing those domains, so we can't match on them. Let' focus on patterns inside the coinhive script that we can detect even in minified/obfuscated code.

src/messages/layout.js
```diff
@@ -84,3 +84,12 @@ export const MOZILLA_COND_OF_USE = {
     See https://www.mozilla.org/en-US/about/legal/acceptable-use/ for more
     details.`),
 };
+
+export const COINMINER_USAGE_DETECTED = {
+  code: 'COINMINER_USAGE_DETECTED',
+  message: _('Firefox add-ons are not allowed to run coin miners.'),
+  description: _(oneLine`We do not allow coinminer scripts to be run inside
```
wagnerand: Nit: Let's use consistent wording: `coin miner` vs `coinminer` (vs `coin-miner`) 😄  (I prefer the latter.)

src/miner_blocklist.js
EnTeQuAk: Don't ask me why I included them like that, I didn't mean to 😆  I think I wanted to find a pattern that might not issue false-positives for adblockers but probably failed along the way.

wagnerand:
r+ once all tests are green. Thank you, @EnTeQuAk !

PR #1774, opened 2018-01-19
Generated the first set of po files

ravneette:
In Reference to #1535

This PR adds the initial po files generated according to the current code base for languages present in addons-frontend.


ravneette:
Generated using the create-locales script.

ravneette:
Haha. Yes that's what I thought. 
Had to file this PR thrice since my browser would crash every time I tried.

PR #1785, opened 2018-01-23
Fixed Jed gettext function

ravneette:
In Reference to #1535

This PR fixes the gettext function exported in utils.js. Right now it is a stub which returns the string as such.


ravneette:
So the next step here should be to have the user export an environment variable like
```
LINTER_LOCAL_LANGUAGE
```
which we can capture in utils.js to understand which language the user wants the errors and warnings in.

I am going through addons-frontend to understand how exactly we route the data from the extracted json files to jed.

I see something here : 

https://github.com/mozilla/addons-frontend/blob/master/src/core/i18n/utils.js#L254

but I'm a little stuck here.

Any pointers @EnTeQuAk ?

EnTeQuAk:
> I am going through addons-frontend to understand how exactly we route the data from the extracted json files to jed.

Yeah, so the language in addons-frontend comes from the users request. You can see that in the `rederApp` call in https://github.com/mozilla/addons-frontend/blob/master/src/core/client/base.js#L38-L79

And the most relevant call to fetch the locale data is here:

https://github.com/mozilla/addons-frontend/blob/master/src/core/client/base.js#L82-L94

Imho that should translate into a `require('locale/${locale}/messages.js')` which exposes the `domain` and `locale_data` for jed. Does that make sense?

ravneette:
So I used 
```
require('locale/${locale}/messages.js')
```

and Localization worked :smile: 

The command to run linter with localization is as follows : 
```js
LOCALE_LANG=de bin/addons-linter path/to/addon 
```
Although before merging this I think we need to export an i18n variable from linter.js which takes care of localization instead of the functions that are being returned by utils.js

Am I right on that @EnTeQuAk ?

EnTeQuAk:
Kinda, the `locale = process.env.LOCALE_LANG || 'en-US';` part should be moved to the `Linter` class in `linter.js` since we need one language to be configurable for the lifetime of the linter.

Preferably somewhere in `Linter.config` so that it's similarly configured like the `scanFile` option.

This will also make it easier to unittest, which will be a requirement for this PR to be merged, given that we're now able to unittest something :)

ravneette:
Sounds good. On it!

ravneette:
So we want just the language to be exported from linter.js, or the gettext and sprintf functions as well?

EnTeQuAk:
Both. I think it's fine to rely and use a global variable  (that holds
the locale and the Jed instance configured with the correct i18ndata)
that get's configured/set by the Linter instance.
We always have a linter instance so that should work. 

Does that make sense? 


On Mon, Feb 12, 2018, at 5:24 PM, Ravneet wrote:
> So we want just the language to be exported from linter.js, or the
> gettext and sprintf functions as well?> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[1], or mute the thread[2].> 

Links:

  1. https://github.com/mozilla/addons-linter/pull/1785#issuecomment-364976351

ravneette:
okay I'm a little stuck on how to access that global variable in files like /src/messages/manifestjson.js

So we need to add two new parameters here : https://github.com/mozilla/addons-linter/blob/master/src/linter.js#L46 (locale and jed)

Also I'll remove https://github.com/mozilla/addons-linter/blob/master/src/utils.js#L128 and https://github.com/mozilla/addons-linter/blob/master/src/utils.js#L136 from utils.js. 

Now we need to access the linter.config.jed and linter.config.locale variables in files like /src/messages/manifestjson.js, but these files dont include linter as such.

Should I be including linter at the top of these files?

ravneette:
Nevermind, I think I figured it out.

src/utils.js
```diff
@@ -126,14 +126,17 @@ export function getVariable(context, name) {
  *
  */
 export function gettext(str) {
-  return str;
+  let locale = getConfig().argv.locale;
+  let i18ndata = require(`./locale/${locale}/messages.js`)
+  let jed = new Jed(i18ndata);
+  return jed.gettext(str);
 }
```
EnTeQuAk: ~~I think I would put `i18ndata` (as `I18N_DATA` or so) at the global level (just like `SOURCE_MAP_RE` above) for various reasons:~~

**Update:** So I talked to @rpl and we came up that the least magic implementation indeed would be to just rely on `process.ENV` for now, as you originally did and ignore the fact that we might want a cli option.

The problems around the cli option are quite huge and resulted in me scratching up https://gist.github.com/EnTeQuAk/001fb42cab37b7f20a4d9f6df8030a63 which is something we might not want to do but it's been fun.

So, let's summarize my finalized thoughts on this:

 * relying on `process.ENV` is simple and allows `Jed` to be globally initialized in `utils.js` from the very beginning
 * No need to overwrite the currently used locale, anywhere. We can just use what's defined by the environment
 * re-using the environment variables isn't a blocker to make it useful for addons-server, it's just 1-2 lines more code which is reasonable and doesn't justify any use of magic

The problems with the cli argument are:

 * We'd need to make our strings that get returned by `gettext` be translated on **first access** instead of directly during the `gettext` call, otherwise they would always be English 'cause that's what we initialize by default
 * That would mean the use of `Proxy` and other laziness methodologies which might not be worth it.

Does that make sense?

EnTeQuAk:
I think I would put i18ndata (as I18N_DATA or so) at the global level (just like SOURCE_MAP_RE above) for various reasons:

Update: So I talked to @rpl and we came up that the least magic implementation indeed would be to just rely on process.ENV for now, as you originally did and ignore the fact that we might want a cli option.

The problems around the cli option are quite huge and resulted in me scratching up https://gist.github.com/EnTeQuAk/001fb42cab37b7f20a4d9f6df8030a63 which is something we might not want to do but it's been fun.

So, let's summarize my finalized thoughts on this:

    relying on process.ENV is simple and allows Jed to be globally initialized in utils.js from the very beginning
    No need to overwrite the currently used locale, anywhere. We can just use what's defined by the environment
    re-using the environment variables isn't a blocker to make it useful for addons-server, it's just 1-2 lines more code which is reasonable and doesn't justify any use of magic

The problems with the cli argument are:

    We'd need to make our strings that get returned by gettext be translated on first access instead of directly during the gettext call, otherwise they would always be English 'cause that's what we initialize by default
    That would mean the use of Proxy and other laziness methodologies which might not be worth it.

Does that make sense?

ravneette:
Yup that makes sense, and makes things easier!

ravneette:
> Update: So I talked to @rpl and we came up that the least magic implementation indeed would be to just rely on process.ENV for now, as you originally did and ignore the fact that we might want a cli option.

So we would expect the user to always pass in a locale variable when he is running addons-linter?

>The problems around the cli option are quite huge and resulted in me scratching up https://gist.github.com/EnTeQuAk/001fb42cab37b7f20a4d9f6df8030a63 which is something we might not want to do but it's been fun.

I'm still going through this file to understand what we're doing there.

I'm still a little unclear as to what was the problem with using cli arguments, but I do get that it was a little messy.

So I'm to revert back to using process.env variable in utils.js for now right? And write the tests for testing the same?

EnTeQuAk:
> So we would expect the user to always pass in a locale variable when he is running addons-linter?

Kinda, on a usual shell those environment variables are already set. There's a few things to note, because there are various environment variables controlling the actually used locale:

Something like ```

const env = process.env
const language = env.LANG || env.LANGUAGE || env.LC_ALL || env.LC_MESSAGES;
```

will be a good idea. Or using something like https://github.com/sindresorhus/os-locale which additionally takes care of our windows users (which we have, primarly through web-ext).

> I'm still going through this file to understand what we're doing there.

It's full of dark magic, that's for certain.

> I'm still a little unclear as to what was the problem with using cli arguments, but I do get that it was a little messy.

The problem is that all our messages are defined are being defined on a module level with a call to `gettext`, which currently returns the translated message based on the configured locale directly.

That means that all strings in those message-constants are already translated at the time of import.

Now, we ususally configure the cli argument a bit later than that and act on it much later which leads to us needing to change the strings in the message-constants at a much later point in time during the addons-linter process.

There's a techique called laziness which I tried to implement in that gist, which essentially delays the call to `Jed` until the message get's evaluated *again* and not directly during import time. This now menas that we most certainly have the configured locale from the cli argument.

Maybe that makes it a bit more clear?

In any case, none of that is what we want right now since it just adds much more complexity to the system which we don't need.

**Comment by ravneette** [comment]
Yup finally got the gist of it!
Thanks @EnTeQuAk

**Comment by EnTeQuAk** [comment]
> Yup finally got the gist of it!
> Thanks @EnTeQuAk

Sure thing, let me know if you run into any issues and if I can be of any help!

File: .eslintrc [hunk 2]
```diff
@@ -2,7 +2,8 @@
   "rules":{
     "no-console": 0,
     // Inherited from airbnb via eslint-config-amo
-    "function-paren-newline": 0
+    "function-paren-newline": 0,
+    "import/no-dynamic-require": 0
```
EnTeQuAk: Let's maybe not generalize this but make use of a `// eslint-disable-next-line import/no-dynamic-require` comment right before the line you're going to use that dynamic import.

You're already doing that so adding the ignore here shouldn't have any impact.

src/utils.js
```diff
@@ -126,14 +127,23 @@ export function getVariable(context, name) {
...
+  const locale = process.env.LANG || process.env.LANGUAGE || process.env.LC_ALL || process.env.LC_MESSAGES || 'en-US';
+  let i18ndata = {};
+  try {
+    // eslint-disable-next-line global-require, import/no-dynamic-require
+    i18ndata = require(`./locale/${locale}/messages.js`);
+  } catch (err) {
+    log.info('Initialize locales using extract-locales command');
+  }
+  const jed = new Jed(i18ndata);
+  return jed.gettext(str);
```
EnTeQuAk: I actually wonder why you moved this into the gettext function again. We have the language set on a process basis which means that it won't change during the whole life-time of the addons-linter process.

What's happening here now is that *every* `gettext` call re-initializes `Jed` which might actually lead to quite some slow-down. It's not super massive but moving that into our global state will probably speed things up.

src/utils.js
```diff
@@ -126,14 +127,23 @@ export function getVariable(context, name) {
  *
  */
 export function gettext(str) {
-  return str;
+  const locale = process.env.LANG || process.env.LANGUAGE || process.env.LC_ALL || process.env.LC_MESSAGES || 'en-US';
```
EnTeQuAk: Let's move this into a utility function, e.g something called `getLocale` so that it's easier to re-use later if needed.

src/utils.js
```diff
@@ -126,14 +127,23 @@ export function getVariable(context, name) {
...
+  const jed = new Jed(i18ndata);
+  return jed.gettext(str);
 }
 
 /*
  * An sprintf to use with gettext. Imported from Jed for when we have a proper
  * l10n solution.
  */
-export const { sprintf } = jed;
+export const { sprintf } = Jed;
```
EnTeQuAk: This probably won't work (needs a test, more on that later) since it doesn't re-use the initialized `jed` instance from inside `gettext` and will only work if you make the `const jed = Jed(i18ndata);` from earlier global and use `jed.sprintf` here, so this will become `export const sprintf = jed.sprintf;` or so (sorry, untested).

.eslintrc
ravneette: Ya I did commit that. Forgot to push I think

src/utils.js
ravneette: Yes that is right @EnTeQuAk 

But If I make the jed variable global to prevent re initialization, a few tests fail in which the global jed variable comes in as undefined inside the gettext function.

ravneette:
So the reason I shifted everything inside the gettext function was that in case I defined them as global variables, a few tests failed where inside the function
```js
function gettext(str) {
  return jed.gettext(str);
}
```
the jed variable became undefined. 

I think the reason for this might be that the function was some how called directly and global variable was not initialized at all. 

Which is why I made https://github.com/mozilla/addons-linter/pull/1785/commits/e8a6bb889761754a8165bfca629289b6b1517b59 commit to shift everything inside the gettext function. 

I'll probably file for another PR with everything defined globally so we can understand which tests are failing and why.

ravneette:
So after going through a few links, I found the best way of doing this in the following blog post
http://luetkemj.github.io/170421/mocking-modules-in-jest/

This is actually a pretty common issue, mentioned here : https://github.com/facebook/jest/issues/936

I've implemented the same. Now a few tests are failing because the lib variable is not getting recognized inside the geti18nData function.

This is similar to the issue I mentioned in https://github.com/mozilla/addons-linter/pull/1785#issuecomment-368210782

src/utils.js
```diff
@@ -120,20 +122,48 @@ export function getVariable(context, name) {
...
+  return osLocale.sync();
+}
+
+export function getLocaleDir(locale) {
+  return `./locale/${locale}/messages.js`;
+}
+
+export const lib = {
+  getLocaleDir,
+};
```
ravneette: This is introduced because of the approach mentioned here : http://luetkemj.github.io/170421/mocking-modules-in-jest/

EnTeQuAk:
> the jed variable became undefined.
> [...]

I'm debugging this right now and will get back to you as soon as possible. I fear we're running into some weird recursive imports or something like that. Not sure I fully understand what's going on yet but I will soon. Stay tuned! Also looking into these test failures, that "undefinedness" (is that even a word?) bothers me :thinking:

EnTeQuAk:
Alrighty, I played with it a bit in https://github.com/mozilla/addons-linter/commits/jed

* moved all jed/i18n/gettext/sprintf functionality into a `i18n` object
* untangled circlar imports (this is where your undefined errors came from), so moved `ESLINT_OVERWRITE_MESSAGE` out of `const.js` into `messages/javascript.js` so that we don't have an import from `utils` inside `const.js` anymore
* Simplified mocking again in gettext test

This seems to be working nicely, but is a bit more work unfortunately. Let's see if you can rebase / merge that into your branch? That'd be a big step forward.

What do you think?

ravneette:
So I rebased your commits into my branch and brought the PR up to date with the current master. The Pull Request looks real clean thanks to the changes you made.

Finally No `undefined` variables :dancer: 

Looking at the tests that are failing I think these are from files unrelated to the PR. Have we broken something? Looking into this now.

EnTeQuAk:
> Looking at the tests that are failing I think these are from files unrelated to the PR. Have we broken something? Looking into this now.

Yeah, I think this is a temporary schema import failure again, Appveyor works perfectly fine it seems. So don't invest too much time into it :)

EnTeQuAk:
Ah! I accidentally removed the `tmp/.gitkeep` file, can you restore that? `touch tmp/.gitkeep` and `git add` again should restore that. That might help getting the travis tests running, currently there isn't a `tmp` folder and we don't create it explicitly.

ravneette:
Ah! Makes sense.

ravneette:
Restored the file. Let's wait for Travis to finish.

Meanwhile any suggestions to add some other tests for testing i18n @EnTeQuAk ?

PR #1790, opened 2018-01-26
Async/await refactoring tests/rules/js folder

hameleonka:
Refactors test files in rules/js folder.


tests/rules/javascript/test.event_listener_fourth.js
```diff
@@ -90,6 +81,9 @@ describe('event_listener_fourth', () => {
       });
   });
```
hameleonka: Can't see where in the file event-listener-fourth.js we are not allowing window identifier.
 And why all the other  tests are called on window and don't report this warning but here it's not allowed.
 Also why 'should not allow test' doesn't report any warnings.

tests/rules/javascript/test.event_listener_fourth.js
```diff
@@ -101,33 +95,29 @@ describe('event_listener_fourth', () => {
...
```
hameleonka: Can't see where in the file event-listener-fourth.js we are not allowing window identifier.
 And why all the other  tests are called on window and don't report this warning but here it's not allowed.
 Also why 'should not allow test' doesn't report any warnings.

tests/rules/javascript/test.no_eval.js
```diff
@@ -5,6 +5,9 @@ import { DANGEROUS_EVAL } from 'messages';
 // These rules were mostly copied and adapted from
 // https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-eval.js
 // Please make sure to keep them up-to-date and report upstream errors.
+
+// ??? should I renew validCodes and invalidCodes as
+// they are now in the source? There is some difference.
```
hameleonka: Should I renew validCodes and invalidCodes as they are now in the source? There is some difference.

tests/rules/javascript/test.no_implied_eval.js
```diff
@@ -136,30 +135,30 @@ describe('no_implied_eval', () => {
       message: [expectedErrorMessage],
       description: [NO_IMPLIED_EVAL.description],
     },
+
+    // ??? there is also 'gives the correct node when dealing with nesting' in the source
```
hameleonka: There is also code for 'gives the correct node when dealing with nesting' in the source

tests/rules/javascript/test.opendialog_nonlit_uri.js
```diff
@@ -4,18 +4,17 @@ import * as messages from 'messages';
 
 
 describe('opendialog_nonlit_uri', () => {
-  it('should provide notice on non-literal uris passed to openDialog', () => {
+  // ??? why we can't pass non-literal uris?
+  it('should provide notice on non-literal uris passed to openDialog', async () => {
```
hameleonka: Why we can't pass non-literal uris?

tests/rules/javascript/test.opendialog_remote_uri.js
```diff
@@ -4,7 +4,8 @@ import * as messages from 'messages';
 
 
 describe('opendialog_remote_uri', () => {
-  it('should warn on remote uris passed to openDialog', () => {
+  // ??? why we can't use remote uris?
+  it('should warn on remote uris passed to openDialog', async () => {
```
hameleonka: Why we can't use remote uris?

tests/rules/javascript/test.widget_module.js
```diff
@@ -2,7 +2,7 @@ import { VALIDATION_WARNING } from 'const';
 import JavaScriptScanner from 'scanners/javascript';
 import * as messages from 'messages';
 
-
+// ?? where is the rule for this test?
```
hameleonka: Where is the rule for this test?

tests/rules/javascript/test.mozindexeddb_property.js
```diff
@@ -7,34 +7,32 @@ describe('mozindexeddb_property', () => {
...
+    expect(linterMessages[0].code).toEqual(
+      messages.MOZINDEXEDDB_PROPERTY.code
+    );
+    expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
   });
 
+  // ??? the code for Identifier node actually cathes
+  // var foo = "mozIndexedDB" without calling window[foo]
   it('should warn when mozIndexedDB is used as a literal', () => {
     const code = 'var foo = "mozIndexedDB"; var myDatabase = window[foo];';
```
hameleonka: Can't see in mozindexeddb.js how it catches var foo = "mozIndexedDB".

tests/rules/javascript/test.event_listener_fourth.js
```diff
@@ -5,129 +5,109 @@ import { VALIDATION_WARNING } from 'const';
...
       foo.addEventListener("click", function(){}, false, false);`;
     const jsScanner = new JavaScriptScanner(code, 'badcode.js');
 
-    return jsScanner.scan()
-      .then(({ linterMessages }) => {
-        expect(linterMessages.length).toEqual(0);
-      });
+    const { linterMessages } = await jsScanner.scan();
+    expect(linterMessages.length).toEqual(0);
   });
```
hameleonka: The question stays: Can't see where in the file event-listener-fourth.js we are checking and not allowing window identifier.
Also why 'should not allow test' doesn't report any warnings.

tests/rules/javascript/test.event_listener_fourth.js
```diff
@@ -5,129 +5,109 @@ import { VALIDATION_WARNING } from 'const';
...
       foo("click", function(){}, false, false);`;
     const jsScanner = new JavaScriptScanner(code, 'badcode.js');
 
-    return jsScanner.scan()
-      .then(({ linterMessages }) => {
-        expect(linterMessages.length).toEqual(0);
-      });
+    const { linterMessages } = await jsScanner.scan();
+    expect(linterMessages.length).toEqual(0);
   });
```
hameleonka: The question stays: Can't see where in the file event-listener-fourth.js we are checking and not allowing window identifier.
Also why 'should not allow test' doesn't report any warnings.

tests/rules/javascript/test.no_eval.js
```diff
@@ -24,13 +24,11 @@ describe('no_eval', () => {
   ];
```
hameleonka: I've checked source codes with the ones in the file. The additional codes that are present in the source code turned out to be invalid codes in test.no_implied_eval.js so I haven't added them.

EnTeQuAk:
So I looked through the rules and it seems most are indeed still being usable for WebExtensions. Only the `mozIndexedDB` seems to be from the old SDK which could be removed (so `mozindexeddb-property.js` and `mozindexeddb.js`).

@wagnerand can you confirm that suspicion?

EnTeQuAk:
Alrighty, so from our latest IRC discussion: `mozIndexedDB` related rules can be removed but the rest seems to be useful and should stay. If you're up for it, freel free to remove all that code @hameleonka 🎉

hameleonka:
created an issue #1814, will make a separate PR

tests/rules/javascript/test.event_listener_fourth.js
hameleonka: @rpl @EnTeQuAk just wanted to remind that we've talked about this test, that something maybe wrong here, but we still haven't figured it out.

PR #1795, opened 2018-01-30
Update firefox 58 schema import.

EnTeQuAk:


EnTeQuAk:
meh, I forgot about tests. I'll update those tomorrow morning and cherry-pick them into the tag

mstriemer:
I sorted everything out with the 59 import, should be a patch up very soon for that. We might be able to use that instead.

mstriemer:
The 59 schema import is in #1798. I removed any updates to the download part of firefox-schema-import.js so these changes can land instead of the ones I made locally.

EnTeQuAk:
Closing for now, I'll resubmit this with the mozilla-unified part only.

PR #1799, opened 2018-01-31
Initialized the yarn command for extraction

ravneette:
In continuation to https://github.com/mozilla/addons-linter/pull/1739
This PR adds a new yarn command, 
```
yarn extract-locales
```

This is still a work in progress. 
The documentation over at grunt : https://gruntjs.com/configuring-tasks#options
states that we can override the options parameter, but this does not seem to work. 

Am I missing something?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "start": "grunt start",
     "build": "grunt build",
     "eslint": "eslint bin/* .",
+    "extract-locales": "webpack --verbose --bail --display-error-details --progress --colors --config webpack.l10n.config.babel.js",
     "test": "jest --runInBand --watch 'tests/.*'",
     "test-coverage": "jest --runInBand --coverage --watch 'tests/.*'",
     "test-once": "jest --runInBand",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -53,6 +53,7 @@ module.exports = {
   resolve: {
     extensions: ['.js', '.json'],
     modules: [
+      'src',
       'node_modules',
     ],
   },


ravneette:
https://github.com/mozilla/addons-linter/pull/1799/files#diff-d7b33e47ff28e5dfb75c9f895df393ccR25 is where I override the options parameter.

ravneette:
@EnTeQuAk any pointers?

muffinresearch:
Are we using grunt for any specific reason? In general we've been moving away from using grunt.

ravneette:
@muffinresearch I tried doing it the way it is done in addons-frontend (using yarn), but it threw errors which I mentioned here : https://github.com/mozilla/addons-linter/issues/1535#issuecomment-354556534

So I used grunt exactly the way it is used for the other commands.

EnTeQuAk:
> @muffinresearch I tried doing it the way it is done in addons-frontend (using yarn), but it threw errors which I mentioned here : #1535 (comment)

I just commented on that, at least how I solved it recently. It might be worth trying that out again, adding `src` to the resolved modules and fallback to using pure yarn again?

At least that's preferable given that we really want to get rid of grunt at some point.

ravneette:
Yes that did work @EnTeQuAk 
Updated the PR

EnTeQuAk:
Thanks a lot! I'll give this a shot today and see how it goes.

EnTeQuAk:
Works like a charm, I'm going to merge this to extract updated messages before a new release tomorrow. Thanks!

PR #1801, opened 2018-01-31
Rules for messages.json

### LINKED ISSUE #1666: Fail on invalid translations
In bug https://bugzilla.mozilla.org/show_bug.cgi?id=1417103 we found a case where a key in the JSON file had been typo'd: mssage instead of message.

This caused Firefox to fail to load the extension with an unhelpful "Extension is invalid" message. Let's try and catch that sooner in the linter and in AMO and raise a fatal error (not a warning) until Firefox changes.

Here's a diff to sort tabs by url to trigger it:

https://github.com/andymckay/sort-tabs-by-url

```
sort-tabs-by-url gd                                                                                            runtime  ✭ ✱ 
diff --git a/_locales/en/messages.json b/_locales/en/messages.json
index 6ed622a..1c903d1 100644
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -3,6 +3,6 @@
     "message": "Sort tabs by URL"
   },
   "extensionCommand": {
-    "message": "Sort tabs"
+    "mssage": "Sort tabs"
   }
 }
```

freaktechnik:
Fixes #1666

This adds a parser for `messages.json` files. The parser is called from the JSON scanner. It has a JSON schema for the `messages.json` files, which is currently just in the schema's directory, since it is neither imported nor a patch of an existing schema. I'm pretty sure that's not where it should be, though.

It also warns about missing placeholders and when using reserved message names (which are potentially inaccessible).

I think the naming regexp as it is documented is not enforced like this by Firefox, so it may be demoted to a warning, since the extension will still work, even with characters other than `[A-Za-Z0-9_@]` in message names. I don't know about placeholders, not something I have violated myself.

The errors are kind of annoying currently, since you have no position or reference to which message they happen at. I haven't really found precedent for how to handle that though.

And lastly, the documentation links should probably be shortened, so they're actually clickable.


tests/scanners/test.json.js
```diff
@@ -25,4 +26,18 @@ describe('JSONScanner', () => {
         expect(err.message).toEqual('Explode!');
       });
   });
+
+  it('should use special parser for message.json', () => {
```
freaktechnik: *`messages.json`

src/parsers/messagesjson.js
```diff
@@ -0,0 +1,105 @@
...
+      // We've parsed the JSON; now we can validate the manifest.
+      this._validate();
+    }
+  }
+
+  errorLookup(error) {
+    // This is the default message.
+    let baseObject = messages.JSON_INVALID;
+
+    // This is the default from webextension-manifest-schema, but it's not a
```
freaktechnik: Looking through the code I'm not even sure if this is used, as the additional properties are both custom error codes.

src/messages/messagesjson.js
```diff
@@ -0,0 +1,40 @@
...
+  code: 'INVALID_PLACEHOLDER_NAME',
+  message: _('Placeholder name contains invalid characters'),
+  description: _('Placeholder name contains characters that are not allowed (https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/i18n/Locale-Specific_Message_reference#placeholder_name).'),
+};
+
+export const NO_PLACEHOLDER_CONTENT = {
+  code: 'NO_PLACEHOLDER_CONTENT',
+  message: _('Placeholder is missing the content property'),
+  description: _('Placeholders need a content property defining the replacement of it (https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/i18n/Locale-Specific_Message_reference#placeholder_content)'),
+};
```
EnTeQuAk: All these links should be shortened a bit imho to make the message more readable. You can shorten the URL via bit.ly and use that name via the mzl.la domain, e.g https://mzl.la/2E5eGtL

src/parsers/messagesjson.js
EnTeQuAk: I'm not sure I follow, which code do you mean and which additional properties?

src/messages/messagesjson.js
freaktechnik: Ok, I thought you needed special privileges to create mzl.la bitly links.

src/parsers/messagesjson.js
EnTeQuAk: Ah, I think I agree, you're not re-using `error` and `error.message` unless I'm blind. So yes, that tidying isn't needed imho

freaktechnik: The code after this comment, which was copied over from the manifest parser (as the comment shows...)

src/parsers/messagesjson.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return Object.assign({}, baseObject, overrides);
+  }
+
+  hasPlaceholder(message, placeholder) {
+    const messageObj = this.parsedJSON[message];
+    return 'placeholders' in messageObj &&
+           placeholder in messageObj.placeholders;
+  }
```
EnTeQuAk: You should use `hasOwnProperty` here instead of `in` to avoid searching the whole prototype chain.

src/schema/messages.json
```diff
@@ -0,0 +1,51 @@
...
+      "type": "object",
+      "patternProperties": {
+        "^[A-Za-z0-9_@]+$": {
+          "$ref": "#/types/i18nMessage"
+        }
+      },
+      "additionalProperties": false
+    }
+  }
+}
```
EnTeQuAk: Great idea!

cc @mstriemer on this as our "hero of schema" - I also wasn't able to find a schema for these messages files in mozilla central (neither https://hg.mozilla.org/mozilla-central/file/tip/toolkit/components/extensions/schemas/ nor https://hg.mozilla.org/mozilla-central/file/tip/browser/components/extensions/schemas) - maybe this is something we should push to upstream?

src/messages/messagesjson.js
EnTeQuAk: As far as I know they're by default generated for everything that points to a mozilla domain so they're validated to some degree.

src/parsers/messagesjson.js
EnTeQuAk: Ah yeah, as said before - I don't think we need it since `error` isn't re-used if I'm right.

src/schema/messages.json
freaktechnik: As I noted in the pull request, the format of message names is not currently enforced in Firefox, I haven't checked in Chrome, but I assume that's where the limitation comes from. I've checked the source, and Firefox does enforce placeholder names: https://dxr.mozilla.org/mozilla-central/source/toolkit/components/extensions/ExtensionCommon.jsm#1672

It also seems to replace missing placeholders with an empty string, probably something worth noting somewhere, but not necessarily in the linter (but the page it links to)

freaktechnik:
While looking through the messages.json parser of Firefox, I've found out, that message names are always normalized to lower case. I know the JSON parser already checks for duplicate keys, but I assume that is case-sensitive?

src/messages/messagesjson.js
pdehaan: Also, if possible, we should omit the /en-US/ portion of the URL. MDN should serve the user the docs in their best matching language.

freaktechnik:
@EnTeQuAk just so I understand this correctly:
>  There's one thing though I'd like to see. I feel like messagesJSONParser or other messages references can lead to some confusion so I'd like to prefix all that with locale, e.g LocaleMessagesJSONParser etc.
>
> For absolutely no reason it also took some time for me to read "parsers/messagesjson" correctly as my brain was reading "parsers/manifestjson" and went bonkers why we'd re-implement all that 😆

You'd only like to see that for the class names etc. and not for the file names? Else the filenames would end up being something like `parsers/localemessagesjson`...

EnTeQuAk:
> You'd only like to see that for the class names etc. and not for the file names? Else the filenames would end up being something like parsers/localemessagesjson...

Honestly, I'm not so sure. I think the file could easily be `parsers/locale-messagesjson.js` imho or `locale-messages-json.js` even but I have a hard time deciding which I like better

freaktechnik:
This commit adds handling for case-insensitivity, as RJSON only checks for case-sensitive duplicates. I've re-used the `JSON_DUPLICATE_KEY` error code though.

EnTeQuAk:
@freaktechnik this looks great now! There's just one test failing for coverage, so it'd be cool if you have the time to writeup some more tests that test the missing pieces, see https://coveralls.io/builds/15353372/source?filename=src%2Fparsers%2Flocale-messagesjson.js for more details.

freaktechnik:
Let me know if I should squash the commits.

EnTeQuAk:
Thanks for the update @freaktechnik, I just squashed and merged the PR so that I can put together a release that'll go live on AMO on Thursday.

Thank you so much for all that work!

PR #1815, opened 2018-02-08
Removes mozIndexedDB rules and related tests, fixes #1814

### LINKED ISSUE #1814: remove all `mozIndexedDB` related rules
from conversation [here](https://github.com/mozilla/addons-linter/pull/1790#issuecomment-363809708)

hameleonka:
@EnTeQuAk please check this one, but if it possible please merge only after async/await, I'm afraid there will be some conflicts


rpl:
@hameleonka @EnTeQuAk it seems that removing both the rule implementation code and the related tests and localized strings for the deprecated rule in the same pull requests makes particularly easy to trigger a conflict with the mozilla-pontoon commits, also the high number of locale files in the same pull request with code changes makes harder to read and review the pull requests, I'm wondering if it would be better to only remove the deprecated rule implementation code and its tests in this PR and clean up the locales in a second ones.

How that sounds to you?

EnTeQuAk:
Yeah, let's not remove the locales just now. We'll update them in a separate step together with a general extraction step where new messages are also added and old ones removed etc.

EnTeQuAk:
There's still one ` locale/sq/LC_MESSAGES/messages.po` file in the PR, otherwise this looks good to me.

Issue #1789, opened 2018-01-25
Fluent Linter should use fluent-syntax, not fluent


zbraniecki:
Fluent linter seems to be using `fluent` package according to:
 - https://github.com/mozilla/addons-linter/blob/master/src/parsers/fluent.js
 - https://github.com/mozilla/addons-linter/blob/master/package.json#L96

The `fluent` package is a runtime Fluent implementation meant for performance, not conformance. We have another package, called `fluent-syntax` that provides a better parser, with better errors, full AST and a serializer - all of that intended for tools like linters.

EnTeQuAk:
Oh, that's a great hint, thanks a lot! That should be ported to use the
Syntax package then.

On Thu, Jan 25, 2018, at 7:23 PM, Zibi Braniecki wrote:
> Fluent linter seems to be using fluent package according to:


>  * https://github.com/mozilla/addons-linter/blob/master/src/parsers/fluent.js
>  * https://github.com/mozilla/addons-linter/blob/master/package.json#L96> The fluent package is a runtime Fluent implementation meant for
> performance, not conformance. We have another package, called fluent-
> syntax that provides a better parser, with better errors, full AST and
> a serializer - all of that intended for tools like linters.> — You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub[1], or mute the
> thread[2].> 

Links:

  1. https://github.com/mozilla/addons-linter/issues/1789

wagnerand:
Since `fluent-syntax` requires at least node 8.9.0, we can't migrate to it before we drop support for node 6.

EnTeQuAk:
> Since fluent-syntax requires at least node 8.9.0, we can't migrate to it before we drop support for node 6.

it does too? That's sad :( Well, then we'll have to wait for a bit I suppose.

Callek:
This issue was discovered as a blocker for the work happening in https://bugzil.la/1438246 and per IRC discussion there may be a path forward here.

EnTeQuAk:
Steps forward:

* [ ] Add an example implementation that uses fluent-syntax for fluent parsing and error extraction
* [ ] Try out their compat build for node 6 compatibility

Since this seems to be a blocking part of  https://bugzil.la/1438246 I pushed the priority and started working on it.

EnTeQuAk:
Added an upstream [patch](https://github.com/projectfluent/fluent.js/pull/164) to relax their engines configuration a bit, the compat build works fine with node6. 

[...]

PR #1826, opened 2018-02-12
fix(package): update fluent to version 0.6.3

EnTeQuAk:
Closes #1800

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "esprima": "3.1.3",
     "extract-text-webpack-plugin": "3.0.2",
     "first-chunk-stream": "2.0.0",
-    "fluent": "0.4.1",
+    "fluent": "0.6.3",
     "glob": "7.1.2",
     "jed": "1.1.1",
     "pino": "4.10.4",


wagnerand:
@EnTeQuAk Looks like we can't upgrade to fluent 0.6 for now since it requires at least node 8.9.

EnTeQuAk:
> @EnTeQuAk Looks like we can't upgrade to fluent 0.4.2 or higher for now since it requires at least node 8.9 (projectfluent/fluent.js#95).

Indeed, but according to https://github.com/mozilla/addons-linter/issues/1789 we might want to use a different package long-term anyway. So it might make sense to look into that at some point.

wagnerand:
Ok, but that same PR drops node 6 support for `fluent-syntax` as well. So, as long as we are still supporting node 6, I'm not sure it's worth the effort to migrate from one outdated package version to another.

EnTeQuAk:
> Ok, but that same PR drops node 6 support for fluent-syntax as well. So, as long as we are still supporting node 6, I'm not sure it's worth the effort to migrate from one outdated package version to another.

Jupp, I don't think it is. :panda_face: We might drop node6 support once it's maintenance LTS window starts in April in coordination with what web-ext does but definitely not before that.

PR #1842, opened 2018-02-15
Async/await refactoring test.firefox-schemas-import.js and test.utils.js fixes #1536

### LINKED ISSUE #1536: [tracker] Convert code-base to async/await
The code-base is heavily promise based. This issue is to convert all the promise-based code to instead use async/await.

This is a large task and will require being broken down into smaller pieces.

* [ ] Check and update babel transpilation steps to be async/await aware
* [ ] Convert src code to use await/async (leaving the tests as promise based initially)
* [ ] Once the src is converted and the existing test suite passes modify the tests to use async/awit too.

hameleonka:
@rpl this should be the final PR in async/await series

### DIFF
--- a/tests/schema/test.firefox-schemas-import.js
+++ b/tests/schema/test.firefox-schemas-import.js
@@ -1081,15 +1081,13 @@ describe('firefox schema import', () => {
       removeDir(outputPath);
     });
 
-    it('rejects if there is no inputPath or version', () => {
-      return fetchSchemas({}).then(
-        () => expect(false).toBeTruthy(),
-        (err) => expect(err.message).toEqual(
-          'inputPath or version is required'
-        ));
+    it('rejects if there is no inputPath or version', async () => {
+      await expect(
+        fetchSchemas({})
+      ).rejects.toThrow('inputPath or version is required');
     });
 
-    it('downloads the firefox source and extracts the schemas', () => {
+    it('downloads the firefox source and extracts the schemas', async () => {
       const cwd = 'tests/schema';
       const schemaPath = 'firefox';
       const tarball = tar.create({ cwd, gzip: true }, [schemaPath]);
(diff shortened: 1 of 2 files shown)


tests/schema/test.firefox-schemas-import.js
```diff
@@ -1152,15 +1146,12 @@ describe('firefox schema import', () => {
...
       expect(fs.readdirSync(outputPath)).toEqual([]);
-      return fetchSchemas({ inputPath: 'mozilla-central.tgz', outputPath })
-        .then(() => {
-          expect(true).toBeFalsy();
-        }, () => {
-          expect(true).toBeTruthy();
-        });
+      await expect(
+        fetchSchemas({ inputPath: 'mozilla-central.tgz', outputPath })
+      ).rejects.toThrow();
```
rpl: I'm wondering if we could also assert something about what error is expected to be raised here.

@EnTeQuAk what do you think? is it worth it to make this assertion more detailed and also verify the expected error message?

tests/schema/test.firefox-schemas-import.js
```diff
@@ -1171,17 +1162,14 @@ describe('firefox schema import', () => {
...
-        .then(() => {
-          expect(true).toBeFalsy();
-        }, () => {
-          // Manually remove the tar file since it doesn't get cleaned up.
-          fs.unlinkSync('tmp/FIREFOX_AURORA_54_BASE.tar.gz');
-          expect(true).toBeTruthy();
-        });
+      await expect(
+        fetchSchemas({ version: 54, outputPath })
+      ).rejects.toThrow();
```
rpl: Same here (if we opt to make the assertions of the previous testcase to also check the error message).

tests/schema/test.firefox-schemas-import.js
```diff
@@ -1171,17 +1162,14 @@ describe('firefox schema import', () => {
...
+      // Manually remove the tar file since it doesn't get cleaned up.
+      fs.unlinkSync('tmp/FIREFOX_AURORA_54_BASE.tar.gz');
```
rpl: I'm wondering if it would be better to move the unlinking of the temporary files created in this test file in an "teardown" step (https://facebook.github.io/jest/docs/en/setup-teardown.html).

@EnTeQuAk how that sounds to you?

tests/schema/test.firefox-schemas-import.js
```diff
@@ -1202,12 +1190,9 @@ describe('firefox schema import', () => {
...
       expect(fs.readdirSync(outputPath)).toEqual([]);
-      return fetchSchemas({ version: 54, outputPath })
-        .then(() => {
-          expect(true).toBeFalsy();
-        }, () => {
-          expect(true).toBeTruthy();
-        });
+      await expect(
+        fetchSchemas({ version: 54, outputPath })
+      ).rejects.toThrow();
```
rpl: Same here (if we opt to make the assertions of the previous testcases to also check the error message).

EnTeQuAk:
:tada: :man_dancing: :star2: :fireworks:

Issue #1343, opened 2017-06-12
followup on new jest based test infrastructure


EnTeQuAk:
I just merged the first part to run our tests with Jest. Lets do a few cleanups:

* test.cli-process should be some kind of integration test
* ~~use sinon sandbox similar to what addons-frontend does (https://github.com/muffinresearch/addons-frontend/blob/bdcb7858594d5c09294f468e8c733151e1a4a8e4/tests/setup.js#L64-L74)~~ - https://github.com/mozilla/addons-linter/pull/1369
* check for docs in need of an update
* https://github.com/mozilla/addons-linter/pull/1334#discussion-diff-121077807R132
* unify `.toBe` and `.toEqual` tests
* check and make sure that we don't have any tests that are needlessly `toBeTruthy` or `toBeFalsey`

muffinresearch:
> check and make sure that we don't have any tests that are needlessly toBeTruthy or toBeFalsey

I would go as far as to suggest we avoid using those expect matchers entirely. We don't allow falsey/truthy comparisons in code via linting so I think it's best to avoid them in tests too.

rpl:
Follows some additional notes related to the creation of a new integration testsuite for the addons-linter.

## Goals

1. integration tests on the addons-linter bin scripts:

   - a minimal set of smoke tests on the addons-linter bin script running as a whole and without any mocks, to ensure that everything is working as expected once the project sources are running using its real npm dependencies
   - test on the cli interface (e.g. porting the test.cli-process from the unit tests to the integration tests)
   - test the firefox-schema-import script
   - test the build-locales / create-locales / merge-locales scripts
   
2. running the addons-linter on some real extension:

   - run the addons-linter on a small set of real extensions submitted on AMO and/or on some of the example extensions from the mnd/webextensions-examples repo, to ensure that the addons-linter behavior is consistent across its new releases (and catch some of the new potential regressions that can be introduced as a side-effect of adding new rules to the addons-linter)
   
## Implementation notes

- the new integration test suite can still make use of jest as the test runner and use its test assertions
  - Link to some existent projects which uses jest in its integration tests:
    - The jest project itself uses jest to run its integration test suite: https://github.com/facebook/jest/blob/23eec748db0de7b6b5fcda28cc51c48ddae16545/integration-tests/__tests__/cli-handles-exact-filenames.test.js#L24-L41      
[...]

PR #1867, opened 2018-02-26
adds integration tests example #1343

hameleonka:
@rpl please check this out, it's integration tests setup and first test case

### DIFF
--- a/tests/integration/runLinter.js
+++ b/tests/integration/runLinter.js
@@ -1,0 +1,22 @@
+import path from 'path';
+import { exec } from 'child_process';
+
+
+export const BIN_PATH = path.join(__dirname, '../../bin/');
+export function runLinter(args = [], options = {}) {
+  return new Promise((resolve) => {
+    options.env = options.env || {}; // eslint-disable-line no-param-reassign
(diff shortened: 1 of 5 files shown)


tests/integration/runLinter.js
```diff
@@ -0,0 +1,22 @@
+import path from 'path';
+import { exec } from 'child_process';
+
+
+export const BIN_PATH = path.join(__dirname, '../../bin/');
+export function runLinter(args = [], options = {}) {
```
rpl: it would be better to make this test helper a bit more generic, just enough to be able to use it run also the other bin scripts (so that we can use the integration test suite also for the bin scripts that are used to update the locales and the Firefox API schema files), and then also rename the module name (e.g. maybe it could be named `helpers.js` instead of `runLinter.js`, like the one used by the unit tests).

wagnerand:
Hey @hameleonka, are you still working on this PR?

EnTeQuAk:
Closing this PR in favor of https://github.com/mozilla/addons-linter/pull/1939 where @rpl is following up on all the work done in this PR

PR #1873, opened 2018-02-28
fixes relative path in content script rule

### LINKED ISSUE #1803: Error when attempting to find content scripts
Recently I was trying to upload a new version of my add-on but I kept getting the error "Content script file could not be found.". However, the scripts still run how and when they're supposed to and don't get any errors while debugging with the browser. So Inside my main folder I have one main JavaScript file and several which are executed by the main script. 
- hackbar.js (main file which calls the other js files and handles all of the events)
- polyglot.js, post.js, redirect.js and custom.js (These are all executed using ``browser.tabs.executeScript``)

Just a quick note so it makes more sense, all of these files are in the same folder. 
Calling one of the scripts like ``var auto = browser.tabs.executeScript({file: 'polyglot.js'});`` works fine, however when ran like this I get prompted when uploading that says ``Content script file could not be found.`` (but when testing on my machine and another it works fine). I can get the error to go away by adding "hackbar/" to the front of it (e.g. ``var auto = browser.tabs.executeScript({file: 'hackbar/polyglot.js'});``) but adding this breaks the script and the script is no longer executed and I'm presented with ``Error: An unexpected error occurred`` in debugging.

### What happened?
Add-on linter cannot detect my script files

I expected it to upload and pass normally as it has in the previous versions
Here is a copy of the add-on that _should_ be passing but is receiving the ``Content script file could not be found.`` error.
[hackbar_error.zip](https://github.com/mozilla/addons-linter/files/1692529/hackbar_error.zip)
Here is the copy of the add-on that _is passing_ but the scripts are no longer executed
[hackbar_no_error.zip](https://github.com/mozilla/addons-linter/files/1692531/hackbar_no_error.zip)

pardeepshokeen:
Fixes #1803 
      PRs open for the same issue.
* [ ] Add a description of the the changes introduced in this PR.

### DIFF
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -1,8 +1,11 @@
-import { isBrowserNamespace, normalizePath } from 'utils';
+import * as path from 'path';
+
+import { isBrowserNamespace } from 'utils';
 import { CONTENT_SCRIPT_NOT_FOUND, CONTENT_SCRIPT_EMPTY } from 'messages/javascript';
 
 export default {
   create(context) {
+    const dirname = path.dirname(context.getFilename());
     const existingFiles = context.settings.existingFiles || {};
     return {
       MemberExpression(node) {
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
Thanks a lot @pardeepshokeen this looks great! Do you have the time to look into writing a test for your code changes? That'll help test and understand these changes later.

pardeepshokeen:
hi @EnTeQuAk . Since most of the test cases are already written in tests, I have added tests when file path is relative to the current file.

PR #1883, opened 2018-03-06
Add support for static theme linting

### LINKED ISSUE #1587: Support Static Themes in the linter
See mozilla/addons/issues/501 and connected issues for background.
We need the linter to know about Static Themes.  There are a few mandatory fields:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/theme

As just having `theme:` in your manifest makes it suddenly not an extension but a theme we might want to warn about having other stuff in the manifest/zip that is now ignored too.

EnTeQuAk:
* Adds initial support to lint and detect static themes
* Adds additional property validation on top of regular schema validation

Fixes #1587

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -111,28 +111,13 @@ export default class ManifestJSONParser extends JSONParser {
     // that are just warnings should be added to this array.
     const warnings = [messages.MANIFEST_PERMISSIONS.code];
     const isStaticTheme = this.parsedJSON.hasOwnProperty('theme');
-    let packageType = 'addon';
+    let validate = validateAddon;
 
-    const manifestValidators = {
-      static_theme: [
-        isStaticTheme,
-        validateStaticTheme],
-      langpack: [
-        this.isLanguagePack,
-        validateLangPack],
-      addon: [
-        !isStaticTheme && !this.isLanguagePack,
-        validateAddon],
-    };
-
-    Object.keys(manifestValidators).forEach((id) => {
-      if (manifestValidators[id][0]) {
-        packageType = id;
-      }
-    })
-
-
-    const validate = manifestValidators[packageType][1];
+    if (isStaticTheme) {
+      validate = validateStaticTheme;
+    } else if (this.isLanguagePack) {
+      validate = validateLangPack;
+    }
 
     this.isValid = validate(this.parsedJSON);
     if (!this.isValid) {
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -7,7 +7,9 @@ import { oneLine } from 'common-tags';
 import probeImageSize from 'probe-image-size';
 import upath from 'upath';
 
-import { validateAddon, validateLangPack } from 'schema/validator';
+import {
+  validateAddon, validateLangPack, validateStaticTheme
+} from 'schema/validator';
 import { getConfig } from 'cli';
 import { MANIFEST_JSON, PACKAGE_EXTENSION, CSP_KEYWORD_RE, IMAGE_FILE_EXTENSIONS, LOCALES_DIRECTORY, MESSAGES_JSON } from 'const';
 import log from 'logger';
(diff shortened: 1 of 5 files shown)


EnTeQuAk:
@mstriemer can I get a second look regarding the schema updates maybe?

`additionalProperties` should probably be included in an upstream patch. If no one disagrees I'm going to file a bug and try to submit a patch for it. I'd assume that the same can be said for the `WebExtensionLangpackManifest` type which allows additional properties in the schema too but I think that's not a great idea.

src/parsers/manifestjson.js
```diff
@@ -108,7 +110,14 @@ export default class ManifestJSONParser extends JSONParser {
     // Not all messages returned by the schema are fatal to Firefox, messages
     // that are just warnings should be added to this array.
     const warnings = [messages.MANIFEST_PERMISSIONS.code];
-    const validate = this.isLanguagePack ? validateLangPack : validateAddon;
+    const isStaticTheme = Object.prototype.hasOwnProperty.call(this.parsedJSON, 'theme');
```
eviljeff: I don't know about the code style in addons-linter... but if this was addons-server I'd be asking should this be set as a  property on the object, like `isLanguagePack` (and maybe in the same place, rather than in `_validate`)?

tests/helpers.js
```diff
@@ -104,6 +104,24 @@ export function validLangpackManifestJSON(extra) {
...
+    manifest_version: 2,
+    name: 'My Static Theme',
+    version: '1.0',
+    theme: {
+      images: {
+        headerURL: 'weta.png',
+      },
+      colors: {
+        accentcolor: '#adb09f',
+        textcolor: '#000',
```
eviljeff: Maybe add some extra `xxxcolor` to test the other styles of colour?  Any valid css colour property should work - in particular you need `rgb(r g, b, a)` style to do partial transparency.

tests/helpers.js
```diff
@@ -104,6 +104,24 @@ export function validLangpackManifestJSON(extra) {
...
 
+
+export function validStaticThemeManifestJSON(extra) {
+  return JSON.stringify(Object.assign({}, {
+    manifest_version: 2,
+    name: 'My Static Theme',
+    version: '1.0',
+    theme: {
+      images: {
+        headerURL: 'weta.png',
```
eviljeff: Maybe out of scope but there are Chrome properties as well (see docs/schema) which are different but we're supposed to support (addons-server doesn't right now, tbh)

EnTeQuAk: Yeah, I think that's out of scope. We are linting whatever the schema describes so I think that should be fine.

tests/helpers.js
EnTeQuAk: I'll add some more tests.

src/parsers/manifestjson.js
EnTeQuAk: Hmm, I'm torn here, but might be a good idea.

tests/helpers.js
rpl: :+1: I also agree

src/schema/validator.js
```diff
@@ -69,17 +72,47 @@ const _validateLangPack = validator.compile({
...
   $ref: '#/types/WebExtensionLangpackManifest',
 });
+
 export const validateLangPack = (...args) => {
   const isValid = _validateLangPack(...args);
   validateLangPack.errors = filterErrors(_validateLangPack.errors);
   return isValid;
 };
 
+
```
rpl: Nit, we could probably remove one of these two empty lines here.

src/schema/validator.js
```diff
@@ -69,17 +72,47 @@ const _validateLangPack = validator.compile({
...
+const _validateStaticTheme = validator.compile({
+  ...merge(
```
rpl: Nit, An inline comment that explains "what we are doing here and why" could be helpful (e.g. why we merge these object and why we make a new object with a copy of the properties of the resulting object here).

tests/helpers.js
```diff
@@ -104,6 +104,24 @@ export function validLangpackManifestJSON(extra) {
   }, extra));
 }
 
+
```
rpl: Nit, we could probably remove one of these two empty lines.

mstriemer:
It looks like there is already an `additionalProperties` line in the schema [1] but since Firefox only warns on additional properties the schema importer is likely removing it. I'm not sure what the right thing to do for that here is. Manually adding it in might be the easiest.

[1] https://searchfox.org/mozilla-central/rev/99df6dad057b94c2ac4d8b12f7b0b3a7fdf25d04/toolkit/components/extensions/schemas/theme.json#535

PR #1888, opened 2018-03-08
Refactor ftl parsing to use fluent-syntax/compat.

### LINKED ISSUE #1789: Fluent Linter should use fluent-syntax, not fluent

EnTeQuAk:
* Also fixes package.json to use and define correct webpack dependencies
* Updates yarn.lock

~~Do not merge yet, blocked by https://github.com/projectfluent/fluent.js/pull/164~~

Fixes #1789

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "eslint-plugin-no-unsafe-innerhtml": "1.0.16",
     "esprima": "3.1.3",
     "first-chunk-stream": "2.0.0",
-    "fluent": "0.4.1",
+    "fluent-syntax": "^0.6.5",
     "glob": "7.1.2",
     "jed": "1.1.1",
     "os-locale": "2.1.0",
--- a/src/parsers/fluent.js
+++ b/src/parsers/fluent.js
@@ -1,4 +1,5 @@
-import { _parse as parseFluent } from 'fluent';
+// Use compat build for node6 compatibility
+import { FluentParser as FluentSyntaxParser } from 'fluent-syntax/compat';
 
 import * as messages from 'messages';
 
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -7,6 +7,8 @@ script:
   - yarn run test-ci
   - yarn run build
   - yarn run lint
+before_install:
+  - yarn config set ignore-engines true
```
EnTeQuAk: Temporary until https://github.com/projectfluent/fluent.js/pull/164 get's merged

stasm: Isn't this actually a better temporary solution to this problem than projectfluent/fluent.js#164? Assuming the longer term solution is to drop the support for Node 6 in `addons-linter`.

src/parsers/fluent.js
```diff
@@ -18,30 +19,47 @@ export default class FluentParser {
     this.isValid = null;
   }
 
+  getLineAndColumnFromSpan(span) {
```
stasm: This looks like a job for `lineOffset` and `columnOffset` exported by `fluent-syntax` :)

https://github.com/projectfluent/fluent.js/blob/fee9cbf347c4edecaaf60f757da6c515110f2356/fluent-syntax/src/index.js#L17-L36

.travis.yml
EnTeQuAk: No, it's not since the linter is being installed by users via web-ext - https://github.com/mozilla/web-ext and a strict engine in `package.json` doesn't allow the user to install the linter via npm or yarn.

src/parsers/fluent.js
EnTeQuAk: Ah, I didn't see those helpers yet. The code is quite similar, I'll use the upstream version then, thanks for the hint!

package.json
```diff
@@ -79,7 +79,8 @@
     "request": "2.83.0",
     "sinon": "4.4.2",
     "tar": "4.4.0",
-    "webpack": "3.10.0",
+    "webpack": "4.0.1",
```
LinusU: It seems like it was actually `webpack` 3.x that was used before, `webpack-grunt` doesn't have webpack 4.x release supported yet (it's merged into master though so should only be a matter of time 🎉)

I also addressed this in #1889, but maybe this has all the changes required to go to webpack 4.x?

EnTeQuAk: Just merged or going to merge the removal of `webpack-grunt` dependency so we'll be able to update to webpack 4.x soon. But step by step…

I'll back out the package.json changes here anyway.

EnTeQuAk:
Rebased to master, removed earlier hacks, rebased to fluent-syntax 0.6.5 which adds node6 support.

tests/parsers/test.fluent.js
```diff
@@ -20,33 +20,6 @@ choose-download-folder-title =
...
-                  name: 'accusative',
-                  type: 'sym',
-                },
-                val: 'Foo2',
-              },
-            ],
-          },
-        ],
-      },
-    });
```
EnTeQuAk: I primarily removed these since I essentially only copy the upstream tree into `parsedData` and that tree is huge and would bloat up our tests. There is one test that verifies things at a very basic level so I thought that has to suffice.

src/parsers/fluent.js
```diff
@@ -19,29 +23,33 @@ export default class FluentParser {
...
+    const parser = new FluentSyntaxParser();
+    const resource = parser.parse(this._sourceString);
+
     this.parsedData = {};
 
-    Object.keys(entries).forEach((id) => {
-      this.parsedData[id] = entries[id];
-    });
+    resource.body.forEach((entry) => {
+      if (entry.type === 'Junk') {
```
tofumatt: If `fluent-syntax` provides a constant for this it'd be nice, magic constants just make me scared.

tests/parsers/test.fluent.js
```diff
@@ -65,29 +38,26 @@ key69
...
-      },
-      key69: {
-        attrs: {
-          accesskey: 'R',
-          label: 'Reconnect to &syncBrand.shortName.label;…',
-        },
-        val: undefined,
-      },
-    });
+    expect(parser.parsedData.key67.attributes[0].value.elements[0].value).toEqual(
```
tofumatt: These tests read foreign to me without context but I have no idea how fluent works... I'm gonna trust you on them 😄

src/parsers/fluent.js
EnTeQuAk: Unfortunately not, I might push a follow-up PR to clean this up a bit and think about better tests.

EnTeQuAk:
I'm going to merge this, if there's any more comments I'm very happy to update the implementation but to land this this week I'm merging and releasing a new version of the linter for now.

It's not breaking the tests and actually fixing something so it can't be too bad :see_no_evil:

src/parsers/fluent.js
zbraniecki: It provides a type you can test against `instanceof` - https://github.com/projectfluent/fluent.js/blob/master/fluent-syntax/src/ast.js#L240

rpl: :+1: on importing the `Junk` class and turning this into `entry instanceof Junk` in a follow up as @zbraniecki pointed out.

PR #1890, opened 2018-03-08
Fix webpack build by inlining the dynamic require call.

EnTeQuAk:
This also removes more grunt dependencies and starts building directly
with webpack.


package.json
```diff
@@ -7,8 +7,7 @@
     "addons-linter": "bin/addons-linter"
   },
   "scripts": {
-    "start": "grunt start",
```
EnTeQuAk: Given that we use `jest` these days this isn't really needed anymore. At least, I haven't used it since I joined the linter project so I don't see why it should stay.

LinusU: Seems like this command currently just does `webpack --watch ...`, or am I missing anything? Could probably be replaced with `npm run build -- --watch`...

LinusU:
Nice 👍

package.json
EnTeQuAk: Kinda, but even that isn't really needed since ages now. We use jest which transpiles / builds the modules dynamically for us.

.travis.yml
```diff
@@ -5,6 +5,7 @@ node_js:
 - '8'
 script:
   - yarn run test-ci
+  - node bin/build-locales
```
rpl: It would be nice if we could make "build-locales" an implicit part of "npm run build", e.g. something like the following should work on both linux and osx but I'm not sure that windows is going to like it:

```diff
diff --git a/package.json b/package.json
index b90703d..e301799 100644
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "addons-linter": "bin/addons-linter"
   },
   "scripts": {
-    "build": "webpack --verbose --bail --display-error-details --progress --colors --config webpack.config.js",
+    "build": "npm run build-locales && npm run build-src",
+    "build-locales": "node ./bin/build-locales",
+    "build-src": "webpack --verbose --bail --display-error-details --progress --colors --config webpack.config.js",
     "eslint": "eslint bin/* .",
     "extract-locales": "webpack --verbose --bail --display-error-details --progress --colors --config webpack.l10n.config.babel.js",
     "test": "jest --runInBand --watch 'tests/.*'",
``` 

anyway, I think that we can defer it for a follow up (to better evaluate the approach and possibly one that would work also on windows), in the meantime fixing the locales loading with this PR like it is now.

In the meantime, how about adding a note about the new "build-locales" build step? (e.g. in the README.md, https://github.com/mozilla/addons-linter/blob/master/README.md#development, and/or in the CONTRIBUTING.md, https://github.com/mozilla/addons-linter/blob/master/CONTRIBUTING.md#build-addons-linter)

EnTeQuAk: > anyway, I think that we can defer it for a follow up (to better evaluate the approach and possibly one that would work also on windows), in the meantime fixing the locales loading with this PR like it is now.

In the meantime, how about adding a note about the new "build-locales" build step? (e.g. in the README.md, https://github.com/mozilla/addons-linter/blob/master/README.md#development, and/or in the CONTRIBUTING.md, https://github.com/mozilla/addons-linter/blob/master/CONTRIBUTING.md#build-addons-linter)

Jupp, I'll file an issue about it

PR #1919, opened 2018-03-19
Remove explicit property linting from messages.json parsing.

### LINKED ISSUE #1918: Messages validator is too strict excluding dot form keys

Extension with messages like:
```
	"options.Parameter.Name": {
		"message": "Nazwa"
	},
```

### What happened?
The linter is complaining about the message.

It was fine last time I submitted my extension. It works fine in Firefox. It validates as JSON. You can find similar keys in e.g. VSCode settings... I see no reason why "." would be an invalid character.

EnTeQuAk:
Fixes #1918

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -162,7 +162,9 @@ export const MESSAGES_JSON = 'messages.json';
 export const LOCALES_DIRECTORY = '_locales';
 
 // This is a string, since it has to be matched globally on a message string.
-export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@\\.]+)\\$';
+// This should match
+// https://searchfox.org/mozilla-central/rev/3abf6fa7e2a6d9a7bfb88796141b0f012e68c2db/toolkit/components/extensions/ExtensionCommon.jsm#1711
+export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-zA-Z0-9_@]+)\\$';
 
 // yauzl should trow error with this message in case of corrupt zip file
 export const ZIP_LIB_CORRUPT_FILE_ERROR = 'end of central directory record signature not found';
--- a/src/const.js
+++ b/src/const.js
@@ -162,7 +162,7 @@ export const MESSAGES_JSON = 'messages.json';
 export const LOCALES_DIRECTORY = '_locales';
 
 // This is a string, since it has to be matched globally on a message string.
-export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@]+)\\$';
+export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@.]+)\\$';
 
 // yauzl should trow error with this message in case of corrupt zip file
 export const ZIP_LIB_CORRUPT_FILE_ERROR = 'end of central directory record signature not found';
--- a/src/const.js
+++ b/src/const.js
@@ -162,7 +162,7 @@ export const MESSAGES_JSON = 'messages.json';
 export const LOCALES_DIRECTORY = '_locales';
 
 // This is a string, since it has to be matched globally on a message string.
-export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@.]+)\\$';
+export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@\\.]+)\\$';
 
 // yauzl should trow error with this message in case of corrupt zip file
 export const ZIP_LIB_CORRUPT_FILE_ERROR = 'end of central directory record signature not found';
--- a/src/schema/messages.json
+++ b/src/schema/messages.json
@@ -25,13 +25,7 @@
           "type": "string"
         },
         "placeholders": {
-          "type": "object",
-          "patternProperties": {
-            "^[A-Za-z0-9_@]+$": {
-              "$ref": "#/types/i18nPlaceholder"
-            }
-          },
-          "additionalProperties": false
+          "type": "object"
         }
       },
       "required": [
@@ -39,13 +33,7 @@
       ]
     },
     "WebExtensionMessages": {
-      "type": "object",
-      "patternProperties": {
-        "^[A-Za-z0-9_@]+$": {
-          "$ref": "#/types/i18nMessage"
-        }
-      },
-      "additionalProperties": false
+      "type": "object"
     }
   }
 }
--- a/src/schema/messages.json
+++ b/src/schema/messages.json
@@ -41,7 +41,7 @@
     "WebExtensionMessages": {
       "type": "object",
       "patternProperties": {
-        "^.+$": {
+        "^.*$": {
           "$ref": "#/types/i18nMessage"
         }
       },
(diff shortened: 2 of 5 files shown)


src/const.js
```diff
@@ -162,7 +162,7 @@ export const MESSAGES_JSON = 'messages.json';
 export const LOCALES_DIRECTORY = '_locales';
 
 // This is a string, since it has to be matched globally on a message string.
-export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@]+)\\$';
+export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@\\.]+)\\$';
```
rpl: I don't think that we should change this regular expression (as well as the one in  src/schema/messages.json at line 30), because the placeholder are not validated by Firefox when the extension is installed (and so it will not raise an error), but a similar regular expression is used to apply the placeholder here:

https://searchfox.org/mozilla-central/rev/3abf6fa7e2a6d9a7bfb88796141b0f012e68c2db/toolkit/components/extensions/ExtensionCommon.jsm#1711

and so the placeholders that don't match that regexp will be just ignored.

Nevertheless, it seems to me that we should add `A-Z` to the ranges of characters matched by this regular expression (the placeholder matches besides its "case", then the matched placeholder is normalized to lowercase and retrieved from the map of the placeholders, also is also normalized to lowercase).

src/schema/messages.json
```diff
@@ -41,7 +41,7 @@
     "WebExtensionMessages": {
       "type": "object",
       "patternProperties": {
-        "^[A-Za-z0-9_@]+$": {
+        "^.+$": {
```
rpl: :+1: this doesn't seem to have any particular check in Firefox and it looks like it is the exactly the property that #1918 would like to be free from restrictions applied by linter that are not actually applied on Firefox.

tests/parsers/test.locale-messagesjson.js
```diff
@@ -92,22 +92,20 @@ describe('LocaleMessagesJSONParser', () => {
     expect(warnings[1].message).toEqual(messages.MISSING_PLACEHOLDER.message);
   });
 
-  it('should be invalid if bad message name', () => {
+  it('should not be an error if bad message name', () => {
```
rpl: I'm a bit unsure about the "if bad messagename" from this test description, how about "should not be invalid on non alphanumeric message name"?

tests/parsers/test.locale-messagesjson.js
```diff
@@ -92,22 +92,20 @@ describe('LocaleMessagesJSONParser', () => {
...
+    expect(localeMessagesJSONParser.isValid).toEqual(true);
     const { errors } = addonsLinter.collector;
-    expect(errors.length).toEqual(1);
-    expect(errors[0].code).toEqual(messages.INVALID_MESSAGE_NAME.code);
-    expect(errors[0].message).toEqual(messages.INVALID_MESSAGE_NAME.message);
+    expect(errors.length).toEqual(0);
   });
 
-  it('should be invalid if bad placeholder name', () => {
+  it('should not be an error if bad placeholder name', () => {
```
rpl: As I wrote in the first message, I think that we should still raise validation errors for the invalid placeholder names, which would be ignored by Firefox and produce unexpected behaviors at runtime (well, we expect that the invalid placeholder is going to be ignored, but an extension developer that is not aware of this rule will be confused by this behavior and Firefox will not log any warning about it).

tests/parsers/test.locale-messagesjson.js
```diff
@@ -120,11 +118,27 @@ describe('LocaleMessagesJSONParser', () => {
...
-    expect(localeMessagesJSONParser.isValid).toEqual(false);
+    expect(localeMessagesJSONParser.isValid).toEqual(true);
     const { errors } = addonsLinter.collector;
-    expect(errors.length).toEqual(1);
-    expect(errors[0].code).toEqual(messages.INVALID_PLACEHOLDER_NAME.code);
-    expect(errors[0].message).toEqual(messages.INVALID_PLACEHOLDER_NAME.message);
+    expect(errors.length).toEqual(0);
+  });
+
+  it('should support dot in placeholder name', () => {
```
rpl: The dot is not a valid character in the placeholder names and it will be ignored, and so this additional test should just be removed.

asamuzaK:
For a note:
[Locale-specific message reference - Mozilla | MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/i18n/Locale-Specific_Message_reference "Locale-specific message reference - Mozilla | MDN")

> The name can include the following characters:
> 
> * A-Z
> * a-z
> * 0-9
> * _ (underscore)
> * @

src/const.js
EnTeQuAk: > and so the placeholders that don't match that regexp will be just ignored.

Interesting, I'll test this given that we have user reports that these messages "simply work in Firefox"

> Nevertheless, it seems to me that we should add A-Z to the ranges of characters matched by this regular expression (the placeholder matches besides its "case", then the matched placeholder is normalized to lowercase and retrieved from the map of the placeholders, also is also normalized to lowercase).

That's a good call!

tests/parsers/test.locale-messagesjson.js
EnTeQuAk: Ah ha! I think I simply wanted to :fire: too much, not checking the message name but keeping the check for placeholder sounds alright. I'll fix the PR

tests/schema/test.locale-messagesjson.js
```diff
@@ -20,12 +20,11 @@ describe('messages', () => {
     expect(validateLocaleMessages.errors[0].dataPath).toEqual('/foo/message');
   });
 
-  it('should fail on invalid message name', () => {
+  it('should not fail on invalid message name', () => {
```
rpl: How about changing this test description to something like: "should not validate the message name"?
it may be a bit confusing to call "invalid" something that doesn't make the validation to fail.

We should also add an empty message name now that the regular expression allows it.

rpl: Also, coveralls still complains about this line: https://coveralls.io/jobs/34824362/source_files/652389568#L47 (thanks @EnTeQuAk for finding it, it was driving me a bit crazy :-))

If there is no way anymore for a message name to be invalid, it is probably reasonable to remove the branch and the related constant?

src/const.js
```diff
@@ -162,7 +162,9 @@ export const MESSAGES_JSON = 'messages.json';
 export const LOCALES_DIRECTORY = '_locales';
 
 // This is a string, since it has to be matched globally on a message string.
-export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-z0-9_@]+)\\$';
+// This should match
+// https://searchfox.org/mozilla-central/rev/3abf6fa7e2a6d9a7bfb88796141b0f012e68c2db/toolkit/components/extensions/ExtensionCommon.jsm#1711
+export const MESSAGE_PLACEHOLDER_REGEXP = '\\$([a-zA-Z0-9_@]+)\\$';
```
freaktechnik: I hadn't included the uppercase letters because I passed the ignore case param to the RegExp constructor, but I guess better safe than sorry ;)

PR #1923, opened 2018-03-20
Fix absent content script detection.

### LINKED ISSUE #1920: Can't find these JS files
Can't find these JS files
But files exist
`chrome.tabs.executeScript(tabs[i].id, {
	file: 'files/script/preload.js'
});
chrome.tabs.executeScript(tabs[i].id, {
	file: 'files/script/postload.js'
});`

Error from your test system - `Content script file could not be found`

In past uploads, errors didn't arise. The files didn't change. Only one file was exchanged. And after your update, boom, got an error. I have to make a fast fix, but i can't. Thx! 😡

EnTeQuAk:
Fixes #1920

This essentially always tries to resolve the `file` based on `dirname`
but properly normalizes absolute paths.

I hope this fixes most of the edge-cases we've seen.

### DIFF
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -6,6 +6,7 @@ import { CONTENT_SCRIPT_NOT_FOUND, CONTENT_SCRIPT_EMPTY } from 'messages/javascr
 export default {
   create(context) {
     const dirname = path.dirname(context.getFilename());
+
     const existingFiles = context.settings.existingFiles || {};
     return {
       MemberExpression(node) {
@@ -41,10 +42,13 @@ export default {
           }
           let normalizedName = path.resolve(dirname, fileValue);
           if (path.isAbsolute(fileValue)) {
-            normalizedName = path.join(path.resolve('.'), path.normalize(fileValue));
+            normalizedName = path.join(dirname, path.normalize(fileValue));
           }
           let existingFileNames = Object.keys(existingFiles);
-          existingFileNames = existingFileNames.map((fileName) => path.resolve(fileName));
+
+          existingFileNames = existingFileNames.map((fileName) => {
+            return path.resolve(dirname, fileName);
+          });
 
           // If file exists then we are good.
           if (existingFileNames.includes(normalizedName)) {
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -40,10 +40,20 @@ export default {
             });
             return;
           }
-          let normalizedName = path.resolve(dirname, fileValue);
-          if (path.isAbsolute(fileValue)) {
-            normalizedName = path.join(dirname, path.normalize(fileValue));
+
+          // We can't reliably validate relative file names because their
+          // behavior depends on the current page and the behavior for
+          // Firefox and Chrome differntiates too. So we choose only to
+          // validate absolute file paths to avoid false positives.
+
+          if (!path.isAbsolute(fileValue)) {
+            return;
           }
+
+          const normalizedName = path.join(
+            dirname,
+            path.normalize(fileValue));
+
           let existingFileNames = Object.keys(existingFiles);
 
           existingFileNames = existingFileNames.map((fileName) => {
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -7,9 +7,9 @@ export default {
   create(context) {
     const existingFiles = Object.keys(
       context.settings.existingFiles || {}
-    ).map((fileName => {
+    ).map((fileName) => {
       return path.resolve('/', fileName);
-    }));
+    });
 
     return {
       MemberExpression(node) {
--- a/src/rules/javascript/content-scripts-file-absent.js
+++ b/src/rules/javascript/content-scripts-file-absent.js
@@ -5,9 +5,8 @@ import { CONTENT_SCRIPT_NOT_FOUND, CONTENT_SCRIPT_EMPTY } from 'messages/javascr
 
 export default {
   create(context) {
-    const dirname = path.dirname(context.getFilename());
-
     const existingFiles = context.settings.existingFiles || {};
+
     return {
       MemberExpression(node) {
         if (!node.object.object || !isBrowserNamespace(node.object.object.name)) {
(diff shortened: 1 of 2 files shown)


src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -39,12 +40,24 @@ export default {
...
+
+          // We can't reliably validate relative file names because their
+          // behaviour depends on the current page and the behaviour for
+          // Firefox and Chrome differentiates too. So we chose only to
+          // validate absolute file paths to avoid false positives.
+          if (!path.isAbsolute(fileValue)) {
+            return;
           }
+
+          const normalizedName = path.join(
```
rpl: I'm wondering if now that we are only going to validate absolute paths (and so we are sure that the file name is going to start with '/') we could simplify this checks by just resolving the file as relative to '/', e.g.:

```js
path.relative("/", "/dir/file.js") // returns "dir/file.js"

path.relative("/", "/dir/../dir/./file.js") // returns "dir/file.js"

path.relative("/", "/dir/../../../.././file.js") // returns "file.js"
```

and then just check if the file is in the existingFileName map without any need to convert them info absolute urls.
(I also double-checked that what we do internally in Firefox is basically the same kind of resolution here: https://searchfox.org/mozilla-central/rev/877c99c523a054419ec964d4dfb3f0cadac9d497/toolkit/components/extensions/ext-tabs-base.js#661)

This way we could simplify this to something like:

```js
const normalizedName = path.resolve('/', fileValue);

if (Object.keys(existingFiles).includes(normalizedName)) {
  ...
}
```

which seems also easier to read.

EnTeQuAk: That's tons cleaner. I'll update the rule :smiley:

EnTeQuAk: That worked mostly, I still need to normalize the `existingFiles` object which is why I kept it but we're now resolving everything to `'/'` and avoid the explicit join on `dirname`.

That already looks a bit better.

I wasn't able to not normalize the `existingFiles` because they're all relative paths so we have to make them absolute which seemed like the more reasonable approach.

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -15,10 +15,8 @@ function createJsScanner(code, validatedFilename, existingFiles = {}) {
 describe('content_scripts_file_absent', () => {
   it('should show an error when content script is missing', async () => {
     const nonExistentFiles = [
```
rpl: Nit, with just one scenario to test here, I'm not sure that it is still worth that this test still iterates on a `nonExistentFiles` array, we could just remove the `nonExistentFiles.forEach` and the `nonExistentFiles` and make this test a bit simpler to read.

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -15,10 +15,8 @@ function createJsScanner(code, validatedFilename, existingFiles = {}) {
 describe('content_scripts_file_absent', () => {
   it('should show an error when content script is missing', async () => {
```
rpl: Nit, it could be reasonable to mention in the test description that the content script missing path has to be "absolute" (e.g. something like "should show an error for a missing absolute content script path")

tests/rules/javascript/test.content_scripts_file_absent.js
```diff
@@ -27,12 +25,26 @@ describe('content_scripts_file_absent', () => {
...
 
       const { linterMessages } = await jsScanner.scan();
+
       expect(linterMessages.length).toEqual(1);
       expect(linterMessages[0].code).toEqual(CONTENT_SCRIPT_NOT_FOUND.code);
       expect(linterMessages[0].type).toEqual(VALIDATION_ERROR);
     });
   });
 
+  it('should not show an error when relative content script is missing', async () => {
```
rpl: Nit, we could probably merge this test with the one at line 68 (especially given that the new version of this rule never validates relative content script paths, without any difference between being in a top level js file or a js file from a sub directory).

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -39,17 +39,26 @@ export default {
             });
             return;
           }
-          let normalizedName = path.resolve(dirname, fileValue);
-          if (path.isAbsolute(fileValue)) {
-            normalizedName = path.join(path.resolve('.'), path.normalize(fileValue));
+
+          // We can't reliably validate relative file names because their
```
rpl: :+1: Thanks for the additional inline comment, it will be helpful when we will get back to this and we need to recall the reasons why we chose this strategy.

The difference between Firefox and Chrome is not exactly the reason why we chose this strategy, but I guess that is still worth to mention it.

Nit, how about tweaking it a bit? e.g. "... because they are resolved as relative to the current page url on Firefox and the rule itself doesn't know the path of the html file (or files) where the js file is going to be loaded, and so we chose to validate only the absolute file paths to avoid false positive. (Also note that Firefox and Chrome behave differently when resolving relative content script paths used in a tabs.executeScript API call)."

src/rules/javascript/content-scripts-file-absent.js
```diff
@@ -39,17 +39,26 @@ export default {
...
+          // validate absolute file paths to avoid false positives.
+          if (!path.isAbsolute(fileValue)) {
+            return;
           }
-          let existingFileNames = Object.keys(existingFiles);
-          existingFileNames = existingFileNames.map((fileName) => path.resolve(fileName));
+
+          const normalizedName = path.resolve('/', path.normalize(fileValue));
+
+          const existingFileNames = Object.keys(existingFiles).map((fileName) => {
```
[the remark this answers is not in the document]
EnTeQuAk: Fixed in 5ce5b91

tests/rules/javascript/test.content_scripts_file_absent.js
EnTeQuAk: Fixed in 9e67fa4

src/rules/javascript/content-scripts-file-absent.js
EnTeQuAk: > The difference between Firefox and Chrome is not exactly the reason why we chose this strategy, but I guess that is still worth to mention it.

It's not necessarily but that difference helps understanding what developers might prefer and what scenarios might happen more often in the real world.

Anyhow, fixed in 473b1a6

tests/rules/javascript/test.content_scripts_file_absent.js
EnTeQuAk: Fixed in 473b1a6

PR #1926, opened 2018-03-21
Get rid of getConfig() calls for ManifestJSONParser

### LINKED ISSUE #1925: Get rid of calls of getConfig() in ManifestJSONParser
The constuctor of ManifestJSONParser uses getConfig() to get default argument value for selfHosted and isLanguagePack, so if caller doesn't specify these values, getConfig()  incidentally parses command line arguments, thus if the command does not specify any argument,  yargs output 'Not enough non-option arguments: got 0, need at least 1' error unfortunately.

You can easily confirm the error to run below script on the top of addons-linter source directory.

```
#!/usr/bin/env node

require('./dist/addons-linter')
  .createInstance({
    config: {
      logLevel: 'debug',
      _: ['./tests/fixtures/webextension_scan_file/'],
    },
    runAsBinary: true,
  }).run();
```

hiikezoe:
Fixes #1925.

### DIFF
--- a/tests/test.yargs-options.js
+++ b/tests/test.yargs-options.js
@@ -1,0 +1,11 @@
+import { getDefaultConfigValue } from 'yargs-options';
+
+describe('getDefaultConfigValue()', () => {
+  it('should return the default value', () => {
+    expect(getDefaultConfigValue('self-hosted')).toEqual(false);
+  });
+
+  it('should return undefined for unknown option', () => {
+    expect(getDefaultConfigValue('unknown')).toBeUndefined();
+  });
+});
--- a/src/cli.js
+++ b/src/cli.js
@@ -2,6 +2,7 @@ import yargs from 'yargs';
 import { oneLine } from 'common-tags';
 
 import log from 'logger';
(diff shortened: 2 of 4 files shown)


hiikezoe:
gah, cwd option was not supported in node v6?

hiikezoe:
I gave up including the test case.

tests/test.yargs-options.js
```diff
@@ -0,0 +1,11 @@
+import { getDefaultConfigValue } from 'yargs-options';
+
+describe('getDefaultConfigValue()', () => {
+  it('should return the default value', () => {
+    expect(getDefaultConfigValue('self-hosted')).toBeFalsy();
```
EnTeQuAk: Let's change this to use `.toEqual(false)` if we are expecting this or `undefined` etc depending on what value we expect. `falsy` can be many things and explicit is better than implicit.

hiikezoe: Thanks.  That's good to know. I didn't know that what 'toBeFalsy()' exactly does.

Issue #1898, opened 2018-03-13
Import Firefox 60 schemas


mstriemer:
Firefox 60 is now in Beta, the schemas should be updated to support any new or changed APIs.

AlexandraMoga:
Could you please add STR or the label "qa:not needed"?
Thanks

mstriemer:
It would be nice to get some QA on this but the changes are quite broad so I'm not sure that's realistic.

PR #1934, opened 2018-03-26
Import Firefox 60 schemas again

### LINKED ISSUE #1893: Update the API schema with "downloads" and "downloads.open" as optional permissions
Submit an add-on that has “downloads" and "downloads.open” as optional permissions.

### What happened?
The add-on validation has failed and some errors are displayed.

The "downloads" and "downloads.open" to be recognized as optional permissions at submit.

Reproduced on AMO -dev with FF60 on Win 7 64-bit.
optionalpermissions

mstriemer:
Fixes #1893, Supports #1898.

The last import must have pulled an initial build instead of the latest beta. I think there is only one beta tag on mozilla-central so that would be the one from a couple weeks before the actual code freeze.

We'll want to update the import script to use the changes from #1795.

### DIFF
--- a/src/schema/imported/devtools.json
+++ b/src/schema/imported/devtools.json
@@ -1,7 +1,7 @@
 {
   "id": "devtools",
   "permissions": [
-    "devtools"
+    "manifest:devtools_page"
   ],
   "allowedContexts": [
     "devtools",
--- a/src/schema/imported/dns.json
+++ b/src/schema/imported/dns.json
@@ -1,0 +1,95 @@
+{
+  "id": "dns",
+  "description": "Asynchronous DNS API",
+  "permissions": [
+    "dns"
+  ],
+  "functions": [
+    {
+      "name": "resolve",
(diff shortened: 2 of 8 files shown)


tofumatt:
I’m on my phone but this seems fine to me. r+

src/schema/imported/devtools.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "devtools",
   "permissions": [
-    "devtools"
+    "manifest:devtools_page"
```
mstriemer: I'm going to ask aswan about this. We might want to overwrite this change, it looks like internally it ends up adding the devtools pref but I don't understand what the deal with this is.

mstriemer: I discussed this on IRC and this basically means including the "devtools_page" manifest property with implicitly request the "devtools" permission. I tested an extension requesting these permissions and they are working as expected.

PR #1939, opened 2018-03-29
test: run integration test suite in a production-like environment

rpl:
This pull request is build on top of #1867 (many kudos to @hameleonka for getting it started) and applies some additional tweaks to allow the integration test suite to be optionally executed in a "production-like environment"[1], and then run the new npm script (currently named "test-integration:production") on both travis and appveyor (as well as in a local development environment so that we can more easily investigate locally failures related to it).

[1]: Basically packing the built addons-linter as it would be when we release a new version on npm, and then unpacking it in a temporary directory and install only the dependencies that would be available when a user installs the released version from npm, similarly to the current hack used in the travis.yml here: https://github.com/mozilla/addons-linter/blob/2cf2439df54ab959aa7307f62970346b7dd609cf/.travis.yml#L12-L21


muffinresearch:
@rpl hi - when you have a moment could you rebase this and resolve the conflicts then maybe we can ask @EnTeQuAk for a review?

EnTeQuAk:
Yeah, give it a tight squeeze and squash & merge it baby! :laughing:

PR #1941, opened 2018-03-30
chore(package): update webpack to version 4.4.1

rpl:
This pull request supersedes #1927, which fails on travis because webpack 4 also requires webpack-cli to be explicitly installed as a separate dev dependency.

This pull request also set the new webpack `mode` config property to `'none'`, which seems to restore most of the default behaviors of webpack3, otherwise the build step raise a number of `ModuleConcatenation bailout: Module is not an ECMAScript module` warning messages (and an additional warning message which complains about the missing `mode` config property and warns that webpack 4 will defaults to the new `mode: 'production'` behavior when the explicit `mode` is missing).

I also briefly tested locally that the new webpack4 bundle is able to run successfully from the addons-linter bin script (and that the `npm run extract-locales` script is able to extract the i18n strings as expected, given that it is also based on a webpack config and it could have been affected by a webpack upgrade).

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
     "request": "2.83.0",
     "sinon": "4.4.6",
     "tar": "4.4.1",
-    "webpack": "3.10.0"
+    "webpack": "4.4.1",
+    "webpack-cli": "2.0.13"
   },
   "dependencies": {
     "ajv": "6.3.0",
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
Webpack just released https://github.com/mozilla/addons-linter/pull/1948 version 4.5 but let's land this here first and upgrade to anything even more recent later.

PR #1983, opened 2018-05-09
Import Firefox 61 schema.

### LINKED ISSUE #1972: Add FF60 new namespace proxy.settings
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1455755

> **proxy.settings is not supported**
> 
> **Warning:** This API has not been implemented by Firefox.

Example:
https://addons.mozilla.org/en-US/developers/upload/9066b64a76224d14925992d914426f9c

### LINKED ISSUE #1974: Import Firefox 61 schema
Firefox 61 beta will be released on May 9th, let's merge the schema for the first time once this happens.

EnTeQuAk:
Fixes #1974
Fixes #1972

### DIFF
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -355,6 +355,37 @@
         }
       }
     },
+    "WebExtensionDictionaryManifest": {
+      "$merge": {
+        "source": {
+          "$ref": "manifest#/types/ManifestBase"
+        },
+        "with": {
+          "type": "object",
+          "description": "Represents a WebExtension dictionary manifest.json file",
+          "properties": {
+            "homepage_url": {
+              "type": "string",
+              "format": "url",
+              "preprocess": "localize"
+            },
+            "dictionaries": {
+              "type": "object",
+              "patternProperties": {
+                "^[a-z]{2}[a-zA-Z-]*$": {
+                  "type": "string",
+                  "format": "strictRelativeUrl",
+                  "pattern": "\\.dic$"
+                }
+              }
+            }
+          },
+          "required": [
+            "dictionaries"
+          ]
+        }
+      }
+    },
     "ThemeIcons": {
       "type": "object",
       "properties": {
--- a/src/schema/imported/browser_settings.json
+++ b/src/schema/imported/browser_settings.json
@@ -25,6 +25,16 @@
         }
       ]
     },
+    "closeTabsByDoubleClick": {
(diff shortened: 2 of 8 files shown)


src/schema/imported/manifest.json
```diff
@@ -355,6 +355,37 @@
         }
       }
     },
+    "WebExtensionDictionaryManifest": {
```
EnTeQuAk: This doesn't have any special handling yet, I opened https://github.com/mozilla/addons-linter/issues/1984 to track that work.

EnTeQuAk:
Hmm, fair point. @wagnerand, does it suffice to train reviewers in this
or should I implement a more suffisticated message here?
This would probably mean a simple custom rule that detects and warns but
let's the developer decide what to do.
I think we can implement tgatbjn a separate step.


On Wed, May 9, 2018, at 9:52 PM, Luca Greco wrote:
> *@rpl* approved this pull request.


> Just an additional side note:


> Bug 1455755[1] has recently moved the
> browser.browserSettings.proxyConfig to browser.proxy.settings, this
> change is also on Firefox 60.> Once we release the new linter version with this change, the addon
> developers are going to get a failure if the linter detects that the
> old namespace is used in an extension, which is good, but the error
> will not mention that the API still exists and it has been actually
> moved into the proxy namespace, and so it would be nice to communicate
> this change (e.g. point it out to the reviewers, which are very likely
> to receive questions about this, and/or the addon developers on AMO).> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].> 

Links:

  1. https://bugzilla.mozilla.org/show_bug.cgi?id=1455755
  2. https://github.com/mozilla/addons-linter/pull/1983#pullrequestreview-118880483

wagnerand:
Development questions are out of scope for reviewers either way, so I don't have a strong opinion on that.

Another thing that comes to my mind is how do we support extensions for 52ESR using this API?

EnTeQuAk:
> Another thing that comes to my mind is how do we support extensions for 52ESR using this API?

There's a "deprecated API" feature in the browser. We might want to make use of that. It's all just a warning so it doesn't block submission or updates anyway. Does that work?

EnTeQuAk:
I'm merging this and open another PR that adds the `browser.browserSettings.proxyConfig` to `browser.proxy.settings` move into the deprecated API list.

Issue #2019, opened 2018-06-04
Lock files in this repository


LinusU:
When starting work on an unrelated issue I found that this repository has both a `package-lock.json` file, and a `yarn.lock` file.

Furthermore, none of these files seems to be in sync with the `package.json` file.

I think that trying to have more than one lock file is impossible, without a special tool that syncs between them. Trying to run both `npm install` and `yarn` to keep them in sync will most likely result in subtle differences.

It would be nice if we could 1) decide on one of them, then 2) make a PR that removes the other file, and updates the chosen one, and finally 3) make Travis run `npm ci` or `yarn --pure-lockfile` so that the build will fail if there is a PR that doesn't update the correct lock-file.

LinusU:
I'd be happy to submit a PR for this if I can get an indication of whether `yarn` or `npm` would be preferred ☺️

wagnerand:
I would be in favor of moving back to `npm`. Any objections, @EnTeQuAk?

EnTeQuAk:
No, more recent versions of npm are very powerful and fast enough so I'm all in favor of simplifying this and moving back to npm.

AlexandraMoga:
@EnTeQuAk please let me know if this requires any testing, otherwise add the "qa:not needed" label
Thanks

PR #2020, opened 2018-06-04
Move doctoc to dev-dependencies

LinusU:
The `doctoc` dependency is only used to generate the `CONTRIBUTORS.md` file, and thus doesn't need to be installed in order to consume this package. Thus, this PR moves it from `dependencies` to `devDependencies`.

I did not update any of the lock-files since both are outdated, see #2019.


EnTeQuAk:
Perfect thanks a lot for your contribution!

PR #2047, opened 2018-06-15
Import Firefox 61b14 schema.

### LINKED ISSUE #2046: Import Firefox 61 Beta 14 Schema

EnTeQuAk:
Fixes #2046

This also syncs yarn.lock

### DIFF
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -604,6 +604,9 @@
         },
         {
           "$ref": "#/types/MatchPatternRestricted"
+        },
+        {
+          "$ref": "#/types/MatchPatternUnestricted"
         }
       ]
     },
@@ -617,6 +620,15 @@
         {
           "type": "string",
           "pattern": "^file:///.*$"
+        }
+      ]
+    },
+    "MatchPatternUnestricted": {
+      "description": "Mostly unrestricted match patterns for privileged add-ons. This should technically be rejected for unprivileged add-ons, but, reasons. The MatchPattern class will still refuse privileged schemes for those extensions.",
+      "anyOf": [
+        {
+          "type": "string",
+          "pattern": "^resource://(\\*|\\*\\.[^*/]+|[^*/]+)/.*$|^about:"
         }
       ]
     },
--- a/yarn.lock
+++ b/yarn.lock
@@ -78,23 +78,6 @@
     lodash "^4.2.0"
     to-fast-properties "^2.0.0"
 
-"@mrmlnc/readdir-enhanced@^2.2.1":
-  version "2.2.1"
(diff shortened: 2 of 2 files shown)


src/schema/imported/manifest.json
```diff
@@ -620,6 +623,15 @@
         }
       ]
     },
+    "MatchPatternUnestricted": {
+      "description": "Mostly unrestricted match patterns for privileged add-ons. This should technically be rejected for unprivileged add-ons, but, reasons. The MatchPattern class will still refuse privileged schemes for those extensions.",
```
rpl: as also mentioned in the description above, this additional match pattern is meant to be used only by privileged add-ons and internally Firefox will refuse to use the privileged schemes for unprivileged add-ons.

Looking at the patch and the test cases (e.g. https://hg.mozilla.org/releases/mozilla-esr60/rev/6ae65bc81d98#l9.47), Firefox will not raise any warning or error when an "unrestricted match pattern" is specified by unprivileged add-ons, but it is going be ignored in those cases.

In my opinion the linter should probably raise a linting error when unprivileged add-ons submitted to AMO are using this pattern, so that third party developers can be aware that those patterns will not be actually be matched for their add-ons.

But it is probably also a good idea to do not raise it as a linting error when the extension has the `mozillaAddons` permission, because some of the system add-ons may be using the addons-linter as part of their CI jobs, e.g. screenshot system addon has the addons-linter in its package.json scripts:
- https://github.com/mozilla-services/screenshots/blob/60e45248bd1ea78519940e5f7d32153a2b82ef92/package.json#L91

src/schema/imported/manifest.json
```diff
@@ -620,6 +623,15 @@
         }
       ]
     },
+    "MatchPatternUnestricted": {
```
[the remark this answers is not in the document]
EnTeQuAk: Nice find! This is indeed quite bad to look at.

src/schema/imported/manifest.json
EnTeQuAk: > In my opinion the linter should probably raise a linting error when unprivileged add-ons submitted to AMO are using this pattern, so that third party developers can be aware that those patterns will not be actually be matched for their add-ons.

> But it is probably also a good idea to do not raise it as a linting error when the extension has the mozillaAddons permission, because some of the system add-ons may be using the addons-linter as part of their CI jobs, e.g. screenshot system addon has the addons-linter in its package.json scripts:

Sounds good, I'll add a helper that verifies this and update the PR. Not sure this will be ready to land today then though but I don't think that's going to be an issue.

EnTeQuAk:
Closing this for now.

PR #2070, opened 2018-06-25
Rename 'prepublish' to 'prepublishOnly'.

EnTeQuAk:
This got deprecated in npm 6, see
https://github.com/npm/npm/issues/10074 for much much more details.

This will also run snyk just once in our CI environment instead of twice.


EnTeQuAk:
> We need to keep in mind that snyk might try and add prepublish in future PRs again, I am not sure whether they have updated that logic yet.

We'll see, I'll keep an open eye on it.

Issue #1775, opened 2018-01-19
Add ES module support


asamuzaK:
### Describe the problem and steps to reproduce it:
This is a feature request.

Module scripts are enabled on Nightly. [1428002 - Enable &lt;script type=&quot;module&quot;&gt; in nightly builds](https://bugzilla.mozilla.org/show_bug.cgi?id=1428002 "1428002 - Enable &lt;script type=&quot;module&quot;&gt; in nightly builds")

addons-linter should support it.

### Anything else we should know?
(Please include screenshots and any relevant files.)

asamuzaK:
Extension for test.
[webextensions_test-0.0.1.zip](https://github.com/mozilla/addons-linter/files/1645320/webextensions_test-0.0.1.zip)

asamuzaK:
See also #1732

asamuzaK:
Extension for test.
[webextensions_test-0.0.2.zip](https://github.com/mozilla/addons-linter/files/1645339/webextensions_test-0.0.2.zip)

addons-linter currently reports JS_SYNTAX_ERROR.

muffinresearch:
This will need checking for support upstream in eslint. If it has it we should be able to enable it. If it doesn't we'll need to file an issue upstream.

asamuzaK:
eslint supports es modules.
set config
```
"env": {
  "parserOptions": {
     "sourceType": "module"
  }
}
```

The problem is that you need to set that config *before* linting.



https://eslint.org/docs/user-guide/configuring

asamuzaK:
node.js decided to use `.mjs` file extension for es modules, browsers decided to use `script="module"` which means file extension doesn't matter, so `.js` files can be normal script or module.

So I suppose addons-linter need to look into the file and detect It's normal script or module?

asamuzaK:
Refs
* [Intent to ship: Module scripts (ES6 modules)](https://groups.google.com/forum/m/#!msg/mozilla.dev.platform/DZ7YctNoUUk/c2w8Ue8QCgAJ)
* [1438139 - Enable JS modules by default for all builds](https://bugzilla.mozilla.org/show_bug.cgi?id=1438139)

robinst:
Firefox 60 has been released with modules enabled by default (see [release notes](https://developer.mozilla.org/en-US/Firefox/Releases/60#JavaScript)):

[...]

PR #2078, opened 2018-06-27
Add ES6 module support

EnTeQuAk:
This implements ES6 module support.

This parses JavaScript files and tries to figure out their sourceType
from how the AST is structured.

We default to `module` but if there are syntax errors fall back to
`script`. We do not read the `<script type="module"` part of HTML files
anymore.

This may not be perfect but it's pretty close to what we want. We may catch too much but can always be more restrictive once this new code proves useful in production.

Fixes https://github.com/mozilla/addons-linter/issues/1775

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -1,4 +1,4 @@
-import { extname } from 'path';
+import path from 'path';
 
 import columnify from 'columnify';
 import chalk from 'chalk';
@@ -294,12 +294,12 @@ export default class Linter {
   getScanner(filename) {
     if (filename.match(constants.HIDDEN_FILE_REGEX) ||
         filename.match(constants.FLAGGED_FILE_REGEX) ||
-        constants.FLAGGED_FILE_EXTENSIONS.includes(extname(filename)) ||
+        constants.FLAGGED_FILE_EXTENSIONS.includes(path.extname(filename)) ||
         filename.match(constants.ALREADY_SIGNED_REGEX)) {
       return FilenameScanner;
     }
 
-    switch (extname(filename)) {
+    switch (path.extname(filename)) {
       case '.css':
         return CSSScanner;
       case '.html':
@@ -473,7 +473,7 @@ export default class Linter {
       await this.scanFiles(filesWithoutJSLibraries);
 
       this.print(deps._console);
-      // This is skipped in the code coverage because the
+      // This is skipped in code coverage because the
       // test runs against un-instrumented code.
       /* istanbul ignore if */
       if (this.config.runAsBinary === true) {
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -1,7 +1,7 @@
-import path from 'path';
-
 import ESLint from 'eslint';
 import { oneLine } from 'common-tags';
+import espree from 'espree';
+import vk from 'eslint-visitor-keys';
 
 import {
   ESLINT_RULE_MAPPING,
@@ -10,6 +10,9 @@ import {
 import * as messages from 'messages';
 import { rules } from 'rules/javascript';
 import { ensureFilenameExists } from 'utils';
+
+
+const ECMA_VERSION = 2019;
 
 
 export function excludeRules(excludeFrom = {}, excludeWhat = []) {
@@ -52,7 +55,10 @@ export default class JavaScriptScanner {
     _ruleMapping = ESLINT_RULE_MAPPING,
     _messages = messages,
   } = {}) {
-    const cli = new _ESLint.CLIEngine({
+    this._ESLint = ESLint;
+    this.sourceType = this.detectSourceType(this.filename);
+
+    const configDefaults = {
       baseConfig: {
         env: {
           es6: true,
@@ -64,8 +70,13 @@ export default class JavaScriptScanner {
           existingFiles: this.options.existingFiles,
         },
       },
+      // It's the default but also shouldn't change since we're using
+      // espree to parse javascript files below manually to figure out
+      // if they're modules or not
+      parser: 'espree',
       parserOptions: {
-        ecmaVersion: 2019,
+        ecmaVersion: ECMA_VERSION,
+        sourceType: this.sourceType,
       },
       rules: _ruleMapping,
       plugins: ['no-unsafe-innerhtml'],
@@ -78,9 +89,12 @@ export default class JavaScriptScanner {
       patterns: ['!bower_components/*', '!node_modules/*'],
 
       filename: this.filename,
+
       // Avoid loading the addons-linter .eslintrc file
       useEslintrc: false,
-    });
+    };
+
+    const cli = new _ESLint.CLIEngine(configDefaults);
 
     const rulesAfterExclusion = excludeRules(_rules, this.disabledRules);
     Object.keys(rulesAfterExclusion).forEach((name) => {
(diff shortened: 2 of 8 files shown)


src/linter.js
```diff
@@ -671,4 +689,32 @@ export default class Linter {
       });
     }
   }
+
+  /*
+    Analyze all HTML files in the package to figure out
+    if we should enable `sourceType: module` instead of parsing the file
+    as a regular script for a particular file.
+  */
+  async _extractJsModules() {
```
rpl: @EnTeQuAk I'm wondering if parsing the html files is really worth it, especially given that it seems that we are always going to resort to "try to re-parse the js file if an error related to ES6 import and export syntax has been raised while parsing the file in the default parsing mode" (at least based on the current implementation of the `verifyParserOptionsAndGetSourceCode` method), and so I'm not really sure that the complexity it adds to the change is properly balanced by useful issues that this may be able to catch (and it is clear that is going to miss a good chunk of them, e.g. tag scripts injected at runtime, es6 modules imported by other modules, and the dynamic imports once officially supported).

src/linter.js
```diff
@@ -1,9 +1,10 @@
-import { extname } from 'path';
+import path from 'path';
 
 import columnify from 'columnify';
 import chalk from 'chalk';
 import Dispensary from 'dispensary';
 import { oneLine } from 'common-tags';
+import cheerio from 'cheerio';
```
rpl: Nit, if we are going to keep the "HTML parsing" part of these changes, let's move this import after the one for `chalk` (it seems to me that the other imports groups are being kept in alphabetic order).

src/scanners/javascript.js
```diff
@@ -78,83 +81,152 @@ export default class JavaScriptScanner {
...
+      eslintScopeManager: true,
+      filePath: config.filename,
+    });
+
+    let ast;
+
+    try {
+      ast = espree.parse(this.code, config.parserOptions);
+    } catch (exc) {
+      if (exc.message.startsWith("'import' and 'export'")) {
```
rpl: Having to resort to check the error message makes me sad :-( 
(and I'm sure that it makes you sad as well ;-))
and it also feels a bit fragile.

I suspect that this is also going to parse twice any js file that uses the import/export syntax in an invalid way because it also catch `"'import' and 'export' may only appear at the top level"` (instead of only catching `"'import' and 'export' may appear only with 'sourceType: module'"`).

If we can't avoid to match the error message, I would probably prefer if we check if the error `endsWith` `"may appear only with 'sourceType: module'"`.

Given that we are parsing the source, why we can't do the opposite?
- parse the source with parserOptions.sourceType = 'module'
- look into the ast for "top level import or export" statements:
  if there are => eslint should use `sourceType: "module"` for this file
  if there aren't => eslint should use `sourceType: "script"` for this file\

how that sounds to you?

tests/unit/scanners/test.javascript.js
```diff
@@ -118,6 +117,17 @@ describe('JavaScript Scanner', () => {
     expect(linterMessages.length).toEqual(0);
   });
 
+  it('should support es6 modules via html module and ast-check', async () => {
+    const addonLinter = new Linter({
+      _: ['tests/fixtures/webextension_es6_module.zip'],
```
rpl: We should also test this on an unpacked extension 
(e.g. with the HTML parsing part of this pull request, we would need to verify that the code that compute the module paths works correctly for unpacked extension, because it may be affected by the difference with the platform filesystem, e.g. the different separator used on windows and the "case insensitivity" of the osx filesystem).

Do you mind to include the fixture test extension unpacked in the pull request? 
(which would also make it easier to review the content of the fixture extension, and to change it in future pull requests if we need to cover some additional scenarios around the es6 modules) 
and then pack it as a zip as part of the test setup to verify the behaviors for a packed extension submitted on AMO.

src/linter.js
EnTeQuAk: I think I agree, I'm going to remove it. It does set the proper `sourceType` upfront but at the cost of parsing all html files in the add-on, even though there may not be any es6 modules anywhere.

It's probably best to add some eslintrc fetching in a later addition instead where people can simply declare their add-on as "uses es6 modules" and we're going to use that information.

src/scanners/javascript.js
EnTeQuAk: Hmm, let me try that for a moment. This might work, given that this would only parse it and avoid any rules being executed upfront and issue wrong warnings.

tests/unit/scanners/test.javascript.js
EnTeQuAk: Will do!

src/scanners/javascript.js
EnTeQuAk: So this works more or less… except for scenarios where `acorn` raises issues because it does indeed verify `eval` isn't being used in a `strict` context which happens when we parse as `sourceType: module`.

I am now parsing with `sourceType: module` by default and falling back to `script` which seems reasonable. I don't think it matters much whether we do it this way or the other way around.

src/linter.js
```diff
@@ -362,8 +365,14 @@ export default class Linter {
         // TODO: Bring this in line with other scanners, see:
         // https://github.com/mozilla/addons-linter/issues/895
         collector: this.collector,
+        // This is used by the JavaScript scanner to determine the contents
+        // of background-pages to enable module support
+        io: this.io,
+        parsedJSON: this.addonParsedJSON,
```
EnTeQuAk: Oh, I can remove all this now! Completely forgot about it.

src/linter.js
```diff
@@ -362,8 +365,14 @@ export default class Linter {
...
         // list of disabled rules for js scanner
         disabledLinterRules: this.config.disableLinterRules,
+        // list of ES6 JavaScript modules
+        javaScriptEs6Modules: this.javaScriptEs6Modules,
```
EnTeQuAk: Eh, I must have been super sleepy, sorry I'll clean this up.

src/scanners/javascript.js
```diff
@@ -64,6 +66,7 @@ export default class JavaScriptScanner {
           existingFiles: this.options.existingFiles,
         },
       },
+      parser: 'espree',
```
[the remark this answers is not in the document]
ExE-Boss: Yes, [`espree` is the default parser](https://eslint.org/docs/user-guide/configuring#specifying-parser).

src/scanners/javascript.js
```diff
@@ -152,9 +158,73 @@ export default class JavaScriptScanner {
...
+
     return {
       linterMessages: this.linterMessages,
       scannedFiles: this.scannedFiles,
     };
   }
+
+  _getSourceType(node) {
+    const possibleImportExportTypes = [
+      'ImportDeclaration', 'ImportSpecifier', 'ImportDefaultSpecifier',
```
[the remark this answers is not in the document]
ExE-Boss: Also sort them alphabetically if possible.

src/scanners/javascript.js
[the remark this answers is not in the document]
EnTeQuAk: I thought since we're actually relying on espree to parse the files to figure out module support I'd set the parser here explicitly.

tests/fixtures/webextension_es6_module/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "applications": {
+    "gecko": {
+      "id": "webExtTest@asamuzak.jp"
```
ExE-Boss: Ant reason why this isn’t `@webextension-guid` like every other current test fixture?

EnTeQuAk: Fixed, I think I simply didn't change it from the example that I fetched from one of the issues.

EnTeQuAk:
I'm merging this as most of the recent changes were simply following comments and were very small cleanups and added tests.

I'm happy to add follow-up PRs but since this has been open for way too long already I'll go ahead and merge this.

PR #2081, opened 2018-06-28
fix(package): update shelljs to version 0.8.2

wagnerand:
Closes #2066


EnTeQuAk:
You'll actually have to update the lock file if you're updating the dependencies. Otherwise, this will continue to fail.

I can take this on if you don't have the time.

PR #2123, opened 2018-08-02
Validate langpacks without command line flag

### LINKED ISSUE #2120: Validate extensions as language packs if language pack properties are found in the manifest
It looks like the only way for the linter to validate language pack properties is through the use of the `--langpack` command line flag. This is a problem, because addons-server does not pass that flag to the linter.

I believe the linter should detect the type on its own, like it does for static themes, by looking for the `langpack_id` property in the manifest. 

To prove that it doesn't work on AMO right now: I uploaded https://addons-dev.allizom.org/en-US/developers/addon/not-a-real-polski-language-pac/ which uses an invalid `langpack_id`, I didn't receive a validation error. Locally, when validating it I don't receive an error either, unless I manually use the `--langpack` flag.

diox:
Fix #2120

### DIFF
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -69,21 +69,9 @@ export const validateAddon = (...args) => {
   return isValid;
 };
 
-const _validateLangPack = validator.compile({
-  ...schemaObject,
-  id: 'langpack-manifest',
-  $ref: '#/types/WebExtensionLangpackManifest',
-});
-
-export const validateLangPack = (...args) => {
-  const isValid = _validateLangPack(...args);
-  validateLangPack.errors = filterErrors(_validateLangPack.errors);
-  return isValid;
-};
-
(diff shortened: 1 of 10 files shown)


src/schema/validator.js
```diff
@@ -113,6 +101,34 @@ export const validateStaticTheme = (...args) => {
   return isValid;
 };
 
+// Like with static themes, we don't want additional properties in langpacks.
+// The only difference is, this time, there is no additional schema file, we
+// just need to reference WebExtensionLangpackManifest and merge it with the
+// object that has additionalProperties: false.
+const _validateLangPack = validator.compile({
```
diox: That seems to work, but let me know if I messed it up :)

EnTeQuAk: Looks good to me and works nicely.

Ideally this should all be upstream but it's a completely different story and time-intensive.

PR #2139, opened 2018-08-11
Search whole file for ES6 module detection

### LINKED ISSUE #2138: Some modules are detected as scripts
Following #2078, addons-linter should detect whether a .js file is a module or a script.
Run addons-linter on an extension with a module whose first statement is not an import or export statement.

### What happened?
The module is still detected as a script, so it reports a syntax error on any import or export statements.

It should be detected as a module.

psimonyi:
_getSourceType was not walking the whole AST, so it would only detect
import/export statements if they were the first in the file.  This fixes
the recursion to walk the whole AST.

Fixes #2138

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -190,7 +190,9 @@ export default class JavaScriptScanner {
 
         if (Array.isArray(child)) {
           for (let j = 0; j < child.length; ++j) {
-            return this._getSourceType(child[j]);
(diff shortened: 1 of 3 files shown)


caitmuenster:
Thanks, @psimonyi! We've added your contribution to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition).

Would you be interested in creating a profile on [mozillians.org](https://mozillians.org)? I'd be happy to vouch fro you.

PR #2175, opened 2018-09-04
Upgrade babel 7

### LINKED ISSUE #2177: Upgrade to Babel 7
https://babeljs.io/docs/en/v7-migration

EnTeQuAk:
Upgrade to Babel 7.

https://babeljs.io/docs/en/v7-migration

This tries to be as backwards compatible with what we generated with babel 6 as possible.

Babel 7 deprecated stages in favour of explicitly defining what features/syntaxes you're using.

Fixes #2177


EnTeQuAk:
> As a side note, I'm wondering if we could move (even in a followup of this PR) all the scripts in the bin/ directory that are not meant to be used by the users (basically all of them besides the one named addons-linter :-P) into a different subdir (e.g. maybe a new one named tasks/?), so that we could move @babel/polyfill and @babel/register into the dev dependencies and just use an eslint comment to avoid that linting error when importing a dev dependency, what do you think?.

I completely agree. I'll file an issue

EnTeQuAk:
Rebased to master, waiting for CI to pass then I'll merge

Issue #1905, opened 2018-03-14
Validation error "/version" should match format "versionString" should say what the versionString format is 


diox:
Initially filed as https://github.com/mozilla/addons-server/issues/7778

Original description of the issue written by @arantius:

I tried to automatically submit a nightly build for signing and self-serving.  The (partial) result was:

    {
      "guid": "{e4a8a97b-f2ed-450b-b12d-ee082ba24781}",
    ...
      "validation_results": {
        "errors": 1,
        "success": false,
    ...
        "messages": [
          {
            "uid": "ba64e9016ce74252b49bc45f074ef667",
            "dataPath": "/version",
            "tier": 1,
            "message": "\"/version\" should match format \"versionString\"",
            "type": "error",
            "id": [
              "JSON_INVALID"
            ],
            "description": "Your JSON file could not be parsed."
          },
    ...
      "version": "2018.03.09.nightly.e7c1c74"
    }

The [webext docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/version) say "see the Version format page."  The [version format](https://developer.mozilla.org/en-US/docs/Mozilla/Toolkit_version_format) page says "A version string consists of one or more version parts, separated with dots. Each version part is itself parsed as a sequence of four parts: \<number-a\>\<string-b\>\<number-c\>\<string-d\>. Each of the parts is optional."

Okay, I've got five parts. 

1. number a (rest, optional, omitted)
2. number a (rest, optional, omitted)
3. number a (rest, optional, omitted)
4. number a (optional, omitted), string b, (rest, optional, omitted)
5. number a (optional, omitted), string b, (rest, optional, omitted)

But, rejected.  And I don't know what to fix, because I'm already compliant with the docs.

(I can make an educated guess that you actually want me to follow [chrome's rules](https://developer.chrome.com/extensions/manifest/version), "One to four dot-separated integers identifying the version of this extension." -- except I've had non-numeric characters work in the past so I think that's not _quite_ it.)

arantius:
Interesting, moving the issue here has helped me find:

https://github.com/mozilla/addons-linter/blob/053bb7a515d3c1d6fd8832da21ff3c5fa77bee34/src/schema/formats.js#L12

The actual code (I think) which does this validation.  Which seems oddly restrictive, but I can reverse engineer into something that I can generate.

Out of curiosity:

* Why a max of 4 parts?
* Why no leading zeros?

[...]

PR #2181, opened 2018-09-07
Import Firefox 63 schema

EnTeQuAk:
* Import Firefox 63 schema
* Add new `manifestShortcutKey` format that got added in the 63 import


src/schema/deepmerge.js
```diff
@@ -32,11 +29,16 @@ function oldArrayMerge(target, source, optionsArgument) {
...
+  return [...target, ...source].filter(
+    (element, index, array) => array.indexOf(element) === index
+  );
+}
+
+export const deepmerge = (a, b) => {
+  return merge(a, b, { arrayMerge: mergeSimpleArrays });
+};
+
+export const deepmergeWithComplexArrays = (a, b) => {
```
rpl: So, I'm thinking about other possible names that we could use to make it a bit clearer what the expected behavior would be, I think that "complex" and "simple" arrays are not making it clear enough what is the outcome that we expect from these two functions.

How about using the following names?

- `deepmerge` -> `deepmerge`
  this would be the only actual deepmerge, and the array are going to be concatenated

- `deepmergeWithComplexArray` -> `patch` or `deeppatch` (or `deepPatch` or `mergePatch`)
  This will be the "patching" helpers used to apply our "addons linter patches" to the imported schemas, and this is always "patching" the first array using the items in the second array, instead of "concatenating" them
 
- `mergeSimpleArrays` -> `concatArrays` (or `concatenateArrays`)
- `mergeComplexArrays` -> `patchArrays`

@EnTeQuAk How it sounds to you?

src/schema/firefox-schemas-import.js
```diff
@@ -250,12 +250,12 @@ inner.updateWithAddonsLinterData = (firefoxSchemas, ourSchemas) => {
       schemas[ourSchema.id || namespace] = {
         ...firefoxSchema,
         file,
-        schema: merge(firefoxSchema.schema, ourSchema),
+        schema: deepmergeWithComplexArrays(firefoxSchema.schema, ourSchema),
```
rpl: Let's add an inline comment here and at line 258 to briefly explain that here we are using a different algorithm to merge the two objects because we are actually "patching" the imported schema with some "addons-linter"-specific tweaks.

src/schema/formats.js
```diff
@@ -109,3 +109,62 @@ export function imageDataOrStrictRelativeUrl(value) {
 }
 
 export const isUnresolvedRelativeUrl = isStrictRelativeUrl;
+
+export function manifestShortcutKey(value) {
+  // Partially taken from Firefox directly via
```
rpl: I guess that we should also mention here how we will have to handle updates on the validation of the shortcut keys:

now that the allowed keys are not part of the schema files, "changes to the allowed shortcut keys" becomes something that we need to check manually when we update the schema files, by looking if there are any changes to the validation code on the Firefox side and be sure to replicate them here.

src/schema/imported/manifest.json
```diff
@@ -230,6 +230,10 @@
                       "preprocess": "localize"
                     }
                   }
+                },
+                "hidden": {
```
rpl: I'm wondering if we need to "guard" this "hidden" property, I think that this isn't something that a non-system addon should actually be allowed to use.

src/schema/validator.js
```diff
@@ -80,9 +82,9 @@ export const validateAddon = (...args) => {
 // structure and can use object-destructuring at the root level
```
rpl: Nit, I guess that this comment could use some tweak now that we have a better picture of why this parts need to use a different "merging" algorithm.

tests/unit/schema/test.commands.js
```diff
@@ -32,7 +32,7 @@ describe('/commands', () => {
     validateAddon(manifest);
     assertHasMatchingError(validateAddon.errors, {
       dataPath: '/commands/up/suggested_key/mac',
-      message: /should match pattern/,
+      message: 'should match format "manifestShortcutKey"',
```
rpl: uhm... I'm starting to think that the change to the manifest's shortcut keys used in the manifest is going to produce validation errors that may not be always immediately "actionable", the previous version was emitting a "monstrous" validation error that wasn't super-readable but it still listed the allowed patterns in a way that the addon developer could (more or less :-)) use to understand which are the allowed keys.

We may need to think how we can produce some better error messages now that the validation is being done from the js code instead of patterns included in the schemas.

src/schema/imported/manifest.json
EnTeQuAk: I can't find any documentation on it so I'm unsure tbh, I think this can be a follow-up as well.

tests/unit/schema/test.commands.js
EnTeQuAk: This is really hard tbh and there's also https://github.com/mozilla/addons-linter/issues/1905 which links to the same issue.

Personally, I'd like to tackle all that in a separate issue that may require restructuring how messages are returned etc. Especially since most of these advanced errors will require us to link the user to an external page which isn't really user-friendly at the moment and really painful.

The problem here, specifically is, the validator cannot return a custom error message because ajv doesn't allow that. I looked at that briefly and it's quite hard to customize and may even require patching ajv and working together with upstream to get custom error messages somehow into custom validation-formats.

I'll file a follow-up issue with more details if that works for you?

src/schema/deepmerge.js
EnTeQuAk: I completely agree, fixed in dc639e2c

src/schema/firefox-schemas-import.js
EnTeQuAk: Done in dc639e2c

src/schema/formats.js
EnTeQuAk: Done in dc639e2c

src/schema/validator.js
EnTeQuAk: done in dc639e2c

tests/unit/schema/test.formats.js
```diff
@@ -56,6 +57,52 @@ describe('formats', () => {
...
+      expect(manifestShortcutKey('F8')).toEqual(true);
+    });
+
+    it('Doesnt accept duplicate key', () => {
+      expect(manifestShortcutKey('Alt+Alt')).toEqual(false);
+    });
+
+    it('Doesnt accept shift with non-function key', () => {
+      expect(manifestShortcutKey('Shift+F8')).toEqual(true);
+      expect(manifestShortcutKey('Shift+Alt')).toEqual(false);
```
rpl:
```suggestion
      expect(manifestShortcutKey('Shift+Home')).toEqual(false);
```
Because 'Alt' will be recognize as a modifier, and so without a key the function is going to early exit as described in the other review comments above.

src/schema/imported/manifest.json
rpl: :+1: sure, that's totally fine, I mentioned it mainly to don't forget that I notice it and that we should investigate if the addon-linter should do something to ensure that it is handled correctly.

tests/unit/schema/test.commands.js
rpl: Absolutely, I agree that we can't (and we shouldn't) solve it as part of this PR, especially given that this PR has already been "exciting" enough :-)

EnTeQuAk:
> Have you already filed the followup issues?

On it, will do that in a bit.

wagnerand:
🎉

PR #2195, opened 2018-09-24
Throw an error for dictionaries without an id

### LINKED ISSUE #2188: Force webextension dictionaries to provide an id
When dealing with an extension without an `id`, Firefox extracts it from the signature. However, [it doesn't do this for dictionaries](https://bugzilla.mozilla.org/show_bug.cgi?id=1489106#c8). Therefore, we need to make sure all webextension dictionaries provide an explicit `id` themselves.

diox:
Fix #2188

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -223,6 +223,12 @@ export default class ManifestJSONParser extends JSONParser {
     }
 
     if (this.parsedJSON.dictionaries) {
+      const {applications: {gecko: {
+        id: addonId = null} = {}} = {}} = this.parsedJSON;
+      if (!addonId) {
+        this.collector.addError(messages.MANIFEST_DICT_MISSING_ID);
+        this.isValid = false;
+      }
       const numberOfDictionaries = Object.keys(this.parsedJSON.dictionaries)
         .length;
       if (numberOfDictionaries < 1) {
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -120,3 +120,4 @@ Rules are sorted by severity.
 | `MANIFEST_DICT_NOT_FOUND`   | error    | A dictionary file defined in the manifest could not be found         |
 | `MANIFEST_MULTIPLE_DICTS`   | error    | Multiple dictionaries found                                          |
 | `MANIFEST_EMPTY_DICTS`      | error    | Empty `dictionaries` object                                          |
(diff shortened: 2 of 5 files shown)


src/parsers/manifestjson.js
```diff
@@ -223,6 +223,12 @@ export default class ManifestJSONParser extends JSONParser {
     }
 
     if (this.parsedJSON.dictionaries) {
+      const {applications: {gecko: {
+        id: addonId = null} = {}} = {}} = this.parsedJSON;
```
EnTeQuAk: This took quite some time to understand tbh and I still don't quite get it :man_shrugging:

we already have two other places that traverse their way through `parsedJson.applications.gecko` so at this point I think it may make sense to do something like

```javascript
const geckoData = (this.parsedJSON.applications || {}).gecko || {}
```

to have some default values and re-use that in https://github.com/mozilla/addons-linter/pull/2195/files#diff-baa432d9354e3de899d28a2638d1499eR257 and following lines.

What do you think?

EnTeQuAk:
There are a few prettier issues though

```
/home/travis/build/mozilla/addons-linter/src/messages/manifestjson.js
  248:2  error  Insert `;`  prettier/prettier
/home/travis/build/mozilla/addons-linter/src/parsers/manifestjson.js
  226:14  error  Replace `applications:·{gecko:·{⏎········id:·addonId·=·null}·=·{}}·=·{}` with `⏎········applications:·{·gecko:·{·id:·addonId·=·null·}·=·{}·}·=·{},⏎······`  prettier/prettier
/home/travis/build/mozilla/addons-linter/tests/unit/helpers.js
  120:12  error  Insert `,`  prettier/prettier
/home/travis/build/mozilla/addons-linter/tests/unit/parsers/test.manifestjson.js
  1367:24  error  Replace `gecko:·{}` with `·gecko:·{}·`  prettier/prettier
```

You can simply run `npm run prettier` to run it through all or `npm run prettier-dev` only for modified files.

src/parsers/manifestjson.js
diox: Sure, that works for me. I can also use a `try`/`catch` if you prefer.

diox: Eh actually we have a `getAddonId()` function that does it with a `try`/`catch`. I'll use that.

PR #2225, opened 2018-10-05
 fix(package): update pino to version 5.6.4 

EnTeQuAk:
Closes #2197


EnTeQuAk:
Closing as upgrading pino requires some more work for which we currently don't have the time.

PR #2262, opened 2018-10-30
Remove greenkeeper-lockfile

mstriemer:
Greenkeeper supports updating package-lock.json without greenkeeper-lockfile now, so we no longer need this.


rpl:
\o/ Thanks Mark!
(related greenkeeper blogpost: https://blog.greenkeeper.io/announcing-native-lockfile-support-85381a37a0d0)

PR #2271, opened 2018-11-08
fix: Detect unsupported images used in static themes manifest

### LINKED ISSUE #2051: [Static Themes] Unsupported image formats are not detected by the linter at static theme submission
STR:

1. Log in to AMO -dev
2. Submit a static theme and make sure to incude an unsupported image format under `'images'` in your manifest -  i.e. a `.tif` file
3. Observe the validation results

Actual result:
There are no errors generated when a static theme package contains an unsupported image format

Expected result:
The linter should detect when an image format doesn't match the static theme requirements specified in the guideline - [ST guideline](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/theme)

Notes:
- when a theme with an unsupported image file is submitted, a preview will be generated but the image file can't be inspected in Rev Tools and the image is not applied in the browser when the theme is installed
- here is an example:
Rev Tools
https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/st-theme-zip-a
Listing page
https://addons-dev.allizom.org/en-US/firefox/addon/st-theme-zip-a/
- reproduced on AMO -dev with FF60, WIn10x64

rpl:
Fixes #2051 

This PR adds the following 5 new validations rules (also added to the docs/rules.md file in this PR):
- errors:
  - MANIFEST_THEME_IMAGE_NOT_FOUND (Theme images missing)
  - MANIFEST_THEME_IMAGE_CORRUPTED (The linter is unable to read their mime types)
  - MANIFEST_THEME_IMAGE_WRONG_EXT (Theme images must have one of the supported file extensions)
  - MANIFEST_THEME_IMAGE_WRONG_MIME (Theme images mime type must be a supported format)

- warnings:
  - MANIFEST_THEME_IMAGE_MIME_MISMATCH (the format is supported, as well as the file extension, but the mime doesn't match the file extension the linter expects) 

**TODO list**
- [x] should we support webp as a valid static theme image?
 - opted to currently disallow it as [it is not yet officially supported based on the relevant MDN doc page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme#Image_formats)   
- [ ] ~~double-check the additional tests are enough to bring back the code coverage to the current status~~ why coveralls is not in the list of the checks? :-(


src/parsers/manifestjson.js
```diff
@@ -392,6 +394,87 @@ export default class ManifestJSONParser extends JSONParser {
     return Promise.all(promises);
   }
 
+  validateStaticThemeImages() {
+    const promises = [];
+    const themeImages = this.parsedJSON.theme && this.parsedJSON.theme.images;
+
+    const validateThemeImage = async (imagePath, manifestPropName) => {
```
mstriemer: Could this be put onto `ManifestJSONParser` instead of being created in the call to `validateStaticThemeImages`?

src/parsers/manifestjson.js
```diff
@@ -392,6 +394,87 @@ export default class ManifestJSONParser extends JSONParser {
...
+        this.collector.addError(
+          messages.manifestThemeImageWrongExtension({ path: _path })
+        );
+        this.isValid = false;
+        return;
+      }
+
+      try {
+        const info = await getImageMetadata(this.io, _path);
+        if (
```
mstriemer: This is definitely a style thing, so up to you, but I'd expect this to be more like:

```js
if (!STATIC_THEME_IMAGE_MIMES.includes(info.mime)
  || info.mime === 'image/webp') {
```

src/const.js
```diff
@@ -109,6 +109,19 @@ export const IMAGE_FILE_EXTENSIONS = [
   'svg',
 ];
 
+// Map the image mime to the expected file extensions
+// (used in the the static theme images validation).
+export const IMAGE_MIMES_TO_EXTENSIONS = {
```
mstriemer: I feel like this name could be worded a bit better, but it seems tough without making it really long. If you can come up with something better I wouldn't be opposed to it ;). I think MIME should be singular and EXTENSIONS might be too overloaded for add-ons code.

Maybe `IMG_MIME_TO_FILE_EXTS` or `MIME_TO_FILE_EXTENSIONS`?

src/parsers/manifestjson.js
rpl: sure thing, I've extracted this method in the updated version of this PR.

src/parsers/manifestjson.js
rpl: eh... to be honest I was aiming for something like that, but prettier strongly disagrees :-)
we have been fighting a bit each other (prettier and I) but it quickly started to look as a "quite pointless battle"  and I decided that I can live happy enough with the version that makes "prettier" happy :-)

PR #2285, opened 2018-11-20
add browser_specific_settings as alias for applications in manifest

### LINKED ISSUE #2280: Add support for 'browser_specific_settings' as an alias to 'applications' in the manifest
Following https://github.com/mozilla/addons-server/issues/9989, let's add support for `browser_specific_settings` to the linter as well.

I haven't checked in detail what needs to be updated, but there are a few code paths that touch `applications`. We should probably also have at least one test that uses `browser_specific_settings`.

Note: This should already be in the schema, if I understand correctly, added as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1262005.

eviljeff:
First pr so likely horribly wrong.  Is supposed to fix #2280.  

(Also, prettier rules? I thought pep8 rules were arbitrary! 😱)

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -70,6 +70,12 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
+      if (
+        this.parsedJSON.browser_specific_settings &&
+        !this.parsedJSON.applications
+      ) {
+        this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
+      }
       this.selfHosted = selfHosted;
       this.isLanguagePack = Object.prototype.hasOwnProperty.call(
         this.parsedJSON,
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -70,10 +70,7 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
-      if (
-        this.parsedJSON.browser_specific_settings &&
-        !this.parsedJSON.applications
-      ) {
+      if (this.parsedJSON.browser_specific_settings) {
         this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
       }
       this.selfHosted = selfHosted;
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -70,7 +70,10 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
-      if (this.parsedJSON.browser_specific_settings) {
+      if (
+        this.parsedJSON.browser_specific_settings &&
+        this.parsedJSON.browser_specific_settings.gecko
+      ) {
         this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
       }
       this.selfHosted = selfHosted;
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -46,7 +46,7 @@ describe('ManifestJSONParser', () => {
     const addonLinter = new Linter({ _: ['bar'] });
     const json = validManifestJSON({
       applications: undefined,
-      browser_specific_settings: {gecko: { id: 'foo@baa' } },
+      browser_specific_settings: { gecko: { id: 'foo@baa' } },
     });
     const manifestJSONParser = new ManifestJSONParser(
       json,
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -42,6 +42,21 @@ describe('ManifestJSONParser', () => {
     expect(metadata.version).toEqual(null);
   });
 
+  it('should see browser_specific_settings as alias of applications', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const json = validManifestJSON({
+      applications: undefined,
+      browser_specific_settings: {gecko: { id: 'foo@baa' } },
+    });
+    const manifestJSONParser = new ManifestJSONParser(
+      json,
+      addonLinter.collector
+    );
+    expect(manifestJSONParser.isValid).toEqual(true);
+    const metadata = manifestJSONParser.getMetadata();
+    expect(metadata.id).toEqual('foo@baa');
+  });
+
   describe('id', () => {
     it('should return the correct id', () => {
       const addonLinter = new Linter({ _: ['bar'] });
--- a/src/messages/javascript.js
+++ b/src/messages/javascript.js
@@ -151,8 +151,8 @@ function temporaryAPI(api) {
     message: i18n._(`"${api}" can cause issues when loaded temporarily`),
     description: i18n._(oneLine`This API can cause issues when loaded
       temporarily using about:debugging in Firefox unless you specify
-      applications > gecko > id in the manifest. Please see:
-      https://mzl.la/2hizK4a for more.`),
+      applications|browser_specific_settings > gecko > id in the manifest.
+      Please see: https://mzl.la/2hizK4a for more.`),
   };
 }
 
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -101,8 +101,10 @@ export const PROP_VERSION_TOOLKIT_ONLY = {
 export const MANIFEST_UPDATE_URL = {
   code: 'MANIFEST_UPDATE_URL',
   message: i18n._('"update_url" is not allowed.'),
-  description: i18n._(oneLine`"applications.gecko.update_url" is not allowed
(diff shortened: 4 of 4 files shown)


src/parsers/manifestjson.js
```diff
@@ -70,6 +70,12 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
+      if (
+        this.parsedJSON.browser_specific_settings &&
+        !this.parsedJSON.applications
+      ) {
+        this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
+      }
```
ExE-Boss: Actually, in Firefox, `browser_specific_settings` takes precedence over `applications` if both are specified.

rpl: @ExE-Boss is right (thanks for pointing it out!)

@eviljeff here is the relevant part of the implementation on the Firefox side:

```
  let bss = (manifest.browser_specific_settings && manifest.browser_specific_settings.gecko)
      || (manifest.applications && manifest.applications.gecko) || {};
  if (manifest.browser_specific_settings && manifest.applications) {
    logger.warn("Ignoring applications property in manifest");
  }
```

from https://searchfox.org/mozilla-central/rev/55895c49f55073d82d977cb74ec1d3a71ae4b25f/toolkit/mozapps/extensions/internal/XPIInstall.jsm#443-444

tests/unit/parsers/test.manifestjson.js
```diff
@@ -42,6 +42,21 @@ describe('ManifestJSONParser', () => {
     expect(metadata.version).toEqual(null);
   });
 
+  it('should see browser_specific_settings as alias of applications', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const json = validManifestJSON({
+      applications: undefined,
+      browser_specific_settings: {gecko: { id: 'foo@baa' } },
```
rpl: @eviljeff this is [the line that is currently triggering the eslint failure on travis](https://travis-ci.org/mozilla/addons-linter/jobs/457538319#L1133-L1134) (and I guess that on appveyor the failure is likely the same).

it is related to prettier :-), it looks that it just wants a space between `{` and `gecko`.

rpl:
> (Also, prettier rules? I thought pep8 rules were arbitrary! scream)

hehe, happy that I'm not the only one that from time to time engage a `"thank you, but I actually want this formatted like it is"`-fight with prettier :-)

src/parsers/manifestjson.js
eviljeff: https://github.com/mozilla/addons-linter/pull/2285/commits/0f73e1805aaca5a9f1dfb1f9d3214a7ff0a922f9

tests/unit/parsers/test.manifestjson.js
eviljeff: https://github.com/mozilla/addons-linter/pull/2285/commits/0f73e1805aaca5a9f1dfb1f9d3214a7ff0a922f9

src/parsers/manifestjson.js
```diff
@@ -70,6 +70,9 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
+      if (this.parsedJSON.browser_specific_settings) {
+        this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
```
rpl: @eviljeff Given that on the firefox side the `browser_specific_settings` override the `applications` property only if it also contains the `gecko` property, I think that we should do the same on the linter side.

As an example, if an extension manifest.json have the `browser_specific_settings` property with a sub-property related to another browser, e.g. edge, and also the `applications` property with the 'gecko' property, on the Firefox side we will be using the content of `applications`, and so the linter should possibly do the same.

src/parsers/manifestjson.js
```diff
@@ -70,6 +70,9 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
+      if (this.parsedJSON.browser_specific_settings) {
```
rpl: Besides overriding the content of `applications` with `browser_specific_settings`, I'm thinking that we should also emit a linting error if both `applications.gecko` and `browser_specific_settings.gecko` exists in the same extension manifest.

@eviljeff (based on what we are actually doing on the AMO server side) what do you think? do you see any reason to don't block the submissions of extensions with both the manifest properties?

wagnerand: It would be best to follow Firefox' behavior as close as possible. It would be a bad experience for developers if they successfully tested their add-on in Firefox (`about-debugging` or `web-ext`) and then are presented with an error when they try to submit to AMO.

wagnerand: > @wagnerand there are already some cases where the extension install successfully on Firefox but linting errors are reported when checked with the addons-linter, e.g. if the manifest contains a content script which is missing from the resources actually included in the extension package.

Can you file a bug about that, please? The linter shouldn't behave differently than Firefox.

I see your point about potentially surprising behavior, but I think using a warning (and not an error) would be best here.

rpl: > I see your point about potentially surprising behavior, but I think using a warning (and not an error) would be best here.

Sounds good to me, if a linting warning is the behavior that we agree on, then I'm also ok with it:
it matches Firefox behavior and the linter will not be totally silent if it met an extension with this kind of potential issue.

> > @wagnerand there are already some cases where the extension install successfully on Firefox but linting errors are reported when checked with the addons-linter, e.g. if the manifest contains a content script which is missing from the resources actually included in the extension package.
> 
> Can you file a bug about that, please? The linter shouldn't behave differently than Firefox.

Filing a new issue to discuss and evaluate the removal of that linting rule sounds good to me, but to be honest I'm not sure that we should remove it completely, one of the reasons why we have that linting rule is that in the past some addon developers were confused about "why suddenly their extension's content script wasn't working anymore", and it was often due to issues with the file names (e.g. a file got renamed and it wasn't updated in the manifest to reflect the new file name).

I agree that the addons-linter should behave as similar as possible to Firefox, but in some cases Firefox is less strict about "which issues should prevent the extension from install successfully" to make it easier to test and investigate these issues or for "performance" reasons/concerns (e.g. when checking it explicitly at runtime in Firefox may be too expensive). 

On the other hand the role of the addons-linter (at least in my own opinion) is to provide some additional guarantees that the extension is going to work as it should for the user that installs it from AMO (at least in the cases where it is actually possible in practice).

wagnerand: I agree with what you are saying and I believe warnings would be the right level for those things, where we do more checks than Firefox. We don't need to remove such rules entirely.

eviljeff: Can you file a follow-up issue for adding a warning for having both properties? - I've not got capacity to keep working on this to add something extra right now. If you think it needs to be landed as a single patch I'll leave this issue for someone else to finish.

src/parsers/manifestjson.js
eviljeff: https://github.com/mozilla/addons-linter/pull/2285/commits/93e6b9cfe8f5e044e63cac09253c3953357fe0fc

src/parsers/manifestjson.js
rpl: @eviljeff sure thing, I'll deal with the follow up, I just filed it as #2294.

src/parsers/manifestjson.js
```diff
@@ -70,6 +70,12 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
+      if (
```
rpl: @eviljeff we should move this into `_validate`, after the manifest has been validated using the schemas imported from Firefox, otherwise every schema validation error/warning related to the `browser_specific_settings` will be logged twice: once for `browser_specific_settings` and then again for `applications` ( because we have already overwritten "applications" with the content of "browser_specific_settings").

If you prefer I can apply this change for you, or you may apply the following diff and update this PR and I'll squash and merge it right away: 

```
index 6314f6e..30636cb 100644
@@ -70,12 +70,6 @@ export default class ManifestJSONParser extends JSONParser {
       };
     } else {
       // We've parsed the JSON; now we can validate the manifest.
-      if (
-        this.parsedJSON.browser_specific_settings &&
-        this.parsedJSON.browser_specific_settings.gecko
-      ) {
-        this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
-      }
       this.selfHosted = selfHosted;
       this.isLanguagePack = Object.prototype.hasOwnProperty.call(
         this.parsedJSON,
@@ -174,6 +168,13 @@ export default class ManifestJSONParser extends JSONParser {
       });
     }
 
+    if (
+      this.parsedJSON.browser_specific_settings &&
+        this.parsedJSON.browser_specific_settings.gecko
+    ) {
+      this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
+    }
+
     if (this.parsedJSON.content_security_policy) {
       this.validateCspPolicy(this.parsedJSON.content_security_policy);
     }
```

PR #2309, opened 2018-12-05
Move internal scripts to separate folder

### LINKED ISSUE #2278: Transpile before publishing

I'm getting the following warning when installing addons-linter:

```js
npm WARN @babel/register@7.0.0 requires a peer of @babel/core@^7.0.0-0 but none is installed. You must install peer dependencies yourself.
```

It seems like this project is using `@babel/register` to transform the source code into compatible JS at _runtime_. It's seems like babel is only used in the `bin/firefox-schema-import` file, which means that my library usage of this is probably safe from crashing because of missing `@babel/core`. However, there are still some drawbacks with this approach. Apart from the warning that's a bit annoying/scary, I also get to install a lot of babel-dependencies which won't be used at all.

Would it be possible to simply transpile the code before publishing to npm? I think that a simple `"prepublish": "babel src --output-dir lib"`, `"files": ["lib/"]`, and changing the require to point to `lib/` instead of `src/` in `bin/firefox-schema-import` would suffice.

LinusU:
Fixes #2278 

This PR moves internal scripts to another folder from `bin/`, where the public scripts will live alone now. It also updates the babel dependencies to be in `devDependencies`, since they are only used in the internal scripts.


LinusU:
Hehe, a linting error, on it 👌

LinusU:
```text
FAIL tests/integration/addons-linter/integration.run-linter-no-params.js
  ● Basic integration test › linter should fail if ran without params with explanation
    expect(string).toContain(value)
    
    Expected string:
      ""
    To contain value:
      "Not enough non-option arguments: got 0, need at least 1"
       5 |     const { exitCode, stderr } = await executeScript('addons-linter');
       6 |     expect(exitCode).toBe(1);
    >  7 |     expect(stderr).toContain(
       8 |       'Not enough non-option arguments: got 0, need at least 1'
       9 |     );
      10 |   });
      
      at Object._callee$ (tests/integration/addons-linter/integration.run-linter-no-params.js:7:20)
      at tryCatch (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
```

Hmm, not sure what's going on here, the tests are passing for me locally 🤔

LinusU:
Ah, I know what it is 👌

Removing the babel package as dependencies also removed `regenerator-runtime`, since there was no dependency on that, which was still used. Adding that to the normal dependencies now...

LinusU:
There we go, @EnTeQuAk, should be good to go now 👍

PR #2323, opened 2018-12-13
chore(package): Moved snyk and shelljs to devDependencies

rpl:
I've been looking a bit into a "slimming cure" to reduce the amount and size of the dependencies that web-ext is currently installing in "production" mode and it looks that the addons-linter is (unsurprisingly) one of the bigger ones.

By taking a look into which are the bigger npm dependencies on the addons-linter, I noticed that the winner is: **snyk**

Based on `cost-of-modules` output snyk introduces 344 childrens npm modules, and +43.67M is the resulting impact in size.

Snyk doesn't seem to be really necessary as a "production" dependency, and so we should be able to just move it to the devDependencies and spare our users a good amount of download time (and it should also be helpful to speed up web-ext's travis CI jobs).

While we are on the "slimming cure"-theme, we can also move shelljs to the devDependencies, as the only scripts that are using it are only used during development (and they have been moved out of the 'bin/' directory in #2309).


rpl:
> Shouldn't this require a re-generation of package-lock.json?

@EnTeQuAk We definitely need to re-generate it, in this PR or as part of the one that release the new addons-linter version on npm.

I'm fine with any of the two approaches, let me know if you prefer that I include the re-generate package-lock.json into this PR and I'll update the patch right now.

EnTeQuAk:
If possible I'd prefer it to be in this PR I think


On Fri, Dec 14, 2018, at 11:08 AM, Luca Greco wrote:
>> Shouldn't this require a re-generation of package-lock.json?


> @EnTeQuAk[1] We definitely need to re-generate it, in this PR or as
> part of the one that release the new addons-linter version on npm.> I'm fine with any of the two approaches, let me know if you prefer
> that I include the re-generate package-lock.json into this PR and I'll
> update the patch right now.> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
Links:

  1. https://github.com/EnTeQuAk
  2. https://github.com/mozilla/addons-linter/pull/2323#issuecomment-447278207

PR #2363, opened 2019-01-16
Port to fluent-syntax 0.10.0, switch to warnings.

### LINKED ISSUE #2305: Update fluent-syntax to pick up bug fixes.

AMO refused to sign language packs for en-US 'cause the files triggered a bug in fluent-syntax 0.7. See https://bugzilla.mozilla.org/show_bug.cgi?id=1511988 for details

### What happened?

Beta 1 busted

Not bust :-)

I tested the different versions, 0.8 had a different bug, 0.9.0 is fine.

CC @callek

### LINKED ISSUE #2313: Please switch fluent lint errors to not block addon signing
We would like fluent files in language packs to still give messages about issues but not actually block language pack signing/publishing.

I'm flagging @Pike and @flodolo to comment and give this a thumbs up from a localization team perspective.

Some background:
For the past few releases and early on we've had issues that were flagged by AMO as errors with fluent, however these issues are not blockers for our users; and when they strike they block a release process (chemspill included, though luckily its never blocked in that urgent of a time) but means these issues put AMO developers and admins in a critical path for releases, liable to be paged or be asked to have sudden deploys. Which is counter to what we all want.

Some of the notable issues we've hit with fluent on AMO:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1511988 - which spawned issue #2305
    * This one was an issue in validator version...
* https://bugzilla.mozilla.org/show_bug.cgi?id=1502234
    * This one was an actual issue in files contained in the language packs `however` the way fluent (ftl) works is that if these strings were invalid they wouldn't get used and it would fallback to english anyway, which would not harm the build
    * Additionally l10n team signs off on translation validity externally before we take a changeset for a release
    * In this particular case it was for a .ftl file that wasn't even used in the Gecko version of the build that we were shipping for (all .ftl files that exist are shipped as part of the language packs, even if they are for "newer" trains)
* Back in march we also had issue #1901 which burned some cycles.

EnTeQuAk:
Fixes #2313
Fixes #2305

Closes #2321

### DIFF
--- a/tests/unit/parsers/test.fluent.js
+++ b/tests/unit/parsers/test.fluent.js
@@ -12,7 +12,7 @@ describe('FluentParser', () => {
     const parser = new FluentParser(
       `
 choose-download-folder-title =
-  {
+  { $faz ->
     *[nominative] Foo
      [accusative] Foo2
   }`,
@@ -28,13 +28,13 @@ choose-download-folder-title =
     const addonLinter = new Linter({ _: ['bar'] });
     const parser = new FluentParser(
       `
-key67
+key67 =
     .label = Sign In To &syncBrand.shortName.label;…
     .accesskey = Y
-key68
+key68 =
     .label = Sync Now
     .accesskey = S
-key69
+key69 =
     .label = Reconnect to &syncBrand.shortName.label;…
     .accesskey = R`,
       addonLinter.collector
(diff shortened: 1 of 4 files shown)


tests/unit/parsers/test.fluent.js
```diff
@@ -28,13 +28,13 @@ choose-download-folder-title =
...
-key67
+key67 =
     .label = Sign In To &syncBrand.shortName.label;…
     .accesskey = Y
-key68
+key68 =
     .label = Sync Now
     .accesskey = S
-key69
+key69 =
```
EnTeQuAk: Changes in the actual spec, https://github.com/projectfluent/fluent/blob/master/spec/CHANGELOG.md

PR #2408, opened 2019-01-30
Fix broken test on Windows

wagnerand:


tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -29,7 +29,7 @@ describe('firefox schema import', () => {
   // Skip the Firefox schema import tests on windows.
   if (process.platform === 'win32') {
     // eslint-disable-next-line jest/no-disabled-tests
-    test.skip('skip test suite on Windows');
+    this.skip();
```
rpl: It looks that the way to skip the test on Jest (which uses jasmine internally), is to call `pending();` instead of `this.skip()`:
```diff
diff --git a/tests/unit/schema/test.firefox-schemas-import.js b/tests/unit/schema/test.firefox-schemas-import.js
index 780e815..586e16a 100644
--- a/tests/unit/schema/test.firefox-schemas-import.js
+++ b/tests/unit/schema/test.firefox-schemas-import.js
@@ -29,7 +29,7 @@ describe('firefox schema import', () => {
   // Skip the Firefox schema import tests on windows.
   if (process.platform === 'win32') {
     // eslint-disable-next-line jest/no-disabled-tests
-    this.skip();
+    pending();
     return;
   }
```

We also have to tweak the eslint config, otherwise eslint will raise a "'pending' is not defined" linting error, because it doesn't know about pending. The following change fixed the eslint error for me locally:

```diff
diff --git a/tests/.eslintrc b/tests/.eslintrc
index d24a7e2..11cbab1 100644
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -1,6 +1,7 @@
 {
   "env": {
     "jest": true,
+    "jest/globals": true,
   },
   "extends": [
     "plugin:promise/recommended",
```

wagnerand: Thanks! Fixed in 50c4a1e5861593c73293ce79e32077735dbaecc0

PR #2410, opened 2019-01-30
Configure renovate, force exact package-lock versions.

### LINKED ISSUE #2409: Configure renovate
Get inspiration from addons-frontend

EnTeQuAk:
This also force-replaces any http urls in package-lock to be https
again.

Fixes #2409

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,20 @@
 {
+  "timezone": "Europe/Berlin",
   "extends": [
-    "config:base"
+    "config:base",
+    ":maintainLockFilesWeekly",
+    ":prNotPending"
+  ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": "before 4pm on Tuesday"
+  },
+  "packageRules": [
+    {
+      "packagePatterns": [
+        "*"
+      ],
+      "rangeStrategy": "pin"
+    }
   ]
 }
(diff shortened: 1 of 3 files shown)


renovate.json
```diff
@@ -1,5 +1,20 @@
 {
+  "timezone": "Europe/Berlin",
```
muffinresearch: Any particular reason for Berlin - maybe UTC would be more universal?

renovate.json
```diff
@@ -1,5 +1,20 @@
...
   "extends": [
-    "config:base"
+    "config:base",
+    ":maintainLockFilesWeekly",
+    ":prNotPending"
+  ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "schedule": "before 4pm on Tuesday"
```
wagnerand: Could we move this to end of Monday or early Tuesday, so we have enough time to deal with potential breakings?

willdurand: I think the default is on Monday morning (European time), which I find convenient.

renovate.json
EnTeQuAk: Indeed, fixed in 7f775ffb

renovate.json
EnTeQuAk: early Tuesday makes sense, that way we can make a new release of dispensary on Monday ideally.

Fixed in 7f775ffb

PR #2412, opened 2019-01-30
Remove eslint-plugin-amo

wagnerand:
I still don't know how to update package-lock.json properly though...


wagnerand:
Ah, looks like *renovate* will do that for us during the weekly lockfile maintenance.

EnTeQuAk:
> Ah, looks like renovate will do that for us during the weekly lockfile maintenance.

Correct, for removing dependencies that works I guess.

PR #2417, opened 2019-01-31
Update to node 10

### LINKED ISSUE #2416: Update to node 10
Let's update to node 10.

wagnerand:
Fix #2416.


EnTeQuAk:
This will drop compatibility with web-ext, as long as web-ext doesn't switch to node 10 we shouldn't do the same, or at least be very certain that we don't have any major updates coming in that web-ext might need (e.g schema updates)

EnTeQuAk:
I think we can attempt to merge this once the Firefox 66 schema update landed.

wagnerand:
I also don't mind waiting for web-ext (if they upgrade sooner, rather than later :) )

package.json
```diff
@@ -6,6 +6,9 @@
   "bin": {
     "addons-linter": "bin/addons-linter"
   },
+  "engines": {
+    "node": ">=10.15"
+  },
```
EnTeQuAk: Let's not enforce this requirement. We are advertising compatibility but as of right now we don't break compatibility with earlier nodejs versions so we shouldn't force users to do so if they dare to use node 8 with the linter.

wagnerand:
I am closing this since web-ext is not moving to node 10 anytime soon.

PR #2419, opened 2019-01-31
Import Firefox 66 beta3 schema.

### LINKED ISSUE #2418: Import Firefox 66 beta schemas

EnTeQuAk:
Fixes #2418

### DIFF
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -104,7 +104,7 @@
                   "enum": [
                     "spanning"
                   ],
-                  "onError": "warn"
+                  "default": "spanning"
                 },
                 "background": {
                   "anyOf": [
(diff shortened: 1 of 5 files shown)


src/schema/imported/manifest.json
```diff
@@ -104,7 +104,7 @@
                   "enum": [
                     "spanning"
                   ],
-                  "onError": "warn"
+                  "default": "spanning"
```
rpl: Just as a side note, this change means that extensions with "not_allowed" or "split" as the requested incognito mode will start to get errors when submitted on AMO.

The error message we currently report for this validation error is unfortunately not super useful, as it doesn't currently say which are the allowed values (currently just "spanning"):
```
Code           Message                           Description                           File   Line   Column
JSON_INVALID   "/incognito" should be equal to   Your JSON file could not be parsed.
               one of the allowed values
```

Shane checked with Mike how many of such extension we have, and they should be very few (I recall there were less then 10).

Anyway, Firefox will also raise an error an refuse to install the extension, and so the linter and AMO behavior is going to be consistent with Firefox.

wagnerand:
That test is flaky, I have seen it failing before. Restarting should fix it.

rpl:
> That test is flaky, I have seen it failing before. Restarting should fix it.

@wagnerand good, that's what I suspected, thanks for confirming it! 
do you know if it is flaky only on appveyor or also on travis?

wagnerand:
I don't remember exactly, sorry.

PR #2567, opened 2019-05-03
Remove unused dependencies

wagnerand:
I can't find usage of any of these packages.
@EnTeQuAk could you double check those are actually unused?


package.json
```diff
@@ -64,15 +64,13 @@
     "po2json": "0.4.5",
     "postcss": "7.0.14",
     "probe-image-size": "4.0.0",
-    "regenerator-runtime": "0.13.2",
```
EnTeQuAk: I think is was a dependency of `@babel/plugin-transform-runtime` and used by many other packages. I suppose we pinned it because of some dependency problems in the past so it should be safe.

Let's double check by building an add-ons linter build and try to run it.

package.json
```diff
@@ -64,15 +64,13 @@
...
     "relaxed-json": "1.0.3",
     "semver": "6.0.0",
     "source-map-support": "0.5.12",
     "strip-bom-stream": "3.0.0",
     "tosource": "1.0.0",
     "upath": "1.1.2",
     "whatwg-url": "7.0.0",
-    "xmldom": "0.1.27",
```
EnTeQuAk: indeed unused

package.json
```diff
@@ -109,10 +107,8 @@
     "gh-pages": "2.0.1",
     "gunzip-maybe": "1.4.1",
     "hashish": "0.0.4",
-    "isparta": "4.1.1",
```
EnTeQuAk: :+1:

package.json
```diff
@@ -109,10 +107,8 @@
...
     "jest": "24.7.1",
     "jest-raw-loader": "1.0.1",
-    "json-loader": "0.5.7",
```
EnTeQuAk: imho this is being used for imports like `import badwords from './badwords.json';` but hmm… looking a the package and reading "⚠️ Since webpack >= v2.0.0, importing of JSON files will work by default. You might still want to use this if you use a custom file extension. See the v1.0.0 -> v2.0.0 Migration Guide for more information" - I guess this is safe to remove :laughing:

package.json
```diff
@@ -121,7 +117,6 @@
     "prettier": "1.17.0",
     "pretty-quick": "1.10.0",
     "raw-loader": "2.0.0",
-    "request": "2.88.0",
```
EnTeQuAk: :+1:

wagnerand:
Thanks! I will look into the test failure later, this has low prio at the moment.

package.json
wagnerand: Tested and verified that the linter still works.

wagnerand: Actually, we do need this for the integration test to pass...

PR #2609, opened 2019-05-24
Fixed #1931

### LINKED ISSUE #1931: Local modules could over shadow node_modules, preventing imports

Add an import like this to any local module, such as for [`logger.js`](https://github.com/mozilla/addons-linter/blob/cc2cb5b1f6ca0e72924972c030066ecd40d42fc9/src/logger.js):

```js
import log from 'logger';
```

Install an npm package by the [same name](https://www.npmjs.com/package/logger), just as an example.

### What happened?

There is no way to import the `logger` module from `node_modules` because `src` is first in the [webpack resolve config](https://github.com/mozilla/addons-linter/blob/cc2cb5b1f6ca0e72924972c030066ecd40d42fc9/webpack.config.js#L56)

Local modules should not overshadow `node_modules`.

You could move all modules under `src/` to `src/linter/` to enable a namespaced import like:

```js
import log from 'linter/logger';
```

devcer:
Changes:
* Moved cli.js, collector.js, const.js, libraries.js, linter.js, logger.js, message.js, miner_blocklist.js, utils.js, yargs-options.js from src/ to src/linter
* Fixed all the paths related to above change
* Fixed all `npm run lint` issues identified while fixing paths

The test cases are however failing with an unidentified issue even after fixing paths. Hope you could help me in fixing it.

Fixes #1931


muffinresearch:
@devcer Thanks for contributing, unfortunately we've decided to re-think this approach as per https://github.com/mozilla/addons-linter/issues/1931#issuecomment-505935343 so I'm going to close this out.

caitmuenster:
@devcer would you like to take a look at other [good first issues](https://addons-pm.herokuapp.com/contrib/good-first-bugs/?dir=desc&sort=updatedAt) and see if you'd like to pick up one of those instead?

PR #2617, opened 2019-05-29
Fix RESERVED_FILENAME message construction.

EnTeQuAk:
In 3888a7d I introduced reserved
filename scanning and forgot to define a `description` property on the
`RESERVED_FILENAME` message.

The `description` is needed to be properly exposable to the JSON and
text exports.

### DIFF
--- a/tests/unit/scanners/test.filename.js
+++ b/tests/unit/scanners/test.filename.js
@@ -159,7 +159,7 @@ describe('Reserved file names', () => {
       const { linterMessages } = await filenameScanner.scan();
       expect(linterMessages.length).toEqual(1);
       expect(linterMessages[0].code).toEqual(messages.RESERVED_FILENAME.code);
-      expect(linterMessages[0].code).toEqual(messages.RESERVED_FILENAME.code);
+      expect(linterMessages[0].message).toEqual('Reserved filename found.');
       expect(linterMessages[0].description.startsWith('Files whose names are reserved')).toEqual(true)
       expect(linterMessages[0].file).toEqual(filePath);
     });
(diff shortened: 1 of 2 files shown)


src/messages/manifestjson.js
```diff
@@ -134,10 +134,9 @@ export function manifestPropMissing(property) {
   };
 }
 
-export const MANIFEST_ICON_NOT_FOUND = 'MANIFEST_ICON_NOT_FOUND';
```
muffinresearch: I think these constants are used in tests so they should continue to be used.

muffinresearch:
> A message must always have a description set, this validates this. The validation is a bit helpless for messages that are exposed through a function but for the specific issue - that RESERVED_FILENAME was broken it works.

I'm not sure I completely follow this description - could you clarify what the branch is aiming for?

EnTeQuAk:
> Ah I think I see it now looking further at the code, this is adding a test for the description which is causing the newly added rule to break.

I re-did the PR completely in  df2e450 and made the diff a bit more manageable, especially given it's a regression fix only which should be cherry-picked into this weeks release.

tests/unit/scanners/test.filename.js
```diff
@@ -159,6 +159,8 @@ describe('Reserved file names', () => {
       const { linterMessages } = await filenameScanner.scan();
       expect(linterMessages.length).toEqual(1);
       expect(linterMessages[0].code).toEqual(messages.RESERVED_FILENAME.code);
+      expect(linterMessages[0].message).toEqual('Reserved filename found.');
+      expect(linterMessages[0].description.startsWith('Files whose names are reserved')).toEqual(true)
```
muffinresearch: These two lines could directly assert against `RESERVED_FILENAME.message` and `RESERVED_FILENAME.description` couldn't they?

EnTeQuAk: Then I actually wouldn't match/notice the `undefined` `description` property which is what I'm after here.

muffinresearch: Ah yes, fair enough 👍

PR #2638, opened 2019-06-11
Support CRX3 files

### LINKED ISSUE #2621: CRX3 files fail

Linting a .crx file that is in CRX3 (rather than CRX2) format fails.

This stems from the package dependency on `crx-parser`. I've submitted a PR that fixes this issue (https://github.com/shyiko/node-crx-parser/pull/2) but considering the library hasn't been published in 4 years I'm not optimistic that it'll get accepted and published.

If there's not a response from the original package maintainer, I'm considering publishing my fork to npm and making a PR to addons-linter to change the dependency, but wanted to get feedback from the addons-linter community before going that route.

Rob--W:
Fixes #2621

This is a minimal CRX parser that supports CRX2 and CRX3 files. It is based on https://github.com/Rob--W/crxviewer/blob/35d3d8af52b2775ac5296a42537e10f2b5711945/src/lib/crx-to-zip.js#L27-L49 . The CRX3 test fixture was generated with Chromium 74, based on the source of the existing `extension.crx` fixture.


EnTeQuAk:
Thanks a lot for the pull request! I'll have a look shortly.

Just so that I don't forget, there's also some CRX code on addons-server that might need updating then: https://github.com/mozilla/addons-server/blob/master/src/olympia/files/utils.py#L1167

Rob--W:
I've updated this PR by adding 3 tests to raise the test coverage.

kewisch:
Something to consider, this reads the entire zip into a buffer/RAM. Would it possibly make sense to use some of the code from https://github.com/thejoshwolfe/yauzl/blob/master/examples/unzip.js to instead read from the middle of the file skipping the header?

kewisch:
Oh I guess it wasn't different before, something to fix later then.

Rob--W:
@kewisch The use of `lazyEntries` doesn't help us here. This option is designed to make it easier for API consumers to uncompress one zip entry at a time. Regardless of the option, the zip file itself must be fully in memory in order to list its content (in theory, only the last part of the zip file is needed, since the linter is going to read other files anyway, we can as well keep the whole zip in memory. If one is only interested in file names in the zip file, then they can use https://github.com/Rob--W/zipinfo.js).

addons-linter is prone to excessive peak memory usage though, because it does eagerly read all files in parallel: https://github.com/mozilla/addons-linter/blob/c76579f88a22849a291cfd53fc11f6f11eac3ef5/src/linter.js#L596-L598
This is not necessarily a bad thing - if there is enough memory, then it is better to have a higher peak memory use in favor of faster linting results.

kewisch:
@Rob--W @EnTeQuAk what is the status of this issue? I'd love to get this fixed so we can get extensiontest.com working again.

EnTeQuAk:
>  This is not necessarily a bad thing - if there is enough memory, then it is better to have a higher peak memory use in favor of faster linting results.

We are limiting memory-size on the server-side which is why the linter doesn't read files > 4MB size for example.

tests/.eslintrc
```diff
@@ -11,6 +11,7 @@
     "sinon": true,
   },
   "plugins": [
+    "jest",
```
EnTeQuAk: :+1:

Rob--W: Oops. I messed up when rebasing. I guess that https://github.com/mozilla/addons-linter/pull/2639 can be closed now.

EnTeQuAk: I figured it'd be okay to just merge with this PR too.

PR #2639, opened 2019-06-12
Reference "jest" plugin in tests/.eslintrc

Rob--W:
Without referencing the plugin, trying to lint individual files in the tests/ directory fails with the following error:

```
$ ./node_modules/.bin/eslint tests/unit/io/test.crx.js
Error: path/to/addons-linter/tests/.eslintrc:
        Environment key "jest/globals" is unknown
```

### DIFF
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -11,6 +11,7 @@
     "sinon": true,
   },
   "plugins": [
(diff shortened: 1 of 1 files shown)


Rob--W:
This was accidentally included in #2638, which has just been merged.

PR #2643, opened 2019-06-13
Limit extensions id to 255 characters max (matches addons-server database limit)

### LINKED ISSUE #2642: Disallow submissions with a guid > 255 characters
- Submit an add-on that has an `id` in its manifest that obey the `@foo` format but is longer than 255 characters 

### What happened?
- Validation passes, but you get a 500 error (https://sentry.prod.mozaws.net/operations/olympia-dev/issues/5942419/?query=is:unresolved)

- You should get a validation error, because even though we now accept long ids (old change from 0e30d41a3ce0340e52fb3a44c0908d743d56f9d3 and cleaned up in 7dc2b45d01b0c3ec0cd22b312b3f17c0129414c9), the column in the database has a 255 character limit.

(Please include a link to the page, screenshots and any relevant files.)
- The limit should be applied even to unlisted add-ons. Maybe it would be simpler to just do this in the linter ?

diox:
Fixes #2642

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -102,6 +102,22 @@ describe('ManifestJSONParser', () => {
       });
     });
 
+    it('should fail on id longer than 255 characters', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        applications: { gecko: { id: '@' + 'a'.repeat(255) } },  // 256 chars.
+      });
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector
+      );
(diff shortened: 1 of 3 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -102,6 +102,22 @@ describe('ManifestJSONParser', () => {
       });
     });
 
+    it('should fail on id longer than 255 characters', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        applications: { gecko: { id: `@${'a'.repeat(255)}` } }, // 256 chars
```
EnTeQuAk: Mind documenting here, that we don't have to care about unicode validation as this is handled by the schema as well?

Issue #2259, opened 2018-10-30
Prepare to prevent use of the deprecated theme property names in new webextension themes submissions


rpl:
As part of [Bug 1472740 Remove aliased property names from themes](https://bugzilla.mozilla.org/show_bug.cgi?id=1472740), we should coordinate the deprecation of these "aliased theme property names" with the addons-linter, so that we can prevent new theme submissions from using the deprecated aliased properties.

rpl:
> @rpl Are we able to write a migration script which can convert themes hosted on AMO to use the non-aliased names?

@msujaws you mean a script to run locally or a conversion that is meant to happen on the AMO server side?

Looping in @eviljeff, he likely knows more about our plans related to the conversion of the existing LWT themes.

msujaws:
I mean a script that will run on AMO, changing these themes to use the non-aliased names, and which will in-turn trigger updates for Firefox clients to get the non-aliased version of the theme.

eviljeff:
> our plans related to the conversion of the existing LWT themes.

This is about webextension based themes that have been uploaded by users, rather than LWTs, afaik.

@msujaws can you log an issue on mozilla/addons-server?  A script is _possible_ but I wouldn't want to commit to something that may take significant dev & ops time without it going through Product, etc.

EnTeQuAk:
Marking as "in progress" given that there's already a PR up for this, that only needs some updating.

AlexandraMoga:
@rpl can we already verify that the properties mentioned in https://github.com/mozilla/gecko-dev/commit/5b052494ce948cff59bbb1692919df7ea19b1220 are no longer supported by the linter?

PR #2694, opened 2019-07-17
fix(themes): Log validation errors on theme manifests using the deprecated LWT aliases

### LINKED ISSUE #2463: Turn MANIFEST_THEME_LWT_ALIAS into a linting error once support for these aliases is removed in Firefox

As a follow up for #2259, once the support for these LWT theme aliases is going to be removed from Firefox, we should turn the `MANIFEST_THEME_LWT_ALIAS` (which is being added as part of #2283) from a linting warning to a linting error.

Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1472740#c10, the linter should start to mark this as a linting error when Firefox 69 reached the beta channel.

rpl:
This PR fixes #2463, follow up for #2259 (which started to log validation warning for these deprecated LWT aliases).

This is also related to [Bug 1472740](https://bugzilla.mozilla.org/show_bug.cgi?id=1472740), which tracks the changes on the Firefox side.

Some notes about this change:

- `toolbar_text` is not deprecated anymore (See [Bug 1472740 Comment 24](https://bugzilla.mozilla.org/show_bug.cgi?id=1472740#c24) for a rationale)
- The LWT aliases properties will be ignored on Firefox (starting from Firefox >= 70 if the patch is not going to be uplifted to 69), as it happens for any other unknown theme manifest property
- [The patch attached to Bug 1472740 ([D37891])](https://phabricator.services.mozilla.com/D37891) is going to update the deprecation message included in the schema to mention explicitly that the deprecated properties are being ignored (but it will not prevent the installation, as it also happens for any other unknown theme manifest property)

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -2334,7 +2334,7 @@ describe('ManifestJSONParser', () => {
     });
 
     describe('deprecated LWT aliases', () => {
-      it('does add validation warnings for LWT alias manifest properties marked as deprecated', () => {
+      it('does add validation errors for LWT alias manifest properties marked as deprecated', () => {
         const linter = new Linter({ _: ['bar'] });
         const manifest = validStaticThemeManifestJSON({
           theme: {
@@ -2344,6 +2344,9 @@ describe('ManifestJSONParser', () => {
             colors: {
               accentcolor: '#000',
               textcolor: '#000',
+              // This is not a deprecated property anymore, but it is still
+              // part of this test to ensure we don't raise linting errors
+              // for it.
               toolbar_text: '#000',
             },
           },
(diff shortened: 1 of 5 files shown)


rpl:
added @wagnerand as an additional reviewer, to also have an opinion from an "add-on review" point of view.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2358,34 +2361,30 @@ describe('ManifestJSONParser', () => {
 
         expect(manifestJSONParser.isValid).toBeFalsy();
 
-        expect(errors).toEqual([]);
+        expect(warnings).toEqual([]);
 
-        const actualWarnings = warnings.map((warn) => {
+        const actualErrors = errors.map((warn) => {
```
wagnerand: Nit: Could rename `warn` as well?

rpl: he, yeah, I missed that one :-)

PR #2727, opened 2019-08-01
Adding Snyk as development dependency.

### LINKED ISSUE #2702: `snyk` should be a development dependency
`snyk` has been added as a production dependency for this package when it appears to only be used for development purposes.

### What happened?
I upgraded [`web-ext`](https://github.com/mozilla/web-ext) to version `3.1.1` from `3.0.0`. I noticed a lot of additional transitive dependencies were added to my `package-lock.json`, and most of those new dependencies are a result of `snyk`.

I expected a number of updated packages in my `package-lock.json`, but not a lot of new ones.

`snyk` only appears to be used for its `snyk-protect` command, which is used by `addons-linter`'s `prepare` script. According to [npm's documentation](https://docs.npmjs.com/misc/scripts.html), `prepare` is only run for development purposes:

- when running `npm install` locally, e.g. when developing on `addons-linter`
- before packing/publishing the package to the npm registry
- when installing git dependencies (which I assume to mean installing this package as a git dependency)

It would appear that none of these cases would occur when a consumer of the package is installing it as a dependency, either directly or through `web-ext`.

I've opened an issue for this is `dispensary` as well: https://github.com/mozilla/dispensary/issues/608

rbrishabh:
Fixes #2702 
Adding Snyk as a development dependency and removing it from production dependencies.
Let me know if anything else needs to be done!

Thanks! =)


wagnerand:
Hey there, looks like the JSON file is not valid. I don't want to give too many hints upfront, but please let me know if you are stuck.

rbrishabh:
I am sorry! I tried to get ahead of myself by not running locally. 

Here we go :)

wagnerand:
Thanks, that looks better. Could you also update the version number? I just merged another PR that updates snyk to the latest version. Thank you!

rbrishabh:
> 
> 
> Looks great, thank you!

Yes! I would be around to solve more bugs! If I can be of service anywhere, just tag me :)
Welcome
!

PR #2741, opened 2019-08-12
Add codecov in place of coveralls

muffinresearch:
Replace coveralls with codecov.

Looks like coveralls is back, but we use codecov elsewhere so we might want to make this switch anyway 🤷‍♂️


muffinresearch:
Looks like coveralls is back, but we use codecov elsewhere so we might want to make this switch anyway 🤷‍♂️

PR #2747, opened 2019-08-15
Refactor deprecated handling.

### LINKED ISSUE #2462: Look into handling the deprecated keyword used in the API schemas

While working on the changes needed for #2283 we noticed that the "deprecated" keyword is not currently being handled, because 'src/schema/validator.js' is actually registering an handler for a  `"deprecated" format`, whereas it is actually a `keyword` and not a `format` (there is actually no `"deprecated" format` used in our WebExtensions API schemas):

https://github.com/mozilla/addons-linter/blob/590d2752a6fbcb9896b6b4f9e85e5d7d7c5770b1/src/schema/validator.js#L35

### What happened?

As part of #2283 we are going to ignore any "deprecated" keyword that is not related to the LWT theme aliases, otherwise we will likely introduce regressions related to deprecation errors that were not being collected.

It seems reasonable to take another look to the new linting errors that may be collected if we do handle the "deprecated" keyword for the rest of the API schemas, and then decide if any of them should be actually part of the collected linting errors/warnings.

### LINKED ISSUE #2556: Warn about deprecated proxy APIs
The linter should begin warning that the following APIs will be deprecated in Firefox 71:
- proxy.register
- proxy.unregister
- proxy.onProxyError

This issue can be tracked in [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1545811).

EnTeQuAk:
Refactor how we're handling deprecated APIs by introducing a whitelist of APIs that are being warned about in case they are defined in the schema.

Add storage.managed, identity.getRedirectURL, runtime.onMessageExternal and runtime.onConnectExternal ot TEMPORARY_APIs along the way.

This PR doesn't handle the `unsupported` keyword for `deprecated` entites yet as I'm unsure if we should rather warn or indeed error in case of that. I think we can work on that in a separate PR.

Fixes #2556
Fixes #2462


EnTeQuAk:
>     * it seems to me that (with the current version of this PR) we are going to losing the ability of associating an url (and an additional description) to the error/warning (and in my opinion we should keep this ability, because it helps to point the addon developer to some docs on MDN to know how to deal with a deprecated API or manifest property)

Indeed, I'll continue to support that ability to overwrite the error message. I initially thought the messages in the schema were good enough but you're right, the linter will most certainly have to warn / inform users very differently.

>     * I would prefer if we could still keep the warnings on those unsupported APIs that we were already warning about in the current version of the addons-linter, I don't care of all the APIs marked as unsupported, but the ones in that list are also deprecated on Chrome and we are never going to implement them (in the schema there are way more marked as "unsupported", some of them are not currently supported on Firefox that may even be implemented in the future)


Ah, you're specifically the ones that are marked as "unsupported" but they have been supported and now aren't and never will be in the future. Yeah, let me try to re-add them to the list.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2401,19 +2401,29 @@ describe('ManifestJSONParser', () => {
...
+          code: messages.MANIFEST_FIELD_DEPRECATED.code,
+          message: messages.MANIFEST_FIELD_DEPRECATED.message,
+          file: messages.MANIFEST_FIELD_DEPRECATED.file,
+        };
         const expectedErrors = [
           {
             ...commonErrorProps,
             dataPath: '/theme/images/headerURL',
+            description:
+              'Please use theme.images.theme_frame, this alias will be removed in Firefox 69.',
```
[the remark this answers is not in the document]
EnTeQuAk: Indeed, as mentioned in my answer, I'll re-add the ability to provide custom error messages. I initially thought the schema provided messages would be fine but it'll be something we need from the start and it'll be harder to add it later.

tests/unit/rules/javascript/test.deprecated_browser_apis.js
```diff
@@ -0,0 +1,19 @@
+import { VALIDATION_WARNING, DEPRECATED_JAVASCRIPT_APIS } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+
+import { validMetadata } from '../../helpers';
+
+describe('deprecated browser APIs', () => {
+  DEPRECATED_JAVASCRIPT_APIS.forEach((api) => {
+    it(`should return deprecation warning when ${api} is used`, async () => {
+      const fakeMetadata = { addonMetadata: validMetadata({}) };
+      const code = `chrome.${api}(function() {});`;
```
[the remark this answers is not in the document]
EnTeQuAk: Indeed, I think I just went on and copy&pasted that part. Let me try to remove it, it shouldn't change the tests-outcome anyway.

src/messages/manifestjson.js
```diff
@@ -298,12 +298,10 @@ export function corruptIconFile({ path }) {
   };
 }
 
-export const MANIFEST_THEME_LWT_ALIAS = {
-  code: 'MANIFEST_THEME_LWT_ALIAS',
-  message: i18n._('This theme LWT alias has been removed in Firefox 70.'),
-  description: i18n._(
-    'See https://mzl.la/2T11Lkc (MDN Docs) for more information.'
```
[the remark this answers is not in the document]
EnTeQuAk: So, yeah, a map is what I'm after right now.

Ideally, we'll be able to overwrite both, `message` and `description` too (`description` is being taken from the schema by default) so that we can provide a proper user experience.

src/const.js
```diff
@@ -144,24 +152,15 @@ export const FLAGGED_FILE_MAGIC_NUMBERS = [
 // Based on the above, this is how deep we need to look into a file.
 export const FLAGGED_FILE_MAGIC_NUMBERS_LENGTH = 4;
 
-export const DEPRECATED_APIS = [
```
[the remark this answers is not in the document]
EnTeQuAk: So… actually, for things that existed and are now unsupported - and as such most certainly won't be supported in the future - how are they recognizable in the schema. They should be `deprecated` *and* `unsupported` at the same time right? Or is that just a coincidence. Can we rely on that?

EnTeQuAk:
I think I fixed most or all of your concerns, mind a re-review?

src/const.js
```diff
@@ -121,11 +122,21 @@ export const MIME_TO_FILE_EXTENSIONS = {
...
 
-// List of the "schema data paths" of the deprecated static theme's LWT aliases.
-export const DEPRECATED_STATIC_THEME_LWT_ALIASES = [
-  '/theme/images/headerURL',
-  '/theme/colors/accentcolor',
-  '/theme/colors/textcolor',
+// Mapping of "schema data paths" of the deprecated properties that we
+// issue warnings for.
+// If the value is `null` we will be using the `deprecated` message
+// from the schema. Otherwise `code`, ?message` and `description` will be taken
```
rpl: Typo:
```
 ?message` => `message`
```

src/const.js
```diff
@@ -142,24 +153,15 @@ export const FLAGGED_FILE_MAGIC_NUMBERS = [
...
-  'tabs.getAllInWindow',
-  'tabs.getSelected',
-  'tabs.onActiveChanged',
-  'tabs.onSelectionChanged',
-  'tabs.sendRequest',
-];
-
 // These are APIs that will cause problems when loaded temporarily
 // in about:debugging.
+// APIs listed here should be defined in https://mzl.la/31p4AMc
```
rpl: at the MDN page linked here there are also a couple of other APIs (in particular: the native messaging API methods, `runtime.connectNative` and `runtime.sendNativeMessage`, and the `pkcs11` API namespace), should we add those here too?

src/parsers/manifestjson.js
```diff
@@ -203,12 +204,29 @@ export default class ManifestJSONParser extends JSONParser {
...
+        baseObject = messages[DEPRECATED_MANIFEST_PROPERTIES[error.dataPath]];
+
+        if (baseObject === null) {
+          baseObject = messages.MANIFEST_FIELD_DEPRECATED;
+        }
+
+        let errorDescription = baseObject.description;
+
+        if (errorDescription === null) {
+          const parsedHtml = cheerio.load(error.message);
```
rpl: sanitize the deprecation message sounds like a good idea to me, but I was wondering if cheerio.load may fail (e.g. by throwing on invalid message) and so if we should then prevent that by:

- run some validation of the deprecated messages while we are importing the schema
- or pre-sanitize all the deprecated messages as part of the schema importing

given that we already have a schema importing (and transformation) step, pre-sanitize the deprecated messages as part of that step may be good (even if cheerio.load would not throw on an invalid HTML fragment included in these messages).

src/schema/browser-apis.js
```diff
@@ -11,46 +11,68 @@ const schemas = schemaList.reduce(
...
+    getObjectProperty(schema, property) || getArrayProperty(schema, property);
+
+  if (schemaItem !== null && schemaItem.unsupported !== undefined) {
     return false;
   }
-  return (
-    hasObjectProperty(schema, property) || hasArrayProperty(schema, property)
-  );
+
+  return schemaItem !== null;
```
rpl: It is true that [when a schema item contains a defined `unsupported` property then it is also set to true](https://searchfox.org/mozilla-central/search?q=%22unsupported%22%3A&case=false&regexp=false&path=schemas%2F), but given that it also always a boolean we may use `... && schemaItem.unsupported) {...}` here.

It may also good to extend the check to also verify that `schemaItem` is not `undefined` (e.g. by just changing `schemaItem !== null` into `schemaItem != null`).

Could we turn these checks (from line 73 to line 77) into a single line return statement?

```
return schemaItem && !schemaItem.unsupported;
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2401,19 +2401,36 @@ describe('ManifestJSONParser', () => {
           return { code, dataPath, file, message, description };
         });
 
-        const commonErrorProps = { ...messages.MANIFEST_THEME_LWT_ALIAS };
+        const commonErrorProps = {
```
rpl: These doesn't seem to be `commonErrorProps` anymore, as they are then overwritten a few lines below for every one of the items in the `expectedErrors` array.

At a first glance it seems that the previous definition of `commonErrorProps` (` { ...messages.MANIFEST_THEME_LWT_ALIAS };`) should still work.

am I reading it wrong?

tests/unit/rules/javascript/test.deprecated_browser_apis.js
```diff
@@ -0,0 +1,19 @@
+import { VALIDATION_WARNING, DEPRECATED_JAVASCRIPT_APIS } from 'const';
+import JavaScriptScanner from 'scanners/javascript';
+
+import { validMetadata } from '../../helpers';
+
+describe('deprecated browser APIs', () => {
+  DEPRECATED_JAVASCRIPT_APIS.forEach((api) => {
+    it(`should return deprecation warning when ${api} is used`, async () => {
+      const fakeMetadata = { addonMetadata: validMetadata({}) };
+      const code = `chrome.${api}();`;
```
rpl: Nit, we may also include a call that uses `browser.${api}();` and explicitly check that we get 2 warning messages for that API.

src/messages/javascript.js
```diff
@@ -124,26 +124,12 @@ export const UNSUPPORTED_API = {
   description: i18n._('This API has not been implemented by Firefox.'),
 };
 
-function deprecatedAPI(api) {
-  return {
-    code: apiToMessage(api),
-    message: i18n._(`"${api}" is deprecated or unimplemented`),
-    description: i18n._(oneLine`This API has been deprecated by Chrome
-      and has not been implemented by Firefox.`),
```
rpl: Could we use a mechanism similar to the one we are now using in this PR to customize the deprecated warning message for the LWT aliases to also customize the message for the API that used to be pointed as unimplemented? 
(I mean only the following ones: 'app.getDetails', 'extension.onRequest', 'extension.onRequestExternal', 'extension.sendRequest', 'tabs.getAllInWindow', 'tabs.getSelected', 'tabs.onActiveChanged', 'tabs.onSelectionChanged', 'tabs.sendRequest').

This deprecated APIs are bit different from deprecations like the proxy API, because they were already deprecated in chrome when we started to work on the WebExtensions API and so they have never been implemented, personally I would prefer their warning message to possibly still state that. 

(I don't see it as a blocker, but it would be nice and at a first glance it seems to me that this PR shouldn't be changed too much to achieve that)

src/const.js
EnTeQuAk: Ah yes, I wasn't quite sure if the pkcs11 namespace would simply work like that and wanted to put that into a separate issue.

Let me file a follow-up issue to take a deeper look at the docs and add the remaining APIs to the list if that works for you?

tests/unit/parsers/test.manifestjson.js
EnTeQuAk: Actually, I removed that whole part and simply hardcoded the expected error messages. That makes it much easier to reason about.

src/const.js
rpl: sure, pkcs11 wasn't there and so it is definitely not a blocking issue. 
Looking into it as part of a follow-up issue works for me too.

tests/unit/rules/javascript/test.deprecated_browser_apis.js
```diff
@@ -1,19 +1,36 @@
+/* eslint-disable import/namespace */
```
rpl: This is going to disable the import/namespace rule for the entire file, which may be a bit extreme, not a super big deal given that this is a test case, but isn't it enough to only disable this rule at line 24?

src/parsers/manifestjson.js
rpl: Deferring this to a follow up sounds ok to me (as we discussed over IRC).
have you already filed a follow up issue to pre-sanitize (or validate) the deprecated messages as part of the schema importing? it would be good to link it here.

it may still be good to cover this part with a couple of ` errorLookup` unit tests (in the `test.manifestjson.js` test file).

I also took a quick look to the cheerio npm dependency and the API schemas in the mozilla-central repo to see if there is any deprecation message that could be able to trigger an exception here and it doesn't seem there is any:

it seems that `cheerio.load` may only throw is `error.message` is not a string (and looking at the searchfox results this one is the only one to be a boolean: https://searchfox.org/mozilla-central/rev/9bb55ae4d808fc48afcf93f99da6a685265b86c6/browser/components/extensions/schemas/windows.json#114, it is not from an API property and so it shouldn't be able to trigger an exception here)

rpl: As a side note:
there are some other annoying "message formats" that we inherited from chrome, e.g. `Please use $(ref:tabs.onActivated)`, but it seems that none of the deprecation messages related to manifest properties is using it (https://searchfox.org/mozilla-central/search?q=%22deprecated%22%3A&case=false&regexp=false&path=schemas%2F) and so it shouldn't be actually an issue yet.

src/rules/javascript/index.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
       .default,
     'webextension-unsupported-api': require('./webextension-unsupported-api')
       .default,
+    'webextension-deprecated-api': require('./webextension-deprecated-api')
```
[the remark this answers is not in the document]
EnTeQuAk: As mentioned in IRC, with 53ba0cb I went a similar but different way.

We are now issuing a correct `code` but we are re-using `DEPRECATED_API` for all deprecated api related messages (issued from the `webextension-deprecated-api` rule). That simplifies the patch as we most certainly won't need the ability to differentiate between these different deprecation's anyway. If we do, we can add that later.

src/const.js
```diff
@@ -121,12 +122,39 @@ export const MIME_TO_FILE_EXTENSIONS = {
...
+  'extension.onRequestExternal': 'DEPRECATED_CHROME_API',
+  'extension.sendRequest': 'DEPRECATED_CHROME_API',
+  'tabs.getAllInWindow': 'DEPRECATED_CHROME_API',
+  'tabs.getSelected': 'DEPRECATED_CHROME_API',
+  'tabs.onActiveChanged': 'DEPRECATED_CHROME_API',
+  'tabs.onSelectionChanged': 'DEPRECATED_CHROME_API',
+  'tabs.sendRequest': 'DEPRECATED_CHROME_API',
+
+  // https://github.com/mozilla/addons-linter/issues/2556
+  'proxy.register': null,
```
rpl: Nit, deprecated APIs with null as the value are going to get 'DEPRECATED_API' as the code assigned by default, right? we may set it explicitly (only to make it immediately visible, the actual behavior should not be affected by specify it explicitly vs. leave it as implicit as it is right now).

src/messages/javascript.js
```diff
@@ -124,26 +124,24 @@ export const UNSUPPORTED_API = {
...
-export const TABS_SENDREQUEST = deprecatedAPI('tabs.sendRequest');
+export const DEPRECATED_CHROME_API = {
+  // We are re-using the same code here for consistency and for technical
+  // reasons. We aren't really able to issue different codes from the same
+  // rule, so until we have to, we're going to re-use the `DEPRECATED_API`
+  // code.
+  code: 'DEPRECATED_API',
+  message: null,
+  messageFormat: i18n._('"{{api}}" is deprecated or unimplemented'),
+  description: i18n._(oneLine`This API has been deprecated by Chrome
```
rpl: Nit, this is currently unused (the linting warnings/error produced will have the messageFormat specified here, but the description isn't currently overwritten), it may be worth to leave an inline comment here (and possibly mention a github issue to look into it at some point).

Otherwise, e.g. if we don't think it is worth to be fixed in a followup, could we just remove the unused description property?

EnTeQuAk: Fixed (accidentally) as part of my merge in d7ff5e57, going to rebase before merging anyway.

src/const.js
EnTeQuAk: Great idea, added in b6fe8ead

PR #2791, opened 2019-09-11
Remove unused dependency

antross:
Comments in tests suggest concepts were borrowed from `eslint-plugin-no-unsafe-innerhtml`, but as far as I can tell no reference to it exists in code. The package itself appears to have been renamed to `eslint-plugin-no-unsanitized` anyway. 

Found by searching for how an old version of `eslint` was being included in one of my projects that uses `web-ext`.


antross:
Closing as CI clearly shows I missed how this dependency gets loaded.

PR #2875, opened 2019-10-22
Remove snyk back from `dependencies` and move po2json to devDep…

### LINKED ISSUE #2845: snyk sneaked snyk into dependencies
The PR in #2827 by @snyk-bot moved the "snyk" dependency from `devDependencies` to `dependencies`. #2824 added it back to `devDependencies` (as a part of a version bump, possibly an incorrect merge conflict resolution), and now we have "snyk" both as a devDependency (where it should be) and in `dependencies` (where it should not be).

EnTeQuAk:
Fixes #2845


EnTeQuAk:
> Let's move snyk to dependencies as per snyk/snyk#146

If I understand this correctly, calling `snyk protect` needs to be done by the application that installs the linter as part of their workflow as this requires a separate step in the deployment workflow.

Or am I reading this wrong?

Rob--W:
I suggest to keep `snyk` in `devDependencies`, because none of the patches apply to the production version, see https://github.com/mozilla/addons-linter/issues/2845#issuecomment-540641668

Check whether the following patches are still necessary, and if not (see https://github.com/mozilla/addons-linter/issues/2845#issuecomment-540641668), remove them.
https://github.com/mozilla/addons-linter/blob/master/.snyk#L467-L494

wagnerand:
Yes, we might want to review the current list of patches. However, we should keep it in `dependencies` to ensure we're still protected once a new patch is being added.

.snyk
```diff
@@ -1,495 +1,4 @@
...
-        patched: '2019-07-25T09:43:24.682Z'
-    - babel-gettext-extractor > @babel/core > @babel/helpers > @babel/traverse > @babel/helper-function-name > @babel/template > @babel/types > lodash:
-        patched: '2019-07-25T09:43:24.682Z'
-  SNYK-JS-HTTPSPROXYAGENT-469131:
-    - codecov > teeny-request > https-proxy-agent:
-        patched: '2019-10-04T01:24:25.818Z'
-    - snyk > proxy-agent > https-proxy-agent:
-        patched: '2019-10-04T01:24:25.818Z'
-    - snyk > proxy-agent > pac-proxy-agent > https-proxy-agent:
-        patched: '2019-10-04T01:24:25.818Z'
```
EnTeQuAk: I ran `snyk wizard` to update that file so these patches aren't needed anymore.

EnTeQuAk:
Updated the .snyk file, now snyk is only in `dependencies` and not `devDependencies` anymore (as that'd be duplicate) - https://github.com/mozilla/addons-linter/pull/2875/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R75

Rob--W:
Could you move `snyk` from `dependencies` to `devDependencies` ? There is no use for it as a normal dependency:

 `snyk` is only used by the `prepare` and `preblishOnly` npm hooks. These aren't run when a package is installed as a dependency.
- Even if the hooks were to be run (and `snyk protect` were to be executed), then it would fail, because `snyk protect` requires a `.snyk` file, which is not included in the released `addons-linter` package.

`.snyk` does currently not have any patches, so even if `snyk protect` were to be run, there wouldn't be any useful content. If you really want to force all consumers of `addons-linter` to run `snyk protect`, then it has to be added in a `postinstall` hook. As one of the maintainers of web-ext, I'd rather not have `snyk` as a downstream dependency (anyone who would like to use `.snyk` could add `snyk` to their dependencies at the top level to opt in to `snyk protect`).

wagnerand:
@Rob--W I hear you, but we should stick to snyk's proposed approach, otherwise we'll run into conflicts and bugs again with the next update.

EnTeQuAk:
Alternatively we could simply remove snyk and make it a problem for downstream (web-ext, addons-server etc). But as addons-linter is installed on users PCs regardless of downstream it feels right to have it in here too.

Let's stick to the approach snyk proposes for now and see how that goes. I hear you @Rob--W and want to say, nothing's settled - this is just a "cleanup" to bring us back to a reasonable state and if it turns out it should indeed be a `devDependnecy` we can do that in a next step.

PR #2892, opened 2019-10-29
Import Firefox 71b4 schema

### LINKED ISSUE #2891: Import Firefox 71 schema

EnTeQuAk:
Fixes #2891


rpl (approved):
Looks good.

In particular, we are still going to show linting warning for proxy.register/unregister and proxy.onProxyError, because (even if they are not in the schema anymore) they are still part of the `DEPRECATED_JAVASCRIPT_APIS` mapping table.

But I just noticed that in #2747 we missed to add `proxy.registerProxyScript` in the mapping table, which was also deprecated and it has been removed in this schema import. Would you mind to add it in this pull request or in a separate one?

(As a side note: in the future schema imports we should always double-check that we are adding to that table the APIs we are completely removing, otherwise they are detected as UNSUPPORTED_APIS and the resulting linting message is going to say "This API has not been implemented by Firefox." :-P, or maybe we should try to make use of the `mdn-browser-compat-data` for the the deprecation warnings of APIs not available in the API schema anymore, as it may "scale better" than the mapping table).

(As a side - side note... I also noticed that on the Firefox side we missed to remove onProxyError from the schema in Bug 1443259, sigh, but don't worry about that I'll deal with it).

EnTeQuAk:
> But I just noticed that in #2747 we missed to add proxy.registerProxyScript in the mapping table, which was also deprecated and it has been removed in this schema import. Would you mind to add it in this pull request or in a separate one?

Added in 1a2365b, if travis approves I'll merge.

PR #2912, opened 2019-11-18
Update to node 10

### LINKED ISSUE #2416: Update to node 10

wagnerand:
Fix #2416.


package.json
```diff
@@ -6,6 +6,9 @@
   "bin": {
     "addons-linter": "bin/addons-linter"
   },
+  "engines": {
+    "node": ">=10.15"
```
EnTeQuAk: The specific version feels rather random, afaik 10.x is LTS as of 10.12 - why did we pick that particular version?

wagnerand: There was a reason, but I can't remember it. We can lower it to 10.13 since that was the first real LTS release.

EnTeQuAk: That'd be good, thanks!

PR #2926, opened 2019-11-20
Configure Node policy support for renovate.

EnTeQuAk:
This configures renovate to always configure our nodejs related configs
to support currently active and in maintainance lts versions.

That's not necessarily how we handled this in the past (we dropped
support sometimes before maintainance ended) but I think that's a fair
strategy and the reintroduced support of node 8 will be automatically
dropped in January 2020.

https://docs.renovatebot.com/node/#configuring-support-policy

This will fix occurrences like https://github.com/mozilla/addons-linter/pull/2925 I think

r? @muffinresearch @wagnerand


wagnerand:
Do we also want to add `"travis": { "enabled": true }`?

wagnerand:
> This configures renovate to always configure our nodejs related configs
> to support currently active and in maintainance lts versions.

The current patch actually just includes the latest LTS (and latest in general), so that's fine?

muffinresearch:
Would be good to file issues + propagate these changes to other add-ons renovate enabled repos if they're not doing this already.

EnTeQuAk:
> The current patch actually just includes the latest LTS (and latest in general), so that's fine?

Yiekes, I changed this locally and didn't commit… I blame the missing coffee. I fixed this in b19362b6

EnTeQuAk:
> Would be good to file issues + propagate these changes to other add-ons renovate enabled repos if they're not doing this already.

I'll do that once we verify that renovate does the right thing after this lands.

PR #2931, opened 2019-11-22
Revert "Pin dependencies" and "Update Node.js to v13.2.0"

wagnerand:
This reverts commit f76dac76a694cb03bed0ab508d216b0d01776a01 and cecec0f8341f67aa444b27a093850ccbc6cadf40.


EnTeQuAk:
Curious what renovate does now… I hoped after updating it's config that it'll create a new PR to update the config again but as it seems it didn't do that :thinking:

PR #2945, opened 2019-11-29
feat: support multi-version BCD and ignore versions with flags

### LINKED ISSUE #2944: Support complex compat data
Currently the BCD logic assumes all data to be in the simplest form. However, sometimes the compat data now has arrays because the support level has changed between Firefox versions. Further, versions that are declared to require a flag should be ignored by the checks.

I've come to this conclusion after checking what the code does based on a discussion with @rpl about the userScripts compat data.

freaktechnik:
Fixes #2944 

This makes the compat checker consider versions that require some flag to be set (either runtime or about:config) to not support a feature. Further it adds support for more complex ways of noting compat data where the support level changes between browser versions.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -385,6 +385,22 @@ export function basicCompatVersionComparison(versionAdded, minVersion) {
   return !Number.isNaN(asNumber) && asNumber > minVersion;
 }
 
+function firstStableVersion(supportInfo) {
+  return supportInfo.reduce((versionAdded, supportEntry) => {
+    if (
+      !Object.prototype.hasOwnProperty.call(supportEntry, 'flags') &&
+      (!versionAdded ||
+        !basicCompatVersionComparison(
+          supportEntry.version_added,
+          parseInt(versionAdded, 10)
+        ))
+    ) {
+      return supportEntry.version_added;
+    }
+    return versionAdded;
+  }, false);
+}
+
 export function isCompatible(bcd, path, minVersion, application) {
   const steps = path.split('.');
   let { api } = bcd.webextensions;
@@ -397,7 +413,11 @@ export function isCompatible(bcd, path, minVersion, application) {
   }
   // API namespace may be undocumented or not implemented, ignore in that case.
   if (api.__compat) {
-    const versionAdded = api.__compat.support[application].version_added;
+    const supportInfo = api.__compat.support[application];
+    const versionAdded = Array.isArray(supportInfo)
+      ? firstStableVersion(supportInfo)
+      : !Object.prototype.hasOwnProperty.call(supportInfo, 'flags') &&
+        supportInfo.version_added;
     return !basicCompatVersionComparison(versionAdded, minVersion);
   }
   return true;
--- a/src/utils.js
+++ b/src/utils.js
@@ -385,8 +385,17 @@ export function basicCompatVersionComparison(versionAdded, minVersion) {
   return !Number.isNaN(asNumber) && asNumber > minVersion;
 }
 
+/**
+ * @param {*} supportInfo - bcd support info of a feature
+ * @returns {string|boolean} The first version number to suppor the feature
+ *          or a boolean indicating if the feature is at all supported.
+ */
 function firstStableVersion(supportInfo) {
-  return supportInfo.reduce((versionAdded, supportEntry) => {
+  let supportInfoArray = supportInfo;
+  if (!Array.isArray(supportInfo)) {
+    supportInfoArray = [supportInfo];
+  }
+  return supportInfoArray.reduce((versionAdded, supportEntry) => {
     if (
       !Object.prototype.hasOwnProperty.call(supportEntry, 'flags') &&
       (!versionAdded ||
@@ -414,10 +423,7 @@ export function isCompatible(bcd, path, minVersion, application) {
   // API namespace may be undocumented or not implemented, ignore in that case.
   if (api.__compat) {
     const supportInfo = api.__compat.support[application];
-    const versionAdded = Array.isArray(supportInfo)
-      ? firstStableVersion(supportInfo)
-      : !Object.prototype.hasOwnProperty.call(supportInfo, 'flags') &&
-        supportInfo.version_added;
+    const versionAdded = firstStableVersion(supportInfo);
     return !basicCompatVersionComparison(versionAdded, minVersion);
   }
   return true;
--- a/tests/unit/test.utils.js
+++ b/tests/unit/test.utils.js
@@ -606,4 +606,105 @@ describe('isCompatible', () => {
       )
     ).toBe(true);
   });
+
+  it('should be compatible if version is behind flag and no other version is available', () => {
+    expect(
+      isCompatible(
+        getBCD({
+          foo: {
+            __compat: {
+              support: { firefox: { flags: [], version_added: true } },
+            },
+          },
+        }),
+        'foo',
+        60,
+        'firefox'
+      )
+    ).toBe(true);
+  });
+
+  it('should be compatible if version is behind flag starting at a later release', () => {
+    expect(
+      isCompatible(
+        getBCD({
+          foo: {
+            __compat: {
+              support: { firefox: { flags: [], version_added: '61' } },
+            },
+          },
+        }),
+        'foo',
+        60,
+        'firefox'
+      )
+    ).toBe(true);
+  });
+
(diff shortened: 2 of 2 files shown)


freaktechnik:
Scrolling through the code further, I've also noticed that the code seems to rely on the `applications` property of the manifest, but I think that's normalized for `browser_specific_settings`?

rpl:
> Scrolling through the code further, I've also noticed that the code seems to rely on the `applications` property of the manifest, but I think that's normalized for `browser_specific_settings`?

if `browser_specific_settings.gecko` exists, then `browser_specific_settings` takes precedence over the `applications` property, it basically overwrites it: https://github.com/mozilla/addons-linter/blob/a5b46c52d125c09324a1cbc36d81f5438c622777/src/parsers/manifestjson.js#L293-L298

rpl:
@freaktechnik sorry I did miss to notice that this pull request was still pending a review, I'm going to get back to this asap.

src/utils.js
```diff
@@ -385,6 +385,22 @@ export function basicCompatVersionComparison(versionAdded, minVersion) {
...
+      (!versionAdded ||
+        !basicCompatVersionComparison(
+          supportEntry.version_added,
+          parseInt(versionAdded, 10)
+        ))
+    ) {
+      return supportEntry.version_added;
+    }
+    return versionAdded;
+  }, false);
```
rpl: Nit, @freaktechnik given that I see that the type of the `versionAdded` property (as used in the mdn/browser-compat-data repo) can be both a boolean or a string, and so this function may as well return a boolean or a string, I wouldn't mind if we would make it more explicitly visible by specify in an inline comment what this function is going to return ;-)

src/utils.js
```diff
@@ -385,6 +385,22 @@ export function basicCompatVersionComparison(versionAdded, minVersion) {
   return !Number.isNaN(asNumber) && asNumber > minVersion;
 }
 
+function firstStableVersion(supportInfo) {
+  return supportInfo.reduce((versionAdded, supportEntry) => {
+    if (
+      !Object.prototype.hasOwnProperty.call(supportEntry, 'flags') &&
```
rpl: Should we consider the API still locked behind a flag if there is a flags property but it is an empty array?
I guess that we may not have any actual entry in the `mdn/browser-compat-data` repo that does have an empty array as its `flags` property, but I guess that it may be something still possible to happen in the future (and likely not be noticed until it does trigger a visible validation error to an extension developer).

src/utils.js
```diff
@@ -397,7 +413,11 @@ export function isCompatible(bcd, path, minVersion, application) {
   }
   // API namespace may be undocumented or not implemented, ignore in that case.
   if (api.__compat) {
-    const versionAdded = api.__compat.support[application].version_added;
+    const supportInfo = api.__compat.support[application];
+    const versionAdded = Array.isArray(supportInfo)
+      ? firstStableVersion(supportInfo)
+      : !Object.prototype.hasOwnProperty.call(supportInfo, 'flags') &&
```
rpl: couldn't just turn supportInfo into an array of a single element inside `firstStableVersion` itself (if it isn't already an array) and just call `const versionAdded = firstStableVersion(supportInfo);` here?

tests/unit/test.utils.js
```diff
@@ -606,4 +606,105 @@ describe('isCompatible', () => {
       )
     ).toBe(true);
   });
+
+  it('should be compatible if version is behind flag and no other version is available', () => {
```
rpl: would you mind to briefly describe me the reasoning behind the behavior expected by this set of test cases?

In particular I'm wondering about the following ones:
- compatible with firefox 60, if on all versions it is locked behind a flag
- compatible with firefox 60, if the only version listed is > 60 and it is also locked behind a flag
- incompatible with firefox 60, if on earlier version was locked behind a pref and then it is available without in a later version

rpl: Besides these unit tests that uses fake browser-compat-data entries (which are definitely needed), in my opinion it would be good to also add at least one test case that is using a real entry from the browser-compat-data repository (just as a smoke test to increase the chances that we may notice with a test failure if the format of the compat data is being changed in a way that would break the expected behavior of the addons-linter).

tests/unit/test.utils.js
```diff
@@ -606,4 +606,105 @@ describe('isCompatible', () => {
...
+    expect(
+      isCompatible(
+        getBCD({
+          foo: {
+            __compat: {
+              support: { firefox: { flags: [], version_added: true } },
+            },
+          },
+        }),
```
rpl: Nit, I'm wondering if a small test helper could be useful here to make these new test cases a bit shorter (without having to engage a fight with "prettier" ;-))

src/utils.js
freaktechnik: That's what https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data.schema.json#L20 is for :)

rpl: ah, that's good, I didn't know exactly what the browser-compat-data was linting/testing explicitly and so I wanted to mention it.

I forgot to mention in my last round of comments, but I was thinking that it would be nice if this kind of helpers would be provided by the browser-compat-data npm package itself, because the automation of that github project could ensure that the helper is in sync with any schema change may happen in a future version.

Not a blocking issue, but I wanted to ask you what is your opinion about that, especially given your first hand experience on using the browser-compat-data package as an addons-linter dependency.

freaktechnik: I think that would if anything be a separate package probably, though currently it seems like most users of BCD are rolling their own logic. And most people seem to be using slightly different logic and filters they apply to the data.
Further, these helpers here are very simplistic and ignore a bunch of things that BCD _could_ describe.

tests/unit/test.utils.js
freaktechnik: Right, so this is a bit misleading, because what the rule really cares about in the end is if it's not compatible, because that's when it'll report a warning/error. In the case where it is not officially supported yet there's a different rule that applies no matter the min version set in the manifest.json. The assumption here is that things behind a flag should be caught by that rule.

This also explains the case for why it would be reported as incompatible in the last case. The generic rule thinks this API is supported, however the particular min version the add-on author specified does not yet support the feature, thus this rule springs in action and warns them.

tests/unit/test.utils.js
freaktechnik: I'm not quite sure what part could be done with a helper, so please elaborate.

rpl: > I'm not quite sure what part could be done with a helper, so please elaborate.

@freaktechnik Sure thing!

I was thinking to just a simple test helper to make the single test case a bit less verbose (while keeping "prettier" still happy ;-)), nothing too "fancy".

The helper I was thinking of looks more or less like the following:

```js
  const getBCDFirefoxCompat = (
    firefoxSupportData,
    { apiName = 'foo', appName = 'firefox' } = {}
  ) =>
    getBCD({
      [apiName]: { __compat: { support: { [appName]: firefoxSupportData } } },
    });
```

Then it should be possible to reduce most of the test case to something like:

```js
    expect(
      isCompatible(
        getBCDFirefoxCompat({ version_added: '60' }),
        'foo',
        60,
        'firefox'
      )
    ).toBe(true);
```
or
```js
    expect(
      isCompatible(
        getBCDFirefoxCompat(
          { version_added: '61' },
          { apiName: 'foo', appName: 'firefox_android' }
        ),
        'foo',
        60,
        'firefox_android'
      )
    ).toBe(false);
```
or
```js
    expect(
      isCompatible(
        getBCDFirefoxCompat([
          { version_added: '60' },
          { flags: [], version_added: '59' },
        ]),
        'foo',
        60,
        'firefox'
      )
    ).toBe(true);
```

tests/unit/test.utils.js
rpl: ah, I think that I see now, yeah it is a bit misleading while looking at it in isolation (without taking into account that this `isCompatible` check is only going to add a linting error when it detects an explicit incompatibility with the min version specified in the manifest.json).

Thanks for the additional details.

freaktechnik:
Apologies for taking this long to address feedback. The latest commit should resolve all conversations that are actionable.

EnTeQuAk:
Squashed and merged, thank you so much for that contribution!

EnTeQuAk:
Squashed and merged, thank you so much for that contribution!

PR #2958, opened 2019-12-05
Try to fix renovate `supportPolicy` definition.

### LINKED ISSUE #2957: Fix `supportPolicy` definition in renovate config
We recently enabled automatic handling of nodejs upgrades via renovate but apparently that didn't quite work as expected and it's still pinning the versions to exact versions.

Let's try to fix that.

EnTeQuAk:
Fixes #2957 (hopefully)

This is explicitly setting the `engines` section to `auto` mode to ensure that our global `rangeStrategy` won't get used for nodejs engines so that our `supportPolicy` can be applied accordingly.

(or at least… that's what I hope reading docs and many many renovate.json files on github)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,11 +18,15 @@
         "*"
       ],
       "rangeStrategy": "pin"
-    }
+    },
+    {
+      "depTypeList": ["engines"],
+      "rangeStrategy": "auto"
+    },
+
   ],
   "stabilityDays": 7,
   "node": {
-    "enabled": true,
     "supportPolicy": ["lts"]
   }
 }
--- a/renovate.json
+++ b/renovate.json
@@ -22,8 +22,7 @@
     {
       "depTypeList": ["engines"],
       "rangeStrategy": "auto"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -18,11 +18,14 @@
...
       "rangeStrategy": "pin"
+    },
+    {
+      "depTypeList": ["engines"],
+      "rangeStrategy": "auto"
     }
   ],
   "stabilityDays": 7,
   "node": {
-    "enabled": true,
```
muffinresearch: I guess this is removed since it's essentially the default?

EnTeQuAk: Correct, it's enabled by default. I think

EnTeQuAk:
Now I'm confused… since I configured this renovate doesn't open a PR to correctly configure this. Maybe the `enabled: true` was needed afterall?

According to https://docs.renovatebot.com/node/ it's enabled by default. Is there some kind of dashboard I can look at that may have some logging?

PR #3000, opened 2020-01-10
Use webpack-node-externals to avoid bundling node_modules

Rob--W:
The current approach to excluding external libraries has the following issues:

- Does not support scoped modules (i.e. `@orgname/modname`)
- Does not support module imports with slashes (i.e. `modname/file`)

This can be fixed by replacing the current logic with `webpack-node-externals`.
For more details, see the issues linked from https://github.com/mozilla/web-ext/pull/1814


Rob--W:
CI fails, but when I run the test locally it passes:

```
npm run build
./node_modules/.bin/jest --runInBand --config=jest.integration.config.js "tests/integration/addons-linter"
```

rpl:
> CI fails, but when I run the test locally it passes:
> 
> ```
> npm run build
> ./node_modules/.bin/jest --runInBand --config=jest.integration.config.js "tests/integration/addons-linter"
> ```

to reproduce the same issue you have to run the integration test in production mode ([as they are running on travis](https://travis-ci.org/mozilla/addons-linter/jobs/635483111#L1295)).

I can reproduce the issue locally, and the underlying issue seems to be the 'bin/addons-linter' script throwing a MODULE_NOT_FOUND error.

The underlying issue is [hidden by the way the `bin/addons-linter` script handle that kind of error](https://github.com/mozilla/addons-linter/blob/132558f79eb647535b77b1df01eca46f38ce2f87/bin/addons-linter#L20-L28), the error actually raised is the following one:

```
Error: Cannot find module '@babel/runtime/helpers/interopRequireDefault'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
    at Function.Module._load (internal/modules/cjs/loader.js:506:25)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/zfs-ubik/work/projects/linter/addons-linter/dist/webpack:/external "@babel/runtime/helpers/interopRequireDefault":1:1)
    at __webpack_require__ (/zfs-ubik/work/projects/linter/addons-linter/dist/webpack:/webpack/bootstrap:19:1)
    at Object.<anonymous> (/zfs-ubik/work/projects/linter/addons-linter/dist/addons-linter.js:96:30)
    at Object.module.exports.module.exports (/zfs-ubik/work/projects/linter/addons-linter/dist/addons-linter.js:139:30)
    at __webpack_require__ (/zfs-ubik/work/projects/linter/addons-linter/dist/webpack:/webpack/bootstrap:19:1)
    at /zfs-ubik/work/projects/linter/addons-linter/dist/webpack:/webpack/bootstrap:83:1 code: 'MODULE_NOT_FOUND'
```

you can also trigger it from the github repository using the following STR:
- build the addons-linter by cloning the addons-linter repo and then running `npm ci` and `npm run build`
- remove the `node_modules/` dir and recreate it in production mode (e.g. using `npm ci --production`)
- then run `./bin/addons-linter` without any parameter:
  - expected behavior: addons-linter to print the usage help and the error "Not enough non-option arguments: got 0, need at least 1"
  - actual behavior: the addons-linter detects the error as a missing build and logs the error "You did not build addons-linter yet. Please run `npm install` and `npm run build` or see README.md for more information."

Given the kind of issue, it seems something that can be triggered by replacing the previous webpack config with the new one based on the `webpack-node-externals` plugin (but I didn't have compared yet the bundle built with the previous webpack config vs. the new one, and so I haven't yet any additional detail about how the plugin is actually triggering this issue).

Rob--W:
Thanks @rpl for debugging.

`npm ci --production` excludes devDependencies.

When `webpack-node-externals` is used with the `modulesFromFile: true` option, then anything listed in dependencies, devDependencies, peerDependencies, optionalDependencies is NOT bundled.

I initially thought of restricting this to just `dependencies` to allow other dependencies to be bundled. But I decided against that, because I asssumed that production would never have to rely on `devDependencies`. The CI error shows that this assumption has been violated. And indeed, when I look at `dist/addons-linter.js`, then I see a bunch of `require("@babel/runtime/....")` in the file.

This could be resolved by either bundling it, or moving `@babel/runtime` from `devDependencies` to `dependencies. When bundled, the file size of addons-linter.js is 834k, opposed to 826k when not bundled. The difference in file size is small enough, so I'm going to allow `@babel/runtime` to be bundled.

EnTeQuAk:
@rpl I assigned this to you as you already looked into it. If you don't have the time, do let me know.

webpack.config.js
```diff
@@ -42,7 +31,12 @@ module.exports = {
       },
     ],
   },
-  externals: nodeModules,
+  externals: [
+    nodeExternals({
+      modulesFromFile: true,
+      whitelist: [/^@babel\/runtime\//],
```
rpl: @EnTeQuAk in the web-ext package we do have @babel/runtime as a dependency (from what I recall we did choose to add it as a direct dependency because is technically something that is being used in production), do you have anything against moving it from the devDependencies to the dependencies for the addons-linter too?

In my opinion moving it to the dependencies makes it clear that it is being used, much more clear than whitelisting it in the webpack config here.

If we have reasons that makes us to prefer keeping that unmodified, then I'm fine with whitelisting it here.

EnTeQuAk: I don't have any reason against it, so let's add it to `dependencies`.

Rob--W:
@rpl Just confirmed that the output has the expected dependencies: Instead of inline dependencies, there are a bunch of `@babel/runtime/` imports. Also `ajv/lib/refs/json-schema-draft-06.json` is included as an external dependency instead of inlined.

I also noticed that `regenerator-runtime` is not required any more, because `@babel/runtime/regenerator` is used instead. I think that it is safe to remove, so I will open a new PR for that.

rpl:
@EnTeQuAk This pull request should be ready to go.
@Rob--W has also just added a separate commit (e934288) to remove regerator-runtime (for the reasons he mentioned in https://github.com/mozilla/addons-linter/pull/3000#issuecomment-587569430) and he told me over Matrix that he would like these two commits to be landed without being squashed (e.g. using rebase and merge instead), to make sure that we preserve the two separate commit messages.

PR #3036, opened 2020-02-10
Port firefox schema import to use zip files

### LINKED ISSUE #2955: Port firefox schema import to use zip files
So, it seems that the mozilla-unified downloads are only available as zip and not as .tar.gz anymore.

That'll unblock #2935

EnTeQuAk:
Fixes #2955


EnTeQuAk:
@rpl I haven't rebased the commits yet - wanted to get a first look now that we figured out the problem. I can squash on merge later anyway.

Happy to hear what you think about that PR

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -1112,58 +1111,88 @@ describe('firefox schema import', () => {
...
+      await new Promise((resolve) => {
+        files.forEach((file) => {
+          zipfile.addFile(`${schemaPath}/${file}`, file);
+        });
+        zipfile.outputStream
+          .pipe(fs.createWriteStream('mozilla-central.zip'))
+          .on('close', () => resolve());
+        zipfile.end();
+      });
+
```
rpl: Nits:
- I think that we zipfile.outputStream is the only part that really need to be wrapped in a promise
- it seems that we are using similar code in at least one other test in this same test file, how about create a small test helper and reuse it in both places?

package.json
```diff
@@ -87,6 +87,7 @@
     "@babel/preset-env": "7.8.4",
     "@babel/register": "7.8.3",
     "@babel/runtime": "7.8.4",
+    "archiver": "3.1.1",
```
rpl: `archiver` doesn't seem to be used.

EnTeQuAk: Bad merge, fixed in 9bf03d3f

tests/unit/schema/test.firefox-schemas-import.js
EnTeQuAk: Indeed, fixed in cf90083b

src/schema/firefox-schemas-import.js
```diff
@@ -576,30 +578,18 @@ inner.isBrowserSchema = (_path) => {
...
-      .on('entry', (entry) => {
-        // The updated tar library doesn't always strip null bytes from the
-        // end of strings anymore so we get to do that here.
-        const entryPath = stripTrailingNullByte(entry.path);
-        if (inner.isBrowserSchema(entryPath)) {
-          const filePath = path.join(outputPath, path.basename(entryPath));
-          entry.pipe(fs.createWriteStream(filePath));
-        } else {
-          entry.resume();
+    zipfile
```
rpl: It looks that codecov.io is currently detecting a decreased coverage on this file, from a quick look it seems to me that it may be because we are missing to cover the "error" scenario (the one that trigger the rejection of this promise at line 595-597 of this file):
- https://codecov.io/gh/mozilla/addons-linter/pull/3036/changes#D1-596

It would be nice to cover it if it is not too much work (but I didn't looked if we can easily trigger it, e.g. if we can reasonably create some error condition, or if it would be easier to just make jest to mock the `yauzl` module and force such error condition using a fake `yauzl.open` method).

@EnTeQuAk would you mind to take a quick look into that?
(I'm fine with ignoring the codecov.io if it would require too much more time for this PR to be completed, there is a good chance that codecov.io is complaining just because we removed a good chunk of code from this file and the coverage percentage looks decreased even if it is not)

src/schema/firefox-schemas-import.js
```diff
@@ -576,30 +578,18 @@ inner.isBrowserSchema = (_path) => {
...
+      .on('entry', async (entry) => {
+        if (inner.isBrowserSchema(entry.fileName)) {
+          const filePath = path.join(outputPath, path.basename(entry.fileName));
+          const readStream = await openReadStream(entry);
+          readStream.pipe(fs.createWriteStream(filePath));
```
rpl: While re-reading this part once more, I started to wonder if we should explicitly await for the write streams we are creating here, to be sure that all data in the write streams has been completely flushed and the file closed.

It looks that in the previous version based on the tar file we were already doing this without explicitly waiting the write streams to emit the close event (and that part shouldn't be affected by this change).

have we already tried this branch on a bunch of real mozilla zip files to double-check that the schema are imported successfully as expected?
It would be good to try it once more before merging this patch on master (if we have not already done it) and it may actually help us to double-check if explicitly waiting for the write streams to be closed is actually necessary.

If it turns out that we need to await the write stream close events (or we just want to do it to be absolutely sure), I think that we may achieve it easily enough by collecting all the related promises and then wait for them to be resolved or rejected using something like:

```
const writeStreamsPromises = [];
   ...
   .on('entry', async (entry) => {
      ...
      const destFileStream = fs.createWriteStream(filePath);
      writeStreamsPromises.push(
        new Promise(resolve => destFileStream.on("close", resolve)
      );
      const readStream = await openReadStream(entry);
      readStream.pipe(destFileStream);
      ...
   .on('end', () => 
       Promise.all(writeStreamsPromises).then(resolve, reject);
       ...
```

EnTeQuAk: Oh, good points. I'll look into that more deeply tomorrow or just steal your code-changes.

I tried out the patch on all the recent schema imports - just never finished up fixing the tests so it does work in real-life.

src/schema/firefox-schemas-import.js
EnTeQuAk: I added 2f5a4eb4 for that, let's see if codecov likes it.

tests/unit/schema/test.firefox-schemas-import.js
EnTeQuAk: Added a small cleanup in 06b0863b - the `zipfile.end()` needs to be called inside the promise, the `addFile` calls don't.

src/schema/firefox-schemas-import.js
EnTeQuAk: Fixed in 
bebea50

tests/unit/schema/test.firefox-schemas-import.js
rpl: I guess that the promise that resolves when the zipfile.outputStream fires the "close" event should be created before calling `zipfile.end()` to ensure that the promise is ever resolved, but anyway calling zipfile.end() inside the promise is also fine as well.

src/schema/firefox-schemas-import.js
rpl: thanks! codecov seems to be happy about it ;-)

src/schema/firefox-schemas-import.js
rpl: :heart: thanks!

PR #3048, opened 2020-02-14
Import Firefox 74b2 schema.

### LINKED ISSUE #3046: Import Firefox schema up to 74b2
Import the firefox schema up to the most recent version

EnTeQuAk:
Fixes #3046

### DIFF
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -189,11 +189,18 @@
                       "properties": {
                         "extension_pages": {
                           "type": "string",
-                          "format": "contentSecurityPolicy"
+                          "format": "contentSecurityPolicy",
+                          "description": "The Content Security Policy used for extension pages."
                         },
                         "content_scripts": {
                           "type": "string",
-                          "format": "contentSecurityPolicy"
+                          "format": "contentSecurityPolicy",
+                          "description": "The Content Security Policy used for content scripts."
+                        },
+                        "isolated_world": {
+                          "type": "string",
+                          "format": "contentSecurityPolicy",
+                          "description": "An alias for content_scripts to support Chrome compatibility.  Content Security Policy implementations may differ between Firefox and Chrome.  If both isolated_world and content_scripts exist, the value from content_scripts will be used."
                         }
                       }
                     }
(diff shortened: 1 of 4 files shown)


src/schema/imported/manifest.json
```diff
@@ -189,11 +189,18 @@
...
+                          "format": "contentSecurityPolicy",
+                          "description": "The Content Security Policy used for extension pages."
                         },
                         "content_scripts": {
                           "type": "string",
-                          "format": "contentSecurityPolicy"
+                          "format": "contentSecurityPolicy",
+                          "description": "The Content Security Policy used for content scripts."
+                        },
+                        "isolated_world": {
```
diox: Wait do we have the code supporting this ? We have some validation code in the linter and some permission extraction code in addons-server that deal with content scripts, if we need to support `isolated_world` as an alias then we need some extra code in both to do it.

EnTeQuAk: Indeed, I must have missed that. I'll add issues in both, addons-server and addons-linter but I'm not quite sure if this blocks the actual schema import as we kinda need that first to have the linter and server process that value further.

EnTeQuAk: Actually, once the linter issues the correct warnings for the alias the code in addons-server will continue to work too as it's listening on the addons-linter internal codes which should stay the same (`MANIFEST_CSP`)

EnTeQuAk: Actually, I'm confused by this as https://developer.chrome.com/extensions/content_scripts#declaratively is clearly mentioning `content_scripts`. cc @rpl who might know more

EnTeQuAk: Alright, I talked to Luca about this and I'll try to summarize:

`isolated_worlds` is an alias for `content_scripts` *inside* the `content_security_policy` object, which historically only contained strings but can now be an object as well.

That means we're going to have to update a few checks in the linter.

I'll update the relevant issue #3053 accordingly and that then means that we need that to land along side with this schema import to avoid missing any custom content security policies.

EnTeQuAk:
I ran the import with the `FIREFOX_74_0b9_RELEASE` tag (no updates). Also, the requested validation for `isolated_world` landed in a43e8ae.

I'll file follow-up issues for addons-server to take `isolated_world` into account. 

re-r? @diox

PR #3066, opened 2020-03-02
fix: Validate manifest CSP properties in v3 format

### LINKED ISSUE #3053: Add support for `isolated_world` alias
See https://github.com/mozilla/addons-linter/pull/3048/files#diff-c24fe98257d3c70153e05cf42301ed84R200

rpl:
Fixes #3053
(also also unblock #3048)

This PR introduce some small changes needed to parse and validate the manifest CSP properties when they are expressed in the new v3 format:

```
  "content_security_policy": {
    "extension_pages": "...",
    "content_scripts": "...",
    // or "isolated_worlds" (which is the chrome compatible alias for the "content_scripts" one)
  }
```

Some other details related to this PR:
- the change is also covered by some additional automated tests.
- the validation messages now also contains the name of the CSP property related
  to the validation message, e.g.

```
Code                              Message                                         Description                                                                                 File                                     Line   Column
MANIFEST_CSP_UNSAFE_EVAL          content_security_policy.extension_pages         In most cases the same result can be achieved differently, therefore it is generally        manifest.json                            
                                  allows 'eval', which has strong security and    prohibited                                                                                                                           
                                  performance implications.
```

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -853,6 +853,7 @@ describe('ManifestJSONParser', () => {
             `content_security_policy.${keys[i]}`
           );
         }
+        expect(warnings.length).toBe(3);
       });
     });
 
@@ -977,6 +978,7 @@ describe('ManifestJSONParser', () => {
           `content_security_policy.${keys[i]}`
         );
       }
+      expect(warningsV3.length).toBe(3);
     });
   });
 
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -799,6 +799,7 @@ describe('ManifestJSONParser', () => {
         "default-src http:; worker-src: 'self'",
       ];
 
+      // Manifest v2 formats.
       invalidValues.forEach((invalidValue) => {
         const addonLinter = new Linter({ _: ['bar'] });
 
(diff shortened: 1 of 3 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -813,9 +814,46 @@ describe('ManifestJSONParser', () => {
...
+        expect(manifestV3JSONParser.isValid).toEqual(true);
+        const { warnings } = addonLinter.collector;
+
+        const keys = Object.keys(contentSecurityPolicy);
+        for (let i = 0; i < keys.length; i++) {
+          expect(warnings[i].code).toEqual(messages.MANIFEST_CSP);
+          expect(warnings[i].message).toContain(
+            `content_security_policy.${keys[i]}`
+          );
+        }
```
EnTeQuAk: Unless I'm mistaken we do expect exactly two warnings here, once for `content_scripts` and once for `isolated_worlds` - could we add an assertion for that too?

tests/unit/parsers/test.manifestjson.js
```diff
@@ -874,10 +936,47 @@ describe('ManifestJSONParser', () => {
...
+
+      const manifestV3JSONParser = new ManifestJSONParser(
+        jsonV3,
+        addonLinter.collector
       );
+
+      expect(manifestV3JSONParser.isValid).toEqual(true);
+      const warningsV3 = addonLinter.collector.warnings;
+
+      const keys = Object.keys(contentSecurityPolicy);
```
EnTeQuAk: Same here, can we add an assertion that makes sure that we have exactly two warnings that match the expected csp warnings?

tests/unit/parsers/test.manifestjson.js
```diff
@@ -874,10 +936,47 @@ describe('ManifestJSONParser', () => {
...
+        // Alias for content_scripts.
+        isolated_world: invalidValue,
+      };
+
+      const jsonV3 = validManifestJSON({
+        content_security_policy: contentSecurityPolicy,
+        applications: {
+          // The new content_security_policy syntax is only supported
+          // on Firefox >= 72.
+          gecko: { strict_min_version: '72.0' },
```
EnTeQuAk: Do we actually issue a warning / error if `strict_min_version` is `71.0`? I don't think we do as part of the schema, but maybe as part of mdn-browser-compat-data?

EnTeQuAk:
Looks like there's a slight eslint issue - but that's it. https://travis-ci.org/mozilla/addons-linter/jobs/657327008#L1383

tests/unit/parsers/test.manifestjson.js
rpl: No, the linter adds exactly 3 warnings, one for each of those properties ("extension_pages", "content_scripts" and "isolated_worlds").

It is true that `isolated_worlds` is used as an alias for `content_scripts` in Firefox, but that happens in the implementation side, for an API schema perspective they are 2 separate properties of the content_script_policy object.

Nevertheless I can add an assertion for the total number of warning we expect (which is 3 at the moment).

tests/unit/parsers/test.manifestjson.js
rpl: it is coming from the rules implemented based on the metadata coming from `mdn-browser-compat-data`.

tests/unit/parsers/test.manifestjson.js
rpl: As mentioned in the other comment, I can add an assertion to be sure that we got the expected number of warnings (which is 3 for this test)

rpl:
> Looks like there's a slight eslint issue - but that's it. https://travis-ci.org/mozilla/addons-linter/jobs/657327008#L1383

yeah, in the web-ext repo we run the linting checks as part of the `test` npm script and so I didn't run `npm run lint` explicitly and missed to notice that one.

I've pushed the updated version that fixes the linting error (and also added 7c5738f to add the assertion for the expected number of validation warnings).

rpl:
@EnTeQuAk unrelated to this change, I filed #3067 because I noticed that in #3036 I missed to notice that the fake mozilla-central.zip file created in the test case is being created in the project root dir.

PR #3076, opened 2020-03-12
fix: Migrated from deprecated eslint plugin no-unsafe-innerhtml to no-unsanitized

### LINKED ISSUE #1273: Migrate from eslint-plugin-no-unsafe-innerhtml to eslint-plugin-no-unsanitized
The package `eslint-plugin-no-unsafe-innerhtml` has been renamed and been released as `eslint-plugin-no-unsanitized` on npm. The new package is version 2 and the changelog suggests there are breaking changes so filing this to handle it independently.

rpl:
Fixes #1273

TODOs:
- [x] double-check that the test cases defined in tests/unit/rules/javascript/test.no_unsafe_innerhtml.js  are actually in sync with the ones defined in https://github.com/mozilla/eslint-plugin-no-unsanitized/tree/master/tests/rules
- [x] evaluate if the addons-linter should use a custom configuration to explicitly disable the escapers and unwrappers that eslint-plugin-no-unsanitized is allowing by default (See 4ba0e45)
- [x] (this is being handled by @mozfreddyb in mozilla/eslint-plugin-no-unsanitized#114 and it has been released as part of eslint-plugin-no-unsanitized v3.1.0) unexpected failures in the test file tests/unit/rules/javascript/test.no_eval.js (because [the plugin is currently raising a "Unexpected Callee" error from here](https://github.com/mozilla/eslint-plugin-no-unsanitized/blob/b698b7409ccd5d889308341673121e720110d0bd/lib/rules/method.js#L87), already reported upstream as mozilla/eslint-plugin-no-unsanitized#113), agree with @mozfreddyb how the plugin should handle a node.type `SequenceExpression`:
      - [no_eval › should not allow the use of eval: (0, eval)("foo")](https://travis-ci.org/github/mozilla/addons-linter/jobs/661679854?utm_medium=notification&utm_source=github_status)
      - [no_eval › should not allow the use of eval: (0, window.eval)("foo")](https://travis-ci.org/github/mozilla/addons-linter/jobs/661679854#L273)
      - [no_eval › should not allow the use of eval: (0, window["eval"])("foo")](https://travis-ci.org/github/mozilla/addons-linter/jobs/661679854#L295)
- [x] regenerating package-lock.json ~~(non-blocking) double-check if there is any way to make the package-lock.json file in this PR to only include the changes needed to replace this particular dependency~~


EnTeQuAk:
> double-check if there is any way to make the package-lock.json file in this PR to only include the changes needed to replace this particular dependency

You could backout these changes before the merge and let renovate take care of it, or just split that into a separate PR maybe. Or just leave it as it, should be fine.

rpl:
Hi @mozfreddyb, we are looking into migrating the addons-linter from eslint-plugin-no-unsafe-innerhtml to eslint-plugin-no-unsanitized (#1273) and 3 of the existing addons-linter unit tests are failing because `unsanitized/method` doesn't currently support nodes of type `SequenceExpression` (see this issue description for examples of code snippets that trigger that).

Before proceeding further with investigating it I'd like to hear your opinion about this scenario and how the `unsanitized/method` rule should be handling this scenario. 

(This issue seems to be already filed upstream as mozilla/eslint-plugin-no-unsanitized#113)

mozfreddyb:
I'll give https://github.com/mozilla/eslint-plugin-no-unsanitized/issues/113 a shot, seems doable.

rpl:
@EnTeQuAk (and also @mozfreddyb for his opinion on this topic):
While I was making sure that the test case defined in tests/unit/rules/javascript/test.no_unsafe_innerhtml.js are again in sync with the upstream tests, I was wondering why the addons-linter should silent the warnings produced by the `no-unsanitized` rule if an extension does use one of the escapers or unwrappers allowed by default by the eslint plugin (`Sanitizer.escapeHTML` and `escapeHTML`).

In the end this is rule is marked as a warning and it aims to make the reviewer aware that it may need to check that part of the extension code, and so it seems to me that the addons-linter should be using a custom config (supported by this eslint plugin) to explicitly disable the escapers and unwrappers allowed by default.

I took a look into that and 4ba0e45 contains the changes needed to fix that ([1]) if we agreed that it would be the right behavior for the addons-linter (otherwise I'll just remove that patch from this pull request).

Let me know what do you think about this detail.

[1]: which obviously require some changes to the test cases that are currently expected to be valid

mozfreddyb:
I agree that the config change makes sense for you.
The threat model for the no-unsanitized is weaker than the one you are facing with add-ons:
The purpose of the linter plugin is to detect mistakes that might accidentally pass a code review. It's an additional seat-belt for projects that have process.
The original use case was FirefoxOS we we had the Sanitizer library (now defunct). We've moved away from that and recommend people use DOMPurify instead.

But the plugin does not intend to deal with potentially malicious or actively harmful code, which could define as a no-op (e.g., `const DOMPurify = { sanitize = (s) => s };`)
And to be clear, it's very easy for an extension author to to find millions of bypasses (e.g., `foo['inner'+'HTML'] = evil`). 

P.S: If you want to avoid that, you'll could expand [the existing runtime checks for the System Principal](https://searchfox.org/mozilla-central/rev/171109434c6f2fe086af3b2322839b346a112a99/dom/base/nsContentUtils.cpp#4683) to also include Extension Principals

rpl:
@EnTeQuAk this should now be ready for your review \o/
(and thanks a lot to @mozfreddyb for the help on the eslint plugin side!)

diox:
@rpl is this good to go now ? I see @EnTeQuAk approved it.

rpl:
> @rpl is this good to go now ? I see @EnTeQuAk approved it.

oh, yes it is
(and sorry, I totally missed to notice that Chris did approved it a week ago)

PR #3095, opened 2020-03-30
Fixes #3067 - Removing fake ZIP file after test execution

### LINKED ISSUE #3067: test.firefox-schemas-import.js should not create the fake mozilla-central.zip in the project root dir
A fake mozilla-central.zip file is created as part of the test.firefox-schemas-import.js tests, but this file seems to be currently created in the project test file but, as any asset created or manipulated in the test cases, we should actually create it in a temporary directory and remove it when the test is exiting.

marchiore:
This PR solves the problem that after run test.firefox-schemas-import.js it's keeping the fake zip file in the project root dir.

### DIFF
--- a/tests/unit/schema/test.firefox-schemas-import.js
+++ b/tests/unit/schema/test.firefox-schemas-import.js
@@ -3,6 +3,7 @@ import path from 'path';
 
 import yazl from 'yazl';
 import yauzl from 'yauzl';
+import tmp from 'tmp-promise';
 
 import {
   FLAG_PATTERN_REWRITES,
@@ -21,10 +22,17 @@ import {
   rewriteValue,
 } from 'schema/firefox-schemas-import';
 
-// Get a reference to unlinkSync so it won't get stubbed later
-const { unlinkSync } = fs;
+function prepareTmpDir() {
+  const tmpDir = tmp.dirSync({ mode: '0750', unsafeCleanup: true });
+  const outputPath = `${tmpDir.name}/schema-imported`;
+  const inputPath = `${tmpDir.name}/mozilla-central.zip`;
+  const cleanup = () => tmpDir.removeCallback();
 
-async function createZipFile() {
+  fs.mkdirSync(outputPath);
+  return { outputPath, inputPath, cleanup };
+}
+
+async function createZipFile(inputPath) {
   const cwd = 'tests/fixtures/schema';
   const schemaPath = `${cwd}/firefox`;
   const zipfile = new yazl.ZipFile();
@@ -37,7 +45,7 @@ async function createZipFile() {
 
   await new Promise((resolve) => {
     zipfile.outputStream
-      .pipe(fs.createWriteStream('mozilla-central.zip'))
+      .pipe(fs.createWriteStream(inputPath))
       .on('close', () => resolve());
     zipfile.end();
   });
@@ -49,17 +57,6 @@ describe('firefox schema import', () => {
     // eslint-disable-next-line jest/no-disabled-tests
     pending();
     return;
-  }
-
-  function createDir(dirPath) {
-    fs.mkdirSync(dirPath);
-  }
-
-  function removeDir(dirPath) {
-    fs.readdirSync(dirPath).forEach((file) =>
-      unlinkSync(path.join(dirPath, file))
-    );
-    fs.rmdirSync(dirPath);
   }
 
   describe('rewriteOptionalToRequired', () => {
@@ -1093,18 +1090,20 @@ describe('firefox schema import', () => {
     const schemaFiles = ['manifest.json', 'cookies.json'];
     const firefoxPath = 'tests/fixtures/schema/firefox';
     const ourPath = 'tests/fixtures/schema/updates';
-    const outputPath = 'tests/fixtures/schema/imported';
     const expectedPath = 'tests/fixtures/schema/expected';
+    let tmpDir;
 
     beforeEach(() => {
-      createDir(outputPath);
+      tmpDir = prepareTmpDir();
     });
 
     afterEach(() => {
-      removeDir(outputPath);
+      tmpDir.cleanup();
     });
 
     it('imports schemas from filesystem', () => {
+      const { outputPath } = tmpDir;
+
       importSchemas(firefoxPath, ourPath, outputPath);
       schemaFiles.forEach((file) => {
         expect(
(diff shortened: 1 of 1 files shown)


tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -23,6 +23,7 @@ import {
 
 // Get a reference to unlinkSync so it won't get stubbed later
 const { unlinkSync } = fs;
+const FAKE_ZIP_FILE_PATH = './tmp/mozilla-central.zip'
```
rpl, line 26: @marchiore instead of using a tmp directory relative to the repository (which may not even exist when the tests are being executed), we could use ["tmp-promise"](https://www.npmjs.com/package/tmp-promise) to get a temporary directory.

We do already have "tmp-promise" in the dev dependencies, because it is being used in [one of the scripts defined in this repository](https://github.com/mozilla/addons-linter/blob/d9ddd34615a3a2d8d50386c0b000eff653064fae/tests/integration/run-as-production-env.js#L8)), and so it is not even a new dependency.

The change suggested should also fix the failure triggered by this version in the CI job: https://travis-ci.org/github/mozilla/addons-linter/jobs/668838092#L233

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -37,12 +38,16 @@ async function createZipFile() {
...
+      .pipe(fs.createWriteStream(FAKE_ZIP_FILE_PATH))
       .on('close', () => resolve());
     zipfile.end();
   });
 }
 
+async function deleteZipFile() {
+  fs.unlinkSync(FAKE_ZIP_FILE_PATH);
+}
+
```
rpl, line 50: once we change this patch to use the API promised by "tmp-promise", this may be done using the `cleanup` method described in the "tmp-promise" API docs ([you can take look to the readme file here](https://www.npmjs.com/package/tmp-promise#usage-api-reference))

marchiore: i've used `removeCallback()` can it be that way?

marchiore:
I made the changes and miss closed the PR while i'm trying to squash commits.

I'm reopening the PR, sorry about that.

marchiore:
@rpl thanks for your review!

I've made the suggested changes but when Travis CI run on it i'm getting some weird lint problems like this one: https://travis-ci.org/github/mozilla/addons-linter/jobs/671791232#L1595

I've already tried:

```
npm run eslint
npm run prettier
```
but with no success, how can i fix it? any suggestions?

rpl:
> @rpl thanks for your review!
> 
> I've made the suggested changes but when Travis CI run on it i'm getting some weird lint problems like this one: https://travis-ci.org/github/mozilla/addons-linter/jobs/671791232#L1595
> 
> I've already tried:
> 
> ```
> npm run eslint
> npm run prettier
> ```
> 
> but with no success, how can i fix it? any suggestions?

sure thing:
by executing `npm run prettier-dev` you should be able to fix automatically all the eslint validation errors related to the prettier eslint rule.

After that, there will be two more `object-shorthand` eslint validation errors to fix, which are due to two statements written as:
```await fetchSchemas({ inputPath: inputPath, outputPath });```

which eslint complains about because the coding style rules expect it to be written as:
```await fetchSchemas({ inputPath, outputPath });```

eslint is able to fix those two validation errors automatically and so you should be able to fix both by just running `npm run eslint -- --fix` (the first `--` tells `npm run` that the arguments that follows should be passed to the underlying command being executed by the npm run script).

**Comment by marchiore** [comment]
Thanks again @rpl!

Now i'm with [this lint error](https://travis-ci.org/github/mozilla/addons-linter/jobs/672044107#L1593)

But when i fix it on my dev environment it tell me the oposite:
image

which one is the correct one?

**Comment by rpl** [comment]
> Thanks again @rpl!
> 
> Now i'm with [this lint error](https://travis-ci.org/github/mozilla/addons-linter/jobs/672044107#L1593)
> 
> But when i fix it on my dev environment it tell me the oposite:
> image
> 
> which one is the correct one?

The correct one for us is the one from the eslint config provided in this repo, and enforced by the CI jobs running on travis.

My guess is that the feature/plugin that your editor is using to lint the source code on the fly while you are editing it is not picking up the eslint config from our repo and it is enforcing its own defaults.

I would suggest to just ignore what the editor is complaining about in this case, and just use `npm run lint` to double-check that the changes applied are still passing the linting as configured in this repo as expected.

**Comment by marchiore** [comment]
@rpl Done!

I'll take a look into my dev Lint to see what's going on.

File: tests/unit/schema/test.firefox-schemas-import.js [hunk 3]
```diff
@@ -3,6 +3,7 @@ import path from 'path';
 
 import yazl from 'yazl';
 import yauzl from 'yauzl';
+import { dirSync } from 'tmp-promise';
```
rpl, line 6: Nit, would you mind to import `"tmp-promise"` as `tmp` and then use `dirSync` as `tmp.dirSync`?
we are going to use it in a single place (the `prepareTmpDir`) and using it `tmp.dirSync` would make it immediately clear (at the cost of just 4 more characters).

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -3,6 +3,7 @@ import path from 'path';
 
 import yazl from 'yazl';
 import yauzl from 'yauzl';
+import * as tmp from 'tmp-promise';
```
rpl, line 6: @marchiore 
Nit, I don't think that using `import * as tmp from '...'` here is necessary,
`import tmp from 'tmp-promise';` should work as well (in this case they should be equivalent).

doesn't `import tmp from 'tmp-promise` work for you as expected?

tests/unit/schema/test.firefox-schemas-import.js
marchiore: Done!

rpl: @marchiore I think that I may have commented at the same time when you where writing this comment :-D
and so I guess that you may have missed the new one here https://github.com/mozilla/addons-linter/pull/3095#pullrequestreview-390009513 ;-)

tests/unit/schema/test.firefox-schemas-import.js
marchiore: Changed it!

marchiore:
@rpl Thank you for helping me with this and I am happy to be able to contribute to the project.

Now I'll take a look in another bugs/issues to fix.

caitmuenster:
Thanks so much for the patch @marchiore! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition), and your Mozillians.org profile has been vouched. :) 

Welcome onboard! We look forward to seeing you around!

PR #3123, opened 2020-04-14
Fixes #2158 - Including a new lint rule to check invalid extension name

### LINKED ISSUE #2158: Don't allow use of "Mozilla" or "Firefox" in add-on names
Up until now, we've allowed add-on names to end with "for Firefox" or "for Mozilla". For SEO reasons, we don't want to allow this anymore. Any use of those terms in add-on names should produce an error.

/cc @wagnerand to coordinate policy change when this lands.

marchiore:
This PR Fix #2158 where a new rule was added to validate cases where the words `Mozilla` and `Firefox` were used in the extension name.

I need some help on this to check for localized versions of the name. Is there a place that has already use localized versions of Mozilla and Firefox that i can rely on?

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -6,26 +6,26 @@ Rules are sorted by severity.
 
 ## JavaScript
 
-| Message code | Severity | Description |
-| --- | --- | --- |
-| `KNOWN_LIBRARY` | notice | This is version of a JS library is known and generally accepted. |
-| `OPENDIALOG_NONLIT_URI` | notice | openDialog called with non-literal parameter. |
-| `EVENT_LISTENER_FOURTH` | notice | `addEventListener` called with truthy fourth argument. |
-| `UNEXPECTED_GLOGAL_ARG` | warning | Unexpected global passed as an argument. |
-| `NO_IMPLIED_EVAL` | warning | disallow the use of `eval()`-like methods. |
-| `OPENDIALOG_REMOTE_URI` | warning | openDialog called with non-local URI. |
-| `NO_DOCUMENT_WRITE` | warning | Use of `document.write` strongly discouraged. |
-| `JS_SYNTAX_ERROR` | warning | JavaScript compile-time error. |
-| `UNADVISED_LIBRARY` | warning | This version of a JS library is not recommended. |
-| `DEPRECATED_API` | warning | API is deprecated. |
-| `STORAGE_SYNC` | warning | Temporary IDs can cause issues with `storage.sync`. |
-| `STORAGE_MANAGED` | warning | Temporary IDs can cause issues with `storage.managed`. |
-| `IDENTITY_GETREDIRECTURL` | warning | Temporary IDs can cause issues with `identity.getRedirectURL`. |
-| `RUNTIME_ONMESSAGEEXTERNAL` | warning | Temporary IDs can cause issues with `runtime.onMessageExternal`. |
-| `RUNTIME_ONCONNECTEXTERNAL` | warning | Temporary IDs can cause issues with `runtime.onConnectExternal`. |
-| `BANNED_LIBRARY` | error | This version of a JS library is banned for security reasons. |
-| `INCOMPATIBLE_API` | warning | API not compatible with `applications.gecko.strict_min_version` |
-| `ANDROID_INCOMPATIBLE_API` | warning | API not compatible with Firefox for Android at `applications.gecko.strict_min_version` |
+| Message code               | Severity | Description                                                                            |
+| -------------------------- | -------- | -------------------------------------------------------------------------------------- |
+| `KNOWN_LIBRARY`            | notice   | This is version of a JS library is known and generally accepted.                       |
+| `OPENDIALOG_NONLIT_URI`    | notice   | openDialog called with non-literal parameter.                                          |
+| `EVENT_LISTENER_FOURTH`    | notice   | `addEventListener` called with truthy fourth argument.                                 |
+| `UNEXPECTED_GLOGAL_ARG`    | warning  | Unexpected global passed as an argument.                                               |
+| `NO_IMPLIED_EVAL`          | warning  | disallow the use of `eval()`-like methods.                                             |
+| `OPENDIALOG_REMOTE_URI`    | warning  | openDialog called with non-local URI.                                                  |
+| `NO_DOCUMENT_WRITE`        | warning  | Use of `document.write` strongly discouraged.                                          |
+| `JS_SYNTAX_ERROR`          | warning  | JavaScript compile-time error.                                                         |
+| `UNADVISED_LIBRARY`        | warning  | This version of a JS library is not recommended.                                       |
+| `DEPRECATED_API`           | warning  | API is deprecated.                                                                     |
+| `STORAGE_SYNC`             | warning  | Temporary IDs can cause issues with `storage.sync`.                                    |
+| `STORAGE_MANAGED`          | warning  | Temporary IDs can cause issues with `storage.managed`.                                 |
+| `IDENTITY_GETREDIRECTURL`  | warning  | Temporary IDs can cause issues with `identity.getRedirectURL`.                         |
+| `RUNTIME_ONMESSAGEEXTERNAL`| warning  | Temporary IDs can cause issues with `runtime.onMessageExternal`.                       |
+| `RUNTIME_ONCONNECTEXTERNAL`| warning  | Temporary IDs can cause issues with `runtime.onConnectExternal`.                       |
+| `BANNED_LIBRARY`           | error    | This version of a JS library is banned for security reasons.                           |
+| `INCOMPATIBLE_API`         | warning  | API not compatible with `applications.gecko.strict_min_version`                        |
+| `ANDROID_INCOMPATIBLE_API` | warning  | API not compatible with Firefox for Android at `applications.gecko.strict_min_version` |
 
 ## Markup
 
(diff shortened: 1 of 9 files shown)


wagnerand:
Thank you. Can you please mention the issue this PR fixes in your first comment, or create a new one and mention it, so they are linked together?

src/parsers/manifestjson.js
```diff
@@ -478,6 +479,20 @@ export default class ManifestJSONParser extends JSONParser {
...
     }
+
+    if (this.parsedJSON.name) {
+      const nameLowerCase = String(this.parsedJSON.name).toLowerCase();
+      const nameContainsInvalidWords = NOT_ALLOWED_NAME_WORDS.some((word) =>
+        nameLowerCase.includes(word)
+      );
+
+      if (nameContainsInvalidWords) {
+        this.collector.addError(
```
wagnerand, line 490: This should rather be a warning, as per the discussion in the issue.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -394,6 +394,20 @@ describe('ManifestJSONParser', () => {
...
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector
+      );
+      expect(manifestJSONParser.isValid).toEqual(false);
+      const { errors } = addonLinter.collector;
+      expect(errors[0].code).toEqual(
+        messages.PROP_NAME_MUST_NOT_CONTAIN_MOZILLA_OR_FIREFOX.code
+      );
+    });
```
wagnerand, line 410: Can you also add a test that covers localized names?

src/parsers/manifestjson.js
```diff
@@ -478,6 +479,20 @@ export default class ManifestJSONParser extends JSONParser {
         });
       }
     }
+
+    if (this.parsedJSON.name) {
+      const nameLowerCase = String(this.parsedJSON.name).toLowerCase();
```
wagnerand, line 484: This should already be a string, no? Can you explain why casting to `String` again is necessary?

docs/rules.md
```diff
@@ -6,26 +6,26 @@ Rules are sorted by severity.
...
-| `STORAGE_SYNC`             | warning  | Temporary IDs can cause issues with `storage.sync`.                                    |
-| `STORAGE_MANAGED`          | warning  | Temporary IDs can cause issues with `storage.managed`.                                 |
-| `IDENTITY_GETREDIRECTURL`  | warning  | Temporary IDs can cause issues with `identity.getRedirectURL`.                         |
-| `RUNTIME_ONMESSAGEEXTERNAL`| warning  | Temporary IDs can cause issues with `runtime.onMessageExternal`.                       |
-| `RUNTIME_ONCONNECTEXTERNAL`| warning  | Temporary IDs can cause issues with `runtime.onConnectExternal`.                       |
-| `BANNED_LIBRARY`           | error    | This version of a JS library is banned for security reasons.                           |
-| `INCOMPATIBLE_API`         | warning  | API not compatible with `applications.gecko.strict_min_version`                        |
-| `ANDROID_INCOMPATIBLE_API` | warning  | API not compatible with Firefox for Android at `applications.gecko.strict_min_version` |
+| Message code | Severity | Description |
```
wagnerand, line 9: Could you avoid trimming all the whitespaces and make sure newly added entries are aligned like the others to improve plain-text readability?

src/parsers/manifestjson.js
marchiore: Sure @wagnerand , i made the parse to String because it wasn't recognizing the `toLowerCase` as a function like the image below. is there a best way to do it?

Screen Shot 2020-04-17 at 10 49 04

wagnerand: So, this comes from the test `should collect an error on non-string name value`. It makes sure we fail if the name is not a string. In that case, we don't need to check whether it contains any of the forbidden words.

You would want to check whether the name is a string as well:

```
if (this.parsedJSON.name && typeof this.parsedJSON.name === 'string') {
```

docs/rules.md
marchiore: it happens when i run `npm run prettier-dev` but i've already reverted it.

tests/unit/parsers/test.manifestjson.js
marchiore: @wagnerand thanks for your review!

Before i start coding this i would like to share to you my solution proposal:

First i'll include a new `msgid` in all `messages.po` like:
(is there an automated way to do it? or it has to be manually?)

```
msgid "mozilla"
msgstr "<string for the equivalent value to the language of the country>"
```
```
msgid "firefox"
msgstr "<string for the equivalent value to the language of the country>"
```

Than i will use [this regex](https://github.com/mozilla/addons-linter/blob/e0e1946a5e88e3536386562c514f1f2a9afb8682/src/const.js#L215) to detect if the extension is localizing the extension name in the manifest.

If there is any localized character will loop through files and then check the msgid for the localized string. If match with `msgid "firefox"` or `msgid "mozilla"` will show a warning about invalid extension name.

is this the way?

I thought about this this solution from some tips from @rpl and @freaktechnik on the Add-on matrix channel.

wagnerand: I don't think this is what you want. Have a look at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization to see how localization works in extensions. Also `src/parsers/locale-messagesjson.js´ might be useful, it deals with localization.

marchiore: Thanks @wagnerand for the explanation! Now I understand the problem and update my PR with a solution with i've included a test to validate it too.

marchiore:
@EnTeQuAk thanks for your review!

I made a change in my PR with a solution based on your suggestion, could you take a look at it?

Any suggestion is welcome.

src/parsers/locale-messagesjson.js
```diff
@@ -153,6 +159,29 @@ export default class LocaleMessagesJSONParser extends JSONParser {
         );
       }
 
+      if (this.addonMetadata && typeof this.addonMetadata.name === 'string') {
+        const messageNameRegexp = new RegExp(MANFIEST_MESSAGE_NAME_REGEXP, 'g');
+        const match = messageNameRegexp.exec(this.addonMetadata.name);
```
freaktechnik, line 164: This match should probably be found outside of the loop this is in.

marchiore: I really missed this one. fixed!

src/parsers/locale-messagesjson.js
```diff
@@ -153,6 +166,23 @@ export default class LocaleMessagesJSONParser extends JSONParser {
         );
       }
 
+      if (messageNameMatch) {
+        const messageName = messageNameMatch[1];
+        let nameContainsInvalidWords = false;
+        if (message === messageName) {
```
wagnerand, line 172: If you make this `if` the toplevel one, we avoid some unnecessary work when we are not looking at `name`.

src/parsers/locale-messagesjson.js
```diff
@@ -153,6 +166,23 @@ export default class LocaleMessagesJSONParser extends JSONParser {
...
+          const nameLowerCase = this.parsedJSON[message].message.toLowerCase();
+          nameContainsInvalidWords = NOT_ALLOWED_NAME_WORDS.some((word) =>
+            nameLowerCase.includes(word)
+          );
+        }
+        if (nameContainsInvalidWords) {
+          this.collector.addWarning(
+            messages.PROP_NAME_MUST_NOT_CONTAIN_MOZILLA_OR_FIREFOX
+          );
```
wagnerand, line 181: You probably want to override the `file` here, since the default is `MANIFEST_JSON`, but we are actually in a locale file. The following should work:

```suggestion
          this.collector.addWarning({
            ...messages.PROP_NAME_MUST_NOT_CONTAIN_MOZILLA_OR_FIREFOX,
            file: this.filename,
          });
```

src/parsers/manifestjson.js
```diff
@@ -478,6 +480,23 @@ export default class ManifestJSONParser extends JSONParser {
         });
       }
     }
+
+    if (this.parsedJSON.name && typeof this.parsedJSON.name === 'string') {
+      const messageNameRegexp = new RegExp(MANFIEST_MESSAGE_NAME_REGEXP, 'g');
+      const match = messageNameRegexp.exec(this.parsedJSON.name);
```
wagnerand, line 486: Nit: No need to declare a `const` (or variable) if the value expression is simple and only used once. You can move it inside the `if`.

tests/unit/parsers/test.locale-messagesjson.js
```diff
@@ -266,6 +266,27 @@ describe('LocaleMessagesJSONParser', () => {
...
+      { name: '__MSG_extensionName__' }
+    );
+
+    localeMessagesJSONParser.parse();
+    expect(localeMessagesJSONParser.isValid).toEqual(false);
+    const { warnings } = addonLinter.collector;
+    expect(warnings[0].code).toEqual(
+      messages.PROP_NAME_MUST_NOT_CONTAIN_MOZILLA_OR_FIREFOX.code
+    );
+  });
```
wagnerand, line 288: Could you add another check here that makes sure the matched filename is `messages.json`?

marchiore: Done!

marchiore:
Solved.
It's always good to improve the code! 
Any other changes just tell me.

Thanks!

marchiore:
Solved.
It's always good to improve the code! 
Any other changes just tell me.

Thanks!

marchiore:
Thanks you all for supporting me into this!

Issue #3125, opened 2020-04-16
Travis CI should run prettier check


willdurand:
We should add a travis-ci job job to run prettier and double check that PRs are "prettier-compliant". See: https://github.com/mozilla/addons-frontend/blob/375eee5c2ba0dc24b774e710b0fa301f1ee5b924/package.json#L34

I don't know if we have that already but we could also add a `prettier-dev` npm script to run [pretty-quick](https://www.npmjs.com/package/pretty-quick). This command would only run prettier on changed files and that is both fast and friendly.

ioanarusiczki:
@willdurand Please add some str for qa else mark this qa not needed. 
Thanks!

PR #3132, opened 2020-04-21
Add prettier-ci npm script

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3125

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,6 +13,7 @@ LICENSE
 /locale/
 /node_modules/
 /tests/fixtures/
+/coverage/
 
 # white-list files we want to process
 !*.js
--- a/.prettierrc
+++ b/.prettierrc
@@ -15,6 +15,12 @@
       "options": {
         "parser": "babel"
       }
+    },
+    {
+      "files": "*.md",
+      "options": {
+        "proseWrap": "preserve"
+      }
     }
   ]
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ script:
   - npm run lint
   # run integration tests using an addons-linter binary in a production-like environment
   - npm run test-integration:production
+  - npm run prettier-ci
 notifications:
   slack:
     if: branch = master
--- a/README.md
+++ b/README.md
@@ -141,17 +141,18 @@ Dependencies are automatically kept up-to-date using [renovatebot](https://renov
 
 #### npm scripts
 
-| Script                          | Description                                                                      |
-| ------------------------------- | -------------------------------------------------------------------------------- |
-| npm test                        | Runs the tests (watches for changes)                                             |
-| npm [run] build                 | Builds the lib (used by Travis)                                                  |
-| npm run test-coverage           | Runs the tests with coverage (watches for changes)                               |
-| npm run test-once               | Runs the tests once                                                              |
-| npm run lint                    | Runs ESLint                                                                      |
-| npm run test-coverage-once      | Runs the tests once with coverage                                                |
-| npm run test-integration-linter | Runs our integration test-suite                                                  |
-| npm run prettier                | Automatically format the whole code-base with Prettier                           |
-| npm run prettier-dev            | Automatically compare and format modified source files against the master branch |
+| Script | Description |
+| --- | --- |
+| npm test | Runs the tests (watches for changes) |
+| npm [run] build | Builds the lib (used by Travis) |
+| npm run test-coverage | Runs the tests with coverage (watches for changes) |
+| npm run test-once | Runs the tests once |
+| npm run lint | Runs ESLint |
+| npm run test-coverage-once | Runs the tests once with coverage |
+| npm run test-integration-linter | Runs our integration test-suite |
+| npm run prettier | Automatically format the whole code-base with Prettier |
+| npm run prettier-ci | Run [Prettier][] and fail if some code has been changed without being formatted |
+| npm run prettier-dev | Automatically compare and format modified source files against the master branch |
 
 ### Building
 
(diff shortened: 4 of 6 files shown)


willdurand:
Expected failure: https://travis-ci.org/github/mozilla/addons-linter/jobs/677752845#L1741

willdurand:
@rpl the first adds all we need to run a prettier check in travis ci but I did not run prettier and it found issues (markdown files) 🤷 The second commit fixes those issues. Let me know what you think.

FWIW, in our other projects, we stopped trying to format tables and let Prettier does its thing.

rpl:
> @rpl the first adds all we need to run a prettier check in travis ci but I did not run prettier and it found issues (markdown files) shrug The second commit fixes those issues. Let me know what you think.
> 
> FWIW, in our other projects, we stopped trying to format tables and let Prettier does its thing.

@willdurand 
yeah, fighting against Prettier is pointless, it doesn't listen ;-)

I'm personally fine to let Prettier do its thing, but I was expecting Prettier to actually taking care of formatting the markdown table to adjust it to its content as it seems that it should be actually supported starting from Prettier 1.8, also accordingly to what I see in the section related to the markdown table in this Prettier blog post related to the Prettier 1.8: https://prettier.io/blog/2017/11/07/1.8.0.html

I'm wondering why it isn't doing that, am I reading the blog post wrong?

willdurand:
> I'm personally fine to let Prettier do its thing, but I was expecting Prettier to actually taking care of formatting the markdown table to adjust it to its content as it seems that it should be actually supported starting from Prettier 1.8, also accordingly to what I see in the section related to the markdown table in this Prettier blog post related to the Prettier 1.8: [prettier.io/blog/2017/11/07/1.8.0.html](https://prettier.io/blog/2017/11/07/1.8.0.html)
> 
> I'm wondering why it isn't doing that, am I reading the blog post wrong?

The problem is that Prettier won't format a table if its width exceeds the `printWidth` when `proseWrap` is set to `never`. Guess what [our config](https://github.com/mozilla/addons-linter/blob/66c6760647d84a66c67266dd9dfec8c138543e9e/.prettierrc#L5) contains? See also: https://github.com/prettier/prettier/issues/5651

rpl:
> > I'm personally fine to let Prettier do its thing, but I was expecting Prettier to actually taking care of formatting the markdown table to adjust it to its content as it seems that it should be actually supported starting from Prettier 1.8, also accordingly to what I see in the section related to the markdown table in this Prettier blog post related to the Prettier 1.8: [prettier.io/blog/2017/11/07/1.8.0.html](https://prettier.io/blog/2017/11/07/1.8.0.html)
> > I'm wondering why it isn't doing that, am I reading the blog post wrong?
> 
> The problem is that Prettier won't format a table if its width exceeds the `printWidth` when `proseWrap` is set to `never`. Guess what [our config](https://github.com/mozilla/addons-linter/blob/66c6760647d84a66c67266dd9dfec8c138543e9e/.prettierrc#L5) contains? See also: [prettier/prettier#5651](https://github.com/prettier/prettier/issues/5651)

eheh, that is definitely explaining it.

how about workarounding the issue by using the [Prettier configuration overrides](https://prettier.io/docs/en/configuration.html#configuration-overrides) to override `proseWrap` for the markdown files?

willdurand:
> how about workarounding the issue by using the [Prettier configuration overrides](https://prettier.io/docs/en/configuration.html#configuration-overrides) to override `proseWrap` for the markdown files?

Yeah, I guess we could but can we do this with the current `prettierrc` file format? Or should we switch to a different format?

willdurand:
Also we should do this in the other repos too, for consistency.

rpl:
> > how about workarounding the issue by using the [Prettier configuration overrides](https://prettier.io/docs/en/configuration.html#configuration-overrides) to override `proseWrap` for the markdown files?
> 
> Yeah, I guess we could but can we do this with the current `prettierrc` file format? Or should we switch to a different format?

It seems that we are already using the `override` property in the current prettierrc, and so I think we could just add one more entry to that one restricted to the markdown files, isn't that enough? 

> Also we should do this in the other repos too, for consistency.

sure, web-ext and the webextension-polyfill reports doesn't use prettier yet, and so I guess that the other repos may be the addons-frontend and sign-addon? let's file an issue in our repositories where prettier is already being used and link this one.

willdurand:
> It seems that we are already using the `override` property in the current prettierrc, and so I think we could just add one more entry to that one restricted to the markdown files, isn't that enough?

ugh, sorry about that. I was thinking about the config for addons-frontend.. and I did not notice that the linter had a different config.

willdurand:
@rpl updated, then again sorry. On a second thought, I think changing `proseWrap` for the tables might not be a good idea because it means we'll have to be careful when we'll add new content to the md files (as we can't tell Prettier to format tables and ensure no prose wrap).

rpl:
> @rpl updated, then again sorry. On a second thought, I think changing `proseWrap` for the tables might not be a good idea because it means we'll have to be careful when we'll add new content to the md files (as we can't tell Prettier to format tables and ensure no prose wrap).

As we discussed over chat, I'm not currently really worried about that, and the markdown as reformatted in this last version of the pull request looks pretty good to me.

If it turns out to be a problem we could choose to tweak the prettier config a bit, e.g.
- to only set `"proseWrap": "preserve"` on docs/*.md and not the README.md file
- or add another entry to the overrides property which is specifically selecting the markdown file (or files) for which we don't want to use the `"proseWrap": "preserve"` config

I'm going to accept this patch on my side, if @EnTeQuAk doesn't have any concern (or any pull request that he wants to land before this one), this should be good to go.

Thanks again for taking care of this.

EnTeQuAk:
> I'm going to accept this patch on my side, if @EnTeQuAk doesn't have any concern (or any pull request that he wants to land before this one), this should be good to go.

@willdurand feel free to land this any time

PR #3181, opened 2020-05-25
Upgrade ESLint to 6.x

willdurand:
This patch replaces https://github.com/mozilla/addons-linter/pull/2652.

---

Update 2020.06.10: I updated the code again with less changes and re-enabled all the tests that were skipped in the previous iteration. This patch works locally with or without bundling, and it should be good enough for a review. Thanks!

---

I don't know if that's the right approach but that's something at least. In short, it's tricky to update to 6.x (or even 7.x) because of the way we use both ESLint `cli` and `linter`.

As of v6, we cannot access or modify the `linter` instance of a `CLIEngine` instance (so `cli.linter.defineRule` does not work anymore). We could create our own `Linter` instance but it does not work very well (see: https://github.com/eslint/eslint/issues/12689#issuecomment-567871422 but also https://eslint.org/docs/user-guide/migrating-to-6.0.0#linter-parsers).

We need the `linter` instance to define rules (using `defineRule`) but that's not possible using the `CLIEngine`, which is the recommended Node API for ESLint v6 (we'll have to change that for v7 but that's a different story).

So we cannot use `defineRule`. No problem, we can tell `CLIEngine` to load the rules for us. That's what is done in this patch and it seems to work based on the test suite. The problem with this approach is that we cannot inject fake rule definitions anymore and that's why I _skipped_ a few test cases.

I don't think we can bring this behavior back easily but I dunno, maybe I am missing something obvious (I spent more time than I wished on this patch already).

I have the feeling that we should stop using the `Linter` instance as well as registering the actual rule definitions. Instead, we should configure rule names and paths, and ESLint would load the rules on its own. Hence this patch.

WDYT?


willdurand:
cc @rpl @Rob--W

src/scanners/javascript.js
```diff
@@ -81,7 +63,8 @@ export default class JavaScriptScanner {
         ecmaVersion: ECMA_VERSION,
         sourceType: this.sourceType,
       },
-      rules: _ruleMapping,
+      rules,
+      rulePaths: [path.join(__dirname, '..', 'rules', 'javascript')],
```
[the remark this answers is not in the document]
willdurand: Yeah, if we want to adopt this approach, we'll have to double check that part. I haven't checked the build step yet because I am not too afraid of it, but it's likely broken indeed.

src/scanners/javascript.js
```diff
@@ -90,31 +84,26 @@ export default class JavaScriptScanner {
       // the scan. See: https://github.com/mozilla/addons-linter/issues/1288
       ignore: false,
       patterns: ['!bower_components/*', '!node_modules/*'],
-      // Also, don't ignore dotfiles in scans.
-      dotfiles: true,
```
rpl, line 94: @willdurand did you checked if we do have coverage to be sure that dotfiles are not ignored in the scans?

tests/unit/scanners/test.javascript.js
```diff
@@ -263,9 +218,8 @@ describe('JavaScript Scanner', () => {
     expect(ok).toBeTruthy();
   });
 
+  // eslint-disable-next-line jest/no-disabled-tests
```
rpl, line 221: This doesn't seem needed anymore, this test isn't disabled anymore in this version of the patch.

tests/unit/scanners/test.javascript.js
```diff
@@ -323,8 +261,8 @@ describe('JavaScript Scanner', () => {
 
     const jsScanner = new JavaScriptScanner('', 'badcode.js');
 
-    await jsScanner.scan();
-    expect(jsScanner._rulesProcessed).toEqual(Object.keys(rules).length);
+    await runJsScanner(jsScanner);
+    expect(jsScanner._rulesProcessed).toEqual(16);
```
rpl, line 265: it would be nice it we could get the expected value for jsScanner._rulesProcessed in a programmatic way instead of using directly the value the we currently expect.

if there isn't a way to avoid this, it may be good to leave a comment to explain where that value is coming from and how a new expected value should be determined if it has to be updated in the future.

tests/unit/scanners/test.javascript.js
```diff
@@ -352,114 +290,34 @@ describe('JavaScript Scanner', () => {
         fakeMetadata
       );
 
-      const { linterMessages } = await jsScanner.scan();
+      const { linterMessages } = await runJsScanner(jsScanner);
       expect(linterMessages.length).toEqual(0);
     });
   });
 
+  // eslint-disable-next-line jest/no-disabled-tests
```
rpl, line 298: Same as previous comment, this eslint-disable-next-line doesn't seem to be needed anymore.

webpack.config.js
```diff
@@ -9,12 +11,19 @@ module.exports = {
   // Set the webpack4 mode 'none' for compatibility with the behavior
   // of the webpack3 bundling step.
   mode: 'none',
-  entry: './src/main.js',
+  entry: {
+    'addons-linter': './src/main.js',
```
rpl, line 15: nit, would you mind to move this after the `...glob.sync(...).reduce(...)`?
just to be on the "safe side" and be absolutely sure that there is no change for it to be overwritten by the computed ones by mistake.

webpack.config.js
```diff
@@ -48,4 +57,7 @@ module.exports = {
     modules: ['src', 'node_modules'],
   },
   devtool: 'sourcemap',
+  node: {
+    __dirname: false,
+  },
```
rpl, line 62: [isn't `__dirname: false` the default](https://webpack.js.org/configuration/node/#node__dirname)?

It may be good to put an inline comment here to briefly mention why it is important, given that we are explicitly setting it.

tests/unit/scanners/test.javascript.js
willdurand: oops, sorry

tests/unit/scanners/test.javascript.js
willdurand: :/

tests/unit/scanners/test.javascript.js
willdurand: I added a comment.

webpack.config.js
willdurand: I think it is but last time I checked, it was not working as intended. Let me check again.

src/scanners/javascript.js
willdurand: Good call, let me double check.

webpack.config.js
willdurand: ha well, it works now.. 🤷

willdurand:
> As a side note, it may be good to release this change as part of a major version bump (to be fair I think that we did already discussed briefly about this and we both agreed about that, but I thought to mention it in case I was wrong and we didn't actually talk about it yet).

Yes, that's what we decided last week :D

tests/unit/scanners/test.javascript.js
```diff
@@ -125,6 +125,39 @@ describe('JavaScript Scanner', () => {
...
+  it('should scan node modules', async () => {
+    const code = oneLine`var a;`;
+
+    const jsScanner = new JavaScriptScanner(
+      code,
+      'node_modules/module/code.js'
+    );
+
+    const { linterMessages } = await jsScanner.scan();
+    expect(linterMessages.length).toEqual(0);
```
rpl, line 137: @willdurand Thanks for double-checking this and adding new tests to cover it.

While looking to these new test cases I was wondering:
Are these test going to fail without the related fix applied to src/scanners/javascript.js?
Which kind of error they trigger in that case?

maybe we could put into code something that would trigger a validation error or warning on purpose, and then we could assert the expected linting message to be absolutely sure that the file has been scanned as expected and not been skipped. 

How that sounds to you?

willdurand: Test cases fail indeed. They raise the error mentioned in the original issue:

```
   - Array []
    + Array [
    +   Object {
    +     "code": "File ignored by default.  Use a negated ignore pattern (like \"--ignore-pattern '!<relative/path/to/filename>'\") to override.",
    +     "column": undefined,
    +     "description": null,
    +     "file": ".code.js",
    +     "line": undefined,
    +     "message": "File ignored by default.  Use a negated ignore pattern (like \"--ignore-pattern '!<relative/path/to/filename>'\") to override.",
    +     "sourceCode": undefined,
    +     "type": "warning",
    +   },
    + ]
```

tests/unit/scanners/test.javascript.js
```diff
@@ -144,6 +125,39 @@ describe('JavaScript Scanner', () => {
...
+  it('should scan node modules', async () => {
+    const code = 'el.innerHTML = evilContent';
+
+    const jsScanner = new JavaScriptScanner(
+      code,
+      'node_modules/module/code.js'
+    );
+
+    const { linterMessages } = await jsScanner.scan();
+    expect(linterMessages[0]).toMatchObject({ code: 'UNSAFE_VAR_ASSIGNMENT' });
```
rpl, line 137: @willdurand thanks for making this assertion a bit more explicit! :heart: 

lgtm, r+ renewed

Rob--W:
Thanks @willdurand for the patch and @rpl for the reviews!

Issue #3228, opened 2020-06-30
Only deploy once on Travis-CI


willdurand:
I noticed that the `deploy` step was run twice on Travis because we have two builds (one for each node env). Let's deploy once instead. We can do this by specifying the node version: https://docs.travis-ci.com/user/deployment/#examples-of-conditional-deployment

PR #3229, opened 2020-06-30
Only deploy once on Travis-CI

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3228

---

I chose the lower node version because this is the one we support. It's sad that we can reference the first entry of the top-level `node_js` key but it is not so bad. We don't change those versions very often.

I also added some blank lines because the content of this file is hard to read..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ deploy:
     tags: true
     repo: mozilla/addons-linter
     branch: master
-    node_js: 'v10.21.0'
+    node_js: 'v11.21.0'
 
 after_success: npm run publish-rules > /dev/null 2>&1
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ deploy:
     tags: true
     repo: mozilla/addons-linter
     branch: master
-    node_js: 'v11.21.0'
+    node_js: 'v10.21.0'
 
 after_success: npm run publish-rules > /dev/null 2>&1
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,6 +35,8 @@ deploy:
     tags: true
     repo: mozilla/addons-linter
     branch: master
+    node_js: 'v10.21.0'
```
rpl, line 38: @willdurand did you already checked if there will be a visible failure if at some point we do remove nodejs v10.21.0 from the job matrix (the one at line 3) and we forget to update the node_js version referenced here?

willdurand: No, not sure how to test this.

I don't think we'd forget that but we could mitigate the risk by adding a comment at the top of the file?

rpl: > No, not sure how to test this.

I guess that one easy way to test it is to refer a node_js version that is supported by travis but not listed in the job matrix at line 3.
 
> I don't think we'd forget that but we could mitigate the risk by adding a comment at the top of the file?

Sure, I'll be ok with just adding a comment too

willdurand: > I guess that one easy way to test it is to refer a node_js version that is supported by travis but not listed in the job matrix at line 3.

@rpl looks like travis does not care..

willdurand: I added a comment.

willdurand: Note: I also tried to use the `travis lint` command but it did not raise any error either. It could raise an error at runtime, but we never know with travis..

PR #3231, opened 2020-06-30
Make sure __dirname is false in webpack config

willdurand:
@rpl should we have a travis job that checkouts `web-ext`, then `npm link addons-linter` and finally run web-ext's functional tests?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -57,4 +57,10 @@ module.exports = {
     modules: ['src', 'node_modules'],
   },
   devtool: 'sourcemap',
+  node: {
+    // This is required because the default value does not seem to be `false`.
+    // If this isn't set to `false`, `__dirname` is likely invalid and it
+    // prevents JS rules to be loaded correctly.
+    __dirname: false,
+  },
 };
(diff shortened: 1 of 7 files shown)


rpl:
> @rpl should we have a travis job that checkouts `web-ext`, then `npm link addons-linter` and finally run web-ext's functional tests?

that sounds like a really good idea, it would let us catch this kind of integration issues before releasing a new addons-linter version of github

webpack.config.js
```diff
@@ -57,4 +57,7 @@ module.exports = {
     modules: ['src', 'node_modules'],
   },
   devtool: 'sourcemap',
+  node: {
+    __dirname: false,
+  },
```
rpl, line 62: Nit, let's add a comment so that we don't forget that we absolutely need this to be set to false explicitly here.

willdurand:
@rpl I know you r+ed this patch but I added a script to run web-ext's functional tests. I tested it locally and it works on Travis-CI (I made sure to change the order of the commits to prove that it works). Let me know what you think.

PR #3235, opened 2020-07-03
feat: add thunderbird schemas

ccarruitero:
This is an attempt to add Thunderbird schemas, so will be possible use `addons-linter` to validate Thunderbird extensions.

I added new scripts for list and get Thunderbird source code, move imported firefox schemas into `src/schema/firefox` and refactored a bit `src/schema/firefox-schema-import.js` so can complete import schemas for Thunderbird.

Sadly, the additional test I added is failing because the manifest schema (`src/schema/firefox/manifest.json`) doesn't have new Thunderbird schema permissions.

After looking into `src/schema/firefox-schema-import.js`, I'm not really sure where permissions are added into manifest schema. If you could point me where I have to look in order to solve this, will be very appreciated.


wagnerand:
The add-ons linter for Thunderbird uses its own repository. You'd want to open this PR on https://github.com/thundernest/addons-linter/ .

ccarruitero:
Thanks @wagnerand

Looking into that repository, doesn't have too much activity or a package published yet. So, I think maybe we can figurate a way to integrate other schemas inside this repo or that we can allow add other schemas as a plugin module. That way we don't have to duplicate entire code base. What do you think?

I'll try to ping people in thunderbird's repository to join the conversation

wagnerand:
We have removed other products from AMO components, so I am sure how useful it would be to add this here. You are right, it seems the server behind addons.thunderbird.net still uses this addons-linter, instead of their own. It also uses a very old version of it, so even if we accepted this patch, I'd assume that it would be substantial effort to upgrade to a recent version of this add-ons linter. 

Honestly, I believe you might have a better chance of getting this landed in the addons-linter for Thunderbird. It would be easier for Thunderbird folks to update the server to use their current version + your patch, than their current version + all our patches since then + your patch.

wagnerand:
@ccarruitero Thank you for this work though, I am sure the Thunderbird folks appreciate this as well!

PR #3243, opened 2020-07-09
feat: Imported Firefox 79b5 schema

rpl:
Import updated API schema files from Firefox 79b5:

- allowed localhost hostname/ips in chrome_settings_overrides's search_url
- new tabs.warmup API method
- some additions to the browser.experiments.urlbar API

### DIFF
--- a/src/schema/imported/chrome_settings_overrides.json
+++ b/src/schema/imported/chrome_settings_overrides.json
@@ -25,7 +25,7 @@
                 "search_url": {
                   "type": "string",
                   "format": "url",
-                  "pattern": "^https://.*$",
+                  "pattern": "^(https://|http://(localhost|127\\.0\\.0\\.1|\\[::1\\])).*$",
                   "preprocess": "localize"
                 },
                 "favicon_url": {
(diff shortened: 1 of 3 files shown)


src/schema/imported/chrome_settings_overrides.json
```diff
@@ -25,7 +25,7 @@
                 "search_url": {
                   "type": "string",
                   "format": "url",
-                  "pattern": "^https://.*$",
+                  "pattern": "^(https://|http://(localhost|127\\.0\\.0\\.1|\\[::1\\])).*$",
```
Rob--W, line 28: This pattern is incorrect, I filed a bug in Firefox @ https://bugzilla.mozilla.org/show_bug.cgi?id=1651665

To fix this, add `(:\d+)?(/|$)`

willdurand:
@wagnerand so I wrote a quick extension with the `warmup` API and I only get a warning from the linter (saying that the API isn't available in FF). We'll deploy the upcoming new release next week then (no need to rush).

willdurand:
Replaced by https://github.com/mozilla/addons-linter/pull/3250

PR #3258, opened 2020-07-20
Add longer timeout for web-ext functional tests

wagnerand:
Depends on https://github.com/mozilla/web-ext/pull/1966.

### DIFF
--- a/scripts/webext-test-functional
+++ b/scripts/webext-test-functional
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+export MOCHA_TIMEOUT=60000
+
 # Build addons-linter
 npm install
 npm run build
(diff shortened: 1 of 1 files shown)


scripts/webext-test-functional
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+export MOCHA_TIMEOUT=60000
```
rpl, line 3: Nit, we could also restrict the environment variables to the web-ext functional tests by doing something like the following at line 22: (line 20 in the original version of this bash script):

```
MOCHA_TIMEOUT=60000 npm run test-functional
```

wagnerand: e3a849c9

PR #3289, opened 2020-08-11
feat: Imported Firefox 80b6 schema

rpl:
This schema import contains:

- a small addition to the geckoProfiler ProfilerFeature enum
- the addition manifest entry for the background service worker

On the Firefox side the background.service_worker property is guarded by the `requireBackgroundServiceWorkerEnabled` postprocess function and locked behind a pref disabled by default on all channels.

It may be worth to special case the background.service_worker property as part of the manifest linting and raise an error if it is being used (and file a follow up to remove that once we are going to enable it by default on the Firefox side).

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -314,6 +314,16 @@ export default class ManifestJSONParser extends JSONParser {
           'page'
         );
       }
+      if (this.parsedJSON.background.service_worker) {
+        // TODO: replace this with a call to this.validateFileExistsInPackage
+        // once background.service_worker is supported on the Firefox side and
+        // we can accept it on AMO submissions
+        // (https://github.com/mozilla/addons-linter/issues/3290).
+        this.collector.addError(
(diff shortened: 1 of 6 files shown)


rpl:
@Rob--W I added an additional patch (c4ad6b3) to make sure that we don't accept yet extensions that are using background.service_worker in their manifest.

I'd be ok to move that in a separate PR if you prefer.

src/parsers/manifestjson.js
```diff
@@ -314,6 +314,15 @@ export default class ManifestJSONParser extends JSONParser {
           'page'
         );
       }
+      if (this.parsedJSON.background.service_worker) {
+        // TODO: replace this with a call to this.validateFileExistsInPackage
```
Rob--W, line 318: Any reference to an issue?

rpl: sure, I was going to file the follow up issue and mention it in this comment after double-checking that we agreed on doing this in this PR.

PR #3310, opened 2020-08-26
fix: Align optional_permissions validation to the validation rules already applied to the manifest permissions 

### LINKED ISSUE #3060: Unknown optional_permissions are being classified as errors by the addons-linter and as warnings by Firefox

Additional details about this problem and how to reproduce it are in the issue originally reported on bugzilla: [Bug 1618137 - Management permission not accepted in optional_permissions](https://bugzilla.mozilla.org/show_bug.cgi?id=1618137).

### What happened?

an unknown (or invalid, as in "permission not supported as optional on Firefox")  optional_permissions is being classified by the linter as an error and so it prevents the AMO submission to complete successfully.
On the Firefox side, an unknown or invalid value in the option_permissions is being classified as a warning and just ignored.

it may be reasonable to align the two behaviors and classify this validation issues as warning also when an addon is submitted to AMO.

mentor: @rpl 
more details about the issue and the addons-linter internals to look into in [this comment](https://github.com/mozilla/addons-linter/issues/3060#issuecomment-665718913).

aa-deet-eeya:
Fixes #3060

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -232,6 +232,13 @@ export default class ManifestJSONParser extends JSONParser {
     ) {
       baseObject = messages.MANIFEST_BAD_PERMISSION;
       overrides.message = `Permissions ${error.message}.`;
+    } else if (
+      error.dataPath.startsWith('/optional_permissions') &&
+      typeof error.data !== 'undefined' &&
+      typeof error.data !== 'string'
+    ) {
+      baseObject = messages.MANIFEST_BAD_OPTIONAL_PERMISSION;
+      overrides.message = `Permissions ${error.message}.`;
     } else if (error.keyword === 'type') {
       baseObject = messages.MANIFEST_FIELD_INVALID;
     }
@@ -242,7 +249,11 @@ export default class ManifestJSONParser extends JSONParser {
     const match = error.dataPath.match(
       /^\/(permissions|optional_permissions)\/([\d+])/
     );
-    if (match && baseObject.code !== messages.MANIFEST_BAD_PERMISSION.code) {
+    if (
+      match &&
+      baseObject.code !== messages.MANIFEST_BAD_PERMISSION.code &&
+      baseObject.code !== messages.MANIFEST_BAD_OPTIONAL_PERMISSION.code
+    ) {
       baseObject = messages[`MANIFEST_${match[1].toUpperCase()}`];
       overrides.message = oneLine`/${match[1]}: Unknown ${match[1]}
           "${error.data}" at ${match[2]}.`;
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -239,9 +239,9 @@ export default class ManifestJSONParser extends JSONParser {
     // Arrays can be extremely verbose, this tries to make them a little
     // more sane. Using a regex because there will likely be more as we
     // expand the schema.
-    const match =
-      error.dataPath.match(/^\/(optional_permissions)\/([\d+])/) ||
-      error.dataPath.match(/^\/(permissions)\/([\d+])/);
+    const match = error.dataPath.match(
+      /^\/(permissions|optional_permissions)\/([\d+])/
+    );
     if (match && baseObject.code !== messages.MANIFEST_BAD_PERMISSION.code) {
       baseObject = messages[`MANIFEST_${match[1].toUpperCase()}`];
       overrides.message = oneLine`/${match[1]}: Unknown ${match[1]}
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -239,7 +239,9 @@ export default class ManifestJSONParser extends JSONParser {
     // Arrays can be extremely verbose, this tries to make them a little
     // more sane. Using a regex because there will likely be more as we
     // expand the schema.
-    const match = error.dataPath.match(/^\/(permissions)\/([\d+])/);
+    const match =
+      error.dataPath.match(/^\/(optional_permissions)\/([\d+])/) ||
+      error.dataPath.match(/^\/(permissions)\/([\d+])/);
     if (match && baseObject.code !== messages.MANIFEST_BAD_PERMISSION.code) {
       baseObject = messages[`MANIFEST_${match[1].toUpperCase()}`];
       overrides.message = oneLine`/${match[1]}: Unknown ${match[1]}
@@ -252,7 +254,10 @@ export default class ManifestJSONParser extends JSONParser {
   _validate() {
     // Not all messages returned by the schema are fatal to Firefox, messages
     // that are just warnings should be added to this array.
-    const warnings = [messages.MANIFEST_PERMISSIONS.code];
+    const warnings = [
+      messages.MANIFEST_PERMISSIONS.code,
+      messages.MANIFEST_OPTIONAL_PERMISSIONS.code,
+    ];
     let validate = validateAddon;
 
     if (this.isStaticTheme) {
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -241,7 +241,6 @@
                 },
                 "optional_permissions": {
                   "type": "array",
-                  "default": [],
                   "items": {
                     "allOf": [
                       {
@@ -249,7 +248,8 @@
                       },
                       {
                         "onError": "warn"
-                      }],
+                      }
+                    ],
                     "anyOf": [
                       {},
                       {
@@ -257,6 +257,7 @@
                       }
                     ]
                   },
+                  "default": [],
                   "uniqueItems": true
                 },
                 "web_accessible_resources": {
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -241,6 +241,7 @@
                 },
                 "optional_permissions": {
                   "type": "array",
+                  "default": [],
                   "items": {
                     "allOf": [
                       {
@@ -251,7 +252,7 @@
                       }
                     ]
                   },
-                  "default": []
+                  "uniqueItems": true
                 },
                 "web_accessible_resources": {
                   "type": "array",
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -249,6 +249,11 @@
                       },
                       {
                         "onError": "warn"
+                      }],
+                    "anyOf": [
+                      {},
+                      {
+                        "format": "deprecated"
                       }
                     ]
                   },
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -46,7 +46,7 @@
                     },
                     "uniqueItems": true
                   } ,
-                  "optioanl_permissions": {
+                  "optional_permissions": {
                     "items": {
                       "anyOf": [
                         {},
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -45,6 +45,15 @@
                       ]
                     },
                     "uniqueItems": true
+                  } ,
+                  "optioanl_permissions": {
+                    "items": {
+                      "anyOf": [
+                        {},
+                        { "format": "deprecated" }
+                      ]
+                    },
+                    "uniqueItems": true
                   }
                 }
               }
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -211,6 +211,25 @@ describe('ManifestJSONParser', () => {
       });
     });
 
+    it('should not error if optional_permission is a string (even if unknown)', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        optional_permissions: ['identity', 'fileSystem'],
+      });
+
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector
+      );
+      expect(manifestJSONParser.isValid).toEqual(false);
+      const { warnings } = addonLinter.collector;
+      expect(addonLinter.collector.errors.length).toBe(0);
+      assertHasMatchingError(warnings, {
+        code: messages.MANIFEST_OPTIONAL_PERMISSIONS.code,
+        message: /Unknown optional_permissions "fileSystem"/,
+      });
+    });
+
     it('should error if permission is not a string', () => {
       const addonLinter = new Linter({ _: ['bar'] });
       const json = validManifestJSON({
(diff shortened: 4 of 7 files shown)


tests/unit/test.messages.js
```diff
@@ -11,12 +11,6 @@ describe('Messages', () => {
       it('should have a code set', () => {
         expect(code).not.toEqual(null);
       });
-
-      it(`should have code length <= 25 for ${code}`, () => {
-        // Otherwise the ansi color sequences will be borked
-        // as columnify doesn't handle them when wrapping text.
-        expect(code.length).toBeLessThan(26);
```
aa-deet-eeya, line 18: I took the liberty of scrapping this out.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -338,7 +369,26 @@ describe('ManifestJSONParser', () => {
...
+      const json = validManifestJSON({
+        optional_permissions: ['idle', 'wat'],
+      });
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector
+      );
+      expect(manifestJSONParser.isValid).toEqual(false);
+      const { warnings } = addonLinter.collector;
+      // expect(warnings.length).toEqual(1);
```
aa-deet-eeya, line 383: This was one of the tests that was failing bc the expected warning list is 1 but this case produced 3 warnings. I could use some guidance here @rpl

src/parsers/manifestjson.js
```diff
@@ -239,7 +239,9 @@ export default class ManifestJSONParser extends JSONParser {
     // Arrays can be extremely verbose, this tries to make them a little
     // more sane. Using a regex because there will likely be more as we
     // expand the schema.
-    const match = error.dataPath.match(/^\/(permissions)\/([\d+])/);
+    const match = error.dataPath.match(
+      /^\/(permissions|optional_permissions)\/([\d+])/
```
aa-deet-eeya, line 243: @rpl this one seems more elegant. The previous one was sloppy from my end, I'm sorry.

rpl (changes requested):

tests/unit/parsers/test.manifestjson.js
```diff
@@ -192,24 +192,36 @@ describe('ManifestJSONParser', () => {
...
+            });
+      it(`should not error if ${manifestKey} is a string (even if unknown)`, () => {
+        const addonLinter = new Linter({ _: ['bar'] });
+        const manifestJSONParser = new ManifestJSONParser(
+          json,
+          addonLinter.collector
+        );
+        expect(manifestJSONParser.isValid).toEqual(false);
+        const { warnings } = addonLinter.collector;
+        console.log({ warnings });
```
aa-deet-eeya, line 217: I think somewhere around here the test is failing, I tried debugging but couldn't find much
here's the log btw

● ManifestJSONParser › bad permissions › should not error if permissions is a string (even if unknown)

    expect(received).toBe(expected) // Object.is equality

    Expected: true

    Received: false

      292 |   }; // eslint-disable-line

      293 | 

    > 294 |   let result;

          |       ^

      295 | 

      296 |   function reset() {

      297 |     process.exit = _exit;

      at assertHasMatchingError (tests/unit/helpers.js:294:7)

      at Object.<anonymous> (tests/unit/parsers/test.manifestjson.js:234:45)

  ● ManifestJSONParser › bad permissions › should not error if optional_permissions is a string (even if unknown)

    expect(received).toBe(expected) // Object.is equality

    Expected: true

    Received: false

      292 |   }; // eslint-disable-line

      293 | 

    > 294 |   let result;

          |       ^

      295 | 

      296 |   function reset() {

      297 |     process.exit = _exit;

      at assertHasMatchingError (tests/unit/helpers.js:294:7)

      at Object.<anonymous> (tests/unit/parsers/test.manifestjson.js:234:45)

rpl: This test is failing in the `assertHasMatchingError` helpers, in particular when it does assert that the linting message is matching the one at line 221:

in the previous version of this test case message was a RegExp, in this new version is a string and so the helper would expect the linting message to be strictly equal to this string.

What you want here is to generate a RegExp for the interpolated string:

```js
         assertHasMatchingError(warnings, {
           code: expectedMsgCode,
           message: new RegExp(`Unknown ${manifestKey} "fileSystem"`),
         });
```

This should make the test to pass again as expected, for both permissions and optional_permissions.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -192,24 +192,36 @@ describe('ManifestJSONParser', () => {
...
+        expect(addonLinter.collector.errors.length).toBe(0);
+        assertHasMatchingError(warnings, {
+          code: expectedMsgCode,
+          message: `/Unknown ${manifestKey} "fileSystem"/`,
+        });
       });
-    });
+    }
 
     it('should error if permission is not a string', () => {
```
rpl, line 226: This `should error if permission is not a string` looks like another test that we may want to also run for the optional_permissions manifest key (and reuse the same for look if there are multiple tests in the same describe group that we have to run on both permissions and optional_permissions).

Also the test that follows, `should error if permission is duplicated`, seems also something that we may want to run on the optional_permissions (and it may require some other changes to pass for the optional_permissions, because none of the implementation currently changes in this PR seems to be covering linting errors related to duplicated manifest keys and so I would expect the test to fail on the optional_permissions until we look into that and make sure that the same thing is enforced on optional_permissions as it is at the moment for the permissions manifest key).

rpl (changes requested):

tests/unit/parsers/test.manifestjson.js
```diff
@@ -192,24 +192,36 @@ describe('ManifestJSONParser', () => {
...
-      expect(addonLinter.collector.errors.length).toBe(0);
-      assertHasMatchingError(warnings, {
-        code: messages.MANIFEST_PERMISSIONS.code,
-        message: /Unknown permissions "fileSystem"/,
+    for (const manifestKey of ['permissions', 'optional_permissions']) {
+      const expectedMsgCode =
+        manifestKey === 'permissions'
+          ? messages.MANIFEST_PERMISSIONS.code
+          : messages.MANIFEST_OPTIONAL_PERMISSIONS.code;
+      const json =
```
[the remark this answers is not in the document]
aa-deet-eeya: Sorry, could you explain the 'moving the `const json= ...` back into the test case' a bit more?

If I'm not wrong we want the `applications: { gecko: { strict_min_version: '55.0' } },` thingy in both the cases where we have `permissions` and `optional_permissions` inside `const json= ...` , like 

`
const json =
        manifestKey === 'permissions'
          ? validManifestJSON({
              permissions: ['identity', 'fileSystem'],
              applications: { gecko: { strict_max_version: '55.0' } },
            })
          : validManifestJSON({
              optional_permissions: ['idle', 'fileSystem'],
              applications: { gecko: { strict_max_version: '55.0' } },
            });
`
or am I missing something?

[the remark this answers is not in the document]
rpl: > Sorry, could you explain the 'moving the `const json= ...` back into the test case' a bit more?
> 
> If I'm not wrong we want the `applications: { gecko: { strict_min_version: '55.0' } },` thingy in both the cases

yes, we don't need to condition the `applications: { gecko: { strict_min_version: '55.0' } },` in these cases
and so we can just define `const json` as:

```
      it(`should not error if ${manifestKey} is a string (even if unknown)`, () => {
        const json = validManifestJSON({
          [manifestKey]: ['history', 'fileSystem'],
          applications: { gecko: { strict_min_version: '55.0' } },
        });
        ...
```

tests/unit/parsers/test.manifestjson.js
aa-deet-eeya: `optional_permissions` doesn't have MANIFEST_BAD_PERMISSION error, should optional_permissions have a such type of error tho?
Or should it have a conditional statement for a warning?
Also currently there's no warning for duplicates in `optional_permissions` and for a "not a string value" it gives the following warning 
```
Code                            Message                   Description                                      File            Line   Column
MANIFEST_OPTIONAL_PERMISSIONS   /optional_permissions:    See https://mzl.la/2Qn0fWC (MDN Docs) for more   manifest.json                
                                Unknown                   information.                                                                  
                                optional_permissions                                                                                    
                                "[object Object]" at 2.  
```

rpl: > `optional_permissions` doesn't have MANIFEST_BAD_PERMISSION error, should optional_permissions have a such type of error tho?
> Or should it have a conditional statement for a warning?

We don't currently collect a `MANIFEST_BAD_PERMISSION` error for the optional_permissions because at the moment the related code in `src/parser/manifestjson.js` is only executed when `error.dataPath.startsWith('/permissions')` and so we are currently skipping that check for the `optional_permissions`:
https://github.com/mozilla/addons-linter/blob/5b4399ed3117bc7a8759b5ff5d24f1aa0a4cd1ca/src/parsers/manifestjson.js#L228-L234

Nevertheless, there shouldn't be any reason for that check to not be extended to the `optional_permissions` (eventually by also defining a new `MANIFEST_BAD_OPTIONAL_PERMISSION` so that we can have a different shorten url in the message description).

> Also currently there's no warning for duplicates in `optional_permissions` and for a "not a string value" it gives the following warning

I took a look into where the duplicated entries are detected for the `permissions` manifest property and I'm pretty sure that it is covered by the `uniqueItems` property here:

https://github.com/mozilla/addons-linter/blob/7530c81c27e7d8484b7d76381924489eaf07a5b3/src/schema/imported/manifest.json#L221-L241

You can double-check that it is actually the case by removing (or flipping the value) that `uniqueItems` property and run the test case that you think it should then fail.

To cover this for the optional permissions we should make sure that the entries for the `optional_permissions` in the same file is also going to contain that `uniqueItems` property.

This unfortunately is not enough, because we do update all those schema files in the `src/schema/imported` on every new beta release, and if we do only add the `uniqueItems` property to `src/schema/imported/manifest.json` then we will lose it once we import the schema for Firefox 81 beta.

To make sure that the `uniqueItems` property is going to be preserved over all the next schema imports, we do also need to apply the same kind of change we applied to `src/schema/updates/manifest.json` for the `permissions` property here:
https://github.com/mozilla/addons-linter/blob/6b26308027569cb674350ae30417cf74a878ff99/src/schema/updates/manifest.json#L40-L48

the schemas in `src/schema/updates/` are basically describing the changes to apply on the schema files as imported from Firefox, and so by applying the change to `src/schema/updates/manifest.json` we also ensure that the `uniqueItems` will be preserved when we will import new versions of the schema files from Firefox.

Let me know if some parts of this comment are not clear enough, otherwise give it a try and let me know how it goes (and feel free to update this PR even if there are still issues to figure out and solve, I'll be happy to take a look to the drafted changes, and travis logs if any test does still fail, and provide more guidance based on that)

src/schema/updates/manifest.json
```diff
@@ -45,6 +45,15 @@
                       ]
                     },
                     "uniqueItems": true
+                  } ,
+                  "optioanl_permissions": {
+                    "items": {
+                      "anyOf": [
+                        {},
+                        { "format": "deprecated" }
```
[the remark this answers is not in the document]
rpl: @aa-deet-eeya there is a typo here at line 49: "optioanl_permissions" -> ""optional_permissions"

I did some digging into the git history of this file to double-check what is the actual purpose of `anyOf` part, but unfortunately the pull request that originally added it doesn't provide any additional context about that.
To be honest I'm pretty sure that the `"format": "deprecated"` part of it isn't actually used (and removing that part from "permissions" doesn't actually trigger any test failure), and so at least at a first glance it seems that it is not that useful (even on "permissions" where it is applied from a long long time).

Anyway, I will keep it for now, for consistency between "permissions" and "optional_permissions" (and we will eventually evaluate if we could/should remove it in a separate issue).

I did run the schema import script locally to double-check that the change applied to src/schema/imports/manifest.json because of this additional annotation are all applied as expected (otherwise we would have some unrelated changes in the next schema import after merging this on master), and we would need to apply the following changes to src/schema/imported/manifest.json:

```diff
diff --git a/src/schema/imported/manifest.json b/src/schema/imported/manifest.json
index 8b274bcb..16356cc8 100644
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -241,7 +241,6 @@
                 },
                 "optional_permissions": {
                   "type": "array",
-                  "default": [],
                   "items": {
                     "allOf": [
                       {
@@ -250,8 +249,15 @@
                       {
                         "onError": "warn"
                       }
+                    ],
+                    "anyOf": [
+                      {},
+                      {
+                        "format": "deprecated"
+                      }
                     ]
                   },
+                  "default": [],
                   "uniqueItems": true
                 },
                 "web_accessible_resources": {
```

would you mind to also apply this changes?

In the meantime I'll give another review round to the rest of the changes in this pull request.

src/schema/imported/manifest.json
```diff
@@ -241,17 +241,23 @@
...
+                      }],
+                    "anyOf": [
+                      {},
+                      {
+                        "format": "deprecated"
                       }
                     ]
                   },
-                  "default": []
+                  "uniqueItems": true
```
rpl, line 260: @aa-deet-eeya This isn't yet the same as it will be once we will import the schema, the properties have to be exactly in the order as in the diff I pasted in my previous comment and the `allOf` array formatted as the `anyof` (currently it is not).

Here is a diff that would apply these changes on top of yours:
```diff
diff --git a/src/schema/imported/manifest.json b/src/schema/imported/manifest.json
index 8a0a24e5..16356cc8 100644
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -241,7 +241,6 @@
                 },
                 "optional_permissions": {
                   "type": "array",
-                  "default": [],
                   "items": {
                     "allOf": [
                       {
@@ -249,7 +248,8 @@
                       },
                       {
                         "onError": "warn"
-                      }],
+                      }
+                    ],
                     "anyOf": [
                       {},
                       {
@@ -257,6 +257,7 @@
                       }
                     ]
                   },
+                  "default": [],
                   "uniqueItems": true
                 },
                 "web_accessible_resources": {
```

(you could also just same this into a text file and then apply them using something like `patch -p1 < path/to/file.patch`).

aa-deet-eeya: I fixed this formatting issue :)
Also, could you explain why the position of `"default": [],` needs to be in the same exact position?

rpl: > I fixed this formatting issue :)
> Also, could you explain why the position of `"default": [],` needs to be in the same exact position?

@aa-deet-eeya Sure thing (and my apologies, my previous comments didn't describe that in detail):

as I did briefly mention in my previous comments about this, we do import the API schema files from Firefox beta builds in a semi-automated way and so to prevent the next import to include unrelated changes we have to make sure that the change applied to src/schema/imported/manifest.json is exactly as the one that our script would apply.

If you want to double-check that locally you can execute the following commands: 
```
$ ./scripts/download-import-tag FIREFOX_81_0b7_RELEASE
...
$ git diff src/schema/imported
...
```
(but take into account that it will download a zip file ~600mb big, and so it will take some time and network bandwidth).

The last version of this patch does apply the same kind of change that the script would apply based on what we have added to `src/schema/updates/manifest.json`, and so `git diff` should now show you only some small new changes to the `src/schema/imported/chrome_settings_overrides.json` schema file and no additional changes to `src/schema/imported/manifest.json` (as we want and we expected after your last fix on the change applied in this PR to that file).

rpl (approved):
lgtm \o/

@aa-deet-eeya Thanks a lot for contributing a fix for this issue!

aa-deet-eeya:
Thanks a lot @rpl :)

caitmuenster:
Thanks so much for your contribution, @aa-deet-eeya! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Welcome onboard! I look forward to seeing you around the project.

PR #3351, opened 2020-09-23
chore: Increased jest timeout when running test.linter.js on nodejs 10

rpl:
Travis nodejs 10 jobs is failing intermittently, but consistently on this test file.

The small changes in this PR are making that test file to detect when it is running on nodejs 10 ([using process.release property, see docs here](https://nodejs.org/api/process.html#process_process_release)) and increase the jest timeout.

### DIFF
--- a/tests/unit/test.linter.js
+++ b/tests/unit/test.linter.js
@@ -1,5 +1,6 @@
 /* eslint-disable max-classes-per-file */
 import fs from 'fs';
+import process from 'process';
 
 import { oneLine } from 'common-tags';
 
@@ -20,6 +21,13 @@ import {
   validStaticThemeManifestJSON,
   assertHasMatchingError,
 } from './helpers';
+
+// Increasing jest timeout when running on nodejs 10,
+// due to some intermittency when running this tests
+// on Travis.
+if (process.release.lts === 'Dubnium') {
+  jest.setTimeout(10000);
+}
 
 const fakeCheckFileExists = async () => {
   return {


tests/unit/test.linter.js
```diff
@@ -21,6 +22,13 @@ import {
   assertHasMatchingError,
 } from './helpers';
 
+// Increasing jest timeout when running on nodejs 10,
+// due to some intermittency when running this tests
+// on Travis.
+if (process.release.lts === 'Dubnium') {
```
pdehaan, line 28: Not sure if it's any easier, but you can also do Node version sniffing using `process.version`:

```sh
node -e "console.log(process.version)" # v14.11.0

nvm use 10 # Now using node v10.21.0 (npm v6.14.4)
node -e "console.log(process.version)" # v10.21.0
```

rpl: @pdehaan eh, I was wondering where they did put that :-)

willdurand (approved):

tests/unit/test.linter.js
pdehaan: Even worse answer if you want to check for specific Node versions using **semver** module:

```js
const semver = require("semver");

const version = process.version.replace(/^v/, "");

if (semver.lt(version, "11.0.0")) {
  console.log(`${version} is too old, increase test timeout.`);
} else {
  console.log(`Congrats, ${ version} is neat!`);
}
```

PR #3356, opened 2020-09-28
Update denied libraries

wagnerand:


rpl:
@wagnerand was there any sec vulnerability in DOMPurify 2.1.0 that has been fixed in 2.1.1?

[Based on the release changelog it seems that 2.1.1 is just removing code targeting old browser versions](https://github.com/cure53/DOMPurify/releases/tag/2.1.1), if that is the case do we still want to add DOMPurify 2.1.0 to the version explicitly disallowed?

wagnerand:
Yes, we recently updated our policies to only allow the latest version of DOMPurify: https://github.com/mozilla/addons-linter/blob/master/docs/third-party-libraries.md

rpl (approved):
lgtm

PR #3364, opened 2020-09-29
chore(travis): Split webext functional test in a separate travis job and deploy to npm from node v12 jobs

### LINKED ISSUE #3352: Update Travis CI config for Node 12 and 14
We should update our Travis CI config to:

- run the test suite with Node 14
- build the package with Node 12 as this is the version we are going to use on AMO

We can keep the Node 10 jobs for some more time as it is needed for `web-ext`.

### LINKED ISSUE #3359: Don't build twice during travis runs
We build the project twice, once during `npm run build` and once during `npm run webext-test-functional` (which calls `build` again).

We don't need to do that and should look into avoiding the redundant build.

Note from @rpl: _we could skip that part in the related script (maybe just skip it when running on travis, given that we do know the order of the steps executed and that the linter was already build)_
Note from @willdurand: _We should probably separate the different tasks into different jobs_

rpl:
This PRs contains some changes to the travis config file that should cover #3359 (and partially also #3352, we could add node 14 to cover it completely as part of this PR).

Fixes #3359 
Fixes #3352

TODO:
- [x] double-check that the expected stages are being executed in the travis job created for the PR (and that they are executing what we expect and completing successfully)
- [x] add nodejs 14 to the nodejs version to run the tests on (if we want to completely cover #3552 in this PR)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ language: node_js
 node_js:
   # When updating the value below, make sure to update the `node_js` version in
   # the `deploy` section at the bottom of this file too.
-  - 'v10.21.0'
-  - 'v12.18.1'
+  - '10'
+  - '12'
+  - '14'
 
 script:
   - node scripts/build-locales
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,7 +13,35 @@ script:
   # run integration tests using an addons-linter binary in a production-like environment
   - npm run test-integration:production
   - npm run prettier-ci
```
willdurand, line 15: I think `prettier-ci` could also run in its own job and maybe some other scripts, too. WDYT?

rpl: @willdurand I was thinking about splitting those too but I'm a bit on the fence about it: splitting the jobs would still introduce some additional time due to:
- some lag time on starting a sub-job (likely because the job is going to wait to get a worker node assigned)
- some additional time needed to clone the repo, install node using nvm and install all the dependencies with npm ci

and so opted to only split two sub-jobs (one for the web-ext functional tests and one for the deploy) and leave everything else in the main test sub-job.

what do you think? would it be worth to split prettier-ci in its own job from your perspective?

rpl:
@willdurand I gave a look to the logs in the sub-jobs from the travis job linked to this PR and they look good.
The only one that I can't actually fully double-check is the "npm deploy" sub-job because it doesn't really run much until it gets on master and on a tagged commit, but the predeploy steps should be the right ones (and I based it on the similar job we do use in web-ext and so it should be ok and working as expected :crossed_fingers: ).

Maybe you can give it a look too to that part of the change, a second pair of eye on it would be good (let me know if there is anything that looks suspicious or unclear).

rpl:
@willdurand on a separate topic: do we want to also add node v14 in the travis file as part of this PR?
at the moment the travis config is running:
- main test stesps on both node 10 and node 12
- web-ext functional test only on node 10
- deploy only on node 12

.travis.yml
willdurand: I like to have one job per tool because in most cases, the test suite will take longer than other jobs and jobs run in parallel. I am fine with the current approach, though, it's more a matter of personal preference (I find it easier to see what failed without necessarily reading the logs every time).

willdurand:
> @willdurand on a separate topic: do we want to also add node v14 in the travis file as part of this PR?
> at the moment the travis config is running:

Yeah, I suppose both test suites should be run against node 14 now

.travis.yml
```diff
@@ -4,6 +4,7 @@ node_js:
   # the `deploy` section at the bottom of this file too.
   - 'v10.21.0'
   - 'v12.18.1'
+  - 'v14.12.0'
```
willdurand, line 7: Should we put exact versions or major ones only? I think it'd be easier to only list:

```
- 10
- 12
- 14
```

(if that's possible)

willdurand (approved):
Looks good to me

.travis.yml
wagnerand: Renovate will update them automatically: https://github.com/mozilla/addons-linter/commit/8e9da4aa3bcd37b0a1b2810e8e3dcd9862b6663a

rpl: @willdurand @wagnerand Personally I'm ok with any of the two approaches (either just leave travis to use the last version available or leave renovate bot to manage the exact versions in our travis file), let's just agree on which one we prefer and I'll update the PR accordingly and then merge it.

willdurand: Let's not rely on Renovate for that.

PR #3390, opened 2020-10-08
fix: .eslintignore files should be ignored

### LINKED ISSUE #3389: Clarify relation to eslint, .eslintignore

- Create an extension which has a javascript build step using, for example, Typescript or Rollup.
- Build the extension, outputting files in the, for example, `dist/` directory.
- Add [`eslint`](https://eslint.org/) for source code linting.
- Create an [`.eslintignore` file](https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories) ignoring build artifacts.
- Run `npx addons-linter .`.

### What happened?

(Tableized CLI output.)

| Code | Message | Description | File | Line | Column |
| --- | --- | --- | --- | --- | --- |
File ignored because of a matching ignore pattern. Use "--no-ignore" to override. | File ignored because of a matching ignore pattern. Use "--no-ignore" to override. | dist/my-build-output.js |

Built javascript files are linted for potential addon-related problems. (No code style-related linting.)

```shell
npx addons-linter --version
```

```text
2.7.0
```

Also tested using an installed `addons-linter` (as opposed to using `npx`) and with `web-ext` v5.2.0.

- The relation to `--no-ignore`, which seems to come from the `eslint` CLI, is unclear. The flag does not seem to work with `addons-linter` nor `web-ext`.
- The `npm`-packaged versions of `addons-linter`/`web-ext` do not contain a `.eslintignore` file, so it should not be interfering.
- Presumably unrelated to the files (un)ignored by default by `addons-linter`/`web-ext`, since `dist/` is not one of them.
  - https://github.com/mozilla/addons-linter/blob/2.8.0/src/scanners/javascript.js#L97
  - https://github.com/mozilla/web-ext/blob/5.2.0/src/util/file-filter.js#L40-L47
- Using `--ignore-files` to ignore files when using `addons-linter`/`web-ext` seems more appropriate.
- Using `web-ext --source-dir` from a directory with `.eslintignore` seems to apply the ignores also to the intended source directory.

rpl:
addons-linter shouldn't load the .eslintignore file that may be available in the current working dir.

The strategy used in this PR to make sure that addons-linter's eslint-based javascript scanner doesn't load any .eslintignore file is to configure eslint to use a custom .eslintignore file included in the addons-linter npm package.

Other side notes about the other approaches evaluated:
- There is technically an `ignore` config property that prevents eslint from loading .eslintignore files, but setting `ignore` to `false` has the side effect of ignoring the `ignorePatterns` config (which we need to use to make sure that we don't ignore `node_modules/` and any dotted file)
- It would be nice to move the `ignorePatterns` config into the custom eslintignore file used in this pull request, but the test related to that behaviors are failing (I didn't really looked into why yet, it may be an issue specific to the test case, we may want to double-check that as a follow up) 

Fixes #3389


src/scanners/javascript.js
```diff
@@ -71,6 +71,9 @@ export default class JavaScriptScanner {
       // Avoid loading the addons-linter .eslintrc file
       useEslintrc: false,
 
+      // Avoid loading the .eslintignore file
+      ignore: false,
```
rpl, line 75: The option seems to be the right one, but based on the current test failures and a quick look to the eslint sources it seems that setting ignore to false does have the side effect of making eslint to ignore baseConfig.ignorePatterns and to only use the default ones  (which exclude dot files and node_modules, that are exactly the two patterns that we do want to include with a negated ignore pattern instead).

I need to take another look to see if there is any way to fix it on our side (or if the only option would be to open an upstream issue in the eslint repository).

willdurand (approved):

PR #3252, opened 2020-07-17
Remove trademark restriction on add-on names

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3251

PR #3392, opened 2020-10-09
fix: add JSON filename in results

### LINKED ISSUE #3282: JSON_INVALID lint error returns null file name instead of actual JSON file name
Via https://github.com/mozilla/web-ext/issues/1993

---

### Is this a feature request or a bug?
Bug?

### What is the current behavior?

I was trying to test the webExt API to lint a project and noticed that it seems to send the `webExt.cmd.lint` output directly to STDOUT instead of returning it to my code, so I piped the output to disk (res.json) and noticed that I get an `JSON_INVALID` error because the file is temporarily empty (and in my `sourceDir`, but the error gives me a `null` value for the `file` prop (and I feel like it should know the file path, and lists it in the `emptyFiles[]` array in the `metadata`):

```json
{
    "count": 2,
    "summary": {
        "errors": 2,
        "notices": 0,
        "warnings": 0
    },
    "metadata": {
        "totalScannedFileSize": 233371,
        "emptyFiles": [
            "res.json"
        ],
        "jsLibs": {},
        "unknownMinifiedFiles": []
    },
    "errors": [
        {
            "_type": "error",
            "code": "TYPE_NO_MANIFEST_JSON",
            "message": "manifest.json was not found",
            "description": "No manifest.json was found at the root of the extension. The package file must be a ZIP of the extension's files themselves, not of the containing directory. See: https://mzl.la/2r2McKv for more on packaging."
        },
        {
            "_type": "error",
            "code": "JSON_INVALID",
            "message": "Your JSON is not valid.",
            "description": "Unexpected end of JSON input",
            "file": null
        }
    ],
    "notices": [],
    "warnings": []
}
```

```js
// lint-test.js
const webExt = require("web-ext");

webExt.cmd.lint({
    sourceDir: ".",
    output: "json",
    verbose: false,
    pretty: true,
  });
```

### What is the expected or desired behavior?

If I get a `JSON_INVALID` error, it should list the correct filename instead of `null`.

### Version information (for bug reports)

* **Firefox version**:
* **Your OS and version**: macOS 10.15.6
* Paste the output of these commands:

```sh
> node --version && npm --version && web-ext --version

v14.7.0
6.14.7
5.0.0
```

regseb:
Fixes #3282

The commit https://github.com/mozilla/addons-linter/commit/087cac8e5979f17eb55979d2aa5ea6c046fce261 added the `addonMetadata` parameter to the `JSONParser` and `LocaleMessagesJSONParser` constructors. But it was not added to the [`JSONParser` constructor](https://github.com/mozilla/addons-linter/blob/master/src/scanners/json.js#L30) call.
The `addonMetadata` property is no longer in use, so I removed it from both constructors and the call to the `LocaleMessagesJSONParser` constructor.

### DIFF
--- a/src/scanners/json.js
+++ b/src/scanners/json.js
@@ -27,9 +27,12 @@ export default class JSONScanner extends BaseScanner {
       );
       localeMessagesJSONParser.parse();
     } else {
-      const jsonParser = new JSONParser(json, this.options.collector, {
-        filename: this.filename,
-      });
+      const jsonParser = new JSONParser(
+        json,
+        this.options.collector,
+        this.options.addonMetadata,
+        { filename: this.filename }
+      );
       jsonParser.parse();
     }
 
--- a/tests/unit/scanners/test.json.js
+++ b/tests/unit/scanners/test.json.js
@@ -20,6 +20,21 @@ describe('JSONScanner', () => {
     await expect(jsonScanner.scan()).rejects.toThrow('Explode!');
   });
 
+  it('should report invalid JSON', async () => {
+    const addonsLinter = new Linter({ _: ['foo'] });
+    const jsonScanner = new JSONScanner('{ bar: undefined }', 'baz.json', {
+      collector: addonsLinter.collector,
+    });
+
+    await jsonScanner.scan();
+
+    const { errors } = addonsLinter.collector;
+    expect(errors.length).toEqual(1);
+    expect(errors[0].code).toEqual(messages.JSON_INVALID.code);
+    expect(errors[0].file).toEqual('baz.json');
+    expect(errors[0].message).toEqual(messages.JSON_INVALID.message);
+  });
+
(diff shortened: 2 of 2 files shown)


wagnerand:
I can't find where exactly, but after trademark checks got reverted in https://github.com/mozilla/addons-linter/pull/3252, we decided to keep the `addonMetadata` around. @rpl do you recall that?

willdurand:
> I can't find where exactly, but after trademark checks got reverted in #3252, we decided to keep the `addonMetadata` around. @rpl do you recall that?

Yes, I think we should keep it for now.

rpl:
@wagnerand personally I don't recall (but I haven't been involved in neither #3252 nor #3123).

Given that this is a regression, independently from how we will agree on fixing the issue, we should also add an additional test case along with the fix (clearly it wasn't covered enough by the existing test cases).

willdurand (approved):

PR #1689, opened 2017-12-05
Switch dependency to probe-image-size

asamuzaK:
Fixes #1668, #1669

'sharp' requires Python executable when installing, but Python is not installed by default on Windows, so we should replace 'sharp' to something else.

Fixed SVG icon size mismatch warning too.

Thanks for opening a Pull Request (PR), here's a few guidelines as to what we need in your PR before we review it.

Please delete anything that isn't relevant to your patch.

* [✓] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [✓] Add `Fixes #ISSUENUM` at the top of your PR.
* [✓] Add a description of the the changes introduced in this PR.
* [✓] The change has been successfully run locally.
* [✓] Add tests to cover the changes added in this PR.

Once you have met the above requirements please replace this section with
a `Fixes #ISSUENUM` linking to the issue fixed by this PR along with an
explanation of the changes. Thanks for your contribution!

Issue #1669, opened 2017-11-24
Don't warn about icon size mismatch if icon is SVG


asamuzaK:
### Describe the problem and steps to reproduce it:
(Please include as many details as possible.)
When I uploaded add-on to AMO, it is warned about icon size mismatch.

```
The size of the icon does not match the manifest.

警告: Expected icon at "icon.svg" to be 32 pixels wide but was 128.
```

My manifest.json is like this.
```
  "icons": {
    "32": "icon.svg#gray",
    "64": "icon.svg"
  },
```

It is an SVG file, so size should not matter.

### What happened?
AMO raised warning.

### What did you expect to happen?
No warning.

### Anything else we should know?
(Please include screenshots and any relevant files.)

In [icons - Mozilla | MDN](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/icons "icons - Mozilla | MDN"), it is explained as:

> Even though you can use one file, you still need to specify various size of the icon in your manifest. E.g.:
```
"icons": {
  "48": "icon.svg",
  "96": "icon.svg"
}
```

erosman:
Was this one filled and pushed? 
The error is still present.

asamuzaK:
In addons-linter, yes.
I don't know it is pushed to AMO server.

asamuzaK:
Latest release seems to be [Release 2017.12.07 · mozilla/addons-server](https://github.com/mozilla/addons-server/releases/tag/2017.12.07 "Release 2017.12.07 · mozilla/addons-server") but linter is merged on Dec 08, 2017? [Comparing 2017.12.07...master · mozilla/addons-server](https://github.com/mozilla/addons-server/compare/2017.12.07...master "Comparing 2017.12.07...master · mozilla/addons-server")

asamuzaK:
Confirmed fixed on AMO server too.

AlexandraMoga:
Tried validating an extension which included `.svg` icons with wrong sizes defined in the manifest and no errors/warnings  were thrown. 
Marking this as verified fixed.

erosman:
There are still warnings ....

`https://reviewers.addons.mozilla.org/en-US/firefox/files/browse/817497/`

> Warning: The size of the icon does not match the manifest.
> Expected icon at "images/rikaichamp-blue.svg" to be 48 pixels wide but was 200.
> 
> Warning: The size of the icon does not match the manifest.
[...]

PR #3406, opened 2020-10-17
fix: manifest validator should add a warning instead of an error on non-square svg icons

### LINKED ISSUE #3322: "Icons must be square." for svg icons

I'm trying to upload a new version of [cf-pop](https://addons.mozilla.org/firefox/addon/cf-pop/) to amo.

### What happened?

It says `Icons must be square.` and `Icon at "icons/Cloudflare-gray.svg" must be square.` twice.

It should accept svg icons (like before).

The icon file didn't change. It's exactly the same icon file in the published version.

Yrkan:
Fixes #3322

it's now a warning instead of an error.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -510,8 +510,12 @@ export default class ManifestJSONParser extends JSONParser {
     try {
       const info = await getImageMetadata(this.io, iconPath);
       if (info.width !== info.height) {
-        this.collector.addWarning(messages.iconIsNotSquare(iconPath));
-        this.isValid = false;
+        if (info.mime !== 'image/svg+xml') {
+          this.collector.addError(messages.iconIsNotSquare(iconPath));
+          this.isValid = false;
+        } else {
+          this.collector.addWarning(messages.iconIsNotSquare(iconPath));
+        }
       } else if (
         expectedSize !== null &&
         info.mime !== 'image/svg+xml' &&
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -510,7 +510,7 @@ export default class ManifestJSONParser extends JSONParser {
     try {
       const info = await getImageMetadata(this.io, iconPath);
       if (info.width !== info.height) {
-        this.collector.addError(messages.iconIsNotSquare(iconPath));
+        this.collector.addWarning(messages.iconIsNotSquare(iconPath));
         this.isValid = false;
       } else if (
         expectedSize !== null &&
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -1730,7 +1730,7 @@ describe('ManifestJSONParser', () => {
         expect(warnings.length).toEqual(0);
       });
 
-      it('adds an error if the dimensions of the image are not the same', async () => {
+      it('adds a warning if the dimensions of the image are not the same', async () => {
         const addonLinter = new Linter({ _: ['bar'] });
         const icon32 = 'tests/fixtures/rectangle.png';
         const size32 = 32;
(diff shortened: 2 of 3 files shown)


wagnerand:
@ThamiMemel Can you fix the failing tests?

ThamiMemel:
Yes, sure i'll change it to adds a warning if the dimensions of the image are not the same instead of an error, please correct me if i'm wrong here.

rpl:
@wagnerand @ThemiMemel the way I read the issue described in #3322 is that the addons-linter should only warn that the icon is not square if the image is an svg, but still mark it as an error if it is not an svg (e.g. if it is a png image).

Basically this seems to be an addition to the fix applied for #1669 (which was about not warning about the size not matching the convention that Firefox expects for the icon image size), which was fixed in #1689 by adding a `info.mime !== 'image/svg+xml'` check to the validation that follows this one.

And so in my opinion we should check if the image is an svg or not (using a check similar to that `info.mime !== 'image/svg+xml'`) and:
- add it as a warning if the image is an svg
- still add it as error if the image is not an svg

Then keep the existing test unmodified and add a new test case that verifies that if the size mismatch is triggered by an svg icon image the validation failure is collected as a warning and that as a result the manifest is considered valid and the linting passes.

ThamiMemel:
@rpl Oh my bad, looks like the changes to test and update the validation to work as intended.
Thank you!  i'll send a PR with the changes.

rpl:
> @rpl Oh my bad, looks like the changes to test and update the validation to work as intended.

No worries at all, on the contrary my sincere apologies for missing to notice this PR before, I did notice it today when Andreas did comment on it.

> Thank you! i'll send a PR with the changes.

Update this PR without creating a new pull request.

If you have never done that, there are multiple way to achieve that, e.g. you may choose one of the following strategies (the first is likely the easiest one if you have never used `git rebase -i ...` or `git push ... --force`):

- add more commit to the same branch, where you revert the parts that didn't need to change and apply further changes to the one that needs to be adapted, and then just push the same branch in your github fork
- do an interactive git rebase on the branch  and drop the commit that you don't need, then add new commit with further changes, and then push with --force the same branch in your github fork (--force it is needed because there are not only additional commits and the branch has to be changed to point to the new set of commits in your local branch)
- remove the local branch, and then create a new branch with the same name, make the change that you need and then push with --force in your github fork (--force is is needed for the same reason described above)

ThamiMemel:
@rpl  yeah i was autopiloting when i said i'll send a PR with the changes i meant i'll commit to this PR.

P.S: the commit is already here, please let me know if anything need to be changed.
Thank you again.

src/parsers/manifestjson.js
```diff
@@ -510,8 +510,12 @@ export default class ManifestJSONParser extends JSONParser {
...
       const info = await getImageMetadata(this.io, iconPath);
       if (info.width !== info.height) {
-        this.collector.addError(messages.iconIsNotSquare(iconPath));
-        this.isValid = false;
+        if (info.mime !== 'image/svg+xml') {
+          this.collector.addError(messages.iconIsNotSquare(iconPath));
+          this.isValid = false;
+        } else {
+          this.collector.addWarning(messages.iconIsNotSquare(iconPath));
+        }
```
rpl, line 518: :+1: this change on the implementation side looks the one I was thinking of.

This PR should only be missing a new test case now, would you mind to define a new test case to cover the warning on non square svg icons?

you could start by looking to the test case that you were changing in the previous version of this PR, and based on that create a new test case to cover the new expected behavior.

rpl (changes requested):

src/parsers/manifestjson.js
ThamiMemel: Sure, i have already made one, i'll commit it right now, please correct me if anything is wrong. Thank you.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -1730,6 +1730,31 @@ describe('ManifestJSONParser', () => {
         expect(warnings.length).toEqual(0);
       });
 
+      it('adds a warning if the icon is SVG but the dimensions of the image are not the same', async () => {
```
rpl, line 1733: Nitpick: let's make this test description a bit shorter e.g. `it('does add a warning if the icon is a non square svg', ...)`

rpl (approved):
@ThamiMemel thanks! 

This looks good to me and I'm marking this PR as accepted, it would be nice to shorten the description of the new test as described below as a small nitpick to cover before landing this.

tests/unit/parsers/test.manifestjson.js
ThamiMemel: done.
Thank for the feedback

rpl (approved):

caitmuenster:
Thanks so much for the patch, @Yrkan! 🙌 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Welcome onboard -- we look forward to seeing you around!

Issue #2253, opened 2018-10-23
"homepage_url"s pointing to AMO websites should be restricted during validation


AlexandraMoga:
Follow up for https://github.com/mozilla/addons-server/issues/9012

STR:
1. Log in to AMO dev
2. Create a new extension and add a `"homepage_url"` property pointing to addons-dev.allizom.org
3. Submit your add-on

Actual result:
Add-on passes validation without any errors

Expected result:
Homepage URLs pointing to AMO websites should trigger an error during validation to comply with the current implementation in Dev Hub

Notes:
- reproduced on AMO dev with FF62, Win10x64

![image](https://user-images.githubusercontent.com/31961530/47355321-be0b9200-d6c9-11e8-8c5a-f5414f671318.png)

rbrishabh:
Hey! I would like to tackle this!

ThamiMemel:
i would like to work on this issue

ThamiMemel:
Here's a first draft, most likely subject to multiples changes.
Here are some point probably need to be edited . hopefully someone with more insight into the project can guide me.

- **Restricted domain should be inside some sort of config file ?** 

- **Error code and description could be better ?**

- **Should i make a test case for this ?**

rpl:
> Here's a first draft, most likely subject to multiples changes.
> Here are some point probably need to be edited . hopefully someone with more insight into the project can guide me.
> 
>     * **Restricted domain should be inside some sort of config file ?** 
>     * **Cureently the checking is a simple indexOf maybe a regex is better ?**

It depends from what degree of flexibility this validation rule should have.
e.g. addons-dev.allizom.org is a non production instance of the demo portal, but based on the original issue linked to this one it seems that addons.mozilla.org is the main domain that we were meant to disallow.
But, are those the only urls that we want to disallow? (e.g. should we also disallow localhost?).

@Yrkan this issue has been stale for a while, let's double-check that is still valid and what kind of behavior we did want the implement on the linter side.

> 
>     * **Error code and description could be better ?**

New error codes does also need to be added to docs/rules.md, and there is a test that does cover that and it is currently failing in the draft patch in #3415 (See https://travis-ci.org/github/mozilla/addons-linter/jobs/738802933#L478)


>     * **Should i make a test case for this ?**

yes, any rule should be covered by explicit test cases, to make sure that the expected behaviors are implemented and that they do not regress.

But let's double-check first that this issue is still valid, and (if it is) what are the behaviors that the new rules should implement.

[...]

PR #3415, opened 2020-10-25
fix: Forbid 'homepage_url' pointing to AMO Website #2253

Yrkan:
Related issue : #2253

Here's a first draft, most likely subject to multiples changes.
Here are some point probably need to be edited . hopefully someone with more insight into the project can guide me.

- **Restricted domain should be inside some sort of config file ?** 

- **Cureently the checking is a simple indexOf maybe a regex is better ?**

- **Error code and description could be better ?**

- **Should i make a test case for this ?**


src/messages/manifestjson.js
```diff
@@ -555,3 +555,12 @@ export function permissionFirefoxAndroidUnsupportedByMinVersion(
     file: MANIFEST_JSON,
   };
 }
+
+export const RESTRICTED_HOMEPAGE_URL = {
+  code: 'RESTRICTED_HOMEPAGE_URL',
```
rpl, line 560: The new rule code has to also added to the markdown table in docs/rules.md (like you did also for the new rule you created in #3425).

src/messages/manifestjson.js
```diff
@@ -555,3 +555,12 @@ export function permissionFirefoxAndroidUnsupportedByMinVersion(
...
+  message: i18n._('Linking to AMO is not allowed'),
+  description: i18n._(
+    'Links directing to AMO domain name are not allowed to be used for homepage'
+  ),
```
rpl, line 564: AMO is how we refer to "addons.mozilla.org" internally, but I doubt that it would be immediately understandable from the extension developer that do not already know that, I would replace AMO with the "addons.mozilla.org" (even if technically we would also reject on "addons-dev.allizom.org", but I doubt that will be really a common mistake, it is just a dev instance we use before deploy new AMO versions to the official one).

src/parsers/manifestjson.js
```diff
@@ -797,6 +801,13 @@ export default class ManifestJSONParser extends JSONParser {
     }
   }
 
+  validateHomePageURL(url) {
+    const restrictedDomain = 'addons-dev.allizom.org';
```
rpl, line 805: Lets create an array or set of urls exported by `src/const.js`, `addons-dev.allizom.org` is the url reported by QA because they verify the fixes on the dev server, but we should also (and mainly) reject homepage_url set to 'addons.mozilla.org'.

src/parsers/manifestjson.js
```diff
@@ -797,6 +801,13 @@ export default class ManifestJSONParser extends JSONParser {
...
+    if (url.indexOf(restrictedDomain) !== -1) {
+      this.collector.addError(messages.RESTRICTED_HOMEPAGE_URL);
+    }
```
rpl, line 808: I think that we should also mark the manifest as invalid (by setting this.isValid to false as we do in a few other places in `src/parsers/manifestjson.js`).

rpl (changes requested):
@Yrkan my apologies for the time I needed to actually get to this pull request.

Follows a couple of initial review comments on the current version of this patch, besides what described below this PR does also need new tests to cover the new rule (let me know if you need any additional pointers for that, feel free to push an incomplete test case even if it isn't complete yet or if it is still fails).

docs/rules.md
```diff
@@ -126,6 +126,7 @@ Rules are sorted by severity.
 | `KEY_FIREFOX_ANDROID_UNSUPPORTED_BY_MIN_VERSION`        | warning  | Manifest key not compatible with Firefox for Android at `applications.gecko.strict_min_version` |
 | `PERMISSION_FIREFOX_UNSUPPORTED_BY_MIN_VERSION`         | notice   | Permission not compatible with `applications.gecko.strict_min_version`                          |
 | `PERMISSION_FIREFOX_ANDROID_UNSUPPORTED_BY_MIN_VERSION` | notice   | Permission not compatible with Firefox for Android at `applications.gecko.strict_min_version`   |
+| `RESTRICTED_HOMEPAGE_URL`                               | error    | Linking to AMO domain name in `homepage_url` is not allowed                                     |
```
rpl, line 129: Nit, let's replace AMO with addons.mozilla.org as in the message description.

rpl: There is a conflict with the changes in this file, I guess your git branch has been branched from an older version of the main addons-linter branch, and so this PR needs to resolve that conflict (by either rebase your branch, or merging the last version of the main branch into yours).

Let me know if you needs help with that, I'll be happy to provide some additional guidance about how to do it.

rpl (changes requested):
follows a couple of additional review comments (in addition to the ones in the previous round that are not handled yet).

Feel free to ping me if you get stuck and need help to figure out some unexpected issues you may be facing.

docs/rules.md
Yrkan: @rpl  Sure i'll try to fix the conflict, i made the other changes.
I'll let you if i need help with fixing the conflict.
Thanks.

Yrkan: @rpl Done. please let me know if i missed anything.

docs/rules.md
```diff
@@ -126,6 +126,7 @@ Rules are sorted by severity.
 | `KEY_FIREFOX_ANDROID_UNSUPPORTED_BY_MIN_VERSION`        | warning  | Manifest key not compatible with Firefox for Android at `applications.gecko.strict_min_version` |
 | `PERMISSION_FIREFOX_UNSUPPORTED_BY_MIN_VERSION`         | notice   | Permission not compatible with `applications.gecko.strict_min_version`                          |
 | `PERMISSION_FIREFOX_ANDROID_UNSUPPORTED_BY_MIN_VERSION` | notice   | Permission not compatible with Firefox for Android at `applications.gecko.strict_min_version`   |
+| `RESTRICTED_HOMEPAGE_URL`                               | error    | Linking to addons.mozilla.org in `homepage_url` is not allowed                                  |
```
rpl, line 129: Nit, it doesn't look that we are keeping this list in alphabetic order, but let's move this to the end of the table and to keep the last ones in what I think is the addition order.

package-lock.json
```diff
@@ -6197,7 +6197,8 @@
     "fsevents": {
       "version": "2.2.1",
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.2.1.tgz",
-      "integrity": "sha512-bTLYHSeC0UH/EFXS9KqWnXuOl/wHK5Z/d+ghd5AsFMYN7wIGkUCOJyzy88+wJKkZPGON8u4Z9f6U4FdgURE9qA=="
+      "integrity": "sha512-bTLYHSeC0UH/EFXS9KqWnXuOl/wHK5Z/d+ghd5AsFMYN7wIGkUCOJyzy88+wJKkZPGON8u4Z9f6U4FdgURE9qA==",
+      "optional": true
```
rpl, line 6201: Please remove the change to the package-lock.json file from this PR (in general you should be able to avoid having unrelated changes to the package-lock.json file by using `npm ci` if the PR doesn't require changes to the dependencies listed in the package.json file).

src/parsers/manifestjson.js
```diff
@@ -806,6 +811,15 @@ export default class ManifestJSONParser extends JSONParser {
     }
   }
 
+  validateHomePageURL(url) {
+    RESTRICTED_HOMEPAGE_URLS.forEach((domain) => {
+      if (url.indexOf(domain) !== -1) {
+        this.collector.addError(messages.RESTRICTED_HOMEPAGE_URL);
+      }
+      this.isValid = false;
+    });
```
rpl, line 820: I would slightly prefer `for (let restrictedUrl of RESTRICTED_HOMEPAGE_URLS) {...}` and return earlier if we did already found that we are adding the error.

Also this.isValid should be set to false inside the if block, because we should only mark the manifest as invalid if the homepage url was actually a forbidden one. 

The unit tests should also catch this, and so let's make sure that we do have coverage to catch that mistake (a test case to verify that a non forbidden homepage url should not be marked as invalid).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -172,6 +172,21 @@ describe('ManifestJSONParser', () => {
...
+      });
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector
+      );
+      expect(manifestJSONParser.isValid).toEqual(false);
+      const { errors } = addonLinter.collector;
+      expect(errors.length).toEqual(1);
+      expect(errors[0].code).toEqual('RESTRICTED_HOMEPAGE_URL');
+    });
```
rpl, line 189: Let's also test the opposite as I did mention in another comment (if the homepage url is not forbidden, then there shouldn't be an error and the manifest should be valid).

rpl (changes requested):
@Yrkan Thanks for the updates! Follows a new round of review comments.

Feel free to ping me if any of the following comments isn't clear (or if you have doubts that I can help you with).

Yrkan:
@rpl Thankf for the review. All changes requested have been made.
I wonder if we have to group the new two test cases in a describe block ?

tests/unit/parsers/test.manifestjson.js
```diff
@@ -96,6 +96,33 @@ describe('ManifestJSONParser', () => {
     expect(warnings[0].code).toEqual('IGNORED_APPLICATIONS_PROPERTY');
   });
 
+  it('should fail if "homepage_url" links to AMO domain names', () => {
```
rpl, line 99: > I wonder if we have to group the new two test cases in a describe block ?

:+1: Sure, let's wrap these two into a `describe('homepage_url', ...);`

rpl, line 99: Nit, let's not quote homepage_url here, for consistency with the one that follows.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -96,6 +96,33 @@ describe('ManifestJSONParser', () => {
...
+  it('should not mark non forbidden homepage url as invalid', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const json = validManifestJSON({
+      homepage_url: 'http://test.org',
+    });
+    const manifestJSONParser = new ManifestJSONParser(
+      json,
+      addonLinter.collector
+    );
+    expect(manifestJSONParser.isValid).toEqual(true);
```
rpl, line 123: Let's also assert that errors.length is 0

src/parsers/manifestjson.js
```diff
@@ -806,6 +811,16 @@ export default class ManifestJSONParser extends JSONParser {
     }
   }
 
+  validateHomePageURL(url) {
+    for (const restrictedUrl of RESTRICTED_HOMEPAGE_URLS) {
+      if (url.indexOf(restrictedUrl) !== -1) {
```
rpl, line 816: `if (url.includes(restricteUrl)) {...}`

tests/unit/parsers/test.manifestjson.js
```diff
@@ -96,6 +96,33 @@ describe('ManifestJSONParser', () => {
     expect(warnings[0].code).toEqual('IGNORED_APPLICATIONS_PROPERTY');
   });
 
+  it('should fail if "homepage_url" links to AMO domain names', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const json = validManifestJSON({
+      homepage_url: 'http://addons.mozilla.org',
```
rpl, line 102: we may want to use an url with a path here to also verify as a side-effect that we are not just checking for strictly equality.

To be totally fair, we could provide even better coverage if we do turn these two test cases into a parametrized function that we will use to test more scenarios, something like:

```js
describe('homepage_url', () => {
  function testHomepageUrl(homepage_url, expectValid) {
    ...
    const json = validManifestJSON({ homepage_url });
    ...
    expect(manifestJSONParser.isValid).toEqual(expectValid);
    if (expectValid) {
      // expect errors.length is 0 
    } else {
      // expect errors.length is 1 and the error code the expected one
    }
  }

  const testInvalidUrls = [
    'https://addons.mozilla.org',
    'http://addons.mozilla.org/somepath',
  ];
  for (let invalidUrl of testInvalidUrls) {
    it(`should fail on forbidden homepage_url "${invalidUrl}"`, () => {
      return testHomepageUrl(invalidUrl, false);
    });
  }

  it('should not mark as invalid on homepage_url using a restricted domain', () => {
    return testHomepageUrl("http://test.org", true);
  });
```

rpl (commented):
@Yrkan looks pretty good, follows some additional review comments for a couple more tweaks

tests/unit/parsers/test.manifestjson.js
Yrkan: @rpl Done, sorry for keeping forgetting and using AMO :laughing:.
Let me know about any further changes needed.

rpl (approved):

rpl:
@Yrkan Thanks! This looks great, I just approved it! :tada:

PR #3425, opened 2020-10-27
fix: Add a warning when both applications and browser_specific_settings are being used in the same manifest.json

### LINKED ISSUE #2294: Add a lint warning when both applications.gecko and browser_specific_settings.gecko are being used in the same manifest.json file
As part of #2285, the addons-linter is going to use the `"browser_specific_settings"` manifest property as an alias of the `"applications"` manifest property when `"browser_specific_settings"` contains a `"gecko"` property.

As a follow up of this new behavior, the addons-linter should log a linting warning when both `"browser_specific_settings"` and `"applications"` have a `"gecko"` property in the same manifest.json file
(see https://github.com/mozilla/addons-linter/pull/2285#discussion_r235926818 for a rationale)

Yrkan:
Fixes #2294 
**This PR will be changed with more commits based on the final decision in #2294**
Currently it gives a warning if both `"browser_specific_settings"` and `applications` have a gecko property

- [x] Add tests to cover the changes added in this PR.

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -455,11 +455,13 @@ export const WRONG_ICON_EXTENSION = {
   file: MANIFEST_JSON,
 };
 
-export const DUPLICATE_GECKO_PROPERTY = {
-  code: 'DUPLICATE_GECKO_PROPERTY',
-  message: i18n._('Multiple use of "Gecko" property'),
+export const IGNORED_APPLICATIONS_PROPERTY = {
+  code: 'IGNORED_APPLICATIONS_PROPERTY',
+  message: i18n._(
+    'Usage of both "browser_specific_settings" and "applications"'
+  ),
   description: i18n._(
-    oneLine`the property Gecko is being used in "applications" and "browser_specific_settings" keys at the same time`
+    oneLine`The applications property is being ignored because it is superseded by the browser_specific_settings property which is also defined in your manifest. Consider removing applications.`
   ),
   file: MANIFEST_JSON,
 };
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -458,10 +458,10 @@ export const WRONG_ICON_EXTENSION = {
 export const IGNORED_APPLICATIONS_PROPERTY = {
   code: 'IGNORED_APPLICATIONS_PROPERTY',
   message: i18n._(
-    'Usage of both "browser_specific_settings" and "applications"'
+    '"applications" property overridden by "browser_specific_settings" property"'
   ),
   description: i18n._(
-    oneLine`The applications property is being ignored because it is superseded by the browser_specific_settings property which is also defined in your manifest. Consider removing applications.`
+    oneLine`The "applications" property is being ignored because it is superseded by the "browser_specific_settings" property which is also defined in your manifest. Consider removing applications.`
   ),
   file: MANIFEST_JSON,
 };
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -455,6 +455,15 @@ export const WRONG_ICON_EXTENSION = {
   file: MANIFEST_JSON,
 };
 
+export const DUPLICATE_GECKO_PROPERTY = {
+  code: 'DUPLICATE_GECKO_PROPERTY',
+  message: i18n._('Multiple use of "Gecko" property'),
+  description: i18n._(
+    oneLine`the property Gecko is being used in "applications" and "browser_specific_settings" keys at the same time`
+  ),
+  file: MANIFEST_JSON,
+};
+
 export const NO_MESSAGES_FILE_IN_LOCALES = 'NO_MESSAGES_FILE_IN_LOCALES';
 export function noMessagesFileInLocales(path) {
   return {
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -305,6 +305,15 @@ export default class ManifestJSONParser extends JSONParser {
 
     if (
       this.parsedJSON.browser_specific_settings &&
+      this.parsedJSON.applications &&
+      this.parsedJSON.browser_specific_settings.gecko &&
+      this.parsedJSON.applications.gecko
+    ) {
+      this.collector.addWarning(messages.DUPLICATE_GECKO_PROPERTY);
+    }
+
+    if (
+      this.parsedJSON.browser_specific_settings &&
       this.parsedJSON.browser_specific_settings.gecko
     ) {
       this.parsedJSON.applications = this.parsedJSON.browser_specific_settings;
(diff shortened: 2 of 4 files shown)


src/parsers/manifestjson.js
```diff
@@ -303,6 +303,15 @@ export default class ManifestJSONParser extends JSONParser {
       });
     }
 
+    if (
+      this.parsedJSON.browser_specific_settings &&
+      this.parsedJSON.applications &&
+      this.parsedJSON.browser_specific_settings.gecko &&
+      this.parsedJSON.applications.gecko
+    ) {
+      this.collector.addWarning(messages.DUPLICATE_GECKO_PROPERTY);
```
rpl, line 312: Let's add the warning if both browser_specific_settings and applications are being used without specifically looking if they also include the gecko property (as we discussed and agreed on the issue).

I agree with you that we may try to find a better code for this rule, how about the following ones? 
- `MULTIPLE_BROWSER_SPECIFIC_SETTINGS`
- `DUPLICATED_BROWSER_SPECIFIC_SETTINGS`

@wagnerand do you have some other proposed rule code for this new rule that we may like more than these ones?

rpl (changes requested):
Thanks @yrkan, this looks pretty good. Let's make the rule to match what we discussed and agreed on in the issue (which should require just a few tweaks) and tweak the rule code and message accordingly (I'm proposing a couple of alternative names, let's see if Andreas may also have in mind some other rule codes that we may like more than those, otherwise let's just pick one listed in the inline comment below).

src/messages/manifestjson.js
```diff
@@ -455,6 +455,15 @@ export const WRONG_ICON_EXTENSION = {
   file: MANIFEST_JSON,
 };
 
+export const DUPLICATE_GECKO_PROPERTY = {
+  code: 'DUPLICATE_GECKO_PROPERTY',
+  message: i18n._('Multiple use of "Gecko" property'),
+  description: i18n._(
+    oneLine`the property Gecko is being used in "applications" and "browser_specific_settings" keys at the same time`
```
rpl, line 462: @wagnerand do you think that it may be worth as part of this new message to also suggest to the developer to remove `"applications"` and prefer to use just `"browser_specific_settings"`?

src/parsers/manifestjson.js
wagnerand: @rpl not a strong preference but I am always a fan of being explicit when possible and appropriate. Maybe something along the lines of `BROWSER_SPECIFIC_SETTINGS_AND_APPLICATIONS_DEFINED`. That's a bit long of course, we could cut it down to `BROWSER_SPECIFIC_SETTINGS_APPLICATIONS` or similar.

src/messages/manifestjson.js
wagnerand: @rpl Yes, maybe something like:

_The `applications` key is being ignored because it is superseded by the `browser_specific_settings` key that is also defined in the manifest. Consider removing `applications`_

src/parsers/manifestjson.js
rpl: @wagnerand how about `REDUNDANT_APPLICATIONS_PROPERTY` or `OVERRIDDEN_APPLICATIONS_PROPERTY`?
in the end, also based on the message associated to the rule as suggested in the other comments below, the warning message is going to suggest the user to remove the redundant `"applications"` property and so the rule code would nicely match that. 
(we do not have a limit to the code anymore, we removed it recently, but if we can find a shorter one that we like then I would prefer it).

src/messages/manifestjson.js
rpl: @Yrkan let's use the message suggested by Andreas in https://github.com/mozilla/addons-linter/pull/3425#discussion_r513417669 in the next updates to this PR.

src/parsers/manifestjson.js
wagnerand: To add to the bike-shedding, we could use `IGNORED_APPLICATIONS_PROPERTY`. I know we're not exposing that code to the user, but I like its explicitness, which could help other code contributors. Again, I don't have a super preference.

src/messages/manifestjson.js
wagnerand: And @rpl reminded me that `property` is the proper word for `key`, so it would be:

_The `applications` property is being ignored because it is superseded by the `browser_specific_settings` property which is also defined in your manifest. Consider removing `applications`_.

src/parsers/manifestjson.js
rpl: Agree, even if extension developers will not really care about the code we do choose, I still care that we choose one that is short but also clear and explicit enough.

Personally I like your last proposed code, `IGNORED_APPLICATIONS_PROPERTY`, and I think we can proceed to use that.

@Yrkan, it looks that we do have a "winner" for the "rule code" contest ;-)
let's use `IGNORED_APPLICATIONS_PROPERTY`.

Yrkan: @rpl Great `IGNORED_APPLICATIONS_PROPERTY` it is then. i'll make the modification an request another review

src/messages/manifestjson.js
```diff
@@ -455,6 +455,17 @@ export const WRONG_ICON_EXTENSION = {
   file: MANIFEST_JSON,
 };
 
+export const IGNORED_APPLICATIONS_PROPERTY = {
+  code: 'IGNORED_APPLICATIONS_PROPERTY',
+  message: i18n._(
+    'Usage of both "browser_specific_settings" and "applications"'
```
rpl, line 461: Nitpick: it may be reasonable to use something like `"applications" property overridden by "browser_specific_settings" property` as the short message, because it should enough but also more directly describe the issue being pointed out by the warning.

src/messages/manifestjson.js
```diff
@@ -455,6 +455,17 @@ export const WRONG_ICON_EXTENSION = {
...
+  ),
+  description: i18n._(
+    oneLine`The applications property is being ignored because it is superseded by the browser_specific_settings property which is also defined in your manifest. Consider removing applications.`
```
rpl, line 464: Nitpick: would you mind to quote both `applications` and `browser_specific_settings` in this message? (I mean like `applications` => `"applications"` and the same for `browser_specific_settings`).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -72,6 +72,30 @@ describe('ManifestJSONParser', () => {
     });
   });
 
+  it('warn if both "applications" and "browser_specific_settings" properties are being used', () => {
```
rpl, line 75: Nitpick: "warn if ..." => "should warn if ..."

rpl (changes requested):
@Yrkan this looks great! would you mind to apply the small changes (just nits) described below?
after that this should be good to go! \o/

src/messages/manifestjson.js
Yrkan: @rpl sure and done :smiley:

rpl (approved):
Thanks @Yrkan, this looks great and ready to go! :tada:

Issue #2866, opened 2019-10-17
Use addons-scanner-utils `io` package


willdurand:
[addons-scanner-utils](https://github.com/mozilla/addons-scanner-utils) contains a `io` package used by the new scanners. It is a port of the linter files with some fixes and improvements. This would likely fix #858 :)

EnTeQuAk:
The link seems to be pointing into the wrong direction, you're talking about https://github.com/mozilla/addons-scanner-utils right?

I wonder, is the initial commit a 1:1 port of the linters io package so that the commit log represents all the fixes and improvements made? I'd like to make sure we don't run into any regressions for things that may have worked for the other scanners but won't for the linter.

willdurand:
> The link seems to be pointing into the wrong direction, you're talking about [mozilla/addons-scanner-utils](https://github.com/mozilla/addons-scanner-utils) right?

Err.. edited, thanks!

>  is the initial commit a 1:1 port of the linters io package so that the commit log represents all the fixes and improvements made

It's not 1:1 but close enough I'd say. The one big improvement is disabled by default so that would not affect the linter anyway.

EnTeQuAk:
> It's not 1:1 but close enough I'd say. The one big improvement is disabled by default so that would not affect the linter anyway.

Mind to elaborate a bit more or are you referring to https://github.com/mozilla/addons-scanner-utils/commit/17ce0771f5440477e390822c4e2b3999b579573b?

willdurand:
> Mind to elaborate a bit more or are you referring to [mozilla/addons-scanner-utils@17ce077](https://github.com/mozilla/addons-scanner-utils/commit/17ce0771f5440477e390822c4e2b3999b579573b)?

I am indeed referring to https://github.com/mozilla/addons-scanner-utils/pull/27. As for the rest, it can't be a 1:1 port because it has been rewritten with TypeScript.

diox:
Let's keep this opened and try to do it sometime...

willdurand:
See also: https://github.com/mozilla/addons-linter/issues/2865

AlexandraMoga:
@willdurand would some general linter checks (i.e submitting add-ons of various sizes and types, triggering validation warnings/errors) be enough to verify this issue or is there something more specific to be tested?

willdurand:
I think so, yea

ioanarusiczki:
@willdurand Verified on AMO Stage - FF82(Win10).

PR #3427, opened 2020-10-28
Use `addons-scanner-utils` for IO stuff

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/2866

---

This patch uses the `addons-scanner-utils` (used by other scanners) for the IO layer. There is no other change than that (the `autoClose` feature will be implemented in a follow-up patch).


rpl:
> @rpl FYI, this is needed to hopefully improve the linter execution speed (by disabling an obscure "auto-close" feature that I will describe in a new issue soon). The code in `addons-scanner-utils` is very similar to the code currently living in this project `src/io/`, but it has been rewritten with TypeScript. We're using it in other scanners so it should be fine.

nice! I like the idea of sharing that code instead of having multiple implementations and having to keep them in sync when we do fix bugs or enhance them.

@willdurand I just gave a quick look to the changes in the PR, but I was wondering: would it be reasonable to keep one integration test that runs the linter on a sample of the fixtures (e.g. one crx3, one xpi or zip file and a directory with an unpacked extension) being removed and confirm that we get the expected results?
just as a smoke test to make it easier to catch potential regressions when we do update the new dependency.

willdurand:
> @willdurand I just gave a quick look to the changes in the PR, but I was wondering: would it be reasonable to keep one integration test that runs the linter on a sample of the fixtures (e.g. one crx3, one xpi or zip file and a directory with an unpacked extension) being removed and confirm that we get the expected results?
> just as a smoke test to make it easier to catch potential regressions when we do update the new dependency.

I thought we had that already, no? We have a bunch of tests reading fixture files.

rpl:
> I thought we had that already, no? We have a bunch of tests reading fixture files.

Honestly I haven't checked yet if we do have enough coverage already, I was just pointing it out as something that may be worth to double-check explicitly.

I'll give it a look asap.

willdurand:
> Honestly I haven't checked yet if we do have enough coverage already, I was just pointing it out as something that may be worth to double-check explicitly.
> 
> I'll give it a look asap.

Thanks. `xpi` files and directories are well-covered I think (given the variety of fixtures). Maybe we should add one test for `crx` files, though.

diox (approved):
I'll leave to @rpl to decide if this needs more integration tests to be sure but r+ from me

rpl:
> I'll leave to @rpl to decide if this needs more integration tests to be sure but r+ from me

we discussed about this in the triage meeting today, we agreed to just add coverage for a crx file (at least one crx3 file) and then proceed with merging this as already accepted by diox.

I'll be happy to take a look if Will feels that it would be necessary, otherwise I will just review the follow up PR #3429 on marked as ready for a review.

PR #3502, opened 2020-12-15
Switch to Circle CI

willdurand:
This PR changes a few thing in addition to switching to Circle CI:

1. we build/test on Node 12 (production) and 14 (next). We don't build Node 10 anymore. I think that's fine.
2. `scripts/publish-rules` has been updated for Circle


.circleci/config.yml
```diff
@@ -0,0 +1,119 @@
+# These environment variables must be set in CircleCI UI
+#
+# NPM_TOKEN - A valid NPM token for releases
+# GH_USER   - A GitHub username
+# GH_EMAIL  - The email linked to the username above
+# GH_TOKEN  - A GitHub token linked to the username above
```
willdurand, line 6: @muffinresearch it looks we'd need these `GH_*` env variables too. It's related to the `scripts/publish-rules` script.

muffinresearch: Thanks, those are added.

.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
...
+      - *save_next_build_cache
+      - *make_production_build
+      - run: npm run test-ci
+
+  publish-rules:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: ~/addons-linter
+      - run: npm run publish-rules > /dev/null 2>&1
```
muffinresearch, line 112: I remember adding this to send stdin/stderr to `/dev/null` out of paranoia, not sure if that's still required.

muffinresearch (approved):
Lgtm r+wc

muffinresearch:
One thing I see the lint is showing up some warnings, I'm not sure if that's an issue: 

test__16__-_mozilla_addons-linter

.circleci/config.yml
muffinresearch: Scope is public_repo

willdurand:
> One thing I see the lint is showing up some warnings, I'm not sure if that's an issue:

That's a consequence of `eslint-config-amo` v4. It uses the jest recommended settings now and we do have "tests without assertions" because we're using `sinon` IIRC.

rpl (approved):
lgtm too

(also thanks a lot, your PRs will be useful to me too, as a reference while migrating the other repos I'm maintaining :heart: )

.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
...
+      - *save_build_cache
+      - *make_production_build
+      - *configure_global_npm
+      - run: npm run test-ci
+      # run integration tests using an addons-linter binary in a
+      # production-like environment
+      - run: npm run test-integration:production
+      - run: npm run lint
+      - run: npm run prettier-ci
+      - run: npm run webext-test-functional
```
rpl, line 88: @willdurand I just noticed that the integration tests (and web-ext-functional) do not currently run as part of test-next, that may mean that we may miss to notice some issues with our dependencies when the addons-linter does run on the next nodejs version. If the time needed to run these additional test steps does not make the total time to run the CI jobs too much higher, it may be reasonable to run it also as part of the test-next.

what do you think?

willdurand: Ah yes, sure. We can add more steps to the `next` job. I'll add the two steps you mentioned.

willdurand:
Note: I am not actually sure the publish rules job will work but we'll see soon enough I guess.

.circleci/config.yml
rpl: thanks! the updated version lgtm.

PR #3510, opened 2020-12-18
Fix npm auth command

willdurand:


willdurand:
Interesting, the node 14 (next) job seems to fail sometimes and it looks like a jest timeout at first glance.

rpl:
> Interesting, the node 14 (next) job seems to fail sometimes and it looks like a jest timeout at first glance.

@willdurand That test case doesn't seem to do anything too special (to be honest I think that we may not be exposing mozIndexedDB anymore on more recent Firefox releases, and I doubt that anyone is still using browser-prefixed IndexedDB global now-days, and so that test case may not be really that much needed anymore)

if turns out that is going to happen often, maybe we can try to increase the jest timeout when running on CI (especially if the test cases that timeouts are not going to be always the same one)

.circleci/config.yml
```diff
@@ -123,7 +123,8 @@ jobs:
           at: ~/addons-linter
       - run:
           name: authenticate with registry
-          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/addons-linter/.npmrc
+          # We don't want to expand the token in this file, npm will do it.
+          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/addons-linter/.npmrc
```
willdurand, line 127: Erf. We shouldn't use double quotes... See: https://github.com/mozilla/addons-linter/pull/3511/files

PR #3511, opened 2020-12-18
Fix npm auth command again

willdurand:


rpl (approved):
My bad, I should have spotted it.

If we are gonna stay on circleci it may be also worth to pack a small set of reusable commands in our own shared orb (which seems to be how circleci allows developers to reuse configuration pieces across multiple projects)

willdurand:
> If we are gonna stay on circleci it may be also worth to pack a small set of reusable commands in our own shared orb (which seems to be how circleci allows developers to reuse configuration pieces across multiple projects)

Yep, I was thinking about that, too.

PR #3525, opened 2021-01-02
Respect disable-linter-rules CLI argument

### LINKED ISSUE #3524: disable-linter-rules CLI argument is ignored

I'm using this utility for the first time and the `--disable-linter-rules` CLI argument is not working for me.  I've tried several variations of:
```cmd
npx addons-linter --disable-linter-rules=no-new-func dist
```

### What happened?

The output included warnings for
> DANGEROUS_EVAL
> The Function constructor is eval.

The output would omit any warnings/errors related to the ESLint rule(s) specified in the `--disable-linter-rules` argument.

I'll submit a PR shortly.  Thanks for this great tool!

wimpyprogrammer:
Fixes #3524.

Upon debugging the issue, I noticed that the JavaScript Scanner expects a property named `disabledRules` https://github.com/mozilla/addons-linter/blob/1b8cb482464da5b2d8a8844adb5084fc2aa60d74/src/scanners/javascript.js#L27-L29

but is passed a property named `disabledLinterRules` https://github.com/mozilla/addons-linter/blob/1b8cb482464da5b2d8a8844adb5084fc2aa60d74/src/linter.js#L389

and so the list of disabled rules was always empty.

The JavaScript Scanner is the only Scanner that accepts a list of disabled rules, and I only found one caller that tries to set that option (`linter.js`, fixed in this PR).

-----

I also adjusted the `jest/expect-expect` ESLint rule to recognize `sinon.assert` assertions.  This eliminated ~15 false-positives for that rule (though some false-positives remain for other reasons).

### DIFF
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -20,7 +20,7 @@
       "devDependencies": true
     }],
     "jest/expect-expect": ["warn", {
-      // Allow Jest, Sinon, or helper lib assertions
+      // Allow Jest or Sinon assertions
       "assertFunctionNames": ["expect", "sinon.assert.*"]
     }],
     "promise/always-return": "off",
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -19,6 +19,10 @@
       // Allow dev-dependencies in this directory.
       "devDependencies": true
     }],
+    "jest/expect-expect": ["warn", {
+      // Allow Jest, Sinon, or helper lib assertions
+      "assertFunctionNames": ["expect", "sinon.assert.*"]
+    }],
     "promise/always-return": "off",
     "promise/avoid-new": "off",
     "promise/no-nesting": "off",
--- a/src/linter.js
+++ b/src/linter.js
@@ -386,7 +386,7 @@ export default class Linter {
         // https://github.com/mozilla/addons-linter/issues/895
         collector: this.collector,
         // list of disabled rules for js scanner
-        disabledLinterRules: this.config.disableLinterRules,
+        disabledRules: this.config.disableLinterRules,
         existingFiles: this.io.files,
       });
 
--- a/tests/unit/test.linter.js
+++ b/tests/unit/test.linter.js
@@ -1549,6 +1549,27 @@ describe('Linter.run()', () => {
     expect(result).toEqual(addonLinter.output);
   });
 
+  it('should pass disabled linting rules to Scanner', async () => {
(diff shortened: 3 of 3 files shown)


tests/.eslintrc
```diff
@@ -19,6 +19,10 @@
       // Allow dev-dependencies in this directory.
       "devDependencies": true
     }],
+    "jest/expect-expect": ["warn", {
```
willdurand, line 22: Should probably be an error if the config below is accurate.

willdurand (commented):
Looks good, thanks. I only have a comment about the ESLint config.

tests/.eslintrc
wimpyprogrammer: Unfortunately this configuration doesn't fix all violations of `jest/expect-expect`.  You can see that [the `master` branch has 26 violations](https://app.circleci.com/pipelines/github/mozilla/addons-linter/86/workflows/a1a3c74c-6cdb-4619-8b18-191856073147/jobs/207/parallel-runs/0/steps/0-109), and [this feature branch still has 12 violations](https://app.circleci.com/pipelines/github/mozilla/addons-linter/53/workflows/c4f2a855-10fa-45be-9f53-ed3fa9f1a6b8/jobs/127/parallel-runs/0/steps/0-109).

Most of the remaining `jest/expect-expect` violations have `expect()` assertions that the lint rule doesn't see, like https://github.com/mozilla/addons-linter/blob/247c3b02308eee34c257f2d2cad6a8888d7fe159/tests/unit/parsers/test.manifestjson.js#L2730-L2732 and https://github.com/mozilla/addons-linter/blob/247c3b02308eee34c257f2d2cad6a8888d7fe159/tests/unit/schema/test.background.js#L9-L17  Those could be permitted by adding the helper function names to `assertFunctionNames` like
```json
"assertFunctionNames": ["expect", "sinon.assert.*", "testHomepageUrl", "assertHasMatchingError"]
````

And some of the violations are indeed assertion-less tests like https://github.com/mozilla/addons-linter/blob/247c3b02308eee34c257f2d2cad6a8888d7fe159/tests/unit/scanners/test.html.js#L105-L110

Basically I noticed that my new test was adding an ESLint warning (because it doesn't have `expect()`) and I wanted to avoid that.  But as a first-time contributor, I figure you probably don't want me making involved changes to your ESLint config 😃.  I'm happy to revert the ESLint change if it simplifies this PR.

willdurand: it's probably fine, thanks!

PR #3534, opened 2021-01-05
Fix message when the file is too large to parse

### LINKED ISSUE #1700: Linter should not suggest including large JSON file as separate JSON file
image

prianka01:
Fixes #1700 

The linter should suggest "Consider moving large lists of data out of JavaScript files and into JSON files, or splitting very large files into smaller ones" instead of "including large JSON file as separate JSON file" when the JSON file is too large. The wording is changed in this commit. The [test](https://github.com/mozilla/addons-linter/blob/cc92e134df63d66389ac4d4a0c6d5d534bb28abb/tests/unit/test.linter.js#L1283-L1320) for the corresponding message worked correctly.
Reproduced this issue locally as `addonLinter.collector.errors[0].description` in the test gave the updated rule description message when the JSON file was large.


willdurand:
cc @rpl

rpl (approved):
lgtm, thanks!

PR #3544, opened 2021-01-13
Add Support for Numeric Separators

### LINKED ISSUE #3543: JavaScript syntax error with numeric separator
### What happened?
I ran Add-ons Linter from the command line, and got a validation error:

> There is a JavaScript syntax error in your code; validation cannot continue on this file.

Which points to this line of code, at column 14:
`let tmp = 10_000`.
It seems that the error is caused by the numeric separator.

I expect that Add-ons Linter will not throw any errors, since the use of numeric separators is described on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Numeric_separators, and is supported since Firefox 70, which was released more than a year ago.

michaelts1:
This PR adds support for numeric separators.

Fixes #3543

### DIFF
--- a/tests/unit/scanners/test.javascript.js
+++ b/tests/unit/scanners/test.javascript.js
@@ -150,7 +150,7 @@ describe('JavaScript Scanner', () => {
     expect(linterMessages).toEqual([]);
   });
 
-  it('numeric separators', async () => {
+  it('should support numeric separators', async () => {
     const code = 'const num = 1_0;';
 
     const jsScanner = new JavaScriptScanner(code, 'code.js');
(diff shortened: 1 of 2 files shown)


tests/unit/scanners/test.javascript.js
```diff
@@ -150,6 +150,15 @@ describe('JavaScript Scanner', () => {
     expect(linterMessages).toEqual([]);
   });
 
+  it('numeric separators', async () => {
```
Rob--W, line 153: Prepend "should support".

michaelts1: Thanks, missed that somehow. Fixed.

Rob--W (approved):
Thanks for the patch. LGTM!

PR #3545, opened 2021-01-13
Drop `whatwg-url` dependency

fregante:
`URL` has been part of Node since v6.13 and it's a global since Node 10 (which matches your `engines.node` field)

https://nodejs.org/api/url.html#url_class_url


willdurand (approved):
Thank you.

Issue #3482, opened 2020-12-04
Improve error message when linting fails due to the use of unrecognized JavaScript features


Rob--W:
### Describe the problem and steps to reproduce it:

The addons-linter relies on eslint for validating the source code, and eslint has a very conservative policy to linting. In particular, new JavaScript language features are not supported until they have matured, [by reaching stage 4 in the TC39 process](https://github.com/eslint/eslint#user-content-what-about-experimental-features). We have had several reports about this in the past, and this is not a new issue. What we can do, however, is to provide a more useful description than a plain "syntax error".

For example, submit an add-on that uses public class field syntax, as shown in #3062

### What happened?

An error message complaining about a JavaScript syntax error.

### What did you expect to happen?

An error message complaining about a JavaScript syntax error, plus a reference to documentation and/or a mention that some experimental JavaScript features that aren't an official part of the language specification are not supported yet.

willdurand:
@rpl @Rob--W would something like this work for you?

```
❯ ./bin/addons-linter test-ext
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code              Message                                   Description                                                                        File    Line   Column
JS_SYNTAX_ERROR   JavaScript syntax error (Parsing error:   There is a JavaScript syntax error in your code, which might be related to some    cs.js   2      5
                  Unexpected token =)                       experimental JavaScript features that aren't an official part of the language
                                                            specification and therefore not supported yet. The validation cannot continue on
                                                            this file.
```

rpl:
> @rpl @Rob--W would something like this work for you?
> 
> ```
> ❯ ./bin/addons-linter test-ext
> Validation Summary:
> 
> errors          1
> notices         0
> warnings        0
> 
> ERRORS:
> 
> Code              Message                                   Description                                                                        File    Line   Column
> JS_SYNTAX_ERROR   JavaScript syntax error (Parsing error:   There is a JavaScript syntax error in your code, which might be related to some    cs.js   2      5
>                   Unexpected token =)                       experimental JavaScript features that aren't an official part of the language
>                                                             specification and therefore not supported yet. The validation cannot continue on
>                                                             this file.
> ```

@willdurand I think that  would work if we can be sure that the JavaScript error we are going to show in the message column is going to be the one that does point the developer to the real parsing error.

[...]

PR #3556, opened 2021-01-27
Improve error message when there is a JS syntax error

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3482

---

It'll be awesome once it's ready.

## Output

```
❯ ./bin/addons-linter test-addon
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code              Message                                               Description                                                                                        File    Line   Column
JS_SYNTAX_ERROR   JavaScript syntax error (Parsing as module error:     There is a JavaScript syntax error in your code, which might be related to some experimental       cs.js   1      1
                  Unexpected token = at line: 4 and column: 5)          JavaScript features that aren't an official part of the language specification and therefore not
                  (Parsing as script error: 'import' and 'export' may   supported yet. The validation cannot continue on this file.
                  appear only with 'sourceType: module' at line: 1
                  and column: 1)
```

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -51,7 +51,8 @@ export default class JavaScriptScanner {
     // testing purposes.
     _rulePaths = undefined,
   } = {}) {
-    this.sourceType = this.detectSourceType(this.filename);
+    const detectedSourceType = this.detectSourceType(this.filename);
+    this.sourceType = detectedSourceType.sourceType;
 
     const rules = {};
     Object.keys(_ruleMapping).forEach((ruleName) => {
@@ -121,9 +122,32 @@ export default class JavaScriptScanner {
 
     results.forEach((result) => {
       result.messages.forEach((message) => {
+        let extraShortDescription = '';
+
         // Fatal error messages (like SyntaxErrors) are a bit different, we
-        // need to handle them specially.
-        if (message.fatal === true) {
+        // need to handle them specially. Messages related to parsing errors do
+        // not have a `ruleId`, which is why we check that, too.
+        if (message.fatal === true && message.ruleId === null) {
+          // If there was a parsing error during the sourceType detection, we
+          // want to add it to the short description. We start by adding it to
+          // a temporary variable in case there are other messages we want to
+          // append to the final short description (which will be the `message`
+          // in the final output).
+          if (detectedSourceType.parsingError !== null) {
+            const { type, error } = detectedSourceType.parsingError;
+            extraShortDescription = `(Parsing as ${type} error: ${error})`;
+          }
+
+          // If there was another error, we want to append it to the short
+          // description as well. `message.message` will contain the full
+          // exception message, which likely includes a prefix that we don't
+          // want to keep.
+          const formattedError = message.message.replace('Parsing error: ', '');
+          extraShortDescription = [
+            extraShortDescription,
+            oneLine`(Parsing as ${this.sourceType} error: ${formattedError} at
+              line: ${message.line} and column: ${message.column})`,
+          ].join(' ');
           // eslint-disable-next-line no-param-reassign
           message.message = _messages.JS_SYNTAX_ERROR.code;
         }
@@ -161,6 +185,10 @@ export default class JavaScriptScanner {
         } else {
           shortDescription = code;
           description = null;
+        }
+
+        if (extraShortDescription.length) {
+          shortDescription += ` ${extraShortDescription}`;
         }
 
         this.linterMessages.push({
@@ -210,7 +238,7 @@ export default class JavaScriptScanner {
               return 'module';
             }
           }
-        } else {
+        } else if (child) {
           return this._getSourceType(child);
         }
       }
@@ -220,10 +248,14 @@ export default class JavaScriptScanner {
   }
 
   /*
-    Analyze the source-code by by parsing the source code manually and
-    check for import/export syntax errors.
+    Analyze the source-code by naively parsing the source code manually and
+    checking for module syntax errors in order to determine the source type of
+    the file.
 
-    This returns `script` or `module`.
+    This function returns an object with the source type (`script` or `module`)
+    and a non-null parsing error object when parsing has failed with the default
+    source type. The parsing error object contains the `error` message and the
+    source `type`.
   */
   detectSourceType(filename) {
     // Default options taken from eslint/lib/linter:parse
@@ -233,15 +265,32 @@ export default class JavaScriptScanner {
       ecmaVersion: ECMA_VERSION,
     };
 
-    let sourceType = 'module';
+    const detected = {
+      sourceType: 'module',
+      parsingError: null,
+    };
 
     try {
       const ast = espree.parse(this.code, parserOptions);
-      sourceType = this._getSourceType(ast);
+      detected.sourceType = this._getSourceType(ast);
     } catch (exc) {
-      sourceType = 'script';
+      const line = exc.lineNumber || '(unknown)';
+      const column = exc.column || '(unknown)';
+      let error = `${exc.message} at line: ${line} and column: ${column}`;
+
+      // When there is no line/column, it likely means something went wrong in
+      // our code (`_getSourceType()`) and we should know about it so we append
+      // a comment to hopefully get new bug reports.
+      if (!exc.lineNumber || !exc.column) {
+        error = oneLine`${error}. This looks like a bug in addons-linter,
+          please open a new issue:
+          https://github.com/mozilla/addons-linter/issues`;
+      }
+
+      detected.sourceType = 'script';
+      detected.parsingError = { type: parserOptions.sourceType, error };
     }
 
-    return sourceType;
+    return detected;
   }
 }
--- a/tests/unit/scanners/test.javascript.js
+++ b/tests/unit/scanners/test.javascript.js
@@ -390,6 +390,28 @@ describe('JavaScript Scanner', () => {
     expect(linterMessages[0].message).toEqual('this is the message');
   });
 
+  it('outputs the parsing errors in the message when possible', async () => {
+    const code = `
+      export const FOO = "FOO";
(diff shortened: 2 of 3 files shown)


tests/unit/rules/javascript/test.no_eval.js
```diff
@@ -52,7 +52,10 @@ describe('no_eval', () => {
     },
     {
       code: '(function(exe){ exe("foo") })(eval);',
-      message: ['eval can be harmful.'],
+      // The code snippet above will be parsed as script and we'll get the
+      // parsing as module error in the message. We use
+      // `expect.stringContaining()` to ignore the error message part.
+      message: [expect.stringContaining('eval can be harmful.')],
```
Rob--W, line 58: Can't you be a little bit more precise to make sure that all messages are as expected?

willdurand: I was thinking more about writing tests for the parsing error part (the new code) once we agree on the implementation.

rpl: @Rob--W I'm not sure that is really necessary in this particular test case, the change required in this PR seems to be mainly needed to don't assume as part of the assertion that the only thing that the message will contain is the no-eval one (but the core of the assertion is still intact: "we expect the no-eval rule to have added the expected validation message".

Anyway I don't think we need to block on this particular detail, this PR is still a draft and if needed we can discuss more about these details in the actual review passes.

rpl:
@willdurand looked to this again once more and I confirm that I'm ok with the general approach, my +1 on proceeding to complete this PR.

willdurand:
@rpl so I wasn't very inspired when writing tests but on the other hand, I think coverage is pretty good already. I thought about adding some more tests for the `replace()` call, etc. but the test setup would be quite complex for little value IMO. WDYT? Mind looking at this patch a bit? It's not urgent.

tests/unit/scanners/test.javascript.js
```diff
@@ -435,5 +457,36 @@ describe('JavaScript Scanner', () => {
...
+      const jsScanner = new JavaScriptScanner(code, 'code.js');
+      const detectedSourceType = jsScanner.detectSourceType('code.js');
+
+      expect(detectedSourceType.sourceType).toEqual('script');
+      expect(detectedSourceType.parsingError).toEqual({
+        type: 'module',
+        error: oneLine`Cannot read property 'type' of null at line: (unknown)
+          and column: (unknown)`,
+      });
+    });
```
rpl, line 490: @willdurand while reading this test I was wondering "where is that error coming from?", and based on what I think is happening here it seems to me that this test case is actually leveraging a bug in our `JavascriptScanner`'s `_getSourceType` method:

https://github.com/mozilla/addons-linter/blob/5b722ec64f964d0bcc028dd7a32a3329e6c3444a/src/scanners/javascript.js#L207-L215

If I'm readin it correctly, that exception is likely coming from a call that goes into line 214, and it is due to a child node with a `null` value.

Looking to the AST that esprima should be returning to us on the code snippet `(function () {})();`:
- https://astexplorer.net/#/gist/6bdd2c8a13a959cc2527f1606ef073c2/64a9440ac8d1f15003339fba95fceb40653cfdcc

I think that the null node is actually the `id` property of the FunctionExpression node:
```
{
  type: 'FunctionExpression',
  id: null,
  ...
},
```

This looks like a bug in the `_getSourceType` logic that we should actually fix, e.g. by only recurse if the child node is defined and have the `type` property that we expect from any AST node.

As a side effect, that also means that we can't leverage it for this test, and so my second question is:
- is there any other condition (besides bug in our own code as in this case) that would generate a parsing error that doesn't have line and column properties? (I guess that another similar one may be bugs in eslint parsing itself, still not helpful for coverage).

I think that it is still reasonable to look into the behavior that we'd like the addons-linter to have in case of bugs like this one, but:
- to cover it with a test we may opt for temporarily overriding our `_getSourceType` method or `espree.parse` (e.g. using jest mocking features), this way the test should not break if the bug is being fixed
- we may want to evaluate if we can detect this conditions consistently and report the validation failure with a separate error code specific to "unexpected parsing issue that we would like to be reported to us" (because they are more likely to be bugs that we want to fix and not related to the limitation of the default eslint parser)

Let me know what do you think (and if you think that it would be useful to have a brief chat together over Matrix or Slack about this).

rpl (changes requested):
@willdurand I didn't reviewed all the PR in detail, but I did find a detail (described below) that I'd like to discuss with you (while I'm going to look to the rest of this patch).

tests/unit/scanners/test.javascript.js
willdurand: > is there any other condition (besides bug in our own code as in this case) that would generate a parsing error that doesn't have line and column properties? (I guess that another similar one may be bugs in eslint parsing itself, still not helpful for coverage).

I am not so sure about that. That's the only error I found.

willdurand:
> @willdurand I didn't reviewed all the PR in detail, but I did find a detail (described below) that I'd like to discuss with you (while I'm going to look to the rest of this patch).

@rpl thanks. I updated this PR to fix the issue you raised in the previous review. I added a note about reporting a new issue when there is likely a bug in our code. The rest is unchanged. Let me know what you think.

willdurand:
(ugh, forgot to request a review)

tests/unit/rules/javascript/test.no_eval.js
rpl, line 58: (My apologies, I'm going to go back to discuss again about a point we previously discussed briefly in https://github.com/mozilla/addons-linter/pull/3556/files#r570419337).

I did want to double-check this one more time, and I think that we may actually have to fix this instead of making the test to pass:

- the test case related to this change is currently failing because the error produced by this tests becomes `"eval can be harmful. (Parsing as module error: Binding eval in strict mode at line: 1 and column: 14)"`, which I think it wrong because we are appending a parsing error to the message coming from another eslint rule that does not have anything to do with a parsing error (and also, given that the eslint rule have been able to produce this message, that means that at least we failed to parse it as an module but not as a script)

I think we may want to revent the changes applied to this change and tweak the changes applied to the `JavaScriptScanner` `scan` method to only add the parsing error to a fatal error without a ruleId (that should allow us to avoid to append the "parsing as a module" error to results coming from the eslint rules, and the parsing error doesn't include a ruleId), e.g. I think that something like this should do it and be reasonable enough:

```js
    results.forEach((result) => {
      result.messages.forEach((message) => {
        let extraShortDescription = [];
        // ...
        if (message.fatal === true && message.ruleId == null) {
          // ...
          if (detectedSourceType.parsingError) {
            const { type, error } = detectedSourceType.parsingError;
            extraShortDescription.push(`(Parsing as ${type} error: ${error})`);
          }

          // ...
          const formattedError = message.message.replace('Parsing error: ', '');
          extraShortDescription.push(
            oneLine`(Parsing as ${this.sourceType} error: ${formattedError} at
              line: ${message.line} and column: ${message.column})`
          );
          // eslint-disable-next-line no-param-reassign
          message.message = _messages.JS_SYNTAX_ERROR.code;
        }

        ...

        if (code === _messages.JS_SYNTAX_ERROR.code &&
            extraShortDescription.length) {
          shortDescription += ` ${extraShortDescription.join(' ')}`
        }
```

@willdurand what do you think? does that sounds like a reasonable approach?

rpl (changes requested):
@willdurand apologies for letting you wait, there are a couple more changes that I think would be good to apply to this PR (these should be the last ones, I promise ;-))

Let me know if you think that any of those points may use some discussion over a synchronous channel.

src/scanners/javascript.js
```diff
@@ -220,28 +247,51 @@ export default class JavaScriptScanner {
...
+    checking for module syntax errors in order to determine the source type of
+    the file.
+
+    This function returns an object with the source type (`script` or `module`)
+    and a non-null parsing error object when parsing has failed with the default
+    source type. The parsing error object contains the `error` message and the
+    source `type`.
   */
-  detectSourceType(filename) {
+  detectSourceType(filename, _getSourceType = null) {
```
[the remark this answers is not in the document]
willdurand: yep

tests/unit/rules/javascript/test.no_eval.js
willdurand: awww, I definitely overlooked that part. Yeah, that makes sense, sorry I should have caught that myself!

willdurand: @rpl hey, I see you added a check at the bottom of the snippet above:

```
if (code === _messages.JS_SYNTAX_ERROR.code &&
            extraShortDescription.length) {
```

Is it useful considering that we only update `extraShortDescription` where we set `code` to `_messages.JS_SYNTAX_ERROR.code`? I am thinking that we might want to reuse `extraShortDescription` in the future so maybe we don't need this new condition (and keep adding `extraShortDescription` when it has content)?

willdurand: (I updated this PR with the other changes)

rpl: > @rpl hey, I see you added a check at the bottom of the snippet above:
> 
> ```
> if (code === _messages.JS_SYNTAX_ERROR.code &&
>             extraShortDescription.length) {
> ```
> 
> Is it useful considering that we only update `extraShortDescription` where we set `code` to `_messages.JS_SYNTAX_ERROR.code`? I am thinking that we might want to reuse `extraShortDescription` in the future so maybe we don't need this new condition (and keep adding `extraShortDescription` when it has content)?

@willdurand yeah, I was honestly on the fence on still checking the `extraShortDescription.length` there for similar reasons, I think it should be just fine if we do only check if extraShortDescription.length is not 0 instead.

rpl (approved):
@willdurand lgtm, thanks for covering the last round of changes requested.

PR #3562, opened 2021-02-03
feat: Imported Firefox 86.0b5 API schema

rpl:
This PR includes the following changes to the API schemas:

- increased from `2` to `3` the maximum allowed value for the `manifest_version` manifest property ([Bug 1594234](https://bugzilla.mozilla.org/1594234)) 
- removed `content_scripts` and `isolated_world` sub-properties from the `content_security_policy` manifest property definition ([Bug 1594234](https://bugzilla.mozilla.org/1594234)) 
- added a new CPU experimental feature to the geckoProfiles API schema ([Bug 1679930](https://bugzilla.mozilla.org/show_bug.cgi?id=1679930)) 
- changed to the pattern used to validate `chrome_settings_overrides.search_provider.suggest_url` (which seems to have been added to allow it to be set as empty)
- added some more values to the `PlatformArch` enum ([Bug 1554971](https://bugzilla.mozilla.org/1554971))
- changed definition for focused property in the `windows.create` API schema (to just print a warning when set to `false`, to explicitly state that it isn't actually supported and it is just ignored)

TODO - I think that before merging this PR we should:
- [x] make sure we don't yet accept `manifest_version: 3` on submissions to AMO (we should make it an error for now, until we are ready to start accepting manifest v3 addons)
- [x] double-check if the changes to the `content_security_policy` do require changes to the CSP validation that the addons-linter does internally independently from the API schema imported (which should be in `src/parsers/manifestjson.js`).

Based on what I recall about the changes we did more recently in the code that is validating the csp string there shouldn't be any changes needed, but I don't recall all the details about it and so it would be better to double-check that explicitly.

### DIFF
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -82,6 +82,9 @@
           "version": {
             "description": "Version string must be a string comprising one to four dot-separated integers (0-65535). E.g: 1.2.3.",
             "format": "versionString"
+          },
+          "manifest_version": {
+            "maximum": 2
           }
         }
       }
--- a/src/schema/imported/chrome_settings_overrides.json
+++ b/src/schema/imported/chrome_settings_overrides.json
@@ -47,7 +47,7 @@
                 },
                 "suggest_url": {
                   "type": "string",
-                  "pattern": "^(https://|http://(localhost|127\\.0\\.0\\.1|\\[::1\\])(:\\d*)?(/|$)).*$",
+                  "pattern": "^$|^(https://|http://(localhost|127\\.0\\.0\\.1|\\[::1\\])(:\\d*)?(/|$)).*$",
                   "preprocess": "localize"
                 },
                 "instant_url": {
--- a/src/schema/imported/geckoProfiler.json
+++ b/src/schema/imported/geckoProfiler.json
@@ -180,7 +180,8 @@
         "fileio",
         "fileioall",
         "noiostacks",
-        "audiocallbacktracing"
+        "audiocallbacktracing",
+        "cpu"
       ]
     },
     "supports": {
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -11,7 +11,8 @@
         "manifest_version": {
           "type": "integer",
           "minimum": 2,
-          "maximum": 2
+          "maximum": 2,
+          "postprocess": "manifestVersionCheck"
         },
         "applications": {
           "type": "object",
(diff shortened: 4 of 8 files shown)


rpl:
@Rob--W that are a couple of todo listed in https://github.com/mozilla/addons-linter/pull/3562#issue-567096573 that I think we should look into before calling this PR ready to be merged, I didn't actually looked into those yet but I did already set you as a reviews on this to get your perspective on the API schema import diff (in case you do notice anything else that I may have missed)

Rob--W (changes requested):
The `manifestVersionCheck` option needs to be added at https://github.com/mozilla/addons-linter/blob/5045967ef1383d89ddea8f3c9b0e77838de8c5a6/src/schema/validator.js#L34

The `contentSecurityPolicy` format doesn't require changes yet. We need to re-evaluate when we support manifest v3 or with upcoming changes to support `sandbox.content_security_policy`. To make sure that we don't accidentally import changes when manifest v3 is enabled, consider adding a comment to the (necessary) `manifestVersionCheck` validator.

rpl:
> The `manifestVersionCheck` option needs to be added at
> 
> https://github.com/mozilla/addons-linter/blob/5045967ef1383d89ddea8f3c9b0e77838de8c5a6/src/schema/validator.js#L34

I actually opted for another strategy that I had in mind, which was enforcing the maximum expected manifest_version value  by overriding it `src/schema/updates/manifest.json` (we do some over overriding and merging from the `src/schema/updates/*.json` files and we reapply those overrides on each new schema import.

I added an additional test case in a separate patch to explicitly cover that expected validation error (nearby the similar test case to make sure that manifest_version < than 2 is invalid).
 
> The `contentSecurityPolicy` format doesn't require changes yet. We need to re-evaluate when we support manifest v3 or with upcoming changes to support `sandbox.content_security_policy`. To make sure that we don't accidentally import changes when manifest v3 is enabled, consider adding a comment to the (necessary) `manifestVersionCheck` validator.

In this PR I decided to only tweak the existing manifest v3 test case to don't use manifest properties that are not currently officially supported and have been removed from the schema.

I think that we can evaluate further changes related to the content_security_policy format allowed separately from this particular PR.

src/schema/updates/manifest.json
```diff
@@ -82,6 +82,9 @@
           "version": {
             "description": "Version string must be a string comprising one to four dot-separated integers (0-65535). E.g: 1.2.3.",
             "format": "versionString"
+          },
+          "manifest_version": {
+            "maximum": 2
```
wagnerand, line 87: Do we have to remember adding this manually every time we import the schema? If so, that seems risky.

Rob--W: This looks like a manual override (opposed to the files in `imported`), and there's a unit test that would fail if this constraint were to be dropped. Therefore I have confidence in the direction of this patch.

Luca can confirm if needed.

Rob--W (approved):

src/schema/updates/manifest.json
rpl: @wagnerand no, we don't need to remind overselves to reapply the override, that is what the change in src/schema/updates/manifest.json does (it gets propagate on every schema import).

I did put an inline comment nearby the test case to remind ourselves that we will have to remove that override when we are ready to accept manifest_version 3 extensions on AMO (we may also file an issue to keep open and remind that as well).

wagnerand: Thank you both for clarifying!

PR #3600, opened 2021-03-08
Replace probe-image-size with image-size

fregante:
- `probe-image-size` has 10 dependencies: https://npm.broofa.com/?q=probe-image-size (total [884KB](https://packagephobia.com/result?p=probe-image-size@6.0.0))
- `image-size` only has 2: https://npm.broofa.com/?q=image-size (total [67KB](https://packagephobia.com/result?p=image-size))

The latter is also used 10 times more:

- https://www.npmjs.com/package/probe-image-size
- https://www.npmjs.com/package/image-size

~~This PR is not yet clean, but passes the tests.~~ I plan on sending more PRs to reduce [`addons-linter`’s huge dependency tree](https://npm.broofa.com/?q=addons-linter) if this is welcome


tests/unit/helpers.js
```diff
@@ -15,7 +15,7 @@ export const fakeMessageData = {
   message: 'message',
 };
 
-export const EMPTY_SVG = Buffer.from('<svg viewbox="0 0 1 1"></svg>');
+export const EMPTY_SVG = Buffer.from('<svg viewBox="0 0 1 1"></svg>');
```
fregante, line 18: Due to https://github.com/image-size/image-size/issues/355

An alternative solution is what I dropped in [`42897b0` (#3600)](https://github.com/mozilla/addons-linter/pull/3600/commits/42897b03838b7e83c13e4db7af52ec56088c1d27)

willdurand: > Due to [image-size/image-size#355](https://github.com/image-size/image-size/issues/355)

Would be nice to get fixed in the upstream lib, yeah.

src/parsers/manifestjson.js
```diff
@@ -40,6 +40,32 @@ import {
...
+function getMimeFromExtension(extension) {
+  for (const [currentMime, extensions] of Object.entries(
+    MIME_TO_FILE_EXTENSIONS
+  )) {
+    if (extensions.includes(extension)) {
+      return currentMime;
+    }
+  }
+
+  return undefined;
```
fregante, line 66: I could also replace this with an additional constant map of `{[ext]: mime}`

fregante: Done in 391d9b3

willdurand (commented):
That looks good to me. Let's make sure to use the latest version of `image-size` (which includes your fix)

tests/unit/helpers.js
fregante: Done and this test change was dropped

src/parsers/manifestjson.js
```diff
@@ -40,6 +40,20 @@ import {
...
 import BLOCKED_CONTENT_SCRIPT_HOSTS from 'blocked_content_script_hosts.txt';
 
+async function getStreamImageSize(stream) {
+  const chunks = [];
+  for await (const chunk of stream) {
+    chunks.push(chunk);
+    try {
+      return getImageSize(Buffer.concat(chunks));
+    } catch (error) {
+      /* Not ready yet */
```
willdurand, line 50: What does that mean exactly? Is it that we don't have enough data to detect the image size?

willdurand (commented):
Looks good to me. I think we should update the comment in `getStreamImageSize()`.

src/parsers/manifestjson.js
fregante: That's correct. Unfortunately `getImageSize` will throw an error when it can't determine the size of the file. In this case, it happens repeatedly until it receives enough information.

My guess though is that this information is within the first chunk, at least that's what I experienced in my few tests with the `http` module.

Should it be

```suggestion
      /* The size information isn't available yet */
```

willdurand: > Should it be

yeah

willdurand (approved):

willdurand:
Looks good to me, thanks. I wasn't involved in the initial iteration of that feature so I trust our test suite :)

PR #3602, opened 2021-03-08
Replace heavy dependency (espree)

fregante:
`espree` takes [1.7MB](https://packagephobia.com/result?p=espree) of disk space ([3 dependencies](https://npm.broofa.com/?q=espree)) and is actually currently duplicated as can be seen in the [lockfile](https://github.com/mozilla/addons-linter/blob/f756b4dec94705020170fbfd7149d39ee36940c5/package-lock.json)

It seemed excessive to load a whole JS parser module just to detect the script type, but I understand if this solution isn't considered on par.

It uses Node’s own parser to detect the module type, with 0 dependencies 🎈

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "eslint": "7.21.0",
         "eslint-plugin-no-unsanitized": "3.1.4",
         "eslint-visitor-keys": "2.0.0",
-        "espree": "7.3.1",
         "esprima": "4.0.1",
         "fluent-syntax": "0.13.0",
         "glob": "7.1.6",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "eslint": "7.21.0",
     "eslint-plugin-no-unsanitized": "3.1.4",
     "eslint-visitor-keys": "2.0.0",
-    "espree": "7.3.1",
     "esprima": "4.0.1",
     "fluent-syntax": "0.13.0",
     "glob": "7.1.6",
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -220,8 +220,9 @@ export default class JavaScriptScanner {
 
     This returns `script` or `module`.
   */
-  detectSourceType(filename) {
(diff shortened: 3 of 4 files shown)


fregante:
The regular test is likely failing because this change uses Node’s parser, therefore it completely depends on Node’s version.

willdurand:
mmm I thought we were using `espree` for more than module/script detection. Isn't it the ESLint parser, too. I am sure we need to require it directly to avoid a linting issue (like importing from a lib not present in `package.json` or something)

fregante:
> Isn't it the ESLint parser, too

It's the default, yes, but currently they're 2 separate versions in the lockfile so it's being downloaded and executed twice.



> I am sure we need to require it directly to avoid a linting issue

Is it possible that this issue has been resolved in the meanwhile? Similar to how many Babel transforms aren't needed anymore (#3603)

fregante:
It appears that https://github.com/mozilla/addons-linter/pull/3556 touched the same file here so I'd have to re-implement the same logic, if accepted.

willdurand:
> It's the default, yes, but currently they're 2 separate versions in the lockfile so it's being downloaded and executed twice.

Yeah, I don't think we should have two versions. We should probably keep `espree` as default parser and see if importing it without explicitly requiring it in `package.json` works.

fregante:
> Yeah, I don't think we should have two versions.

That's unavoidable sometimes, versions are managed by npm and the lockfile.

> importing it without explicitly requiring it in `package.json` works.

It might, but if eslint changes/updates it with something incompatible then `addons-linter` will fail to run. If you want to keep `espree`, no changes are needed (other than trying to follow eslint’s version exactly).

willdurand:
> It might, but if eslint changes/updates it with something incompatible then `addons-linter` will fail to run. If you want to keep `espree`, no changes are needed (other than trying to follow eslint’s version exactly).

Ok, I think we want to keep `espree` for now and maybe try to pin the same version used by eslint.

PR #3617, opened 2021-03-15
Use ECMA_VERSION constant from addons-scanner-utils

### LINKED ISSUE #3616: Use ECMA_VERSION constant from addons-scanner-utils
This is to avoid problems when our different tools have different supported JS features.

willdurand:
Fixes #3616


willdurand:
@rpl I requested a review from you so that you can be aware of the change

rpl:
> @rpl I requested a review from you so that you can be aware of the change

thanks!

rpl (approved):

PR #3603, opened 2021-03-08
Drop some babel config

fregante:
It seems to me that using babel in this project goes back quite a bit, and it was just carried over for inertia. Is that correct?

Why is there babel-env? Does this package run in the browser? I ran some tests and here are the results

## Current config

```is
asset addons-linter.js 727 KiB [compared for emit] (name: addons-linter) 1 related asset
asset rules/javascript/webextension-deprecated-api.js 511 KiB [compared for emit] (name: rules/javascript/webextension-deprecated-api) 1 related asset
asset rules/javascript/webextension-unsupported-api.js 469 KiB [compared for emit] (name: rules/javascript/webextension-unsupported-api) 1 related asset
asset rules/javascript/webextension-api-compat-android.js 469 KiB [compared for emit] (name: rules/javascript/webextension-api-compat-android) 1 related asset
asset rules/javascript/webextension-api-compat.js 469 KiB [compared for emit] (name: rules/javascript/webextension-api-compat) 1 related asset
asset rules/javascript/webextension-api.js 458 KiB [compared for emit] (name: rules/javascript/webextension-api) 1 related asset
asset rules/javascript/deprecated-entities.js 85.3 KiB [compared for emit] (name: rules/javascript/deprecated-entities) 1 related asset
asset rules/javascript/global-require-arg.js 84.9 KiB [compared for emit] (name: rules/javascript/global-require-arg) 1 related asset
asset rules/javascript/opendialog-remote-uri.js 84.7 KiB [compared for emit] (name: rules/javascript/opendialog-remote-uri) 1 related asset
asset rules/javascript/opendialog-nonlit-uri.js 84.7 KiB [compared for emit] (name: rules/javascript/opendialog-nonlit-uri) 1 related asset
asset rules/javascript/content-scripts-file-absent.js 45.3 KiB [compared for emit] (name: rules/javascript/content-scripts-file-absent) 1 related asset
asset rules/javascript/event-listener-fourth.js 43.5 KiB [compared for emit] (name: rules/javascript/event-listener-fourth) 1 related asset
runtime modules 793 bytes 4 modules
cacheable modules 735 KiB
  modules by path ./src/schema/ 450 KiB 64 modules
  modules by path ./src/rules/ 22.3 KiB 16 modules
  modules by path ./src/*.js 95.5 KiB 11 modules
  modules by path ./src/messages/*.js 52.5 KiB 9 modules
  modules by path ./src/scanners/*.js 50.8 KiB 8 modules
[...]

PR #3626, opened 2021-03-18
chore: Fix 'Cannot use import statement' error raised when running firefox-schema-import script

rpl:
I did notice this today while looking into some future changes to the addons-linter schemas import related to the ongoing Manifest V3 work on the Firefox side.

We did break it in #3603 by removing that `require('@babel/register')` line from `scripts/firefox-import-schema`.

I did look into other ways to fix it without having to bring back @babel/register but I didn't liked any of them (either because it wasn't working as I initially expected and needed more digging or it was not going looking that great)

Honestly I'm not sure that is worth to spend too much time on figuring out another strategy just for a dev dependency (I would feel different if it was a dependency also for users and dependents packages) used by a script that only the developers of the project are going to use.

In the short run I propose to pick the easiest option, which is bring back the @babel/register devDependency and add back the `require('@babel/register')` call in `scripts/firefox-import-schema`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,7 @@
         "@babel/plugin-proposal-numeric-separator": "7.12.13",
         "@babel/plugin-proposal-throw-expressions": "7.12.13",
         "@babel/preset-env": "7.13.10",
+        "@babel/register": "7.13.8",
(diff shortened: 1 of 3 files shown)


willdurand (approved):
LGTM, thanks

PR #3627, opened 2021-03-18
chore: Support importing Firefox API schema from a local mozilla-central directory

rpl:
A small addition to firefox-schemas-import, meant to be used to make it easier for us to evaluate the impacts of changes to the schema format while they are still in work and not landed yet in Firefox.

I tested it locally by running the script on a local mozilla-central clone with some phabricator revisions still in review applied on it:
```
./scripts/firefox-schema-import ~/mozlab/projects/mc-reviews
```

I haven't added any additional test coverage for the addition, mainly because we are the only consumer of it and if we use it often enough we may just notice and fix it right away if something breaks from time to time, and we may avoid to increase the time needed to the CI jobs to complete.

### DIFF
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -599,19 +599,9 @@ async function fetchSchemasFromDir({ inputPath, outputPath }) {
   );
 
   const promiseLoadSchemaFrom = (baseDir) =>
-    new Promise((resolve, reject) => {
-      fs.readdir(baseDir, { encoding: 'utf-8' }, (err, files) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(
-            files
-              .filter((fp) => fp.endsWith('.json'))
-              .map((fp) => path.join(baseDir, fp))
-          );
-        }
-      });
-    });
+    fs.promises.readdir(baseDir, { encoding: 'utf-8' }).then((files) => 
+      files.filter((fp) => fp.endsWith('.json'))
+           .map((fp) => path.join(baseDir, fp)));
 
   return Promise.all([
     promiseLoadSchemaFrom(toolkitSchemasBaseDir),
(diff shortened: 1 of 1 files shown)


rpl:
Let's see if codecov is fine with it or if it will complains about coverage trend going down :-P

willdurand:
> I haven't added any additional test coverage for the addition, mainly because we are the only consumer of it and if we use it often enough we may just notice and fix it right away if something breaks from time to time, and we may avoid to increase the time needed to the CI jobs to complete.

+1

src/schema/firefox-schemas-import.js
```diff
@@ -578,7 +578,64 @@ inner.isBrowserSchema = (_path) => {
...
+    inputPath,
+    'browser',
+    'components',
+    'extensions',
+    'schemas'
+  );
+
+  const promiseLoadSchemaFrom = (baseDir) =>
+    new Promise((resolve, reject) => {
+      fs.readdir(baseDir, { encoding: 'utf-8' }, (err, files) => {
```
willdurand, line 603: `fs.promises.readdir`?

src/schema/firefox-schemas-import.js
```diff
@@ -578,7 +578,58 @@ inner.isBrowserSchema = (_path) => {
...
+    promiseLoadSchemaFrom(toolkitSchemasBaseDir),
+    promiseLoadSchemaFrom(browserSchemasBaseDir),
+  ]).then((results) => {
+    const allFiles = results.flat();
+    const writeStreamsPromises = allFiles.map((filePath) => {
+      const outFilePath = path.join(outputPath, path.basename(filePath));
+      const destFileStream = fs.createWriteStream(outFilePath);
+      const readStream = fs.createReadStream(filePath);
+      readStream.pipe(destFileStream);
+      // Ensure we're closing all the creted write streams before resolving.
```
willdurand, line 620:
```suggestion
      // Ensure we're closing all the created write streams before resolving.
```

willdurand (approved):
I haven't tried this patch locally but it looks good.

rpl:
> I haven't tried this patch locally but it looks good.

Thanks Will! 

(I'm using these changes locally while preparing more changes to deal with the upcoming additional schema changes related to the manifest_version 3 work)

PR #3633, opened 2021-03-22
feat: experimental manifest v3 schema annotations

### LINKED ISSUE #3290: Remove MANIFEST_FIELD_UNSUPPORTED validation error on background.service_worker once supported and enabled on the Firefox side

As part of #3289 we are making sure that extensions using the `background.service_worker` manifest property are marked as invalid (and a new `MANIFEST_FIELD_UNSUPPORTED` linting error code is associated to the manifest validation error related to that), because the property has been added to the API schemas but it is not yet fully supported by Firefox.

As part of this follow up issue we should, once the feature related to this manifest property is going to be actually supported on the Firefox side:
- remove that validation error added on `background.service_worker` manifest property
- replace it with validating that the worker script is part of the xpi package, as we are already doing for the `background.page` and `background.scripts` properties).
- don't mark the manifest as invalid unconditionally when this property is being used

rpl:
This PR includes an initial draft for the changes we'll need on the linter side for the new API schema annotations related to the `manifest_version: 3` addons:

- add ajv keywords for the new `min_manifest_version`/`max_manifest_version` properties
- `min_manifest_version` and `max_manifest_version` and manifest.json file validation: using manifest properties unsupported for the given addon's manifest_version should make the manifest file invalid as expected
- `min_manifest_version` and `max_manifest_version` and javascript linting rules: we should be collecting the expected linting warning if we detect calls to APIs unsupported on the given addon's manifest_version
- manual and automated testing:
  - add a new CLI option to the addons-linter (e.g. one named `--enable-manifest-version3`) to make sure we don't yet accept `manifest_version: 3` on AMO (but allow developers to run the addons-linter on addons using the new manifest format, to make it easier to run locally experiments on the new manifest format)
  - add a new option to the ManifestJSONParser class, to make it easier to test the new behaviors while the new manifest format isn't yet enabled by default
- also closes #3290 as a side effect of the preparation work in this PR to allow validating mv3 extensions using new command line option, which means that:
   - when this PR lands, we would still not allow the manifest background service worker in a manifest_version 2 extension, and we would allow manifest_version 3 extensions to be submitted to AMO
   - when we will be ready to accept manifest_version 3 extensions submitted to AMO, in the addons-linter we should just need to increase the max_manifest_version set in the yargs-options.js file

NOTE:
- [x] propagate min/max manifest version from top level API namespace down to the extended manifest definitions (e.g. setting "max_manifest_version" to the two top level entries in the `pageAction.json` JSONSchema should make both the `pageAction` API methods/events and the `page_action` manifest property considered unsupported for `manifest_version: 3` extensions, and still valid and supported for `manifest_version: 2` extensions)
- [x] evaluate to change the CLI option to `--manifest-version=N` or `--max-manifest-version=N`
- [x] replace the patch that contains changes related to importing API schema, the current version of those schema is experimental and they have not been landed yet on mozilla-central, they are still being discussed and they may still change based on the review passes (and the assessment of the changes to the linter from this PR)
- [x] double-check the test coverage
- [x] reorganize PR to make easier to follow  and review (ideally splitting it in two or three commits, e.g. one for the manifest schema validation and one for the changes related to the eslint rules and one for the changes to the schema import) 
- [ ] keep in sync with the changes on the Firefox side:
  - D105821 Bug 1693403 support manifest versioning in schema
  - D108270 Bug 1697059 require favicon to be in the extension xpi 
  - D108265 Bug 1687762 version a few APIs removed in MV3 
  - D105846 Bug 1696043 convert schema format for MV3 content_security_policy 
  - D107310 Bug 1696580 Add support for object form of web_accessible_resources 
  - D107746 Bug 1697334 implement matches property in web_accessible_resources


src/schema/validator.js
```diff
@@ -94,9 +152,26 @@ const _validateAddon = validator.compile({
...
+      },
+    },
+  },
+});
+
+export const validateAddon = (
+  manifestData,
+  { enableManifestVersion3 } = {}
+) => {
+  const _validate = enableManifestVersion3 ? _validateAddonMV3 : _validateAddon;
```
rpl, line 172: Note to self: this isn't going to work well enough, because we can only have one schema definition with id "manifest" but using a different one makes the merged extended definitions coming from the other APIs not being applied in the manifest v3 validation.

I've started to rework this and I'll update this PR asap.

willdurand:
>     * add a new CLI option to the addons-linter (e.g. one named `--enable-manifest-version3`) to make sure we don't yet accept `manifest_version: 3` on AMO (but allow developers to run the addons-linter on addons using the new manifest format, to make it easier to run locally experiments on the new manifest format)

I was thinking that we could use a slightly different CLI option: `--manifest-version`, which would default to `2` and allow `2` or `3`. My idea behind that would be to build everything else based on that in order to easily add another manifest version in the future (as opposed to it's either MV2 or MV3).

rpl:
> > ```
> > * add a new CLI option to the addons-linter (e.g. one named `--enable-manifest-version3`) to make sure we don't yet accept `manifest_version: 3` on AMO (but allow developers to run the addons-linter on addons using the new manifest format, to make it easier to run locally experiments on the new manifest format)
> > ```
> 
> I was thinking that we could use a slightly different CLI option: `--manifest-version`, which would default to `2` and allow `2` or `3`. My idea behind that would be to build everything else based on that in order to easily add another manifest version in the future (as opposed to it's either MV2 or MV3). But feel free to ignore this comment :p

:+1: That is definitely reasonable, and I was also thinking that changing that cli option from boolean to number would make it more useful in the long run.

What I was thinking of is actually to opt for an `--max-manifest-version=N` CLI option, e.g. because in the future we may introduce another `--min-manifest-version=N` CLI option once we are going to prepare to sunset the manifest version 2 extensions.

At the moment I decided to defer changing that for now, while I'm still exploring the various scenarios (and assess we have to do as part of the JSONSchema imports and what should be done at runtime).

I'm going to add it to the TODO list in the pull request description in the meantime, so that I'll not forget to look into that once I'm happy enough with everything else in this PR.

src/schema/validator.js
rpl: Refactored this in 413f14f and verified that this strategy does work as expected (while verifying that the page_action manifest property mv3 deprecation). I'll add some more test coverage to explicitly cover the part that wasn't working in the previous approach (which wasn't making any test to fail yet).

rpl:
I'm not sure why codecov doesn't want to update the code coverage comment after it did added that comment when I opened this PR (in theory codecov should be updating the comment when I push more commits in this branch).

Anyway, with the last additions to the test, the coverage report is back to green: https://app.codecov.io/gh/mozilla/addons-linter/compare/3633

rpl:
@willdurand I've just force pushed on this branch a reorganized version of this PR, spitted into 4 commits:

- In 1403086 there are the changes related to the manifest parsing and validation (which is basically the main purpose of the ajv-based JSONSchema validator):
  - refactored src/schema/validator to be able to use different schema data (which is used to be able to validate manifest_version 2 and manifest_version 3 extensions manifests, but also to make it easier to unit test the schema validation without making the test to depend on actual usage in the schema data imported from Firefox)
  - added two new ajv custom keywords for min_manifest_version and max_manifest_version
  - changes to the manifest parser's errorLookup to generate the linting warning and error messages (which will have to include the version range from the validation error reported by the min/max_manifest_version custom keywords)
  - changes to the Collector's isDuplicateMessage to make sure we report linting messages related to the same dataPath if they have different a linter message code
- In c529ceb there are the changes (and test coverage using jest mocks) to the eslint webextensions api rules to report linting errors on API calls related to API unsupported in `manifest_version: 3` exensions
- In 0803d50 there are the changes to the firefox schema import, in particular to propagate a min/max_manifest_version property from a top level api namespace definition that extends the manifest types to the api specific manifest extended types definitions (this isn't needed for the eslint rules, but it is needed to make ajv-based validator custom keywords to be triggered by the manifest fields merged into the manifest form the api-specific json schemas)
- In e954868 there are the changes to the rules.md file (I kept this separate in case I have to change the message codes based on what we will agree on as part of the review).

Hopefully these 4 separate commits are self-contained enough to let you more easily see the role of those changes, obviously I still expect that you may have question about parts that may not be clear enough yet and/or about details we may have to discuss a bit to get to an agreement and tweak the changes accordingly.

Let me know if you think that having a chat over a synchronous channel may be helpful.

docs/rules.md
```diff
@@ -100,7 +100,8 @@ Rules are sorted by severity.
 | `MANIFEST_FIELD_REQUIRED`                               | error    | A required field is missing                                                                     |
 | `MANIFEST_FIELD_INVALID`                                | error    | A field is invalid                                                                              |
 | `MANIFEST_FIELD_DEPRECATED`                             | error    | A field is deprecated                                                                           |
-| `MANIFEST_FIELD_UNSUPPORTED`                            | error    | A manifest field is not supported                                                               |
+| `MANIFEST_FIELD_UNSUPPORTED`                            | error    | A manifest field is not supported (or not supported for the extension manifest_version)         |
+| `MANIFEST_PERM_UNSUPPORTED`                             | error    | A manifest permission is not supported for the extension manifest_version                       |
```
willdurand, line 104: nit: maybe we can have `MANIFEST_PERMISSION_UNSUPPORTED`

src/const.js
```diff
@@ -73,7 +73,10 @@ export const LOCAL_PROTOCOLS = ['chrome:', 'resource:'];
 
 export const MANIFEST_JSON = 'manifest.json';
 
-export const VALID_MANIFEST_VERSION = 2;
+export const MANIFEST_VERSION_VALID = 2;
```
willdurand, line 76: It's the top of the patch so I will probably find out later but this constant does not speak to me. The other constants make sense but I don't know what that one should do.

If it is about validation and we want to introduce new manifest versions, I would have expected an array instead. OTOH we have min/max already...

src/linter.js
```diff
@@ -263,8 +263,12 @@ export default class Linter {
       _log.info('Retrieving metadata from manifest.json');
       const json = await this.io.getFileAsString(constants.MANIFEST_JSON);
       const manifestParser = new ManifestJSONParser(json, this.collector, {
-        selfHosted: this.config.selfHosted,
         io: this.io,
+        selfHosted: this.config.selfHosted,
```
willdurand, line 267: Should that one go into `validatorOptions` too? Not sure what `validatorOptions` is yet, though.

src/messages/javascript.js
```diff
@@ -126,6 +126,13 @@ export const UNSUPPORTED_API = {
   description: i18n._('This API has not been implemented by Firefox.'),
 };
 
+export const REMOVED_MV2_API = {
+  code: 'UNSUPPORTED_API',
+  message: null,
+  messageFormat: i18n._('"{{api}}" has been removed in manifest_version 3'),
```
willdurand, line 132: nit: if this is a user message, maybe we should write "Manifest Version" instead of `manifest_version`

src/messages/manifestjson.js
```diff
@@ -22,13 +22,39 @@ export const MANIFEST_FIELD_INVALID = {
...
+    description: message,
+    file: MANIFEST_JSON,
+  };
+}
+
+export const MANIFEST_PERM_UNSUPPORTED = 'MANIFEST_PERM_UNSUPPORTED';
+export function manifestPermUnsupported(permissionName, error) {
+  const versionRange = errorParamsToUnsupportedVersionRange(error.params);
+  const messageTmpl = versionRange
+    ? i18n._(oneLine`/%(fieldName)s: "%(permissionName)s" is not supported in
```
willdurand, line 45: Why is the forward slash needed in those messages?

src/parsers/manifestjson.js
```diff
@@ -281,13 +300,13 @@ export default class ManifestJSONParser extends JSONParser {
...
-      /^\/(permissions|optional_permissions)\/([\d+])/
-    );
+    const match = error.dataPath.match(PERMS_DATAPATH_REGEX);
+
     if (
       match &&
       baseObject.code !== messages.MANIFEST_BAD_PERMISSION.code &&
-      baseObject.code !== messages.MANIFEST_BAD_OPTIONAL_PERMISSION.code
+      baseObject.code !== messages.MANIFEST_BAD_OPTIONAL_PERMISSION.code &&
+      baseObject.code !== messages.MANIFEST_PERM_UNSUPPORTED
```
willdurand, line 309: shouldn't that be `messages.MANIFEST_PERM_UNSUPPORTED.code`?

src/parsers/manifestjson.js
```diff
@@ -314,7 +334,8 @@ export default class ManifestJSONParser extends JSONParser {
       validate = validateDictionary;
     }
 
-    this.isValid = validate(this.parsedJSON);
+    this.isValid = validate(this.parsedJSON, this.validatorOptions);
```
willdurand, line 337: Ha, I see now. So `validatorOptions` refers to the schema validator, right? Maybe the name should include "schema" somehow then.

src/parsers/manifestjson.js
```diff
@@ -373,14 +394,12 @@ export default class ManifestJSONParser extends JSONParser {
         );
       }
       if (this.parsedJSON.background.service_worker) {
-        // TODO: replace this with a call to this.validateFileExistsInPackage
-        // once background.service_worker is supported on the Firefox side and
-        // we can accept it on AMO submissions
-        // (https://github.com/mozilla/addons-linter/issues/3290).
```
willdurand, line 379: 🏆

src/schema/browser-apis.js
```diff
@@ -40,16 +46,111 @@ function getArrayProperty(schema, property) {
...
+  const schemaItem =
+    getObjectProperty(schema, property) || getArrayProperty(schema, property);
+
+  let ns_min_mv = schema && schema.min_manifest_version;
+  ns_min_mv = ns_min_mv == null ? MANIFEST_VERSION_MIN : ns_min_mv;
+
+  let prop_min_mv = schemaItem && schemaItem.min_manifest_version;
+  prop_min_mv = prop_min_mv == null ? MANIFEST_VERSION_MIN : prop_min_mv;
+
+  // Return the higher min_manifest_version value between the one set at
```
willdurand, line 93: nit: highest?

src/schema/browser-apis.js
```diff
@@ -59,18 +160,47 @@ export function isDeprecatedApi(namespace, property) {
...
+  );
+  if (
+    manifestVersion < min_manifest_version ||
+    manifestVersion > max_manifest_version
+  ) {
+    // Consider the API unsupported if its schema has a min_manifest_version
+    // greater then the extension manifest version or a max_manifest_version
+    // lower than the extension manifest version.
+    return false;
+  }
```
willdurand, line 203: This whole new addition looks similar to the one above (in `isDeprecatedApi()`).

src/schema/browser-apis.js
```diff
@@ -59,18 +160,47 @@ export function isDeprecatedApi(namespace, property) {
...
     getObjectProperty(schema, property) || getArrayProperty(schema, property);
 
+  const manifestVersion = getManifestVersion(addonMetadata);
+  const min_manifest_version = getMinManifestVersion(
+    namespace,
+    property,
+    addonMetadata,
+    apiSchemas
+  );
+  const max_manifest_version = getMinManifestVersion(
```
willdurand, line 189: Shouldn't that be `getMaxManifestVersion()`?

src/schema/firefox-schemas-import.js
```diff
@@ -299,12 +299,55 @@ function rewriteExtendRefs(definition, namespace, types) {
   return definition;
 }
 
+// This methods does propagates a top level supported manifest version
```
willdurand, line 302: nit: `does propagate` (I think)

src/schema/validator.js
```diff
@@ -19,183 +23,391 @@ import {
...
 
-const _validateAddon = validator.compile({
-  ...schemaObject,
-  id: 'manifest',
-  $ref: '#/types/WebExtensionManifest',
-});
+export class SchemaValidator {
+  // TODO: add jsdoc with the expected validator options.
+  constructor(validatorOptions) {
+    this._options = validatorOptions;
```
willdurand, line 40: nit: do we need underscore prefixes?

src/schema/validator.js
```diff
@@ -19,183 +23,391 @@ import {
...
+      const { _validator, _options } = this;
+      this._addonValidator = this._compileAddonValidator(_validator, _options);
+    }
+
+    return this._addonValidator;
+  }
+
+  get schemas() {
+    return this._options?.schemas ?? schemas;
+  }
```
willdurand, line 88: fancy new syntax 😎

src/utils.js
```diff
@@ -13,6 +13,20 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
...
+  if (min_manifest_version != null || max_manifest_version != null) {
+    return [
+      min_manifest_version ? `< ${min_manifest_version}` : undefined,
+      max_manifest_version ? `> ${max_manifest_version}` : undefined,
+    ]
+      .filter((e) => e !== undefined)
+      .join(' or ');
+  }
+
+  return '';
```
willdurand, line 27: This should probably return `null` even if it does not impact the condition (in one of the files above).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -213,7 +213,7 @@ describe('ManifestJSONParser', () => {
       );
       expect(manifestJSONParser.isValid).toEqual(true);
       const metadata = manifestJSONParser.getMetadata();
-      expect(metadata.manifestVersion).toEqual(VALID_MANIFEST_VERSION);
+      expect(metadata.manifestVersion).toEqual(MANIFEST_VERSION_VALID);
```
willdurand, line 216: Ha. I see now, but I am still confused. The previous constant reads a bit better but I don't there is a single value that is valid these days (or soon). Shouldn't we have a list of valid manifest versions? And maybe a default one?

tests/unit/schema/helpers.js
```diff
@@ -8,3 +10,134 @@ export const validManifest = {
...
+                  { "$ref": "manifest#/types/ExtensionURL" },
+                  { min_manifest_version: 3 },
+                ]
+              },
+            },
+            required: ["service_worker"],
+          },
+        ],
+      },
+      */
```
willdurand, line 41: If this stays in the file, we should reformat it a bit and maybe add a comment above it.

tests/unit/schema/test.browser-apis.js
```diff
@@ -22,6 +35,63 @@ describe('browserApis', () => {
...
+
+      // Unsupported api method in mv3.
+      fakeSchemas = {
+        [fakeNs]: {
+          functions: [{ name: fakeProp, max_manifest_version: 2 }],
+        },
+      };
+      expect(
+        hasBrowserApi(fakeNs, fakeProp, fakeMetadataV3, fakeSchemas)
+      ).toBeFalsy();
```
willdurand, line 90: nit: I'd rather use `.toEqual(false)` or `.toEqual(true)` but I don't mind (`toEqual()` only allows the value that is passed to it)

willdurand (commented):
I have a few comments and questions but nothing major. This looks good to me overall.

I am wondering how some of the `hasBrowserApi()` tests pass, though, because I think there is a mistake (but I am not sure).

Let's have a round of questions/answers before final approval.

willdurand:
Some more info. I tried this patch locally with [amo-info](https://addons.mozilla.org/en-US/firefox/addon/amo-info/) version [1.15.0](https://addons.mozilla.org/firefox/downloads/file/3766843/amo_info-1.15.0-fx.xpi).

```
$ ./bin/addons-linter ~/projects/mozilla/amo-info/dist-zip/amo-info-v1.15.0.zip
Validation Summary:

errors          0
notices         0
warnings        4

WARNINGS:

Code                              Message                                 Description                                                                            File             Line   Column
KEY_FIREFOX_ANDROID_UNSUPPORTE…   Manifest key not supported by the       "strict_min_version" requires Firefox for Android 59.0, which was released before      manifest.json
D_BY_MIN_VERSION                  specified minimum Firefox for Android   version 79 introduced support for "page_action.show_matches".
                                  version
DANGEROUS_EVAL                    The Function constructor is eval.       Evaluation of strings as code can lead to security vulnerabilities and performance     popup/popup.js   1      4578
                                                                          issues, even in the most innocuous of circumstances. Please avoid using `eval` and
                                                                          the `Function` constructor when at all possible.'
UNSAFE_VAR_ASSIGNMENT             Unsafe assignment to innerHTML          Due to both security and performance concerns, this may not be set using dynamic       popup/popup.js   7      45382
                                                                          values which have not been adequately sanitized. This can lead to security issues or
                                                                          fairly serious performance degradation.
DANGEROUS_EVAL                    The Function constructor is eval.       Evaluation of strings as code can lead to security vulnerabilities and performance     popup/popup.js   7      67419
                                                                          issues, even in the most innocuous of circumstances. Please avoid using `eval` and
                                                                          the `Function` constructor when at all possible.'
```

Then I edited the zip file to set `manifest_version` to `3`:

```
$ ./bin/addons-linter ~/projects/mozilla/amo-info/dist-zip/amo-info-v1.15.0.zip
Validation Summary:

errors          6
notices         0
warnings        7

ERRORS:

Code                         Message                                      Description                                                            File            Line   Column
JSON_INVALID                 "/manifest_version" should be <= 2           Your JSON file could not be parsed.
JSON_INVALID                 "/background/scripts" is an invalid          Your JSON file could not be parsed.
                             additional property
MANIFEST_FIELD_REQUIRED      "/background/page" is a required property    See https://mzl.la/1ZOhoEN (MDN Docs) for more information.            manifest.json
MANIFEST_FIELD_UNSUPPORTED   "/background" is in a format not supported   "/background" is in a format not supported in manifest versions > 2.   manifest.json
                             in manifest versions > 2.
MANIFEST_FIELD_REQUIRED      "/background/service_worker" is a required   See https://mzl.la/1ZOhoEN (MDN Docs) for more information.            manifest.json
                             property
JSON_INVALID                 "/background" is not a valid key or has      Your JSON file could not be parsed.
                             invalid extra properties
WARNINGS:

Code                              Message                                 Description                                                                            File             Line   Column
KEY_FIREFOX_ANDROID_UNSUPPORTE…   Manifest key not supported by the       "strict_min_version" requires Firefox for Android 59.0, which was released before      manifest.json
D_BY_MIN_VERSION                  specified minimum Firefox for Android   version 79 introduced support for "page_action.show_matches".
                                  version
UNSUPPORTED_API                   runtime.onMessage is not supported      This API has not been implemented by Firefox.                                          background.js    1      4924
DANGEROUS_EVAL                    The Function constructor is eval.       Evaluation of strings as code can lead to security vulnerabilities and performance     popup/popup.js   1      4578
                                                                          issues, even in the most innocuous of circumstances. Please avoid using `eval` and
                                                                          the `Function` constructor when at all possible.'
UNSAFE_VAR_ASSIGNMENT             Unsafe assignment to innerHTML          Due to both security and performance concerns, this may not be set using dynamic       popup/popup.js   7      45382
                                                                          values which have not been adequately sanitized. This can lead to security issues or
                                                                          fairly serious performance degradation.
DANGEROUS_EVAL                    The Function constructor is eval.       Evaluation of strings as code can lead to security vulnerabilities and performance     popup/popup.js   7      67419
                                                                          issues, even in the most innocuous of circumstances. Please avoid using `eval` and
                                                                          the `Function` constructor when at all possible.'
UNSUPPORTED_API                   tabs.query is not supported             This API has not been implemented by Firefox.                                          popup/popup.js   7      78030
UNSUPPORTED_API                   runtime.sendMessage is not supported    This API has not been implemented by Firefox.                                          popup/popup.js   7      78147
```

I see a lot more information, which makes sense but for `UNSUPPORTED_API`, I am wondering if the messages make sense.

Rob--W:
These "unsupported" warnings don't make sense and should be resolved.

rpl:
> These "unsupported" warnings don't make sense and should be resolved.

those unsupported warnings shouldn't be there, I suspect it is to the wrong call Will pointed in https://github.com/mozilla/addons-linter/pull/3633#discussion_r624943185 (and that does also suggest me that there may be a test case missing, or one that may have the wrong assertions). I'm Looking into it.

docs/rules.md
rpl: Sure, I think I opted for `MANIFEST_PERM_` mainly because it was shorter, but if we prefer to include PERMISSION for consistency with the other rule code related to permission, I'm fine with that.

src/const.js
rpl: @willdurand MANIFEST_VERSION_VALID was meant to go away (I originally renamed it from VALID_MANIFEST_VERSION to MANIFEST_VERSION_VALID for consistency with some of the other consts, but in a later refactoring I actually opted for MANIFEST_VERSION_DEFAULT + MANIFEST_VERSION_MIN / MANIFEST_VERSION_MAX:

- MANIFEST_VERSION_DEFAULT is the default we would use if we didn't got one explicitly and we need exactly one manifest version to refer to, e.g.
  - in src/schema/validator.js: if the current extension being validate does miss a manifest_version property, we will complete the validation considering the manifest_version as if it was set to MANIFEST_VERSION_DEFAULT (but this doesn't mean we wouldn't want for a missing manifest_version mandatory property, that would still happen as expected)
  - in src/schema/browser-api.js: same as for validator.js, we would validate the API calls as it the manifest_version missing was meant to be set to the default version (but we will still collect an error for the missing manifest_version property) 
- MANIFEST_VERSION_MIN and MANIFEST_VERSION_MAX are the default values for the `min/max_manifest_version` json schema definitions that may not have one on their own.

Locally I've just added a fixup patch that removes MANIFEST_VERSION_VALID and does add inline comments on these consts to point out what they mean / used for, that change will be pushed along the next update on this PR.

src/linter.js
rpl: @willdurand I was wondering if "that one" in this comment does refer to the consts from comment https://github.com/mozilla/addons-linter/pull/3633#discussion_r624936948, if that is the case, does https://github.com/mozilla/addons-linter/pull/3633#discussion_r625039218 provide you enough context to see why those consts are not explicitly used inside validatorOptions?

basically, is it clear enough why `MANIFEST_VERSION_MAX` and `getDefaultConfigValue('max_manifest_version')` are not the same thing? otherwise it may be worth to mention that more explicitly some where:
- `MANIFEST_VERSION_MAX` represents the `max_manifest_version` we would use for schema entries that do now have a max_manifest_version on their own (which means "we consider this API still valid at least up to this manifest version)
- `getDefaultConfigValue('max_manifest_version')` represents the maximum manifest_version that we will consider valid while validating an extension (by default is currently set to 2, and only extensions with `manifest_version: 2` would be considered valid for submission to AMO)

src/messages/javascript.js
rpl: @willdurand if for user message we mean that this message is meant for the extension developer, the answer is yes. I personally preferred `manifest_version` here because it was matching the name of the property in the manifest that the error want to point the attention of the developer to.

I'm not sure that 'Manifest Version' would always be read as we expect, e.g. some developer may read it as the "version" property in the Manifest (which is definitely not what this error means).

src/messages/manifestjson.js
rpl: @willdurand honestly I just kept using the same convention used elsewhere, the logic seems to be that `fieldName` is the path to the manifest property that the error refers to, and so as an example:

- if the error is related to the background scripts in the manifest, in the error we translate that into `"/background/scripts"`
- and if the error is related to a permission, in the error we translate that into `"/permissions"` or `"/optional_permissions"`

src/parsers/manifestjson.js
rpl: As for some other messages that are being created using an exported method, `MANIFEST_PERM_UNSUPPORTED` is a string and not an object, and there is a function paired with that rule code `const` named `manifestPermUnsupported` which is the one returning the object with the code property.

There are more that follows this slightly different convention, e.g.:
- `MANIFEST_CSP` and `manifestCsp`
- `MANIFEST_CSP_UNSAFE_EVAL` and `manifestCspUnsafeEval`
- `MANIFEST_ICON_NOT_FOUND` and `manifestIconMissing`

To be honest, I also find the mix of these two slightly different conventions quite confusing, but I think we should choose a convention that we like and use it for all of the rules not just newly ones added, I don't see much value in that.

Given how big this PR was already becoming, I opted to just follow the current conventions as is for the new ones added or changed in this PR.

src/parsers/manifestjson.js
rpl: I can definitely add `schema` in the name of the property, but to be fair the manifestjson parser is the only thing that does actually use the JSONSchema validator and so in the context of this particular class I do think that `this.validatorOptions` may as well be considered as the set of options passed to the related ajv validator instance and so I didn't bothered adding `scheme` to the name.

If you think that it may make it easier to understand what this property does, I'll rename the property (it should be just a couple of changes and so not really a bit deal to change it).

src/linter.js
willdurand: It's perfectly clear.

"that one" was referring to `selfHosted` actually but I found later that it was passed to `validator` and I suggested to rename `validatorOptions` to `schemaValidatorOptions`. It's a bit more verbose but it is also more explicit.

(I should maybe update some comments before submitting my reviews, I kinda want to give how I approached the review but I get that it can be confusing)

src/messages/javascript.js
willdurand: I think it's valid, maybe we should have both then?

```suggestion
  messageFormat: i18n._('"{{api}}" has been removed in Manifest Version 3 (`manifest_version` property)'),
```

src/messages/manifestjson.js
willdurand: Ok, that works for me. I am not too familar with the linter output tbh..

src/parsers/manifestjson.js
willdurand: ok

src/linter.js
rpl: ah, I see now, and yeah, and my willingness to rename `validatorOptions` into `schemaValidatorOptions` just got increased as a side effect :-)

> (I should maybe update some comments before submitting my reviews, I kinda want to give how I approached the review but I get that it can be confusing)

don't worry, I actually find that useful because it provides me additional hints about what wasn't immediately clear while reading the changes gradually without having the full picture yet (and push me to think into ways to make those points more immediately clear) and so I don't actually mind that.

src/messages/javascript.js
rpl: :+1: sgtm

src/schema/browser-apis.js
rpl: I confirm that this was the reason behind the unexpected UNSUPPORTED_API linter warnings that you did mention in https://github.com/mozilla/addons-linter/pull/3633#issuecomment-831129468 

As I was also guessing in https://github.com/mozilla/addons-linter/pull/3633#issuecomment-831140404 , there was also one missing test case or assertion to cover that scenario explicitly (I have applied both locally in a fixup commit that I'm going to push as part of the next update on this PR).

src/utils.js
rpl: My personal preference for this helper would be to keep the same return type in both cases, either when there is a range and when there isn't any.

At the moment the return type is always 'string' and so the caller can just interpolate the return value into a template string as is without bothering to convert the null into an empty string (`${null}` => 'null'`, on the contrary I want it to be an empty string when there is no range).

tests/unit/schema/helpers.js
rpl: ehm... that was actually a fragment of schema data that I kept around temporarily as a reference while I was working on some of those tests that use custom schema data, I should have removed it but I forgot and got into one of the patches. I'll remove it in the next update.

tests/unit/schema/test.browser-apis.js
rpl: that's a good point, I agree that a stronger assertion would be better (at least in most cases, at least the ones when we care the value to be really a boolean). I'll just change the one I added or changed directly for now (but I'm pretty sure there are many more that could also be matching this logic, we can take care of those separately).

src/utils.js
willdurand: :+1:

src/parsers/manifestjson.js
rpl: This reminded me that this PR should mention #3290 (and I think also close it, because those steps would become implicit when this lands).

I'm not marking this comment as resolved yet, because I want to double-check if there was any other reference to #3290 around the sources and see if there is anything left to do from what #3290 does currently describe (eventually filing a separate issue if necessary).

src/schema/browser-apis.js
rpl: @willdurand,
Yeah it is similar, but `isDeprecatedApi` does also have other conditions at the end of it.

In the meantime I have moved the shared check into its own helper function named `isInSupportedManifestVersionRange`, the two checks were similar but with small differences that were not an issue but also not needed, better to have one helper function that we make us sure we do use the exact same code when the check we want is exactly the same one.

As an additional side note, in one of the intermediate versions of this PR I was getting confused by how the `isDeprecatedApi` and `hasBrowserApi` has interacting with each other and I had to fix that, and I also added some inline comments hoping to make that a bit more clear.
To summatize the thing that was confusing me was:

- `hasBrowserApi` has an initial check that basically says: if an API is temporary or is deprecated, then we "have" the API and `hasBrowserApi` should return true to avoid to warn about the same API twice

Tying to clarify that for my "future ourselves" I have added the following two inline comments:

```
export function isDeprecatedApi(...) {
  // If the API has been removed in a certain manifest version, or only
  // available starting from a manifest version, then it is unsupported
  // and not deprecated.
  ...
}

export function hasBrowserApi(...) {
  // We "have" the API if it's deprecated or temporary so we don't double warn.
  ....
  // We don't have the API if the extension manifest_version is outside of the 
  // manifest_version range where the API is actually supported for.
  ....}
  // Or the schema entry for the API has an unsupported property set to true.
  ...
}
```

I'll push an update asap, but I wanted to mention this to you, to double-check explicitly with you if for another engineer perspective (you got extracted as the "lucky one" :-P sorry ;-)) they are of any help.

willdurand: Given that I didn't know about these two functions, it's super helpful!

src/schema/validator.js
rpl: I can, it is more of a warning for us to not access or change that from outside the class (me looking forward to start using the standardized private class fields instead of this kind of "note to self" conventions :-)).

we have a bunch of similar cases like this in this repo (I didn't look to closely, I have just grepped a bit using `egrep "\._" src -r`), if you feel strongly about it I'm definitely ok on removing the _ from this class, but it may as well be reasonable to keep following the same convention we already used elsewhere in this project and then change it consistently across the entire code base when we decided how we like to handle this in the long run (e.g. from a quick look it seems that nodejs 12 may have introduced support for the private fields, if that is enabled by default without any cli flag we may look into migrating to that).

But... I did notice a TODO here related to describing what validatorOptions is expected to contain, I'll take care of that in the meantime.

willdurand: I don't really care. I think I tried to ban that convention over time but not in all repos yet obviously :p If this convention exists in this repo, and you want to keep it here, it's fine.

src/parsers/manifestjson.js
rpl: I double-checked and the same issue doesn't seem to be referenced anywhere else.

rpl:
@willdurand hopefully I didn't miss to apply any of the changes discussed in the previous review round (I actually re-looked to them while marking as resolved one by one, but I may still have missed something), this should be ready for another review pass (I squashed the updates in the patch that introduced each change, hoping to re-evaluate if we can land them separately once the review is complete).

willdurand:
> @willdurand hopefully I didn't miss to apply any of the changes discussed in the previous review round (I actually re-looked to them while marking as resolved one by one, but I may still have missed something), this should be ready for another review pass (I squashed the updates in the patch that introduced each change, hoping to re-evaluate if we can land them separately once the review is complete).

For a user POV, that looks better:

```
$  ./bin/addons-linter amo_info-1.15.0-fx-with-mv3.xpi
Validation Summary:

errors          6
notices         0
warnings        5

ERRORS:

Code                         Message                                               Description                                                            File            Line   Column
JSON_INVALID                 "/manifest_version" should be <= 2                    Your JSON file could not be parsed.
JSON_INVALID                 "/background/scripts" is an invalid additional        Your JSON file could not be parsed.
                             property
MANIFEST_FIELD_REQUIRED      "/background/page" is a required property             See https://mzl.la/1ZOhoEN (MDN Docs) for more information.            manifest.json
MANIFEST_FIELD_UNSUPPORTED   "/background" is in a format not supported in         "/background" is in a format not supported in manifest versions > 2.   manifest.json
                             manifest versions > 2.
MANIFEST_FIELD_REQUIRED      "/background/service_worker" is a required property   See https://mzl.la/1ZOhoEN (MDN Docs) for more information.            manifest.json
JSON_INVALID                 "/background" is not a valid key or has invalid       Your JSON file could not be parsed.
                             extra properties
WARNINGS:

Code                              Message                                               Description                                                                                               File                   Line   Column
KEY_FIREFOX_ANDROID_UNSUPPORTE…   Manifest key not supported by the specified minimum   "strict_min_version" requires Firefox for Android 59.0, which was released before version 79 introduced   manifest.json
D_BY_MIN_VERSION                  Firefox for Android version                           support for "page_action.show_matches".
ALREADY_SIGNED                    Package already signed                                Add-ons which are already signed will be re-signed when published on AMO. This will replace any           META-INF/manifest.mf
                                                                                        existing signatures on the add-on.
DANGEROUS_EVAL                    The Function constructor is eval.                     Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the    popup/popup.js         1      4578
                                                                                        most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all
                                                                                        possible.'
UNSAFE_VAR_ASSIGNMENT             Unsafe assignment to innerHTML                        Due to both security and performance concerns, this may not be set using dynamic values which have not    popup/popup.js         7      45382
                                                                                        been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
DANGEROUS_EVAL                    The Function constructor is eval.                     Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the    popup/popup.js         7      67419
                                                                                        most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all
                                                                                        possible.'
```

willdurand:
@rpl when we set `--min-manifest-version=3` only, we get a JSON error:

```
$ ./bin/addons-linter --min-manifest-version=3 amo_info-1.15.0-fx.xpi
Validation Summary:

errors          6
notices         0
warnings        5

ERRORS:

Code                         Message                                               Description                                                            File            Line   Column
JSON_INVALID                 "/manifest_version" should be <= 2                    Your JSON file could not be parsed.
```

Given that passing `--max-manifest-version=3` alone or in addition to `--min-manifest-version=3` do not output this error, I am thinking that there might be a problem with the min version. I haven't checked yet but I think we should always make sure the `max` version is greater than or equal to the `min` version.

WDYT?

rpl:
> @rpl when we set `--min-manifest-version=3` only, we get a JSON error:
> 
> ```
> $ ./bin/addons-linter --min-manifest-version=3 amo_info-1.15.0-fx.xpi
> Validation Summary:
> 
> errors          6
> notices         0
> warnings        5
> 
> ERRORS:
> 
> Code                         Message                                               Description                                                            File            Line   Column
> JSON_INVALID                 "/manifest_version" should be <= 2                    Your JSON file could not be parsed.
> ```
> 
> Given that passing `--max-manifest-version=3` alone or in addition to `--min-manifest-version=3` do not output this error, I am thinking that there might be a problem with the min version. I haven't checked yet but I think we should always make sure the `max` version is greater than or equal to the `min` version.
> 
> WDYT?

@willdurand

I do agree, it is not a big deal right now, but in the long run it would be good to explicitly show an error if a developer is using these new command line options and the version range requested is invalid, otherwise they may be get confused by the unexpected error they would get in the validation summary.

The annoying part is just find a good place to validate that part of the config got from the cli options and show the error to the user in a nice form, I have drafted the change (didn't add explicit coverage for it yet, that's why codecov is complaining at the moment) and this is what it would log when the version range requested is invalid:

```
$ ./bin/addons-linter --min-manifest-version=3 tmp/amo_info
Invalid manifest version range requested: --min-manifest-version (currently set to 3) should not be greater than --max-manifest-version (currently set to 2).
```

(+ addons-linter to exit the process with an error code).

In d3abfab the check is added in two places:
- src/linter.js - to validate the config and produce a nicer error for the user
- src/schema/validator.js - one more check near by the code that creates the validator instance in case that part may be used on its own (and so the check in src/linter.js not actually hit)

Honestly I don't like that much to call process.exit from inside the Linter class, it is done also in some other methods of that class but it is technically wrong if the addons-linter is used as a dependency (not a super big deal for web-ext because web-ext doesn't do anything with the results got from the addons-linter and the entire web-ext process is expected to exit anyway when the linter call is done, but I find it annoying when a library does exit the process and kind of a "design issue" for a library).

willdurand:
> Honestly I don't like that much to call process.exit from inside the Linter class, it is done also in some other methods of that class but it is technically wrong if the addons-linter is used as a dependency (not a super big deal for web-ext because web-ext doesn't do anything with the results got from the addons-linter and the entire web-ext process is expected to exit anyway when the linter call is done, but I find it annoying when a library does exit the process and kind of a "design issue" for a library).

Could we `throw` in the `Linter` class as well?

rpl:
> > Honestly I don't like that much to call process.exit from inside the Linter class, it is done also in some other methods of that class but it is technically wrong if the addons-linter is used as a dependency (not a super big deal for web-ext because web-ext doesn't do anything with the results got from the addons-linter and the entire web-ext process is expected to exit anyway when the linter call is done, but I find it annoying when a library does exit the process and kind of a "design issue" for a library).
> 
> Could we `throw` in the `Linter` class as well?

@willdurand yes, I was also thinking of something like that. I think that the easiest way to make it slightly better could be to just throw from the Linter class, then catch the rejection from `bin/addons-linter` and turn it into a nicer message, but include the stack trace from the exceptions / rejections when requested (e.g. it would be nice if we could tight that behavior to the --verbose cli option passed on the command line).

In web-ext we do throw using a specific `Error` subclass named `UserError` (or something like that) when we want the error to be caught and then logged without the stack trace, for any other errors we just let them to be logged with all the details (included the stack trace) because they are considered unexpected.

I may also introduce something like that for this error (and we may then extend the same strategy to some other similar scenarios that I'm sure already exist in the linter and not handled in the same way yet, as further polishing follow ups).

wdyt?

rpl:
@willdurand I've updated the last patch (now 8e1ab1c) to use the approach described in https://github.com/mozilla/addons-linter/pull/3633#issuecomment-831979448:
- from src/linter.js we validate the config and throw an instance of AddonsLinterUserError if the manifest version range requested is invalid
- from bin/addons-linter we check if `Linter` instance `run` method did reject with an AddonsLinterUserError error, and we print the error message without a stack if the linter was executed without the `--stack` option (otherwise we just rethrow the error object and the logged error will include the stack too)

I'm going to look into covering these additional changes with tests, but in the meantime feel free to add a comment to let me know what do you think about this new approach.

rpl:
In bceaf09 the updated version of the last patch with a couple of additional tests included.

To be fair I would guess that these new tests are enough to make codecov happy, but we may want to add one more integration test to explicitly test the expected behavior when the invalid manifest version range are passed over the command line (because bin/addons-linter isn't covered only by integration tests and it doesn't contribute to the code coverage reports).

bin/addons-linter
```diff
@@ -7,23 +7,39 @@ global.appRoot = path.relative(process.cwd(), absoluteAppRoot);
...
+    }
 
-  throw err;
-}
+    // Print the error message without the stack if it is a user error
+    // and --stack cli option wasn't part of the cli option passed to the
+    // linter.
+    if (err.name === 'AddonsLinterUserError' && !instance.config.stack) {
+      console.error(err.message);
+      process.exitCode = 1;
```
willdurand, line 39: Maybe add a different exit code here like 2 or 127.

willdurand: Also, question: why do we use `process.exitCode` and not `process.exit(1)`?

src/utils.js
```diff
@@ -13,6 +13,26 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
 
 const SOURCE_MAP_RE = new RegExp(/\/\/[#@]\s(source(?:Mapping)?URL)=\s*(\S+)/);
 
+export class AddonsLinterUserError extends Error {
```
willdurand, line 16: Maybe add some explanations about this special error.

rpl: +1 In 661f1cf I added an inline comment right above the class to provide some details about when this error class is supposed to be used and what kind of special handling is being done on this kind of errors.

bin/addons-linter
rpl: > Maybe add a different exit code here like 2 or 127.

sure sounds good to me too, In 661f1cf I did also add one small integration test while I was here ;-) 

> Also, question: why do we use `process.exitCode` and not `process.exit(1)`?

I was honestly asking myself the same thing, mostly out of curiosity.

My guess was that `process.exitCode + return` may easier to unit test (because the nodejs process would  be forced to exit and so in theory a unit test could require that module and then check that process.exitCode has been set to the expected value)... but I looked back in the history (because I wasn't really working on this repo at the time) but I didn't find a reasoning for preferring this particular form (the change has been introduced in #1492) and the change didn't land with a unit test (which doesn't exclude that maybe it was meant to be a follow up, but I wasn't able to track down if that was the case).

Well, unfortunately this brief historical research didn't got me any additional detail about the reasons behind that :-P, but in the end I don't mind that we use `process.exitCode + return` instead of `process.exit` :-D

rpl:
@willdurand this should be ready for one last review pass (only the last commit, 661f1cf, has been updated)

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -1445,4 +1445,147 @@ describe('firefox schema import', () => {
...
+              properties: {
+                propName: { type: 'object' },
+              },
+            },
+          },
+        },
+      },
+    ];
+
+    for (const testCase of testCases) {
```
willdurand, line 1582: When doing such things, I kinda like to have an extra assertion to make sure that all test cases will be taken into account:

```js
expect.assertions(testCases.length);
```

Otherwise, if you add `return` at the end of the loop code for example, tests would still pass.

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
-import { validManifest } from './helpers';
+import { getValidatorWithFakeSchema, validManifest } from './helpers';
+
+describe('getValidator', () => {
+  it('throws on invalid manifest version range options', () => {
+    expect(() => {
+      const validator = getValidator({
+        minManifestVersion: 3,
+        maxManifestVersion: 2,
+        // No need to cachethis validator instance.
```
willdurand, line 11:
```suggestion
        // No need to cache this validator instance.
```

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
+        forceNewValidatorInstance: true,
+      });
+      // Trigger the validator instance initialization,
```
willdurand, line 14:
```suggestion
      // Trigger the validator instance initialization
```

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
+                },
+              },
+            },
+          },
+        },
+      });
+
+      const isValidMV2 = validator.validateAddon({
+        ...validManifest,
+        // Add a properties that are expected to be triggering
```
willdurand, line 98:
```suggestion
        // Add properties that are expected to be triggering
```

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
+        action: {},
+      });
+
+      expect(validator.validateAddon.errors).toEqual(null);
+      expect(isValidMV3).toBe(true);
+    });
+
+    it('does report a validation error on max_manifest_version', () => {
+      // Create a validator with custom fake schema that
+      // contains an `page_action` manifest property only supported
```
willdurand, line 128:
```suggestion
      // contains a `page_action` manifest property only supported
```

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
+                },
+              },
+            },
+          },
+        },
+      });
+
+      const isValidMV2 = validator.validateAddon({
+        ...validManifest,
+        // Add a properties that are expected to be valid
```
willdurand, line 151:
```suggestion
        // Add properties that are expected to be valid
```

tests/unit/schema/test.schema.js
```diff
@@ -1,10 +1,179 @@
...
+        // with manifest_version 2.
+        manifest_version: 2,
+        page_action: {},
+      });
+      expect(validator.validateAddon.errors).toEqual(null);
+      expect(isValidMV2).toBe(true);
+
+      const isValidMV3 = validator.validateAddon({
+        ...validManifest,
+        // Add a properties that are expected to be triggering
```
willdurand, line 161:
```suggestion
        // Add properties that are expected to be triggering
```

tests/unit/test.utils.js
```diff
@@ -689,3 +691,44 @@ describe('isCompatible', () => {
...
+describe('errorParamsToUnsupportedVersionRange', () => {
+  const testCases = [
+    [{ min_manifest_version: 3 }, '< 3'],
+    [{ max_manifest_version: 2 }, '> 2'],
+    [
+      // This would be actually used anywhere because each
+      // schema validation would be only including one of them
+      // and generate separate validation errors.
+      { min_manifest_version: 2, max_manifest_version: 4 },
+      '< 2 or > 4',
```
willdurand, line 704: This is related to a comment above (about l10n)

tests/unit/test.utils.js
```diff
@@ -689,3 +691,44 @@ describe('isCompatible', () => {
...
+    ],
+    [{}, ''],
+    [null, ''],
+    [undefined, ''],
+  ];
+  for (const [params, expected] of testCases) {
```
willdurand, line 710: Each time we are not using `it.each()` for creating new test cases, we should use `expect.assertions()` I think.

tests/unit/test.utils.js
```diff
@@ -689,3 +691,44 @@ describe('isCompatible', () => {
...
+    const paramsStr = JSON.stringify(params);
+    it(`returns the expected version range string from ${paramsStr}`, () => {
+      expect(errorParamsToUnsupportedVersionRange(params)).toEqual(expected);
+    });
+  }
+});
+
+describe('AddonsLinterUserError', () => {
+  it('should be an instanceof Error', () => {
```
willdurand, line 719:
```suggestion
  it('should be an instance of Error', () => {
```

willdurand (approved):
r+wc

Thank you!

src/utils.js
```diff
@@ -13,6 +13,35 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
...
+
+export function errorParamsToUnsupportedVersionRange(errorParams) {
+  const { min_manifest_version, max_manifest_version } = errorParams || {};
+  if (min_manifest_version != null || max_manifest_version != null) {
+    return [
+      min_manifest_version ? `< ${min_manifest_version}` : undefined,
+      max_manifest_version ? `> ${max_manifest_version}` : undefined,
+    ]
+      .filter((e) => e !== undefined)
+      .join(' or ');
```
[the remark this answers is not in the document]
rpl: I opted for joining the version strings using  `', '`.

tests/unit/schema/test.firefox-schemas-import.js
rpl: I keep forgetting about `it.each`, I actually really prefer to use that for this kind of test cases, I just need to remind myself of its existence :-P

Thanks for pointing it out ;-)

tests/unit/test.utils.js
rpl: :heart: I rewrote this into an `it.each` form

rpl:
@willdurand thanks a lot for doing the last review pass so quickly :heart: 

I applied all the suggested tweaks (and typos to fix) and squashed them into the related patch that was introducing those changes, this way we still have the same 5 "kind of self-contained" separate commits.

I think it may be a good idea to land this using "Rebase and merge" (instead of the default "Squash and merge"), this way be may be able to more easily bisect which part of these set of changes may have introduced a regression if we need to.

(I may also double-check that each of this patch does pass all the tests on its own if we want to be 100% sure that bisecting in the middle of these 5 commits is going to be really possible in practice).

Let me know wdyt about this proposed "landing strategy".

willdurand:
> Let me know wdyt about this proposed "landing strategy".

I don't think we're going to revert this patch anyway so I am fine with rebase + merge (to avoid the additional merge commit)

rpl:
@willdurand I just verified locally that all our unit tests pass on each of the commits in this PR, and so it looks that we can land this using "Rebase and merge" and be able to bisect regressions that a single one of this patches may have introduced if we need to.

And so this should now be ready to land as soon as we want (&& strictly before it may bit rot ;-))

rpl:
Manually merged on the main branch as:
- c396d94c6abacec7ec7a393849a8e00f7dc1449d
- 75cba1b7202db374dec616402af716ca9f92729d
- 6e88dcfc5b9e35e78ac7b2096e7e441b088fb089
- c3d6237c6e4bc8fc2a61e02c4cabd0e725770825
- 6886efcc8ef8a25b01dc699e63e1e0e96e2711fd

PR #3641, opened 2021-03-26
fix: Prefer parsing as a module on js scripts with .mjs extension

### LINKED ISSUE #3639: False positive error for "Cannot use 'import.meta' outside a module" on a module file.
Repro:

Include a module file that is dynamically inserted into the DOM.

Use an es6 module feature such as `import.meta.url`.

image
^^^ Image from validation report of extension being uploaded to AMO

This file is an es6 module and has a `.mjs` extension.  It is loaded via JS with a `type='module'`.

The code is correct and works properly on all browsers including FireFox (of course).

Is there a way to hint the linter to avoid this rejection?  I'm not able to release my new extension version because of this.

rpl:
This PR contains the set of (small) proposed changes for the "more conservative" of the two approaches described in https://github.com/mozilla/addons-linter/issues/3639#issuecomment-806524060 (which defaults to sourceType: 'module' if the filename extension is `.mjs`, even if no import/export statement were being found in the parsed AST).

Fixes #3639

### DIFF
--- a/tests/unit/scanners/test.javascript.js
+++ b/tests/unit/scanners/test.javascript.js
@@ -170,6 +170,15 @@ describe('JavaScript Scanner', () => {
     expect(addonLinter.collector.warnings.length).toEqual(0);
   });
 
+  it('should parse as module .mjs files', async () => {
+    const code = 'const url = import.meta.url;';
+
+    const jsScanner = new JavaScriptScanner(code, 'code.mjs');
+
+    const { linterMessages } = await jsScanner.scan();
+    expect(linterMessages).toEqual([]);
+  });
+
   it('should support optional catch binding', async () => {
     const code = oneLine`
       try {} catch {}
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -271,7 +271,9 @@ export default class JavaScriptScanner {
 
     try {
       const ast = espree.parse(this.code, parserOptions);
-      detected.sourceType = this._getSourceType(ast);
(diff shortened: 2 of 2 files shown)


tests/unit/scanners/test.javascript.js
```diff
@@ -170,6 +170,15 @@ describe('JavaScript Scanner', () => {
     expect(addonLinter.collector.warnings.length).toEqual(0);
   });
 
+  it('should parse as module .mjs files', async () => {
```
willdurand, line 173: Should we consider `.jsm` as module files too? (Sorry this does not necessarily belong here but I wanted to ask)

willdurand (approved):
r+wc, thanks

tests/unit/scanners/test.javascript.js
rpl: > Should we consider .jsm as module files too? (Sorry this does not necessarily belong here but I wanted to ask)

@willdurand I don't think we have to parse `.jsm` as modules, technically the .jsm module files are not ESM modules (at least not at the moment), and so `import`/`export` are not technically valid statement if used in Firefox's `.jsm` module files. 

(If I'm not mistaken we added explicitly the `.jsm` file extension mainly to be sure we don't skip them in the validations rules, because they are used for our built in addons, which may include some of them as part of code used by the experimental APIs that privileged addon can embed, and to prevent other third party extension to be able to skip validation rules in some of their js code by just renaming them into .jsm files).

PR #3642, opened 2021-03-26
chore: Fix intermittent sinon spy callOrder assertion failure

### LINKED ISSUE #2441: Intermittent test failure
```
Linter › should optionally scan selected files
    AssertError: expected spy, spy, spy to be called in order but were called as spy, spy, spy

      590 | 
      591 |   it('should survive even a 1 column terminal', () => {
    > 592 |     const addonLinter = new Linter({ _: ['bar'] });
          |                          ^
      593 |     addonLinter.collector.addError({
      594 |       code: 'WHATEVER_ERROR',
      595 |       message: 'whatever error message',

      at Object.fail (node_modules/sinon/lib/sinon/assert.js:106:21)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:65:16)
      at Object.assertCallOrder [as callOrder] (node_modules/sinon/lib/sinon/assert.js:136:13)
      at Object._callee17$ (tests/unit/test.linter.js:592:26)
      at tryCatch (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:288:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/@babel/runtime/node_modules/regenerator-runtime/runtime.js:114:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
```

rpl:
Fixes #2441

### DIFF
--- a/tests/unit/test.linter.js
+++ b/tests/unit/test.linter.js
@@ -238,11 +238,16 @@ describe('Linter', () => {
     const getFileSpy = sinon.spy(addonLinter, 'scanFile');
 
     await addonLinter.scan();
-    sinon.assert.callOrder(
-      getFileSpy.withArgs('manifest.json'),
-      getFileSpy.withArgs('subdir/test.js'),
-      getFileSpy.withArgs('subdir/test2.js')
-    );
+
+    // There is no guarantee that test.js and test2.js will always
+    // be scanned in the same order (the order will depend from the
+    // order of the files returned by the getFiles method from the
+    // IOBase subclass, part of the addons-scanner-utils) and asserting
+    // a specific calls order here was making this test to fail
+    // intermittently.
+    sinon.assert.calledWith(getFileSpy, 'manifest.json');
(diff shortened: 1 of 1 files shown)


rpl:
@willdurand another option (that I think would also be reasonable given the purpose of this test) is to assert that the call did happen on the 3 files but don't assert the call order at all.

willdurand:
> @willdurand another option (that I think would also be reasonable given the purpose of this test) is to assert that the call did happen on the 3 files but don't assert the call order at all.

Do we really need to assert the order? The name of the test case doesn't mention that the order matters so I'd only assert that we process the 3 files and that's it.

willdurand (approved):

PR #3763, opened 2021-06-14
feat: Imported Firefox 90.0b7 api schemas

### LINKED ISSUE #3738: Deduplicate linting errors collected on invalid manifest_version 3 extension manifests
As an additional follow up to #3721, we would like to deduplicate some of linting errors that are collected when a manifest property is invalid due to the min_manifest_version/max_manifest_version JSONSchema annotations.

At the moment if a manifest property is invalid because the min/max_manifest_version set in the related JSONSchema definition is incompatible with the manifest_version set of the extension, our manifest validation is producing more than a single linting, quoting from https://github.com/mozilla/addons-linter/issues/3721#issuecomment-839869503:

> in terms of the errors being raised, if I change the manifest_version to 3 for one of the exiting extensions, I can see the following additional errors being reported, since the extension is not using some of expected the MV3 patterns:
```
Code                         Message                            Description                                                       File            Line   Column
JSON_INVALID                 "/background/scripts" is an        Your JSON file could not be parsed.
                             invalid additional property
MANIFEST_FIELD_REQUIRED      "/background/page" is a required   See https://mzl.la/1ZOhoEN (MDN Docs) for more information.       manifest.json
                             property
MANIFEST_FIELD_UNSUPPORTED   "/background" is in a format not   "/background" is in a format not supported in manifest versions   manifest.json
                             supported in manifest versions >   > 2.
                             2.
MANIFEST_FIELD_REQUIRED      "/background/service_worker" is    See https://mzl.la/1ZOhoEN (MDN Docs) for more information.       manifest.json
                             a required property
JSON_INVALID                 "/background" is not a valid key   Your JSON file could not be parsed.
                             or has invalid extra properties
```

The fact that the extension has `manifest_version: 3` but also including a `background` sub-property not valid for that manifest_version turns into multiple errors all related to a single actual issue to fix in the manifest.

We would like to deduplicate some of these linting errors to improve the readability of the actual underlying issue.

rpl:
This PR imports the WebExtensions API JSONSchema from Firefox 90.0b7, which includes the following changes:

- Added deprecation message on the `browserSettings`'s `ftpProtocolEnabled`
- changes to `chrome_settings_override`'s `favicon_url`, which will enforce the favicon url to be part of the extension xpi file in manifest_version 3 extensions ([Bug 1697059](https://bugzilla.mozilla.org/1697059))
- `extension.lastError`, `extension.getURL`, `extension.getBackgroundPage`/`runtime.getBackgroundPage` as marked as unsupported for manifest_version 3 extensions ([Bug 1687762](https://bugzilla.mozilla.org/1687762))
- added `"matrix"` to the list of protocols that can be handled by extensions ([Bug 1688030](https://bugzilla.mozilla.org/1688030)
- marked with min/max_manifest_version the schema entries related to the manifest `content_security_policy` ([Bug 1696043](https://bugzilla.mozilla.org/1696043))
- added manifest_version 3 format for the `web_accessible_resources manifest` property ([Bug 1696580](https://bugzilla.mozilla.org/1696580)

The changes to the `web_accessible_resources` format was triggering a test failure (due to the same issue tracked by #3738), and so in addition to the changes to the imported schema, this PR also includes a separate patch with a proposed approach to fix #3738.

The change to the test case related to `background.service_worker` being used in a `manifest_version: 2` extension is related to this set of changes, the resulting messages for a `manifest_version 2` extension does not mention anymore the formats supported in `manifest_version: 3`.

### DIFF
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -25,11 +25,51 @@ import schemas from './imported';
 
 const jsonSchemaDraft06 = require('ajv/lib/refs/json-schema-draft-06');
 
-function filterErrors(errors) {
+function filterErrors(errors, manifestData = undefined) {
   if (errors) {
-    return errors.filter((error) => {
+    let filteredErrors = errors.filter((error) => {
       return error.keyword !== '$merge';
     });
+
+    // Filter out errors related to future manifest versions.
+    if (
+      filteredErrors.length > 0 &&
+      manifestData?.manifest_version === MANIFEST_VERSION_DEFAULT
+    ) {
+      filteredErrors = filteredErrors.filter((error) => {
+        // Omit errors related to a schema fragment that is only allowed
+        // in a future manifest version (and also if its parent schema
+        // is only allowed in future manifest versions).
+        if (
+          error.params?.min_manifest_version > MANIFEST_VERSION_DEFAULT ||
+          error.parentSchema?.min_manifest_version > MANIFEST_VERSION_DEFAULT
+        ) {
+          return false;
+        }
+
+        // Skip anyOf errors if only one entry is actually allowed in the
+        // currently supported manifest versions.
+        if (error.keyword === 'anyOf') {
+          const anyOfSchemaEntries = error.schema.filter((schema) => {
+            return !(
+              'min_manifest_version' in schema &&
+              schema.min_manifest_version > MANIFEST_VERSION_DEFAULT
+            );
+          });
+
+          // Skip if there is only one or no entries (if there is one entry
+          // we can still omit the anyOf, the linting message related to that
+          // entry would already been separately part of the validation errror
+          // if it should have been).
+          if (anyOfSchemaEntries.length <= 1) {
+            return false;
+          }
+        }
+        return true;
+      });
+    }
+
+    return filteredErrors;
   }
   return errors;
 }
@@ -356,6 +396,23 @@ export class SchemaValidator {
         const manifestVersion =
           (rootData && rootData.manifest_version) || MANIFEST_VERSION_DEFAULT;
         const res = condFn(keywordSchemaValue, manifestVersion);
+
+        // If the min/max_manifest_version is set on a schema entry of type array,
+        // propagate the same keyword to the `items` schema, which is needed to
+        // - be able to recognize that those schema entries are also only allowed on
+        //   certain manifest versions (which becomes part of the linting messages)
+        // - be able to filter out the validation errors related to future (not yet
+        //   supported) manifest versions if they are related to those schema entries
+        //   (which happens based on the current or parent schema in the `filterErrors`
+        //   helper method).
+        if (schema.type === 'array') {
+          // NOTE: the alternative would be to do this for the entire set of schema
+          // entries (either while importing them, or when we load the data into the
+          // SchemaValidator class instance).
+          // eslint-disable-next-line no-param-reassign
+          schema.items[keyword] = schema[keyword];
+        }
+
         if (!res) {
           validate.errors = [
             {
@@ -418,7 +475,10 @@ export function getValidator(validatorOptions) {
 export const validateAddon = (manifestData, validatorOptions = {}) => {
   const validator = getValidator(validatorOptions);
   const isValid = validator.validateAddon(manifestData);
-  validateAddon.errors = filterErrors(validator.validateAddon.errors);
+  validateAddon.errors = filterErrors(
+    validator.validateAddon.errors,
+    manifestData
+  );
   return isValid;
 };
 
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -25,53 +25,122 @@ import schemas from './imported';
 
 const jsonSchemaDraft06 = require('ajv/lib/refs/json-schema-draft-06');
 
-function filterErrors(errors, manifestData = undefined) {
-  if (errors) {
-    let filteredErrors = errors.filter((error) => {
-      return error.keyword !== '$merge';
+function isRelevantError({
+  error,
+  manifest_version,
+  allowedManifestVersionsRange,
+}) {
+  // The errors related to the manifest_version are always relevant,
+  // if an error has been collected for it then it is because the
+  // addon manifest_version is outside or the allowed range.
+  if (error.dataPath === '/manifest_version') {
+    return true;
+  }
+
+  const { minimum, maximum } = allowedManifestVersionsRange;
+
+  const errorMinManifestVersion =
+    error.params?.min_manifest_version ??
+    error.parentSchema?.min_manifest_version ??
+    minimum;
+
+  const errorMaxManifestVersion =
+    error.params?.max_manifest_version ??
+    error.parentSchema?.max_manifest_version ??
+    maximum;
+
+  // Omit errors related to a schema fragment that is not relevant
+  // for the given manifest version (and also if its parent schema
+  // is not relevant for the given manifest version).
+  if (
+    manifest_version < errorMinManifestVersion ||
+    manifest_version > errorMaxManifestVersion
+  ) {
+    return false;
+  }
+
+  // An error collected by an `anyOf` schema entry is relevant only if its the schema
+  // entries are relevant for the given addon manifest_version.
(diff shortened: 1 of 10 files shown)


rpl:
- @Rob--W for the changes related to the imported JSONSchema (see c0ee8ec)
- @willdurand for the additional changes related to #3738 (see  705b2a4)

Related to #3738, with the changes as currently included in this PR, the validation on a MV2 extension which wrongly use `background.service_worker` in its manifest produces the following results:

```
Validation Summary:

errors          4
notices         0
warnings        0

ERRORS:

Code                      Message               Description                             File            Line   Column
JSON_INVALID              "/background/servi?   Your JSON file could not be parsed.
                          ce_worker" is an
                          invalid additional
                          property
MANIFEST_FIELD_REQUIRED   "/background/page"    See https://mzl.la/1ZOhoEN (MDN Docs)   manifest.json
                          is a required         for more information.
                          property
MANIFEST_FIELD_REQUIRED   "/background/scrip?   See https://mzl.la/1ZOhoEN (MDN Docs)   manifest.json
                          ts" is a required     for more information.
                          property
JSON_INVALID              "/background" is      Your JSON file could not be parsed.
                          not a valid key or
                          has invalid extra
                          properties
```

Rob--W (approved):
Schema import lgtm. I started to look at the mv3 keywords and thought that Will may be a better person to review this. And then I noticed that you had already asked him for a review of those changes :)

willdurand (commented):
I only took a quick look, I need to think about it a bit more.

src/schema/validator.js
```diff
@@ -25,11 +25,51 @@ import schemas from './imported';
...
+            return !(
+              'min_manifest_version' in schema &&
+              schema.min_manifest_version > MANIFEST_VERSION_DEFAULT
+            );
+          });
+
+          // Skip if there is only one or no entries (if there is one entry
+          // we can still omit the anyOf, the linting message related to that
+          // entry would already been separately part of the validation errror
+          // if it should have been).
```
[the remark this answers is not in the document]
rpl: @willdurand yeah, sorry, I should have added a comment in the PR to provide you a concrete example.

Follows the one I've been looking to,

Preconditions:
- no error filtering (and so omitting all changes currently applied in `filterErrors` by this pull request)
- run only the 1st test case in the test.web_accessible_resources.js test file

Result:
the validation errors collected are the following ones:

- web_accessible_resources should be an array for the max_manifest_version 2 schema entry:
```json
{
  "keyword": "type",
  "dataPath": "/web_accessible_resources",
  "schemaPath": "#/allOf/0/properties/web_accessible_resources/anyOf/0/type",
  ...
  "message": "should be array",
  "schema": "array",
  "parentSchema": {
    "max_manifest_version": 2,
    ...
  },
  "data": "foo.png"
}
```

- web_accessible_resources should be an array also for the min_manifest_version 3 schema entry:
```json
{
  "keyword": "type",
  "dataPath": "/web_accessible_resources",
  "schemaPath": "#/allOf/0/properties/web_accessible_resources/anyOf/1/type",
  ...
  "message": "should be array",
  "schema": "array",
  "parentSchema": {
    "min_manifest_version": 3,
    ...
  },
  "data": "foo.png"
}
```

- The two schema entries that have triggered the validation results described above, are then wrapped into a single `anyOf` entry, given that the property being validated is invalid for both of them, the `anyOf` entry also produces its own validation result:
```json
{
  "keyword": "anyOf",
  "dataPath": "/web_accessible_resources",
  "schemaPath": "#/allOf/0/properties/web_accessible_resources/anyOf",
  ...
  "message": "should match some schema in anyOf",
  "schema": [
    {
      # this is the max_manifest_version 2 entry
      "max_manifest_version": 2,
      "type": "array",
      ...
    },
    {
      # and this is the max_manifest_version 3 entry
      "min_manifest_version": 3,
      "type": "array",
      ...
    }
  ]
  "data": "foo.png"
}
```

This is the validation result that the comment refers to, in some cases there may be other entries part of the `anyOf` that
are valid on all manifest versions (not being marked with min_manifest_version or max_manifest_version) and so I thought to only omit it if the anyOf only have one entry if we don't count the entries related to future manifest versions which are going to be filtered out.

Let me know if this helped to clarify what the comment was meant to explain.

rpl:
> I only took a quick look, I need to think about it a bit more.

@willdurand sure thing, I created this PR sooner rather than later exactly because I thought that part was going to require some discussion to evaluate if the approach as proposed is the one we think can work in the long run (and evaluate other possible approaches if we think it would be necessary).

Let me know if you having a quick sync chat may help.

src/schema/validator.js
```diff
@@ -25,11 +25,51 @@ import schemas from './imported';
...
   if (errors) {
-    return errors.filter((error) => {
+    let filteredErrors = errors.filter((error) => {
       return error.keyword !== '$merge';
     });
+
+    // Filter out errors related to future manifest versions.
+    if (
+      filteredErrors.length > 0 &&
+      manifestData?.manifest_version === MANIFEST_VERSION_DEFAULT
```
willdurand, line 37: shouldn't we use min/max versions here and below instead of the default version? We only filter out eror for "future manifest versions" because the default is set to `2` but would it make sense to filter out errors that are outside the range specific by the user?

rpl: @willdurand and I discussed about this comment in a meeting, and this was a summary of what we discussed:
- the approach in this PR aims to only report errors that are relevant for the manifest_version of the addon being validated ([1])
- the reasoning behind this approach is that:
  - the extension developers would not get wrong the manifest_version, and so the validation errors should only suggest the changes that are relevant to make the addon valid with the manifest_version set in their manifest
  - if an extension developer wants to port its extension to manifest_version 3, they would be changing the manifest_version set in their extension manifest, and at that point the errors related to that schema versions becomes relevant  (and the ones related to the older manifest version are not relevant anymore)
  - the addons-linter cli options --min-manifest-version / --max-manifest-version are meant to allow the developers to set the manifest_version to a value that is not yet allowed by AMO, but it should not change the errors that we are reporting (until the developer decide to actually set a different manifest_version on the extension manifest)

----
[1]: the new version of this patch, which the additions from 17eea76, does not use MANIFEST_VERSION_DEFAULT in these checks anymore, it does use the manifestData.manifest_version instead

src/schema/validator.js
```diff
@@ -241,42 +323,48 @@ export class SchemaValidator {
...
-          })}`
-        );
-      }
-
-      schemaData = deepPatch(this.schemaObject, {
-        types: {
-          ManifestBase: {
-            properties: {
-              manifest_version: { minimum, maximum },
+  get allowedManifestVersionsRange() {
```
willdurand, line 326: When should we prefer a getter over a function? I don't think this necessarily needs to be a (computed) property.

src/schema/validator.js
```diff
@@ -356,6 +444,23 @@ export class SchemaValidator {
...
         const res = condFn(keywordSchemaValue, manifestVersion);
+
+        // If the min/max_manifest_version is set on a schema entry of type array,
+        // propagate the same keyword to the `items` schema, which is needed to
+        // - be able to recognize that those schema entries are also only allowed on
+        //   certain manifest versions (which becomes part of the linting messages)
+        // - be able to filter out the validation errors related to future (not yet
+        //   supported) manifest versions if they are related to those schema entries
+        //   (which happens based on the current or parent schema in the `filterErrors`
+        //   helper method).
```
willdurand, line 455: Would be nice to file an issue for this + add `TODO: <url of the issue>` here.

willdurand (approved):
Honestly, it looks very good so r+wc

src/schema/validator.js
rpl: yeah, honestly I think that I agree with you, this may be actually computed once in the constructor (especially given that it is also validating the resulting range and throwing an error if it turns out to be invalid).

I'll apply that change and update this PR accordingly.

src/schema/validator.js
rpl: :+1: totally, I'm going to file an issue upfront and mention it here in this comment

rpl:
@willdurand I updated the PR based on the last round of review comments, and so it should now be good to go.

Issue #3761, opened 2021-06-11
Use Node 14 as default node version


willdurand:
~~We would like to use Node 14 for AMO on production so we are going to set the default Node version of addons-linter to `14` instead of `12` as it is currently.~~

As discussed with @rpl on Matrix, we're going to continue to use Node 12 as the "default version" so that we can release new versions fully compatible with Node 12 until web-ext is "ready" to deprecate Node 12.

---

For QA: once a new version of the linter has been released and deployed on AMO, please make sure that we can still validate new submissions.

willdurand:
> @willdurand
> I'm thinking of testing this tomorrow at release testing on AMO stage - uploading addons using dev hub.
> I hope it's fine.

I reverted the patch for this issue as we are not ready to build the linter with Node 14. Instead, we'll continue to use Node 12 to build the linter, even if AMO uses Node 14 as runtime (this shouldn't be a problem).

willdurand:
This has been fixed.

ioanarusiczki:
@willdurand I assume this requires testing as mentioned above. Can this be done on AMO dev with linter 5.1.0 ?

willdurand:
> @willdurand I assume this requires testing as mentioned above. Can this be done on AMO dev with linter 5.1.0

I think this has been verified already because the changes aren't new.

ioanarusiczki:
@willdurand Ah ok, thanks!

PR #3787, opened 2021-06-22
Update circle-ci config to test the project against a third node version

willdurand:
We're going to adopt node 14 as the "new" current node version in #3761 but we probably want to keep node 12 compatibility for a while, at least for web-ext, which is why I am proposing a `test-alternate` job that could be use to test against a 3rd Node version. It's currently set to Node 16 but it could be set to Node 12 once the current node version becomes "14" and the next version becomes "16".

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,9 +145,7 @@ jobs:
       # run integration tests using an addons-linter binary in a
       # production-like environment
       - run: npm run test-integration:production
-      # TODO: re-enable this test suite, see:
-      # https://github.com/mozilla/addons-linter/issues/3788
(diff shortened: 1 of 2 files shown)


willdurand:
@rpl it is not urgent and it depends on the web-ext plans too but I am thinking that we might want to continue to run the test suite against Node 12, 14 and 16 in the near future. Our plans to switch to Node 14 by default on AMO is coming soon (hopefully by the end of the week).

willdurand:
@rpl I updated this PR as per our last discussion (a long time ago..)

rpl (approved):
Lgtm

PR #3810, opened 2021-07-16
Use node 14 as default

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3761

---

This PR changes the default node version from `12` to `14` because AMO is going
to use Node 14 as the default production version. The next node version is set
to `16` and we set the alternate Node version to `12` because we want to
support Node 12 for web-ext.

This change also means that the `addons-linter` package (published on npmjs)
will be built with Node `14` and not Node `12` anymore.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ references:
     working_directory: ~/addons-linter
     docker:
       # This is the NodeJS version we run in production.
-      - image: circleci/node:12
+      - image: circleci/node:14
     environment:
       # This environment variable is needed for `scripts/publish-rules`.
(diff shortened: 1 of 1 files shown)


willdurand:
> This change also means that the `addons-linter` package (published on npmjs)
> will be built with Node `14` and not Node `12` anymore.

Given that fact, I am thinking about releasing a **major** version even if Node 12 is still supported. What do you think?

willdurand:
@rpl hey, if we can land this before EOD, that would be great (but if not, it is also fine), thanks!

rpl (approved):
Lgtm

PR #3823, opened 2021-07-23
Update/firefox 91 api schemas

rpl:
This PR imports the WebExtensions API JSONSchema from Firefox 90.0b7, which includes the following changes:

- bookmarks.json - removed two unsupported API methods that are not exposed to extensions on Chrome [Bug 1713674](https://bugzilla.mozilla.org/show_bug.cgi?id=1713674)
- actions.json / browser_action.json / context_menus.json / menus.json - introduced "action" namespace, redefined "browserAction" namespace as an "alias" or the "action" namespace definition [Bug 1706398](https://bugzilla.mozilla.org/show_bug.cgi?id=1706398)
- browser_settings.json - introduces some webRender related browserSettings - [Bug 1714428](https://bugzilla.mozilla.org/show_bug.cgi?id=1714428) and [Bug 1719688](https://bugzilla.mozilla.org/show_bug.cgi?id=1719688)
- geckoProfiler.json - some addition to the geckoProfiler API namespace - [Bug 1715257](https://bugzilla.mozilla.org/show_bug.cgi?id=1715257) and [Bug 1715922](https://bugzilla.mozilla.org/show_bug.cgi?id=1715922)

Besides the updated JSONSchema data imported, this PR contains some other changes required to successfully import the JSONSchema data from Firefox 91 and provide the expected behaviors (which should possibly match the behavior on Firefox, in particular errors that prevent an successfully installation on Firefox should be reported as blocking errors on the addons-linter side):

- The first two patches contains the minimal changes needed to handle a couple of new special cases introduced by the new definitions in browser_action.json in Firefox 91, browse_action.json definitions are similar to the ones in menus.json but different enough to require changes to the import script, in particular:
  - browser_action.json contains more than one schema entry that extend the manifest
  - browser_action.json contains two API namespace, "browserAction" and "action", where browserAction is defined by importing the action namespace, but the imported "action" definition has a "min_manifest_version" property set to 3 while "browserAction" has a "max_manifest_version" property set to 2
  - "browserAction"  imports from "action" in a file named browser_action.json (while menus.json does the opposite, "contextMenus" imports "menus")

- The third patch contains some additional changes to src/schema/validator.js to make sure that:
  - we don't filter out the errors related to top level manifest keys that are only supported on specific manifest versions (which is needed to make the addons-linter to match Firefox behavior on "action" / "browser_action" manifest keys used in a manifest with the wrong "manifest_version"
  - but at the same time that we still filter out multiple errors reported to the same top level manifest property due to multiple anyOf definition related to different manifest version ranges ("web_accessible_resources" is for sure one of those)

- The rest of the patches are importing the schema data from Firefox 91beta5 and adding some additional test coverage for the changes introduces


rpl:
Hey @willdurand, while I look into adding some more tests (and before I add Rob for an additional review of the changes to the schema imported), would you mind give this a round of preliminary feedback on the changes needed on the schema import script and schema validator?

src/schema/firefox-schemas-import.js
```diff
@@ -354,7 +356,14 @@ export function rewriteExtend(schemas, schemaId) {
...
         // Move the $extend into definitions.
-        extendDefinitions[$extend] = rest;
+        //
+        // NOTE: some schema files, like browser_action.json, may contains more than one extends
+        // and so here we use deepmerge to merge the multiple extended definitions instead
+        // of overwriting it with the last one processed.
+        extendDefinitions[$extend] = deepmerge(
+          extendDefinitions[$extend] ?? {},
+          rest
+        );
```
willdurand, line 366: I suppose this is for things like this: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/browser_action.json#50-68

Is there a chance that the `properties` keys are the same in the future? (which would lead to unexpected deep merges I'd suspect).

src/schema/firefox-schemas-import.js
```diff
@@ -473,47 +482,88 @@ export function foldSchemas(schemas) {
   return foldedSchemas;
 }
 
+// Normalize the JSONSchema, gets an array of JSONSchema objects and return an array of
+// normalized JSONSchema objects (a JSONSchema file may contain more than one API namespaces,
```
willdurand, line 486:
```suggestion
// normalized JSONSchema objects (a JSONSchema file may contain more than one API namespace,
```

willdurand: Does the normalization return an array with each item being one API namespace?

src/schema/firefox-schemas-import.js
```diff
@@ -473,47 +482,88 @@ export function foldSchemas(schemas) {
...
-    namespace
-  );
-  const updatedTypes = { ...loadTypes(types), ...extendTypes };
-  return {
-    ...rest,
-    ...extendRest,
-    id: namespace,
-    types: updatedTypes,
-  };
+  // Rewrite manifest extend types using the each api namespace in the file.
```
willdurand, line 520:
```suggestion
  // Rewrite manifest extend types using each api namespace in the file.
```

src/schema/firefox-schemas-import.js
```diff
@@ -473,47 +482,88 @@ export function foldSchemas(schemas) {
...
+  return apiNamespaceSchemas.map((apiScheme) => {
+    let { namespace, types, ...rest } = apiScheme; // eslint-disable-line prefer-const
```
willdurand, line 522:
```suggestion
  return apiNamespaceSchemas.map((apiSchema) => {
    let { namespace, types, ...rest } = apiSchema; // eslint-disable-line prefer-const
```

src/schema/firefox-schemas-import.js
```diff
@@ -473,47 +482,88 @@ export function foldSchemas(schemas) {
...
+    const { types: extendTypes, ...extendRest } = rewriteExtend(
+      manifestExtendSchemas,
+      namespace
+    );
+    const updatedTypes = { ...loadTypes(types), ...extendTypes };
+
+    if (rest.$import) {
+      // menus.json and browser_action.json contains the definition of two API
```
willdurand, line 530:
```suggestion
      // menus.json and browser_action.json contain the definition of two API
```

src/schema/firefox-schemas-import.js
willdurand, line 522: If you extract `$import` here, do you need to (1) use `let` and (2) `delete` it later?

willdurand: You could potentially rename `namespace` to `id` here so that you don't need to do it at the end.

src/schema/firefox-schemas-import.js
```diff
@@ -621,6 +672,13 @@ export function importSchemas(firefoxPath, ourPath, importedPath) {
   const ourSchemas = {
     ...readSchema(ourPath, 'manifest.json'),
     ...readSchema(ourPath, 'contextMenus.json'),
+    // the "action" API namespace is currently part of browser_action.json
```
willdurand, line 675:
```suggestion
    // The "action" API namespace is currently part of browser_action.json
```

src/schema/firefox-schemas-import.js
```diff
@@ -621,6 +672,13 @@ export function importSchemas(firefoxPath, ourPath, importedPath) {
...
+    // along with "browserAction", the following is needed to make sure we
+    // will normalize and import the "action" API namespace in a file named
+    // "src/schema/imported/action.json" (similar to what contextMenus.json
+    // does for the "contextMenus" API namespace which is defined along with
+    // the "menus" API namespaces)
```
willdurand, line 680:
```suggestion
    // the "menus" API namespaces).
```

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -421,23 +425,66 @@ describe('firefox schema import', () => {
...
+    });
+
+    it('resolves the top level $import in API namespace definitions', () => {
+      const schemas = [
+        {
+          namespace: 'action',
+          functions: [{ name: 'setTitle', type: 'function', parameters: [] }],
+          events: [{ name: 'onClicked', type: 'function', parameters: [] }],
+          // min/max_manifest_version are expected to not be imported in the
+          // jsonschema definition importing this one.
```
willdurand, line 456:
```suggestion
          // JSONSchema definition importing this one.
```

(as you wrote it many times above :p )

tests/unit/schema/test.action.js
```diff
@@ -0,0 +1,62 @@
+import cloneDeep from 'lodash.clonedeep';
+
+import { validateAddon } from 'schema/validator';
+
+import { validManifest } from './helpers';
+import { assertHasMatchingError } from '../helpers';
+
+describe('/browser_action', () => {
```
willdurand, line 8: (I suppose we don't do this in this project, I like to use a top-level `describe(__filename, () => {});` in test files).

tests/unit/schema/test.action.js
```diff
@@ -0,0 +1,62 @@
...
+      browser_action: { default_popup: 'popup.html' },
+    };
+    validateAddon(manifest);
+    expect(validateAddon.errors).toBe(null);
+  });
+
+  it('should not be allowed in manifest_version 3', () => {
+    let manifest = cloneDeep(validManifest);
+    manifest = {
+      ...manifest,
```
willdurand, line 23:
```suggestion
    const manifest = {
      ...cloneDeep(validManifest),
```

willdurand: (maybe)

also do you need `cloneDeep` here?

tests/unit/schema/test.action.js
```diff
@@ -0,0 +1,62 @@
+import cloneDeep from 'lodash.clonedeep';
+
+import { validateAddon } from 'schema/validator';
+
+import { validManifest } from './helpers';
+import { assertHasMatchingError } from '../helpers';
+
+describe('/browser_action', () => {
+  it('should be allowed in manifest_version 2', () => {
+    let manifest = cloneDeep(validManifest);
```
willdurand, line 10: I am wondering if it would make sense to have a helper for that like (especially if `cloneDeep` is needed):

```
const manifest = makeValidManifest({
  manifest_version: 2,
  // and other overrides here
});
```

willdurand (approved):
Looks good to me, everything makes sense so far and it is documented. I have many comments but not very important ones.

tests/unit/schema/test.action.js
rpl: yeah, I don't recall that to be used anywhere in this project at the moment.

isn't that redundant? from what I recall the filename would still be reported in the failures output and so at that point wouldn't be mostly the same to don't group the tests into a describe if there is only one and it is named exactly as the file?

Anyway, besides my doubt described above (which is not really a big deal), I don't really mind to use that convention in this new test files if we want to start using it in this project, but at that point I would split this test file in two separate ones, one for `action` and the other for `browserAction`.

tests/unit/schema/test.action.js
rpl: definitely not needed anymore for the top level properties given that we are already creating a new object now, and as long as we don't change any non-toplevel properties coming from the `validManifest` then the cloneDeep would be redundant and not strictly necessary.

tests/unit/schema/test.action.js
rpl: I can definitely add one and use it in the new test added in this PR (but we may leave to a follow up changing all other existing places where we are using `cloneDeep(validManifest)`, I haven't looked but from what I recall I would expect most of the tests in tests/unit/schema may want to be updated to use the new helper once introduced).

src/schema/firefox-schemas-import.js
rpl: > I suppose this is for things like this: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/browser_action.json#50-68

yes it is

> Is there a chance that the properties keys are the same in the future? (which would lead to unexpected deep merges I'd suspect).

it shouldn't happen, but honestly I think that it would be a good idea if we would explicitly check we are not overwriting a property with a different definition, and throw an explicit error if we are.
We will still need to review the imported schema as we are already doing, and we may likely notice that if it happens, but I would still feel more "relaxed" if we throw on importing an unexpected set of schema data, instead of silently doing something potentially unexpected.

I'm going to add an explicit "check and throw on invalid data" here.

src/schema/firefox-schemas-import.js
rpl: yes, that's what this method is expected to do.

to be precise:
- the version without this changes was returning the normalized version of the JSONSchema object **not related to extending the manifest** (which,  if there is one, is expected to be an API)
- the new version does basically the same thing but it does not assume that there will be only one API namespace per file, because that's not completely true (for menus.json and browser_action.json at the moment), + the addition of normalizing the $import explicitly

Given that I'm going to still do some other tweaks based on your review comments, I'm going to take another look to this method to see if there is any other unexpected corner case that we may want to throw an explicit error for.

src/schema/firefox-schemas-import.js
rpl: I'll take another look, I wouldn't exclude I can replace this eslint-disable-line with a few tweaks to the code instead (honestly I just rubberstamped the eslint-disable-line because I wanted to be sure that everything was working before spending some more time tweaking this function to avoid that, but I was just being a bit "lazy"... errr "efficient" :-P)

rpl:
@willdurand I've just update this PR, merged in all typos fixes suggested and added:
- two new checks that throw an explicit errors on unexpected schema data that would result in imported schema data different from what the linter eslint rules and manifest schema validator expects (in particular potentially recursing $import properties and extended types that are overwriting properties with the same name) + new test case to cover these new expected behaviors
- a new `createValidManifest` test helper (for now only used in the new test.action.js test file, we may file a good first bug to rewrite some other schema tests that may be also calling cloneDeep on the validManifest const).

Feel free to give this PR (or just the last patches added from the previous review, if you prefer to just look to the interdiff).

In the meantime I'm going to add Rob for a separate sign-off on the schema data imported.

src/schema/firefox-schemas-import.js
rpl: In the end I haven't added an additional check for this, we are currently assuming that in each JSONSchema file, after filtering out any schema entry related to the manifest namespace, all the remaining entries should be API namespaces, and that should still be a reasonable assumption.

We can re-evaluate later on if we determine that we should be more strict (especially given that at that point it may be more clear what to check to decide if the schema data is not the one we expect).

src/schema/firefox-schemas-import.js
rpl: @willdurand ah, in my comment with a summary of the last round of changes I forgot to mention that I removed this `eslint-disable-line` suppress comment, with just a couple of some small tweaks to this arrow function.

tests/unit/schema/test.action.js
rpl: (I haven't done this, I think we can do it in a follow up if we think it would be valuable, even just to match conventions used in other ones of our projects).

rpl:
@Rob--W in the first part of https://github.com/mozilla/addons-linter/pull/3823#issue-695919205 there is a short summary of the schema changes imported in this PR.

Rob--W (approved):
Schema changes lgtm.

src/schema/firefox-schemas-import.js
```diff
@@ -353,8 +355,29 @@ export function rewriteExtend(schemas, schemaId) {
     (extendSchema.types || []).forEach((type) => {
       const { $extend, id, ...rest } = type;
       if ($extend) {
+        // Throw an explicit error if we are going to deep merge a set of properties which may be
+        // overriding an existing one, instead of extending the type with new ones as expected.
+        if (rest.properties) {
+          const newProps = Object.keys(rest.properties);
+          if (
+            !newProps.every(
+              (k) => extendDefinitions?.[$extend]?.properties?.[k] == null
```
willdurand, line 364:
```suggestion
              (k) => extendDefinitions?.[$extend]?.properties?.[k] === null
```

?

src/schema/firefox-schemas-import.js
```diff
@@ -353,8 +355,29 @@ export function rewriteExtend(schemas, schemaId) {
...
+            )
+          ) {
+            throw new Error(
+              `Unsupported schema format: detected multiple extend schema entries overwriting existing properties while processing "${schemaId}" namespace`
+            );
```
willdurand, line 369: Import `oneLine` from `common-tags`, then re-format more or less like this:

```suggestion
            throw new Error(oneLine`Unsupported schema format: detected
              multiple extend schema entries overwriting existing properties
              while processing "${schemaId}" namespace`);
```

src/schema/firefox-schemas-import.js
```diff
@@ -473,47 +497,94 @@ export function foldSchemas(schemas) {
...
+        // also ignore them while processing the "$import").
+        namespace: importedNamespace,
+        min_manifest_version,
+        max_manifest_version,
+        ...importedRest
+      } = apiNamespaceSchemas.find((schema) => schema.namespace === $import);
+      if (importedRest.$import) {
+        throw new Error(
+          `Unsupported schema format: "${namespace}" is importing "${importedNamespace}" which also includes an "$import" property`
+        );
```
willdurand, line 567: Another location where `oneLine` could be useful

tests/unit/schema/test.firefox-schemas-import.js
```diff
@@ -911,6 +975,47 @@ describe('firefox schema import', () => {
...
+                  min_manifest_version: 3,
+                  $ref: 'ActionManifest',
+                  optional: true,
+                },
+              },
+            },
+            {
+              $extend: 'WebExtensionManifest',
+              properties: {
+                // Fake schema entry which contains a property definition that would overwrite
```
willdurand, line 1001: :+1:

willdurand (approved):

src/schema/firefox-schemas-import.js
rpl: `== null` is true for both `null` and `undefined`, that's why I opted for `==` instead of `===` here

willdurand: I wasn't sure this was expected tbh so maybe add a comment?

rpl: oh, yeah, that can be reasonable to remind ourselves that.

rpl: comment added in one last patch added to this PR.

PR #3835, opened 2021-07-30
Add (skipped) test case for public field declarations

willdurand:
I wrote this test case to determine whether espree v8 alone would be a enough to support public field declarations. This isn't enough, which is why the test case is currently skipped.


willdurand:
mmm I just noticed we already had a skipped test but I didn't grep the code for the right term... "public field declaration" vs "class field".

Issue #3507, opened 2020-12-16
Verify that `publish-rules` actually works when we need it


willdurand:
The new Circle CI config runs this job but it does not push new commits to the `gh-pages` branch all the time it seems, so let's double check when we expect new content to be published.

muffinresearch:
@willdurand noting that token associated with this has never been used so I've revoked it. Please let me know when it should be re-instated.

willdurand:
> @willdurand noting that the `GH_TOKEN` associated with this has never been used so I've revoked it. Please let me know when it should be re-instated.

ok, thanks. I suppose we'll need it next time we introduce a rule? 🤔

willdurand:
Something went wrong lately: https://app.circleci.com/pipelines/github/mozilla/addons-linter/977/workflows/e89f11a7-562c-4adc-8694-a8c5effe2d92/jobs/2210

willdurand:
@muffinresearch could you please re-add a `GH_TOKEN` env var to the addons-linter project on Circle CI? This is needed for the `publish-rules` command. Thanks!

muffinresearch:
Done

muffinresearch:
Leaving open for verification this actually works :)

willdurand:
It sort of works, heh. For some reasons, this repo has a `CIRCLE_PULL_REQUEST` env var defined all the time with a link to a completely different project. This is preventing the script that publishes the rules to execute.

Unsetting this env var allows to publish the updated content (which I did manually) so the command and the config works, except that we have this Circle CI specific problem.

willdurand:
This isn't necessary anymore because of https://github.com/mozilla/addons-linter/issues/3852

PR #3848, opened 2021-08-02
Disable Circle CI on commits pushed to gh-pages

willdurand:
Refs #3507

---

WIP


willdurand:
Let's do https://github.com/mozilla/addons-linter/pull/3851 instead

PR #3851, opened 2021-08-03
Move generated docs to /docs/

willdurand:
Refs #3507 

---

This PR changes the way we publish a website with a list of rules, available at: https://mozilla.github.io/addons-linter/

Instead of using a different branch (`gh-pages`), we now serve the `/docs/` folder on the main branch (`master`) and the generated content lives in git, just like before except that it is not in a different branch anymore. This allows us to remove the `publish-rules` script as well as some env vars on Circle CI (no more `GH_*` env vars seems better in terms of security).

There is a CI check for when the generated content is outdated. In this case, the author of the PR will run `npm run build-rules` and commit the changes. This should not happen frequently but breakages will be more visible, which is a good thing.

Note: once this PR is accepted, a github admin will have to update the "source" in the GH Pages settings to use the `master` branch and the `/docs` folder. After that, we can remove the `gh-pages` branch.


rpl (approved):
Sounds like a reasonable strategy and the changes in this PR lgtm.

Thanks Will.

willdurand:
I filed https://github.com/mozilla/addons-linter/issues/3852 to update the source of the gh pages.

PR #3880, opened 2021-08-30
chore: speed up schema data import script using git partial clones

rpl:
This PR introduces a new script that imports schema updates from gecko-dev using [git partial clone](https://git-scm.com/docs/partial-clone) feature.

The script we are currently using is downloading a zip archive from hg.mozilla.org for a given mercurial tag name, downloading the zip file from hg.mozilla.org takes around: 
- ~720Mb disk space for the zip file
- ~38 minutes for the download to complete (at least from my home office network)

Using git partial clones we can't refer to a specific mercurial tag (because it looks that we are not creating git tags for the mercurial tags for a while in [mozilla/gecko-dev](https://github.com/mozilla/gecko-dev), but we do have the main and beta branches which are in sync with nightly and beta and downloading and importing the schema data takes instead:
- ~11Mb disk space for the git partial clone (that does only include the subset of files we really need)
- ~50 seconds for the download and import to complete (and apparently even less, ~20s, when the scripts runs in a github workflow)

The reduced disk space and time needed to download the needed assets will also help to follow up with automating the process a bit more (e.g. through a github workflow).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ web-ext/
 
 yarn.lock
 .DS_Store
+tmp/
--- a/scripts/download-import-schema-from-gecko-dev
+++ b/scripts/download-import-schema-from-gecko-dev
@@ -1,0 +1,122 @@
+#!/usr/bin/env node
+
+const path = require('path');
+
+const shell = require('shelljs');
+const tmp = require('tmp-promise');
+
+// auto-cleanup when the script is exiting.
+tmp.setGracefulCleanup();
+
+// Fail as soon as a shelljs command fails.
+shell.config.fatal = true;
+
+const GIT_CONFIG = {
+  repo: 'https://github.com/mozilla/gecko-dev',
+  branch: 'beta',
+  paths: [
+    'browser/config/version.txt',
+    'browser/config/version_display.txt',
+    'toolkit/components/extensions/schemas',
+    'browser/components/extensions/schemas',
+    'mobile/android/components/extensions/schemas',
+  ],
+  partialCloneOptions: '--depth 1 --filter=blob:none --no-checkout',
+  tmpDir: tmp.dirSync({
+    prefix: 'gecko-dev--partial',
+    // Optionally allow to keep the temporary directory for debugging purpose.
+    keep: process.env.KEEP_GECKODEV_TMPDIR === '1',
+    // Allow to cleanup the directory along with the git clone created into it.
+    unsafeCleanup: true,
+  }).name,
+};
+
+function processScriptOptions() {
+  const customBranch = process.argv[2];
+  if (customBranch) {
+    if (!['master', 'beta'].includes(customBranch)) {
+      shell.echo(`ERROR: Invalid gecko-dev branch name "${customBranch}"`);
+      return false;
+    }
+    GIT_CONFIG.branch = customBranch;
+  }
+  return true;
+}
+
+function checkPendingSchemaChanges() {
+  const oldFatal = shell.config.fatal;
+  shell.config.fatal = false;
+  const result = shell.exec(`git diff --exit-code src/schema/imported`, {
+    silent: true,
+  });
+  if (result.code === 1) {
+    shell.echo(
+      'ERROR: Unexpected pending changes in src/schema/imported.\n\n' +
+        'Existing changes should be committed or cleared before importing new schema data:\n'
+    );
+    shell.echo(result.stdout);
+    return false;
+  }
+  shell.config.fatal = oldFatal;
+  return true;
+}
+
+function partialClone() {
+  const { branch, partialCloneOptions, repo, tmpDir, paths } = GIT_CONFIG;
+
+  shell.echo(`Partial cloning gecko-dev branch ${branch}`);
+
+  // Creates the partial clone dir.
+  shell.exec(`git clone -b ${branch} ${partialCloneOptions} ${repo} ${tmpDir}`);
+  // Checkout the working tree paths we need for the import.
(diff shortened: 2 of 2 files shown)


rpl:
@willdurand as a side note: I did also experimented a bit with creating a github workflow around this "git partial clone" approach, my current idea is to have the workflow just run on a weekly schedule and to file an issue if either

- there are some schema changes detected: https://github.com/rpl-limited/gh-workflow-tests/issues/8
- or the import script did throw while trying to import the schema data: https://github.com/rpl-limited/gh-workflow-tests/issues/9
- or the CI job seems to be broken: https://github.com/rpl-limited/gh-workflow-tests/issues/14

and no issue at all if there are no schema changes to import.

This kind of workflow doesn't event strictly need to be part of the addons-linter repo, it doesn't need privileges on the addons-linter repo and so it could as well be defined and running on a weekly schedule from a separate repository (even separate organization).

.gitignore
```diff
@@ -35,3 +35,5 @@ web-ext/
 
 yarn.lock
 .DS_Store
+tmp/
+docs/html/
```
willdurand, line 39: I removed this entry in a previous patch: https://github.com/mozilla/addons-linter/commit/2d1c81e8724bce241e234e40d8567a903715b5e8#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L36, why do we need it again here?

scripts/download-import-schema-from-gecko-dev
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env node
+
+const path = require('path');
+
+const shell = require('shelljs');
+
+// Fail as soon as a shelljs command fails.
+shell.config.fatal = true;
+
+const gitClone = {
```
willdurand, line 10: I'd declare it like this:

```
const GIT_CONFIG = {
```

because (1) it is some config and (2) a global variable modified in `processScriptOptions()`

scripts/download-import-schema-from-gecko-dev
```diff
@@ -0,0 +1,130 @@
...
+  if (shell.exec(`git diff --exit-code src/schema/imported`).code === 0) {
+    shell.echo('No changes to the schema data');
+  }
+  shell.config.fatal = oldFatal;
+}
+
+function cleanTmpDir() {
+  const { tmpDir } = gitClone;
+  // Remove the tempDir by default, unless environment variable is set.
+  if (process.env.KEEP_GECKODEV_TMPDIR == null && tmpDir) {
```
willdurand, line 111: Why do we need such an env var? (also this env var should be checked against `'1'` or something because I'd expect `KEEP_GECKODEV_TMPDIR=0` to skip this block)

scripts/download-import-schema-from-gecko-dev
```diff
@@ -0,0 +1,130 @@
...
+  branch: 'beta',
+  paths: [
+    'browser/config/version.txt',
+    'browser/config/version_display.txt',
+    'toolkit/components/extensions/schemas',
+    'browser/components/extensions/schemas',
+    'mobile/android/components/extensions/schemas',
+  ],
+  partialCloneOptions: '--depth 1 --filter=blob:none --no-checkout',
+  tmpDir: path.join('tmp', `gecko-dev--partial-${Date.now()}`),
```
willdurand, line 21: Maybe having `tmpDir` as an absolute path would make sense here (on some systems, it might be better).

Also, could we use the OS "tmp dir" feature?

willdurand (commented):
I have questions/comments but that looks good enough already.

.gitignore
rpl: eh, `docs/html` actually just slipped into this PR by mistake, it wasn't meant to be there, only `tmp/` should have been added to .gitignore.

scripts/download-import-schema-from-gecko-dev
rpl: :+1: sgtm too, I will update the PR accordingly

scripts/download-import-schema-from-gecko-dev
rpl: by default I prefer that this removes the git clone, but for debugging purpose would be useful to be able to turn off that and be able to see the partial clone content and be able to run the import script manually for further investigations.

Given that it is just for debugging purpose I preferred to allow it through an environment variable instead of a command line option.

I'm definitely ok on doing a more strict check on the value set in the environment variables.

scripts/download-import-schema-from-gecko-dev
rpl: sure, I'm not against that, the other script we are currently using is using the ./tmp directory in the addons-linter working tree and so I opted to just stick to that in this alternative script (mainly because of "weekend lazyness" :-P).

Separately I'll need to double-check if in a github workflow I may need to point TMPDIR elsewhere, but that isn't going to be an issue.

scripts/download-import-schema-from-gecko-dev
```diff
@@ -0,0 +1,122 @@
...
+    return false;
+  }
+  shell.config.fatal = oldFatal;
+  return true;
+}
+
+function partialClone() {
+  const { branch, partialCloneOptions, repo, tmpDir, paths } = GIT_CONFIG;
+
+  shell.echo(`Partial Cloning gecko-dev brach ${branch}`);
```
willdurand, line 67:
```suggestion
  shell.echo(`Partial cloning gecko-dev branch ${branch}`);
```

willdurand (approved):
LGTM, thanks

PR #3873, opened 2021-08-23
feat: Import API schema data for Firefox 92

rpl:
Fixes #3872

-----

This PR includes the following changes to the schema:
- downloads API (added cookieStoreId to the parameters for some of the API methods) - [Bug 1669566](https://bugzilla.mozilla.org/1669566)
- host_permissions support in manifest_version 3 extensions - [Bug 1693385](https://bugzilla.mozilla.org/1693385)

Along with some new tests related to the behavior expected on the `host_permissions` manifest property in manifest_version 2 and manifest_version 3 extensions (the change to the downloads API isn't actually used by the eslint rule that validates the API calls and so it didn't need additional coverage).

PR #3891, opened 2021-09-03
fix: Invalid host_permissions should be reported as warning in MV3 and filtered out in MV2

### LINKED ISSUE #3888: Make invalid values set in host_permissions a "severity: warning" linting result instead of an "severity: error" one
For an invalid value set in permissions we explicitly set the severity to warnings, if Firefox is also just warning on invalid host_permissions values then the linter should also do the same.

rpl:
Followup to #3873, Fixes #3888

This PR includes the following changes:
- two new rules code:  `MANIFEST_HOST_PERMISSIONS` (used to report invalid permission as a warning) and `MANIFEST_BAD_HOST_PERMISSION` (reported as error, only used when a permission doesn't have the expected type string), as the ones we already have for the `permissions` and `optional_permissions` manifest keys 
- ensure that invalid host_permissions (the ones that are still in the expected string form) are reported as warnings (as we do already for `permissions` and `optional_permissions`)
- filter out the additional linting error that the top level anyOf schema entry (recently introduced because of the difference in the schema definition for the "permissions" manifest key in manifest_version 2 and manifest_version 3), because that is redundant (the schema entries nested into it have already reported their validation errors) and it would trigger a JSON_INVALID that is reported as an error even when the nested schema entries reported only warnings
- changes to the tests to properly cover the expected behaviors

As a side note, the `ManifestJSONParser`'s `errorLookup` method could use a refactoring, it has never been that much readable, but it is also increasing in complexity and it would be good to reorganize it to make it easier to read and maintain over the time, but it may be reasonable to defer that to a separate follow up (e.g. to keep this incremental change to fix the permissions validations smaller).

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -76,6 +76,15 @@ export const MANIFEST_BAD_OPTIONAL_PERMISSION = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_BAD_HOST_PERMISSION = {
+  code: 'MANIFEST_BAD_HOST_PERMISSION',
+  message: i18n._('The permission type is unsupported.'),
+  // TODO(https://github.com/mozilla/addons-linter/issues/3893): link host_permissions
+  // MDN doc page here once we have created it.
+  description: i18n._('The permission type is unsupported.'),
+  file: MANIFEST_JSON,
+};
+
 export const MANIFEST_PERMISSIONS = {
   code: 'MANIFEST_PERMISSIONS',
   message: i18n._('Unknown permission.'),
@@ -91,6 +100,15 @@ export const MANIFEST_OPTIONAL_PERMISSIONS = {
   description: i18n._(
     'See https://mzl.la/2Qn0fWC (MDN Docs) for more information.'
   ),
+  file: MANIFEST_JSON,
+};
+
+export const MANIFEST_HOST_PERMISSIONS = {
+  code: 'MANIFEST_HOST_PERMISSIONS',
+  message: i18n._('Invalid host permission.'),
+  // TODO(https://github.com/mozilla/addons-linter/issues/3893): link host_permissions
+  // MDN doc page here once we have created it.
+  description: i18n._('Invalid host permission.'),
   file: MANIFEST_JSON,
 };
 
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -216,6 +216,28 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   errorLookup(error) {
+    if (error.dataPath === '/permissions' && error.keyword === 'anyOf') {
+      // With the addition of the schema data for the manifest_version 3
+      // JSONSchema data, permissions has a top level anyOf schema entry
+      // which include the two alternative set of schema definitions
+      // for manifest_version 2 and manifest_version 3, which will produce
+      // one more validation error in addition to the ones reported by the
+      // manifest_version based entries included into it.
+      //
+      // The validation results from the nested entries are being already reported
+      // before the top level anyOf one and so we can ignore this redundant validation
+      // error.
+      const isManifestVersionAnyOf =
+        error.schema &&
+        error.schema.every(
+          (schema) =>
+            'min_manifest_version' in schema || 'max_manifest_version' in schema
+        );
+      if (isManifestVersionAnyOf) {
+        return null;
+      }
+    }
+
     // This is the default message.
     let baseObject = messages.JSON_INVALID;
 
@@ -293,6 +315,13 @@ export default class ManifestJSONParser extends JSONParser {
     ) {
       baseObject = messages.MANIFEST_BAD_OPTIONAL_PERMISSION;
       overrides.message = `Permissions ${error.message}.`;
+    } else if (
+      error.dataPath.startsWith('/host_permissions') &&
+      typeof error.data !== 'undefined' &&
+      typeof error.data !== 'string'
+    ) {
+      baseObject = messages.MANIFEST_BAD_HOST_PERMISSION;
+      overrides.message = `Permissions ${error.message}.`;
     } else if (error.keyword === 'type') {
       baseObject = messages.MANIFEST_FIELD_INVALID;
     }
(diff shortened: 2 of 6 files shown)


src/messages/manifestjson.js
```diff
@@ -76,6 +76,15 @@ export const MANIFEST_BAD_OPTIONAL_PERMISSION = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_BAD_HOST_PERMISSION = {
+  code: 'MANIFEST_BAD_HOST_PERMISSION',
+  message: i18n._('The permission type is unsupported.'),
+  // TODO: File and reference here an issue to update this description with a link to the
```
willdurand, line 82: Is there an issue filed yet? Maybe we should add a link instead of a long comment here.

src/messages/manifestjson.js
```diff
@@ -94,6 +103,15 @@ export const MANIFEST_OPTIONAL_PERMISSIONS = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_HOST_PERMISSIONS = {
+  code: 'MANIFEST_HOST_PERMISSIONS',
+  message: i18n._('Invalid host permission.'),
+  // TODO: File and reference here an issue to update this description with a link to the
```
willdurand, line 109: Same here.

src/parsers/manifestjson.js
```diff
@@ -306,13 +335,28 @@ export default class ManifestJSONParser extends JSONParser {
...
       baseObject.code !== messages.MANIFEST_PERMISSION_UNSUPPORTED
     ) {
       baseObject = messages[`MANIFEST_${match[1].toUpperCase()}`];
       overrides.message = oneLine`/${match[1]}: Unknown ${match[1]}
           "${error.data}" at ${match[2]}.`;
     }
 
+    // Make sure we filter out warnings and errors code that should never be reported
+    // on manifest version 2 extensions.
+    const ignored_on_manifestV2 = [
```
willdurand, line 348:
```suggestion
    const ignoredOnMV2 = [
```

willdurand: (it's mostly about using camelcase)

tests/unit/parsers/test.manifestjson.js
```diff
@@ -272,21 +272,14 @@ describe('ManifestJSONParser', () => {
...
-      );
+      expect(manifestJSONParser.collector.errors).toEqual([]);
     });
 
     it('ignores host_permissions manifest key in MV2', () => {
       const addonLinter = new Linter({ _: ['bar'] });
       const json = validManifestJSON({
         manifest_version: 2,
-        host_permissions: ['*://example.org/*'],
+        host_permissions: ['*://example.org/*', 'foo'],
```
willdurand, line 282: Do you really need to change this test case here?

willdurand (commented):
Looks good, thanks.

An additional question: should we update the (generated) docs?

src/messages/manifestjson.js
rpl: @willdurand We don't have that doc page yet on MDN ([1]), and so this TODO was meant to remind me to file an issue in this repo, to follow up once we got a MDN page to link here.

I'm going to file the issue and update the patch to mention it here.

[1]: we have been recently discussing about MV3-related doc page on MDN, but we did not settle yet on the process we will use to gradually introduce them (e.g. I'm not yet sure if we are going to create these pages and land them while MV3 addons can't be yet submitted to AMO or if they will stay in a more long living branch in a [mdn/content](https://github.com/mdn/content) fork until we start accepting MV3 extensions on AMO)

src/parsers/manifestjson.js
rpl: :+1: I'm not sure why I didn't make it in camelcase :shrug:

tests/unit/parsers/test.manifestjson.js
rpl: the additional 'foo' here is to be sure that we are actually ignoring the host_permissions and not reporting any validation error ('foo' is invalid as an host permission, the other one is not), and so technically we may actually remove the other one (`'*://example.org/*'`) and just keep the invalid one (`'foo'`).

rpl:
> Looks good, thanks.
> 
> An additional question: should we update the (generated) docs?

@willdurand eh that's a good point, they should... which is making me think: shouldn't we check as part of the CI job that the rules docs are updated with the content of rules.md and there is no pending change that should have been part of the PR?

willdurand:
> @willdurand eh that's a good point, they should... which is making me think: shouldn't we check as part of the CI job that the rules docs are updated with the content of rules.md and there is no pending change that should have been part of the PR?

I thought that was the case already...

rpl:
> > @willdurand eh that's a good point, they should... which is making me think: shouldn't we check as part of the CI job that the rules docs are updated with the content of rules.md and there is no pending change that should have been part of the PR?
> 
> I thought that was the case already...

Looking to the CI job, that should be the case:
https://github.com/mozilla/addons-linter/blob/69a4a950c831b1398d550e8fda047deb42587299/.circleci/config.yml#L106-L113

rpl:
> > > @willdurand eh that's a good point, they should... which is making me think: shouldn't we check as part of the CI job that the rules docs are updated with the content of rules.md and there is no pending change that should have been part of the PR?
> > 
> > 
> > I thought that was the case already...
> 
> Looking to the CI job, that should be the case:
> 
> https://github.com/mozilla/addons-linter/blob/69a4a950c831b1398d550e8fda047deb42587299/.circleci/config.yml#L106-L113
> 
> @willdurand I'll double-check why that step isn't making this PR to fail as we assume it should.

bah, if I run the job locally (using circleci cli tools and docker) it does fail as expected if changes to docs/index.html were not committed. I'm going to giving another look into why it doesn't seem to fail on circleci as part of the jobs triggered by the pull requests.

src/messages/manifestjson.js
rpl: Updated inline comment to reference #3893.

src/messages/manifestjson.js
```diff
@@ -94,6 +102,14 @@ export const MANIFEST_OPTIONAL_PERMISSIONS = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_HOST_PERMISSIONS = {
+  code: 'MANIFEST_HOST_PERMISSIONS',
+  message: i18n._('Invalid host permission.'),
+  // TODO(#3893): link host_permissions MDN doc page here only we have created it.
```
willdurand, line 108: I like to use actual links because I can click them in my IDE (vim). Not a big deal, though.

willdurand (approved):
thanks

Issue #3909, opened 2021-09-13
Some validation messages for manifest syntax errors could be more accurate


AlexandraMoga:
### Describe the problem and steps to reproduce it:
Run `addons-linter` on an add-on that has the following syntax errors in manifest.json: 

1. `/browser_specific_settings/gecko/id` has the below non-string format:
- first example - only letters:
```
"browser_specific_settings": {
    "gecko": {
      "id": foobar
```
- second example - numbers+letters:
```
"browser_specific_settings": {
    "gecko": {
      "id": 123foobar
```
2. `/applications.gecko.strict_min_version` not in string format:
```
"applications": {
    "gecko": {
      "strict_min_version": 55.0
    }
```

### What happened?
The following error messages are received:
- for `/browser_specific_settings/gecko/id` only letters: 
```
ERRORS:

Code           Message            Description                                  File            Line   Column
JSON_INVALID   Your JSON is not   Unexpected token o in JSON at position 463
               valid.
JSON_INVALID   Your JSON is not   Unexpected token o in JSON at position 463   manifest.json
               valid.
```
-  for `/browser_specific_settings/gecko/id` numbers+letters:
```
ERRORS:

Code           Message            Description                        File            Line   Column
JSON_INVALID   Your JSON is not   Line 8: Unexpected token ILLEGAL
               valid.
JSON_INVALID   Your JSON is not   Line 8: Unexpected token ILLEGAL   manifest.json
               valid.
```
-  `/applications.gecko.strict_min_version` not in string format: `manifestJson.applications.gecko.strict_min_version.split is not a function`

### What did you expect to happen?
For the `gecko/id` errors we should probably return the _'/browser_specific_settings/gecko/id" should be string'_ error message. For the `/applications.gecko.strict_min_version` error we should also have a validation message stating that this value should be in string format.

willdurand:
I personally think the messages are good enough, it is "standard JSON error"-like. @rpl WDYT?

rpl:
@willdurand I agree on the first two, I expect the extension developers to be able to find where the error pretty quickly based on the JSON_INVALID error reported to them.

But we should fix the last one (`manifestJson.applications.gecko.strict_min_version.split is not a function`) because at the moment we are throwing:

```
manifestJson.applications.gecko.strict_min_version.split is not a function
(node:2850623) UnhandledPromiseRejectionWarning: TypeError: manifestJson.applications.gecko.strict_min_version.split is not a function
[...]

PR #3913, opened 2021-09-14
Prevent error when strict_min_version is not a string value

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3909

---

Thanks to @rpl for the sketch of patch.


willdurand:
(the `test-alternate` job is still broken because of [a Node issue](https://github.com/nodejs/node/issues/40030) FWIW)

rpl (approved):
@willdurand thanks!

PR #3940, opened 2021-09-29
fix: Add an INVALID_ZIPFILE lint error on zip file with invalid chars in filenames

rpl:
This is the addons-linter counterpart of the proposed changes to detect and collect a linting error on zip files with invalid characters in zip entries filenames.

NOTE: this can't and shouldn't be merged until mozilla/addons-scanner-utils#190 has been reviewed, merged and released on npm in a new addons-scanner-utils release. 

With this and the related changes to addons-scanner-utils, the linter should report the issue as follows:
```
$ ./bin/addons-linter ../addons-scanner-utils/src/tests/fixtures/io/archive-with-invalid-chars-in-filenames.zip
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code              Message                 Description                                    File   Line   Column
INVALID_ZIPFILE   invalid characters in   The package is invalid. It may contain
                  fileName:               entries using invalid characters, as an
                  path\to\file.txt        example using '\' as a path separator is not
                                          allowed in Firefox. Try to re-zipping your
                                          add-on package and make sure all entries are
                                          using '/' as the path separator.
```

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -52,17 +52,18 @@ Rules are sorted by severity.
 
 ## Package layout
 
-| Message code               | Severity | Description                                         |
-| -------------------------- | -------- | --------------------------------------------------- |
-| `MOZILLA_COND_OF_USE`      | notice   | Mozilla conditions of use violation.                |
-| `FLAGGED_FILE_TYPE`        | notice   | (Binary) Flagged file type found.                   |
-| `FLAGGED_FILE_EXTENSION`   | warning  | Flagged file extensions found                       |
-| `DUPLICATE_XPI_ENTRY`      | warning  | Package contains duplicate entries                  |
-| `ALREADY_SIGNED`           | warning  | Already signed                                      |
-| `COINMINER_USAGE_DETECTED` | warning  | Firefox add-ons are not allowed to run coin miners. |
-| `BAD_ZIPFILE`              | error    | Bad zip file                                        |
-| `FILE_TOO_LARGE`           | error    | File is too large to parse                          |
-| `RESERVED_FILENAME`        | error    | Reserved filename detected.                         |
+| Message code               | Severity | Description                                                                      |
+| -------------------------- | -------- | -------------------------------------------------------------------------------- |
+| `MOZILLA_COND_OF_USE`      | notice   | Mozilla conditions of use violation.                                             |
+| `FLAGGED_FILE_TYPE`        | notice   | (Binary) Flagged file type found.                                                |
+| `FLAGGED_FILE_EXTENSION`   | warning  | Flagged file extensions found                                                    |
+| `ALREADY_SIGNED`           | warning  | Already signed                                                                   |
+| `COINMINER_USAGE_DETECTED` | warning  | Firefox add-ons are not allowed to run coin miners.                              |
+| `DUPLICATE_XPI_ENTRY`      | error    | Package contains duplicate entries.                                              |
+| `INVALID_XPI_ENTRY`        | error    | Package contains invalid entries (e.g. invalid characters in entries path name). |
+| `BAD_ZIPFILE`              | error    | Bad zip file.                                                                    |
+| `FILE_TOO_LARGE`           | error    | File is too large to parse                                                       |
+| `RESERVED_FILENAME`        | error    | Reserved filename detected.                                                      |
 
 ## Type detection
 
--- a/src/linter.js
+++ b/src/linter.js
@@ -5,6 +5,10 @@ import chalk from 'chalk';
 import Dispensary from 'dispensary';
 import { oneLine } from 'common-tags';
 import { lstat } from 'addons-scanner-utils/dist/io/utils';
+import {
+  DuplicateZipEntryError,
+  InvalidZipFileError,
+} from 'addons-scanner-utils/dist/errors';
 import { Directory, Xpi, Crx } from 'addons-scanner-utils/dist/io';
 
 import { terminalWidth } from 'cli';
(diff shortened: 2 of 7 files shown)


docs/rules.md
```diff
@@ -57,9 +57,10 @@ Rules are sorted by severity.
 | `MOZILLA_COND_OF_USE`      | notice   | Mozilla conditions of use violation.                |
 | `FLAGGED_FILE_TYPE`        | notice   | (Binary) Flagged file type found.                   |
 | `FLAGGED_FILE_EXTENSION`   | warning  | Flagged file extensions found                       |
-| `DUPLICATE_XPI_ENTRY`      | warning  | Package contains duplicate entries                  |
 | `ALREADY_SIGNED`           | warning  | Already signed                                      |
 | `COINMINER_USAGE_DETECTED` | warning  | Firefox add-ons are not allowed to run coin miners. |
+| `DUPLICATE_XPI_ENTRY`      | error    | Package contains duplicate entries                  |
+| `INVALID_ZIPFILE`          | error    | Package is invalid. May contain invalid characters  |
```
willdurand, line 63: What's the difference between invalid and bad zipfile?

src/linter.js
```diff
@@ -98,19 +98,40 @@ export default class Linter {
...
+      // given that we are also collecting an explicit linting error.
+      return false;
+    }
+
+    if (err.message.includes(constants.ZIP_LIB_CORRUPT_FILE_ERROR)) {
       this.collector.addError(messages.BAD_ZIPFILE);
       this.print(_console);
-    } else if (this.config.stack === true) {
+      // TODO: Tests currently expect this to throw. Consider if we want to change that
+      // given that we are also collecting an explicit linting error.
```
willdurand, line 122: This is related to my question above (invalid vs bad).

src/messages/layout.js
```diff
@@ -11,6 +11,23 @@ export const DUPLICATE_XPI_ENTRY = {
     and re-zipping your add-on package and try again.`),
 };
 
+export const INVALID_ZIPFILE = {
+  code: 'INVALID_ZIPFILE',
+  // message will be replaced with the InvalidZipFileError message
+  // got from the addons-scanner-utils dependency when we were reading
+  // the zip file entries (in particular this would be triggered by
+  // a zipfile entry using invalid characters, like '\' as a path
+  // reparator, and the undelying yauzl error message follows the
```
willdurand, line 20:
```suggestion
  // separator, and the underlying yauzl error message follows the
```

src/messages/layout.js
```diff
@@ -11,6 +11,23 @@ export const DUPLICATE_XPI_ENTRY = {
...
+  // format:
+  //   `invalid characters in fileNane: nameOfInvalidTheZipFileEntry`
```
willdurand, line 22:
```suggestion
  //   `invalid characters in fileName: nameOfTheInvalidZipFileEntry`
```

src/messages/layout.js
```diff
@@ -11,6 +11,23 @@ export const DUPLICATE_XPI_ENTRY = {
...
+  message: 'Invalid ZIP file',
+  description: i18n._(oneLine`The package is invalid. It may contain
+    entries using invalid characters, as an example using '\\' as a
+    path separator is not allowed in Firefox. Try to re-zipping your
```
willdurand, line 26:
```suggestion
    path separator is not allowed in Firefox. Try to recreate your
```

src/messages/layout.js
```diff
@@ -11,6 +11,23 @@ export const DUPLICATE_XPI_ENTRY = {
...
+    add-on package and make sure all entries are using '/' as the
```
willdurand, line 27:
```suggestion
    add-on package (ZIP) and make sure all entries are using '/' as the
```

tests/unit/test.linter.js
```diff
@@ -330,6 +330,7 @@ describe('Linter', () => {
     addonLinter.collector.addError = sinon.stub();
     addonLinter.print = sinon.stub();
     const expectedError = new Error('DuplicateZipEntry the zip has dupes!');
+    expectedError.name = 'DuplicateZipEntryError';
```
willdurand, line 333: Could we import `DuplicateZipEntryError` from addons-scanner-utils?

willdurand (commented):
Some early feedback, thanks for taking care of this problem.

docs/rules.md
rpl: bad zipfile seems to be associated to a zip file that is so corrupt that we are unable to open it (at least that is what I gather from the message description)

src/messages/layout.js
rpl: :facepalm:

tests/unit/test.linter.js
rpl: @willdurand we can, but I think we need to update the addons-scanner-utils PR to export it (which would make it cleaner). I'm 100% of with making that change to that PR:

willdurand:
Quick update: `addons-scanner-utils` v5 is on its way 🚀

docs/rules.md
rpl: @willdurand I looked a bit more into what would make BAD_ZIPFILE to be reported and how the error flows from yauzl to addons-linter:

- we recognize it based on the yauzl error message we expect to receive (See `ZIP_LIB_CORRUPT_FILE_ERROR` here https://github.com/mozilla/addons-linter/blob/c5975adb97b5f3c8aabbe8140f3973ffa390f1cf/src/const.js#L220-L221)

- we get an error that includes the ZIP_LIB_CORRUPT_FILE_ERROR string from yauzl's `open` callback parameter (unlike the other yauzl errors that we are now going to handle with the changes in this PR, which are received through 'error' events emitted on the zipfile instance, after it has been at least open successfully)

And so, from my perspective BAD_ZIPFILE, DUPLICATE_XPI_ENTRY and the new INVALID_ZIPFILE looks all like different error conditions (and seems reasonable to use separate error codes):
  - BAD_ZIPFILE: for zipfiles that are corrupted and can't be inspected by the linter
  - DUPLICATE_XPI_ENTRu: for zipfiles that can be opened just fine but are considered invalid because there are multiple entries with the exact same filename
  - INVALID_ZIPFILE: for zipfile that can be opened just fine, but some entries use characters that are considered invalid (and Firefox would refuse to load the zip file, even if the file isn't completely corrupted from a linter perspective and the linter was able to inspect the package)

Nevertheless, INVALID_ZIPFILE doesn't look like a good rule code name, it sounds too similar to BAD_ZIPFILE, and so I think that in this PR I should change the new rule name to something that would make it easier to don't confuse differetn errors between them.

I'm going to rename the new one into INVALID_XPI_ENTRY (insted of INVALID_ZIPFILE), because I think that we can assume that the "error" events emitted by yauzl on the zipfile instance after it has been opened successfully should always (or often enough) be related to invalid entries (instead of the entire zip file to be corrupted and the linter being unable to look into its entries).

willdurand: Excellent, thanks!

src/linter.js
```diff
@@ -98,19 +102,42 @@ export default class Linter {
   }
 
   handleError(err, _console = console) {
-    if (err.message.includes('DuplicateZipEntry')) {
+    // The zip files contains invalid entries (likely path names using invalid
+    // characters like '\\'), the linter can inspect the package but Firefox
+    // would fail to load it.
+    if (err instanceof InvalidZipFileError) {
+      this.collector.addError({
+        ...messages.INVALID_XPI_ENTRY,
```
willdurand, line 110: I do wonder if we should rename the exception in addons-scanner-utils now :D

willdurand (approved):
r+, thanks

src/linter.js
rpl: Honestly I was also tempted to propose you that, but in the end from the linter perspective the fact that the rule name doesn't exactly match the addons-scanner-utils error name is annoying but probably not "annoying enough" to justify releasing a new addons-scanner-utils version just to rename the error class.

PR #3941, opened 2021-09-30
Allow to enforce a strict_min_version for a given permission

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3942

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -226,6 +226,6 @@ export const RESTRICTED_HOMEPAGE_URLS = [
   'addons.mozilla.org',
 ];
 
-// This map should contain a permission name as key and a min Firefox version
-// as value (both string values).
-export const RESTRICTED_PERMISSIONS = {};
+// This map should contain entries with a permission name as key and a min
+// Firefox version as value (both string values).
+export const RESTRICTED_PERMISSIONS = new Map([]);
--- a/src/const.js
+++ b/src/const.js
@@ -225,3 +225,7 @@ export const RESTRICTED_HOMEPAGE_URLS = [
   'addons-dev.allizom.org',
   'addons.mozilla.org',
 ];
+
+// This map should contain a permission name as key and a min Firefox version
+// as value (both string values).
+export const RESTRICTED_PERMISSIONS = {};
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -632,28 +632,31 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   validateRestrictedPermissions() {
-    Object.keys(this.restrictedPermissions).forEach((permission) => {
-      const minVersion = this.restrictedPermissions[permission];
+    const permissionsInManifest = (this.parsedJSON.permissions || []).map(
+      (permission) => String(permission).toLowerCase()
+    );
 
-      if (
-        (this.parsedJSON.permissions || [])
-          .map((perm) => perm.toLowerCase())
-          .includes(permission)
-      ) {
-        const { firefoxMinVersion: minVersionSetInManifest } =
-          this.getMetadata();
+    if (permissionsInManifest.length === 0) {
+      return;
+    }
+
+    const { firefoxMinVersion: minVersionSetInManifest } = this.getMetadata();
+
+    for (const permission of this.restrictedPermissions.keys()) {
+      if (permissionsInManifest.includes(permission)) {
+        const permMinVersion = this.restrictedPermissions.get(permission);
 
         if (
           !minVersionSetInManifest ||
-          mozCompare(String(minVersionSetInManifest), minVersion) === -1
+          mozCompare(String(minVersionSetInManifest), permMinVersion) === -1
         ) {
           this.collector.addError(
-            messages.makeRestrictedPermission(permission, minVersion)
+            messages.makeRestrictedPermission(permission, permMinVersion)
           );
           this.isValid = false;
         }
       }
-    });
+    }
   }
 
   async validateIcon(iconPath, expectedSize) {
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -640,7 +640,9 @@ export default class ManifestJSONParser extends JSONParser {
       return;
     }
 
-    const { firefoxMinVersion: minVersionSetInManifest } = this.getMetadata();
+    const minVersionSetInManifest = String(
+      this.getMetadata().firefoxMinVersion
+    );
 
     for (const permission of this.restrictedPermissions.keys()) {
       if (permissionsInManifest.includes(permission)) {
@@ -648,7 +650,7 @@ export default class ManifestJSONParser extends JSONParser {
 
         if (
           !minVersionSetInManifest ||
-          mozCompare(String(minVersionSetInManifest), permMinVersion) === -1
+          mozCompare(minVersionSetInManifest, permMinVersion) === -1
         ) {
           this.collector.addError(
             messages.makeRestrictedPermission(permission, permMinVersion)
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -628,12 +628,16 @@ export default class ManifestJSONParser extends JSONParser {
       this.validateHomePageURL(this.parsedJSON.homepage_url);
     }
 
+    this.validateRestrictedPermissions();
+  }
+
+  validateRestrictedPermissions() {
     Object.keys(this.restrictedPermissions).forEach((permission) => {
       const minVersion = this.restrictedPermissions[permission];
 
       if (
         (this.parsedJSON.permissions || [])
-          .map((permission) => permission.toLowerCase())
+          .map((perm) => perm.toLowerCase())
           .includes(permission)
       ) {
         const { firefoxMinVersion: minVersionSetInManifest } =
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -7,6 +7,7 @@ import { oneLine } from 'common-tags';
 import getImageSize from 'image-size';
 import upath from 'upath';
 import bcd from '@mdn/browser-compat-data';
+import { mozCompare } from 'addons-moz-compare';
 
 import { getDefaultConfigValue } from 'yargs-options';
 import {
@@ -26,6 +27,7 @@ import {
   FILE_EXTENSIONS_TO_MIME,
   STATIC_THEME_IMAGE_MIMES,
   RESTRICTED_HOMEPAGE_URLS,
+  RESTRICTED_PERMISSIONS,
   PERMS_DATAPATH_REGEX,
 } from 'const';
 import log from 'logger';
@@ -88,6 +90,7 @@ export default class ManifestJSONParser extends JSONParser {
       selfHosted = getDefaultConfigValue('self-hosted'),
       schemaValidatorOptions,
       io = null,
+      restrictedPermissions = RESTRICTED_PERMISSIONS,
     } = {}
   ) {
     super(jsonString, collector, { filename });
@@ -119,6 +122,7 @@ export default class ManifestJSONParser extends JSONParser {
         'theme'
       );
       this.io = io;
+      this.restrictedPermissions = restrictedPermissions;
       this._validate();
     }
   }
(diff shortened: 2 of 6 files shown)


src/parsers/manifestjson.js
```diff
@@ -623,6 +627,33 @@ export default class ManifestJSONParser extends JSONParser {
...
+        (this.parsedJSON.permissions || [])
+          .map((perm) => perm.toLowerCase())
+          .includes(permission)
+      ) {
+        const { firefoxMinVersion: minVersionSetInManifest } =
+          this.getMetadata();
+
+        if (
+          !minVersionSetInManifest ||
+          mozCompare(String(minVersionSetInManifest), minVersion) === -1
```
rpl, line 648: As I describe in a bit more detail in another review comment, I think we may want to consider the scenario where the strict_min_version value set was not valid from a JSONSchema perspective, and in that case don't call mozCompare but also collect the additional RESTRICTED_PERMISSION error.

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,33 @@ export default class ManifestJSONParser extends JSONParser {
...
+    Object.keys(this.restrictedPermissions).forEach((permission) => {
+      const minVersion = this.restrictedPermissions[permission];
+
+      if (
+        (this.parsedJSON.permissions || [])
+          .map((perm) => perm.toLowerCase())
+          .includes(permission)
+      ) {
+        const { firefoxMinVersion: minVersionSetInManifest } =
+          this.getMetadata();
```
rpl, line 644: Nit, we may do this once outside of the loop

rpl (changes requested):
@willdurand looks pretty good to me, but I'm marking it as requested changes (mainly to discuss if the scenario where the strict_min_version format is invalid needs some special handling at it seems to me at a first glance, most of other comments are just nits).

src/const.js
```diff
@@ -225,3 +225,7 @@ export const RESTRICTED_HOMEPAGE_URLS = [
   'addons-dev.allizom.org',
   'addons.mozilla.org',
 ];
+
+// This map should contain a permission name as key and a min Firefox version
+// as value (both string values).
```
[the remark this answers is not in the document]
willdurand: True but we don't necessarily have the use case for it.

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,29 @@ export default class ManifestJSONParser extends JSONParser {
     if (this.parsedJSON.homepage_url) {
       this.validateHomePageURL(this.parsedJSON.homepage_url);
     }
+
+    Object.keys(this.restrictedPermissions).forEach((permission) => {
+      const minVersion = this.restrictedPermissions[permission];
+
+      if (
+        (this.parsedJSON.permissions || [])
```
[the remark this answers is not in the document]
willdurand: Right, we could do that once, above the loop.

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,29 @@ export default class ManifestJSONParser extends JSONParser {
     if (this.parsedJSON.homepage_url) {
       this.validateHomePageURL(this.parsedJSON.homepage_url);
     }
+
+    Object.keys(this.restrictedPermissions).forEach((permission) => {
```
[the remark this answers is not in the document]
willdurand: I am not very familiar with Map but I suppose this is fine. Any reason to use that? Is it only for the `keys()` method?

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,29 @@ export default class ManifestJSONParser extends JSONParser {
...
+      const minVersion = this.restrictedPermissions[permission];
```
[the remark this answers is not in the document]
willdurand: Sounds good

src/parsers/manifestjson.js
willdurand: Yep

src/parsers/manifestjson.js
willdurand: If the strict min version is invalid, an error will be reported (invalid strict min version).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,276 @@ describe('ManifestJSONParser', () => {
...
+          alarms: '88.1.3',
+        },
+      ],
+      [
+        'invalid strict_min_version (version below the strict min)',
+        // manifest props
+        {
+          browser_specific_settings: {
+            gecko: {
+              strict_min_version: '88.1.2',
```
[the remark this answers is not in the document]
willdurand: > include in the test case at least one with a non numeric part, because it seems to be technically supported

Yeah, why not. It should be covered by addons-moz-compare already but adding a test case or two is easy.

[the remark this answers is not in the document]
willdurand: > double-check the scenario where the strict_min_version has an invalid format, to be sure that when we got to validate the permissions our assumption would not make us to break instead of behaving as expected (I think if the strict_min_version is invalid we still want to collect an additional RESTRICTED_PERMISSION lint error)

I don't know about collecting the error for the restricted permission. Maybe? There will definitely be an error reported for the invalid strict min version. I will add new tests anyway.

src/parsers/manifestjson.js
willdurand: I added some test cases. Looks like both errors will be reported.

tests/unit/parsers/test.manifestjson.js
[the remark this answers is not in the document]
willdurand: The added test cases show that we collect two errors when strict_min_version is invalid.

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,36 @@ export default class ManifestJSONParser extends JSONParser {
...
+  validateRestrictedPermissions() {
+    const permissionsInManifest = (this.parsedJSON.permissions || []).map(
+      (permission) => String(permission).toLowerCase()
+    );
+
+    if (permissionsInManifest.length === 0) {
+      return;
+    }
+
+    const { firefoxMinVersion: minVersionSetInManifest } = this.getMetadata();
```
rpl, line 643: Nit: we may stringify `minVersionSetInManifest` once here:
```
const minVersionSetInManifest = String(this.getMetadata().firefoxMinVersion);
```

src/parsers/manifestjson.js
```diff
@@ -623,6 +627,36 @@ export default class ManifestJSONParser extends JSONParser {
...
+
+    for (const permission of this.restrictedPermissions.keys()) {
+      if (permissionsInManifest.includes(permission)) {
+        const permMinVersion = this.restrictedPermissions.get(permission);
+
+        if (
+          !minVersionSetInManifest ||
+          mozCompare(String(minVersionSetInManifest), permMinVersion) === -1
```
rpl, line 651: We should add explicit coverage for a non string min version value set in the manifest, which will make us sure that the linter will still complete successfully in that case (with the expected errors).

At the moment, it looks that removing `String(...)` here wouldn't make any of the existing tests to fail.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,337 @@ describe('ManifestJSONParser', () => {
...
+      (title, { manifestProps, restrictedPermissions }) => {
+        const { parser, linter } = validate({
+          manifestProps,
+          restrictedPermissions,
+        });
+        const { errors } = linter.collector;
+
+        expect(parser.isValid).toEqual(false);
+        expect(errors.length).toEqual(1);
+        expect(errors[0].code).toEqual(messages.RESTRICTED_PERMISSION);
```
rpl, line 3331: I don't think we care that the error collect is only one, as long as we got the expected restricted permission one.
I also prefer if we could use `expect(...).toEqual(expect.arrayContaining([...]))` to assert the errors that we expect are there, because when that isn't the case it produces a better error message (which include both the expected values and the actual array elements that didn't match the expectation), while this one wouldn't provide any useful information to investigate a CI job that is failing this test:

```diff
diff --git a/tests/unit/parsers/test.manifestjson.js b/tests/unit/parsers/test.manifestjson.js
index bda9f550..c9e741c9 100644
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -3327,8 +3327,14 @@ describe('ManifestJSONParser', () => {
         const { errors } = linter.collector;

         expect(parser.isValid).toEqual(false);
-        expect(errors.length).toEqual(1);
-        expect(errors[0].code).toEqual(messages.RESTRICTED_PERMISSION);
+        expect(errors).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              code: messages.RESTRICTED_PERMISSION,
+              file: 'manifest.json',
+            }),
+          ])
+        );
       }
     );
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,337 @@ describe('ManifestJSONParser', () => {
...
+      (title, { manifestProps, restrictedPermissions }) => {
+        const { linter } = validate({
+          manifestProps,
+          restrictedPermissions,
+        });
+        const { errors } = linter.collector;
+
+        // We cannot assert `parser.isValid` because some (unrelated) warnings
+        // could set this prop to `false`.
+        expect(errors.length).toEqual(0);
```
rpl, line 3473: Similarly to my previous comment, I would prefer this to be asserting `expect(errors).toEqual([]);` which should provide a more useful test failure message.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,337 @@ describe('ManifestJSONParser', () => {
...
+      }
+    );
+
+    it.each(['1', 'a'])(
```
rpl, line 3477: as anticipated in another review comment, I think we should also add a non string value, to be sure we have at least one test that would fail if we are not stringifying the strict_min_version value got from the parsed (but not necessarily valid) manifest:

e.g.
```
 it.each(['1', 'a', { notAString: true }])(...
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,337 @@ describe('ManifestJSONParser', () => {
...
+            permissions: ['alarms'],
+          },
+          restrictedPermissions: new Map([['alarms', '1.01']]),
+        });
+        const { errors } = linter.collector;
+
+        expect(parser.isValid).toEqual(false);
+        expect(errors.length).toEqual(2);
+        expect(errors[0].code).toEqual(messages.JSON_INVALID.code);
+        expect(errors[1].code).toEqual(messages.RESTRICTED_PERMISSION);
```
rpl, line 3496: As for the other assertion, I'd prefer this to use arrayContaing/objectContaining, and apparently we'll need to account for the other error to be MANIFEST_FIELD_INVALID when the strict_min_version property is not a string:

```diff
@@ -3491,9 +3497,19 @@ describe('ManifestJSONParser', () => {
         const { errors } = linter.collector;

         expect(parser.isValid).toEqual(false);
-        expect(errors.length).toEqual(2);
-        expect(errors[0].code).toEqual(messages.JSON_INVALID.code);
-        expect(errors[1].code).toEqual(messages.RESTRICTED_PERMISSION);
+        const errorOnInvalidStrictVersion = typeof strict_min_version == "string"
+          ? expect.objectContaining({ code: messages.JSON_INVALID.code })
+          : expect.objectContaining({ code: messages.MANIFEST_FIELD_INVALID.code })
+
+        expect(errors).toEqual(
+          expect.arrayContaining([
+            errorOnInvalidStrictVersion,
+            expect.objectContaining({
+              code: messages.RESTRICTED_PERMISSION,
+              file: 'manifest.json',
+            }),
+          ])
+        );
       }
     );
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3185,4 +3185,337 @@ describe('ManifestJSONParser', () => {
...
+      expect(parser.isValid).toEqual(false);
+      expect(errors.length).toEqual(1);
+      expect(errors[0].code).toEqual(messages.RESTRICTED_PERMISSION);
+      expect(errors[0].message).toEqual(oneLine`The "alarms" permission
+        requires "strict_min_version" to be set to "78.1" or above`);
+      expect(errors[0].description).toEqual(oneLine`The "alarms" permission
+        requires "strict_min_version" to be set to "78.1" or above. Please
+        update your manifest.json version to specify a minimum Firefox
+        version.`);
+    });
```
rpl, line 3519: Similarly reasoning and proposed change to these assertions:

```diff
@@ -3508,14 +3524,19 @@ describe('ManifestJSONParser', () => {
       const { errors } = linter.collector;

       expect(parser.isValid).toEqual(false);
-      expect(errors.length).toEqual(1);
-      expect(errors[0].code).toEqual(messages.RESTRICTED_PERMISSION);
-      expect(errors[0].message).toEqual(oneLine`The "alarms" permission
-        requires "strict_min_version" to be set to "78.1" or above`);
-      expect(errors[0].description).toEqual(oneLine`The "alarms" permission
-        requires "strict_min_version" to be set to "78.1" or above. Please
-        update your manifest.json version to specify a minimum Firefox
-        version.`);
+      expect(errors).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({
+            code: messages.RESTRICTED_PERMISSION,
+            message: oneLine`The "alarms" permission
+              requires "strict_min_version" to be set to "78.1" or above`,
+            description: oneLine`The "alarms" permission
+              requires "strict_min_version" to be set to "78.1" or above. Please
+              update your manifest.json version to specify a minimum Firefox
+              version.`,
+          }),
+        ]),
+      );
     });
   });
 });
```

rpl (approved):
@willdurand lgtm, r+wc, where the requested changes are the following:
- additional coverage for non-string strict_min_version in the manifest
- some changes to the assertion to get better failure messages

as optional nit: convert the manifest strict_min_version into a string only once

Issue #3960, opened 2021-10-11
Cannot pass General Tests to upload add-on because it contains ES modules and class fields


sineway:
The are multiple files, but all with the same «symptoms» 

> JavaScript syntax error (Parsing as module error: Unexpected token = at line: 8 and column: 16) (Parsing as script error: 'import' and 'export' may appear only with 'sourceType: module' at line: 4 and column: 1)

[Source file](https://github.com/sineway/emir/blob/master/api/TimeUnit.js)

sineway:
Is there any solution, similar to «add `type: "module"` to manifest» or something? 
I would like to keep my web extension free of JS processing.

sineway:
Should I close duplicate issue at mozilla/addons?

ioanarusiczki:
@willdurand 
The JavaScript syntax error is no longer present with linter 4.0.0 or on AMO dev. 

now:
![dev](https://user-images.githubusercontent.com/33448286/140743942-ca26508f-6e7d-4f92-8ecb-d4ba8a051124.png)

before:
![stage](https://user-images.githubusercontent.com/33448286/140743956-d9696730-4721-45a7-a522-402e2e49e14d.png)


[timeunit.zip](https://github.com/mozilla/addons-linter/files/7496978/timeunit.zip)

Issue #3741, opened 2021-05-30
Allow `await` in global (module) scope


NiklasGollenstede:
### Describe the problem and steps to reproduce it:

Putting an `await` in a global module scope causes a syntax error, failing the validation (and thus submission on AMO).
[proposal-top-level-await](https://github.com/tc39/proposal-top-level-await) was actually just updated to Stage 4, and works in Firefox 89 (which is to be released any day now) and Chrome.
So as far as mainline Firefox is concerned, this *is* valid.

I guess one could use this as a test file `background.js`:
```js
import foo from 'foo';
await null;
export const bar = 42;
```

### What happened?

Critical validation error:
```
JavaScript syntax error
(Parsing as module error: Cannot use keyword 'await' outside an async function at ...)
(Parsing as script error: 'import' and 'export' may appear only with 'sourceType: module' at ...)
```

### What did you expect to happen?

The JS file should be parsable.

### Anything else we should know?

I had actually been holding back on porting my AMD modules to ES modules precisely because of the lack of an `await` before `export`ing. Now I ported my code, and can't upload it.

Rob--W:
We're relying on `eslint` to parse the code. Since top-level-await has been promoted to stage 4, I expect eslint to follow shortly:

- https://github.com/eslint/eslint/issues/13178
- https://github.com/eslint/eslint/discussions/13934

NiklasGollenstede:
Well, that was quick, I assume you had this on your radar already :+1: 

And true, to use "fancy" syntax with `eslint`, you have to use a different parser, like `babel` or `tsc`.
We'll just have to wait ...

Rob--W:
Depends on https://github.com/eslint/eslint/issues/14632

ioanarusiczki:
@willdurand 
I tested with the addon attached:
I'm seeing the JavaScript syntax error present on AMO stage. 

![stage](https://user-images.githubusercontent.com/33448286/140509943-4683f4d5-d7c6-436d-b1f0-735a6172829d.png)

On AMO dev or linter 4.0.0 I don't see any errors being present:

![dev](https://user-images.githubusercontent.com/33448286/140509981-b72df08d-0122-4a61-b203-c40df8305c66.png)

![linter 4 0 0](https://user-images.githubusercontent.com/33448286/140509999-30f5cbf8-b308-4f41-a78e-17c7e149364a.png)



[background.zip](https://github.com/mozilla/addons-linter/files/7485941/background.zip)

Issue #3062, opened 2020-02-28
Validator fails ES2019 public field declaration


erosman:
ref: 75.0a1 (2020-02-27) (64-bit)

Validator fails ES2019 public field declaration even though Firefox works fine with.

![image](https://user-images.githubusercontent.com/6458876/75563215-38d38280-5a5f-11ea-94e0-2c1d1e8ee5f1.png)

Example:
```JavaScript
class MyClass {

  a = 1;
  b = 2;
  c = 3;

}
```

PS. If support was added to Firefox, which version is it from?

Rob--W:
This is not a "variable outside the constructor". It is a (public) field declaration, supported in Firefox as of version 75 (and Chrome 74+), with the relevant bug being https://bugzilla.mozilla.org/show_bug.cgi?id=1499448 . The documentation is at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields#Public_fields (scroll a bit below to "Public instance fields" for your specific case).

erosman:
Thank you Rob. Title updated.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields#Public_fields  states support Firefox 69 !!

Rob--W:
I stand corrected; Public instance fields (what you're using here) have already been supported since 69 (https://bugzilla.mozilla.org/show_bug.cgi?id=1555464).

I had public *static* fields in mind (syntax is `class {static name = value}`), which differs from what you've reported here, and is supported in Firefox 75+ (https://bugzilla.mozilla.org/show_bug.cgi?id=1535804).

lfilho:
Arrived here for the same reason.
Tried running the `web-ext lint --ignore-files "**/.*js"` in an attempt to let the addons-linter only lint the manifes.json file and leave all the .js files to my local eslint setup, but that `--ignore-files` argument is actually applied to `web-ext` and not only to the `lint` subcommand, failing the lint command completely with: `"content script defined in the manifest could not be found at "content_script.js""`.

Right now I'm just stopping to use `web-ext lint` anymore :-/

Ideally addons-linter would be able to identify an eslint already installed in the user's package.json and use that one instead?

Or maybe a even better idea is instead of having a customer linter here, you would create an eslint plugin/preset for WebExtensions instead?

Anyway, thanks!

lfilho:
[...]

PR #3984, opened 2021-10-23
Upgrade ESLint to v8

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3985
Fixes https://github.com/mozilla/addons-linter/issues/3062
Fixes https://github.com/mozilla/addons-linter/issues/3741
Fixes https://github.com/mozilla/addons-linter/issues/3960

---

This PR upgrades `eslint` to v8, which will allow us to support the 2022 syntax.

## TODO

- [x] update `addons-scanner-utils` when it supports ESLint v8
- [x] update `eslint-config-amo` when it supports ESLint v8
- [x] add support for ESLint v8 in https://github.com/mozilla/eslint-plugin-no-unsanitized


tests/unit/scanners/test.html.js
```diff
@@ -102,6 +102,7 @@ describe('HTML', () => {
     expect(linterMessages.length).toEqual(0);
   });
 
+  // eslint-disable-next-line jest/expect-expect
```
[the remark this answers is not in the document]
willdurand: done

package.json
```diff
@@ -87,15 +88,13 @@
...
-    "babel-eslint": "10.1.0",
     "babel-gettext-extractor": "4.1.3",
     "babel-jest": "27.3.1",
     "babel-loader": "8.2.3",
     "comment-json": "4.1.1",
     "doctoc": "2.1.0",
-    "eslint-config-amo": "4.11.0",
+    "eslint-config-amo": "5.1.0",
     "eslint-plugin-async-await": "0.0.0",
-    "eslint-plugin-promise": "5.1.1",
```
willdurand, line 98: not used it seems?

package.json
```diff
@@ -87,15 +88,13 @@
...
     "babel-core": "7.0.0-bridge.0",
-    "babel-eslint": "10.1.0",
     "babel-gettext-extractor": "4.1.3",
     "babel-jest": "27.3.1",
     "babel-loader": "8.2.3",
     "comment-json": "4.1.1",
     "doctoc": "2.1.0",
-    "eslint-config-amo": "4.11.0",
+    "eslint-config-amo": "5.1.0",
     "eslint-plugin-async-await": "0.0.0",
```
willdurand, line 97: I am questioning the usefulness of this plugin: I don't think it is used in most of the other projects and we do use async/await everywhere now. I would like to remove this dev dependency: WDYT?

willdurand:
The last commit about removing `eslint-plugin-promise` shows that pinning the `eslint-plugin-no-unsanitized` version (using a git commit) fixes the issue for this project (i.e. `eslint-plugin-no-unsanitized` with ESLint v8 support works as intended here).

Removing `eslint-plugin-promise` was "needed" because [this plugin does not support ESLint v8 yet](https://github.com/xjamundx/eslint-plugin-promise/issues/218) and `npm` fails in some cases because of the peer dependency mismatch. That's annoying because we are blocked by this dev dependency now.

Apart from that, I'd like to remove `eslint-plugin-async-await`, which I don't find particularly useful. Thoughts?

willdurand:
re-requesting a review not for the review itself but to make sure you saw the last comment.

willdurand:
I switched to v4.0.0 of the `eslint-plugin-no-unsanitized` plugin.

willdurand:
@rpl for when you have time, this is ready for a review with two caveats:

- `eslint-plugin-promise` removed because it is currently not compatible
- `eslint-plugin-async-await` isn't removed yet but I would like to remove it, see: https://github.com/mozilla/addons-linter/pull/3984#discussion_r740788256

I am not sure when `eslint-plugin-promise` will get proper support for ESLint v8 so I am thinking about removing it here now and adding it back in a different patch when it is compatible? That would allow us to move forward given that `eslint-plugin-promise` is a dev dependency (and not a lib we use in the core of the linter).

package.json
rpl: @willdurand it looks that it does only provides two "code formatting" rules, which I would guess that prettier may also be taking care of. Also the package version is v0.0.0 and last release 5 years ago, definitely :+1:  on my side on getting rid of it.

rpl:
> @rpl for when you have time, this is ready for a review with two caveats:
> 
>     * `eslint-plugin-promise` removed because it is currently not compatible
> 
>     * `eslint-plugin-async-await` isn't removed yet but I would like to remove it, see: [Upgrade ESLint to v8 #3984 (comment)](https://github.com/mozilla/addons-linter/pull/3984#discussion_r740788256)
> 
> 
> I am not sure when `eslint-plugin-promise` will get proper support for ESLint v8 so I am thinking about removing it here now and adding it back in a different patch when it is compatible? That would allow us to move forward given that `eslint-plugin-promise` is a dev dependency (and not a lib we use in the core of the linter).

100% agree on `eslint-plugin-async-await` (https://github.com/mozilla/addons-linter/pull/3984#discussion_r741861581).

For `eslint-plugin-promise` instead, it would be nice to bring it back as soon as we can, I'm ok to remove it to unblock us to upgrade to eslint v8 along with filing a follow up issue to bring it back.

package.json
```diff
@@ -43,9 +43,10 @@
   },
   "homepage": "https://github.com/mozilla/addons-linter#readme",
   "dependencies": {
+    "@babel/eslint-parser": "7.15.8",
```
rpl, line 46: It looks that "babel-eslint" used to be a dev dependency, while "@babel/eslint-parser" is listed as a direct dependency, 

Based on the content of this PR it looks that @babel/eslint-parser it is still only used while running eslint on the addons-linter sources for development purpose, and it may still not be needed as a strict dependency for an addons-linter user.

is there something else that I may be missing to notice?

rpl (changes requested):
@willdurand lgtm, just one detail related to @babel/eslint-parser being listed as a dependency instead of a dev dependency. Everything else looks good to me as is.

package.json
willdurand: No, that was a mistake. Good catch, thanks.

willdurand:
I moved the babel parser to the list of dev dependencies and removed the async-await eslint plugin. I also filed an issue for the eslint-plugin-promise: https://github.com/mozilla/addons-linter/issues/4016

rpl (approved):
Thanks!

Issue #3911, opened 2021-09-13
Search extensions fail validation if search parameter values are localized messages


aminomancer:
I uploaded some extensions to the add-on dev hub as normal. I'll just discuss one example for simplicity's sake.

This extension's manifest is very similar to Firefox's [built-in search extensions](https://searchfox.org/mozilla-central/source/browser/components/search/extensions/amazondotcom/manifest.json), because my intention was to replace the built-in engines with my own engines where I can control the icon. Some of the built-in engines are fine and have x-suggestions URLs (like Amazon) though others lack suggestions when an API is actually available. I've been meaning to file a bug about this on Firefox but it slipped my mind since I don't know what component to file it under.

I also wanted to replace the icons so they look consistent with built-in icons. That is, replace colorful ICO images with SVG files with context properties, so the icon can respond to themes.

Anyway, here is the form of my version of the Amazon extension. As you can see it's nearly identical to the built-in extension.
```
{
    "name": "Amazon (Search)",
    "description": "__MSG_extensionDescription__",
    "manifest_version": 2,
    "version": "1.3",
    "applications": {
        "gecko": {
            "id": "amazondotcom@aminomancer"
        }
    },
    "hidden": true,
    "default_locale": "en",
    "icons": {
        "16": "favicon.svg"
    },
    "web_accessible_resources": ["favicon.svg"],
    "chrome_settings_overrides": {
        "search_provider": {
            "keyword": "@ama",
            "name": "Amazon",
            "search_url": "__MSG_searchUrl__",
            "search_form": "__MSG_searchForm__",
            "suggest_url": "__MSG_suggestUrl__",
            "search_url_get_params": "__MSG_searchUrlGetParams__"
        }
    }
}
```
And here are some of those localized messages:
```
{
  "extensionName": {
    "message": "Amazon.com"
  },
  "extensionDescription": {
    "message": "Amazon.com Search"
  },
  "searchUrl": {
    "message": "https://www.amazon.com/exec/obidos/external-search/"
  },
  "searchForm": {
    "message": "https://www.amazon.com/exec/obidos/external-search/?field-keywords={searchTerms}&ie={inputEncoding}&mode=blended&tag=mozilla-20&sourceid=Mozilla-search"
  },
  "suggestUrl": {
    "message": "https://completion.amazon.com/search/complete?q={searchTerms}&search-alias=aps&mkt=1"
  },
  "searchUrlGetParams": {
[...]

PR #1798, opened 2018-01-30
Firefox schema 59 import

mstriemer:
This adds ajv-merge-patch to support `$import` by rewriting it as a `$merge`. The structure of the manifest.json file changed in Firefox to be a ManifestBase type which gets imported by WebExtensionManifest and WebExtensionLangpackManifest.

The $merge creates some noisy errors, so those are ignored. The update we have for manifest.json had to get split up for ManifestBase and WebExtensionManifest, too.

Fixes #1378.

PR #4002, opened 2021-10-28
Add support and test coverage for /developer/{url,name}

### LINKED ISSUE #4008: Add better support for /developer attrs in manifest.json
As documented in https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/developer, specifically:

> The object, and both of its properties, are optional. The "name" and "url" properties, if present, will override the  author and homepage_url keys, respectively

willdurand:
Fixes #4008

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -624,6 +624,18 @@ export default class ManifestJSONParser extends JSONParser {
       }
     }
 
+    if (this.parsedJSON.developer) {
+      const { name, url } = this.parsedJSON.developer;
+
+      if (name) {
+        this.parsedJSON.author = name;
+      }
+
+      if (url) {
+        this.parsedJSON.homepage_url = url;
+      }
+    }
+
     if (this.parsedJSON.homepage_url) {
       this.validateHomePageURL(this.parsedJSON.homepage_url);
     }
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -61,6 +61,21 @@
                       ]
                     },
                     "uniqueItems": true
+                  },
+                  "developer": {
+                    "properties": {
+                      "url": {
+                        "format": "ignore",
+                        "oneOf": [
+                          {
+                            "format": "url"
+                          },
+                          {
+                            "pattern": "^__MSG_.*?__$"
+                          }
+                        ]
+                      }
+                    }
                   }
                 }
               }
@@ -72,7 +87,6 @@
         "properties": {
           "homepage_url": {
             "format": "ignore",
-            "type": "string",
             "oneOf": [
               {
                 "format": "url"
--- a/tests/unit/schema/test.homepage_url.js
+++ b/tests/unit/schema/test.homepage_url.js
@@ -5,36 +5,34 @@ import { validateAddon } from 'schema/validator';
 import { validManifest } from './helpers';
 import { assertHasMatchingError } from '../helpers';
 
-describe('/homepage_url', () => {
-  const validURLs = [
-    'https://example.com/some/page',
-    'http://foo.com',
-    '__MSG_foo^&#__',
-  ];
+// eslint-disable-next-line jest/no-export
+export const VALID_URLS = [
+  'https://example.com/some/page',
+  'http://foo.com',
+  '__MSG_foo^&#__',
+];
 
-  validURLs.forEach((validURL) => {
-    it(`${validURL} should be valid`, () => {
-      const manifest = cloneDeep(validManifest);
-      manifest.homepage_url = validURL;
-      validateAddon(manifest);
-      expect(validateAddon.errors).toBeNull();
-    });
+// eslint-disable-next-line jest/no-export
+export const INVALID_URLS = ['__MSG_', 'wat'];
+
+describe('/homepage_url', () => {
+  it.each(VALID_URLS)('%s should be valid', (validURL) => {
+    const manifest = cloneDeep(validManifest);
+    manifest.homepage_url = validURL;
+    validateAddon(manifest);
+    expect(validateAddon.errors).toBeNull();
   });
 
-  const invalidURLs = ['__MSG_', 'wat'];
-
-  invalidURLs.forEach((invalidURL) => {
-    it(`${invalidURL} a URI should be invalid`, () => {
-      const manifest = cloneDeep(validManifest);
-      manifest.homepage_url = invalidURL;
-      validateAddon(manifest);
-      assertHasMatchingError(validateAddon.errors, {
-        dataPath: '/homepage_url',
-        message: /should match format "url"/,
-      });
-      assertHasMatchingError(validateAddon.errors, {
-        message: 'should match pattern "^__MSG_.*?__$"',
-      });
+  it.each(INVALID_URLS)('%s should be invalid', (invalidURL) => {
+    const manifest = cloneDeep(validManifest);
+    manifest.homepage_url = invalidURL;
+    validateAddon(manifest);
+    assertHasMatchingError(validateAddon.errors, {
+      dataPath: '/homepage_url',
+      message: /should match format "url"/,
+    });
+    assertHasMatchingError(validateAddon.errors, {
+      message: 'should match pattern "^__MSG_.*?__$"',
     });
   });
 });
--- a/docs/index.html
+++ b/docs/index.html
@@ -558,7 +558,7 @@
 <tr>
 <td><code>RESTRICTED_HOMEPAGE_URL</code></td>
 <td>error</td>
-<td>Linking to addons.mozilla.org in <code>homepage_url</code> is not allowed</td>
+<td>Linking to addons.mozilla.org in <code>homepage_url</code> or <code>developer.url</code> is not allowed</td>
 </tr>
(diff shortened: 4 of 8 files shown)


willdurand:
@wagnerand looks like we didn't support the `developer` object correctly, which is what this PR tries to fix. There will be a small patch to read `developer.url` in addons-server too, assuming that's what we want. Before I finish this patch, WDYT?

wagnerand:
Looks good to me!

src/parsers/manifestjson.js
```diff
@@ -624,6 +624,18 @@ export default class ManifestJSONParser extends JSONParser {
       }
     }
 
+    if (typeof this.parsedJSON.developer === 'object') {
+      const { name, url } = this.parsedJSON.developer;
+
+      if (name) {
+        this.parsedJSON.author = name;
+      }
```
willdurand, line 632: This isn't strictly needed because I don't think we do anything with the `author` yet but it is there to mimic FF's behavior.

willdurand:
(rebased)

src/schema/updates/manifest.json
```diff
@@ -61,6 +61,20 @@
                       ]
                     },
                     "uniqueItems": true
+                  },
+                  "developer": {
+                    "properties": {
+                      "url": {
+                        "oneOf": [
```
rpl, line 68: I see that homepage_url does have a slightly different JSONSchema:
https://github.com/mozilla/addons-linter/blob/4b80f2b21a3aebe4b6bf6ca2e47cf5f781ab1c41/src/schema/updates/manifest.json#L87-L98
(and obviously also the same in the imported one)

In particular, the property `"type": "string"` shouldn't matter because it is already part of the imported schema data (and actually I think it was likely not necessary, and probably not advisable, to overwrite that from the schema updates), but I'm wondering if `"format": "ignore"` does serve any actual purpose (I haven't digged enough to collect some more of the historical reasons, but that seems to have been added in the initial PR https://github.com/mozilla/addons-linter/pull/1798).

The test coverage seems to suggest that it may not really matter, but may be worth to double-check that.

tests/unit/schema/test.homepage_url.js
```diff
@@ -5,36 +5,34 @@ import { validateAddon } from 'schema/validator';
...
-  const validURLs = [
-    'https://example.com/some/page',
-    'http://foo.com',
-    '__MSG_foo^&#__',
-  ];
+// eslint-disable-next-line jest/no-export
+export const VALID_URLS = [
+  'https://example.com/some/page',
+  'http://foo.com',
+  '__MSG_foo^&#__',
```
rpl, line 12: do we have any test that is asserting that if the url is localized and the localized string is one of the restricted ones then we do mark the property as invalid as expected? (I haven't checked if it is abstracted out and not immediately visible)

rpl (commented):
@willdurand this PR looks pretty good to me too, just some small nits and a couple of questions/stuff I think it may be worth to double-check.

src/schema/updates/manifest.json
willdurand: Yeah, `format: ignore` is used to override the `format` of the `homepagee_url`: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/manifest.json#78 (which is set to `url`).

I didn't add `format: ignore` to `/developer/url` because `format` isn't defined: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/manifest.json#296-300

tests/unit/schema/test.homepage_url.js
willdurand: I don't see anything in https://github.com/mozilla/addons-linter/pull/925 or https://github.com/mozilla/addons-linter/pull/3415

willdurand: I don't think we actually support this case at all based on the existing code (in the main branch). I don't see a lot of code around localized strings either, though.

src/schema/updates/manifest.json
rpl: I see, that seems to be one more detail that has been overlooked when we added the JSONSchema for the `developer` manifest properties.

I did some more digging into the "historical context" and here is what I was able to gather:
- `format: "url"` has been added along with the initial definition of `homepage_url` as part of [Bug 1244805](https://bugzilla.mozilla.org/show_bug.cgi?id=1244805), but based on what I can see in https://hg.mozilla.org/mozilla-central/rev/184854a1ca0e the addition of the `homepage_url` seems to have been slipped into that patch, but without any clear connection to the bug and patch it is part of
- then after ~1y the definition of the "developer" manifest property has been added in [Bug 1282977](https://bugzilla.mozilla.org/show_bug.cgi?id=1282977) but without a `format: "url"` (and without putting it into the ManifestBase as it should have been, as we learned recently)

The result is that:
- setting an invalid url in homepage_url makes the addon invalid because it does not pass the `format: "url"` validation
- setting an invalid url in developer.url is ignored, and it does silently override a valid homepage url if one was specified, and the invalid url ends up in about:addons (in the addon details view, as a link with the non validated url set as the tag href)

This seems wrong and something we may want to also fix on the Firefox side, which means that we may add the missing `format: "url"` property in a future import.

At this point, I think it may be reasonable to set `format: "ignore"` is the schema updates for `developer.url` now in preparation for the schema import that is going to import that change at some point (e.g. when we will import 96 schema if we proceed with fixing it on the Firefox side, maybe after running a code scan on the already submitted addon to double-check if any already signed addon have a developer.url property that would prevent the addon to install successfully once we will properly validate that property on the Firefox side).

@willdurand wdyt?

tests/unit/schema/test.homepage_url.js
rpl: yeah, I also just took a look and "preprocess" isn't considered at all at the moment and even besides that it doesn't seems we do have anything in the linter that is  taking care of preprocessing the manifest properties that can technically be localized (and then make sure the resulting manifest values are still valid).

Let's leave that as a separate follow up.

I looked into the existing and closed issues and I couldn't find any related to preprocess localized manifest properties to ensure the localized value is valid from a linter perspective, and so we should file a new one.

src/schema/updates/manifest.json
willdurand: I agree. I added the `format: ignore`. I filed a bug for FF: https://bugzilla.mozilla.org/show_bug.cgi?id=1739602

This part is still unclear to me but I guess we can discuss it in the bug:

> maybe after running a code scan on the already submitted addon to double-check if any already signed addon have a developer.url property that would prevent the addon to install successfully once we will properly validate that property on the Firefox side

tests/unit/schema/test.homepage_url.js
willdurand: > Let's leave that as a separate follow up.

+1

This reminds me of https://github.com/mozilla/addons-linter/issues/3911

src/schema/updates/manifest.json
rpl: @willdurand I think that the `"format": "ignore"` is currently in the wrong place (it should be in the `developer.url` sub-property instead of being set on the `developer` property)

Besides that, I think it may be good to also:
- remove the `"type": "string"` override from `"homepage_url` (it will not make any difference in the imported schemas, but as previously discussed it seems wrong to override the type from the schema updates)
- refresh the imported schema (to avoid the change to the manifest.json schema update to become part of the Firefox 95 schema import, where we didn't actually do any change on this `developer.url` property on the Firefox side, not a big deal but it would avoid confusion and it would be a bit more consistent)

Here is a diff of the changes needed to update both the updates schema and imported schemas as described above (I had a local mozilla-unified close locally, and so I switched that to release 94 and refreshed the schema by re-importing the schemas from that version to be able to get a clean patch with no other changes):

```diff
diff --git a/src/schema/imported/manifest.json b/src/schema/imported/manifest.json
index 785d4620..6a9640e7 100644
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -338,6 +338,7 @@
                     "url": {
                       "type": "string",
                       "preprocess": "localize",
+                      "format": "ignore",
                       "oneOf": [
                         {
                           "format": "url"
diff --git a/src/schema/updates/manifest.json b/src/schema/updates/manifest.json
index 015cdbfa..d53ac422 100644
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -63,9 +63,9 @@
                     "uniqueItems": true
                   },
                   "developer": {
-                    "format": "ignore",
                     "properties": {
                       "url": {
+                        "format": "ignore",
                         "oneOf": [
                           {
                             "format": "url"
@@ -87,7 +87,6 @@
         "properties": {
           "homepage_url": {
             "format": "ignore",
-            "type": "string",
             "oneOf": [
               {
                 "format": "url"
```

Would you mind to apply this patch and update this PR?

rpl (changes requested):

willdurand:
> @willdurand I think that the "format": "ignore" is currently in the wrong place (it should be in the developer.url sub-property instead of being set on the developer property)

🤦  sorry... I applied your diff, thanks

rpl (approved):
lgtm

Issue #3601, opened 2021-03-08
Reduce dependencies


fregante:
`addons-linter` has [280 dependencies](https://npm.broofa.com/?q=addons-linter@2.13.1) ([47MB on disk](https://packagephobia.com/result?p=addons-linter)), which are almost half of [web-ext’s 610 dependencies](https://npm.broofa.com/?q=web-ext@5.5.0) ([72MB on disk](https://packagephobia.com/result?p=web-ext))

That's a lot.

Some dependencies can't be dropped, like eslint’s [108 dependencies](https://npm.broofa.com/?q=eslint@7.21.0) / [12MB](https://packagephobia.com/result?p=eslint), but others can.

I already sent #3600, hopefully I can send more, also improving sub-dependencies:

- [x] #3600 (10 dependencies -> 2)
- [ ] [mozilla/dispensary#864](https://github.com/mozilla/dispensary/issues/864) (80 dependencies -> 0)
- [x] [pierreinglebert/json-merge-patch#18](https://github.com/pierreinglebert/json-merge-patch/pull/18) (13 dependencies -> 1)
	- [x] Sub-dependency of [ajv-validator/ajv-merge-patch#45](https://github.com/ajv-validator/ajv-merge-patch/pull/45)
- [x] [sindresorhus/os-locale/pulls (author:fregante)](https://github.com/sindresorhus/os-locale/pulls?q=is%3Apr+author%3Afregante+drop) (26 dependencies -> 2)
- [x] [saadtazi/firefox-profile-js/pulls (author:fregante)](https://github.com/saadtazi/firefox-profile-js/pulls?q=is%3Apr+author%3Afregante+drop) (60 dependencies -> 12)

fregante:
Also related: https://github.com/mozilla/dispensary/issues/864

willdurand:
@fregante I see dispensary being mentioned here, I'll take a look as well but this week has been pretty busy for me. Thanks for your work!

fregante:
Wow I didn't know you could assign non-members to issues 😅 unless I'm suddenly a Mozilla member without realizing

willdurand:
haha, sorry. If you don't want to be assigned, it's also fine but we use assignees to know that a bug is actively worked on and give attribution during our standups as well.

fregante:
Yeah no worries, I assumed that was the case

fregante:
[...]

Issue #869, opened 2016-08-03
Move dispensary into the linter


tofumatt:
No one other than us seems to use https://github.com/mozilla/dispensary as a dependency. Updating it and the linter at the same time is a bit of a pain, so we should move dispensary into the linter to make updates easier.

ioanarusiczki:
@willdurand In case this can be tested let me know.

willdurand:
@ioanarusiczki yeah, you can submit an extension with a known file from the list: https://github.com/mozilla/addons-linter/blob/master/src/dispensary/hashes.txt

AlexandraMoga:
@willdurand I've tested with a few of the files mentioned in https://github.com/mozilla/addons-linter/blob/master/src/dispensary/hashes.txt and compared the validation results from -dev and prod. There were no differences as far as I could tell.

Here are some examples:

- validation messages when certain `.js` libraries are identified in the archive(*)
![image](https://user-images.githubusercontent.com/31961530/141108681-b8841ad9-2f44-4ccc-89b8-621284727007.png)
 *a note on angular - I've checked a few libraries and for all of them I've received the "Banned js library" error:
![image](https://user-images.githubusercontent.com/31961530/141109076-6c2277d7-ccfc-4188-90c7-a3334ddc70c3.png)

- validation messages for invalid syntax:
![image](https://user-images.githubusercontent.com/31961530/141109347-0a52d5f4-4875-4173-bebe-c73005cc85ea.png)

willdurand:
> @willdurand I've tested with a few of the files mentioned in [`master`/src/dispensary/hashes.txt](https://github.com/mozilla/addons-linter/blob/master/src/dispensary/hashes.txt) and compared the validation results from -dev and prod. There were no differences as far as I could tell.

No difference = good :)

Issue #3995, opened 2021-10-27
npm deprecated warnings on install


willdurand:
We use a bunch of deprecated libraries/versions in this project and we might be able to fix some of them, either by adopting new libraries or getting rid of them.

```
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated gfm.css@1.1.2: Package no longer supported. Please use github-markdown-css instead.
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated codecov@3.8.3: https://about.codecov.io/blog/codecov-uploader-deprecation-plan/
```

Note: `codecov` will likely be updated anyway because we have no other choice.

willdurand:
> npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
> npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

should be fixed by https://github.com/mozilla/addons-linter/issues/869

willdurand:
That has been fixed.

PR #4006, opened 2021-11-01
Merge mozilla/dispensary into this project

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/869
Closes https://github.com/mozilla/dispensary/issues/864
Closes https://github.com/mozilla/addons-linter/issues/3601

---

First commit: This is an almost 1:1 import of mozilla/dispensary, with tests, docs and the CLI (`scripts/dispensary`). Except one test case in `dispensary/test.index.js` (for the cached libraries, rewritten using async/await and jest spyOn), only Prettier has been applied and paths have been adjusted. There is no other code change.

Second commit: Split the logic of the dispensary so that we have a "runtime" piece for addons-linter and another piece (updater) to update the dispensary hashes using `scripts/dispensary`.

Third commit: Simplify the dispensary updater to only do one thing: compute the hashes for the libraries in the json file, and output that to stdout.

Fourth commit: Removes `request` dep, replaced by `node-fetch`.

Fifth commit: Improves test coverage. Should be good now.

Last commit: add npm script to update the hashes.txt file


README.md
```diff
@@ -222,6 +222,38 @@ We deploy to npm automatically using Circle CI. To release a new version, increm
...
+- http://prototypejs.org/
+- https://github.com/facebook/react/releases
+- https://github.com/jashkenas/underscore/releases
+- https://github.com/mozilla/webextension-polyfill/releases
+
+3. On each page, check whether there are newer release versions than what is in `src/dispensary/libraries.json`. Note that some libraries, like react, support several versions, so we need to check each "branch".
+4. For major upgrades, take a quick look at the code changes
+5. Add new versions to `src/dispensary/libraries.json`
+6. Run `scripts/dispensary > src/dispensary/hashes.txt`
+7. Commit the changes in `src/libraries.json`and `src/hashes.txt`
```
willdurand, line 254: TODO: fix paths here

README.md
```diff
@@ -222,6 +222,38 @@ We deploy to npm automatically using Circle CI. To release a new version, increm
...
+- https://download.dojotoolkit.org/
+- https://github.com/cure53/DOMPurify/releases
+- https://github.com/jquery/jquery/releases
+- https://github.com/jquery/jquery-ui/releases
+- https://github.com/moment/moment/releases
+- https://github.com/mootools/mootools-core/releases
+- http://prototypejs.org/
+- https://github.com/facebook/react/releases
+- https://github.com/jashkenas/underscore/releases
+- https://github.com/mozilla/webextension-polyfill/releases
```
willdurand, line 248: This list was updated with a script in mozilla/dispensary. We could potentially do that again, but maybe in a follow-up?

package.json
```diff
@@ -64,11 +64,14 @@
     "image-size": "1.0.0",
     "is-mergeable-object": "1.1.1",
     "jed": "1.1.1",
+    "natural-compare-lite": "1.4.0",
     "os-locale": "5.0.0",
     "pino": "7.0.5",
     "postcss": "8.3.11",
     "relaxed-json": "1.0.3",
+    "request": "2.88.2",
```
willdurand, line 72: Would be good to get rid of `request`, cf. #3995

src/dispensary/index.js
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  // Output command; this is the default and just echoes out all hashes
+  outputCommand(hashes, _console) {
+    _console.log(hashes.join('\n'));
+
+    return Promise.resolve(hashes);
+  }
+
+  // Update command; this gets all hashes and writes them to hashes.txt
```
willdurand, line 73: This command didn't seem documented in https://github.com/mozilla/dispensary/. The `output` one didn't get much documentation either but that's the default. I think we should make `updateCommand` the default, which will simplify the steps required to update the hashes.

In addition, we should remove the code of the `outputCommand` and maybe simplify the config and `run` logic.

src/main.js
```diff
@@ -18,6 +19,7 @@ export function createInstance({
 }
 
 export default {
+  Dispensary,
```
willdurand, line 22: (`scripts/dispensary` uses the compiled bundle)

package.json
```diff
@@ -48,12 +48,12 @@
     "addons-scanner-utils": "5.0.0",
     "ajv": "6.12.6",
     "ajv-merge-patch": "4.1.0",
+    "async": "3.2.2",
```
willdurand, line 51: We add a few new deps here, which isn't worse than before but maybe we could split the code of the `Dispensary` class to avoid this problem. I don't know "how" exactly, though.

fregante:
On second thought, I think this is great and makes https://github.com/mozilla/dispensary/issues/864 much easier. In fact I think you can just move those dependencies to `devDependencies` instead and be done. 

The problem in that issue was how deal with the change of package type on `npm` (from generator to data-only). This PR kills the `dispensary` package on `npm` so that issue no longer exists.

willdurand:
Requesting pre-reviews to answer a few questions:

- does that approach sound good?
- do we want to update this patch to reduce its size (e.g., by dropping the `command` stuff in the `Dispensary` class)?

src/dispensary/index.js
wagnerand: Or maybe have no default to make commands more explicit?

wagnerand:
This looks good to me overall. I'm all for simplifying things, the command stuff was a bit hard to read if you aren't familiar with the code. Let's also try and keep this low effort.

willdurand:
> This looks good to me overall. I'm all for simplifying things, the command stuff was a bit hard to read if you aren't familiar with the code. Let's also try and keep this low effort.

Thanks. Yeah, I am not sure we need to keep the existing CLI stuff. We might not even need `yargs`. We could just have a `Updater` class or something that reads the `libraries.json` and updates `hashes.txt`. This would be called by the new script and nothing more. This would allow us to move most of the dependencies as dev deps too.

willdurand:
> On second thought, I think this is great and makes [mozilla/dispensary#864](https://github.com/mozilla/dispensary/issues/864) much easier. ~In fact I think you can just move those dependencies to `devDependencies` instead and be done.~ Edit: It looks like `_getCachedHashes` needs to be extracted too, which I can do separately.
> 
> The problem in that issue was _how to deal with the change of package type on `npm`_ (from generator to data-only). This PR kills the `dispensary` package on `npm` so that issue no longer exists.

The second commit should do what you're describing here.

wagnerand:
> > This looks good to me overall. I'm all for simplifying things, the command stuff was a bit hard to read if you aren't familiar with the code. Let's also try and keep this low effort.
> 
> Thanks. Yeah, I am not sure we need to keep the existing CLI stuff. We might not even need `yargs`. We could just have a `Updater` class or something that reads the `libraries.json` and updates `hashes.txt`. This would be called by the new script and nothing more. This would allow us to move most of the dependencies as dev deps too.

Ah I see, that makes sense to. There are only two commands, of which the `update` one is the most important, I think I never used the `output` one. Also, I don't recall ever using one of the arguments.

fregante (approved):
> The second commit should do what you're describing here.

Yessss, LGTM. If merged, it would close:

- https://github.com/mozilla/dispensary/issues/864
- https://github.com/mozilla/addons-linter/issues/3601

willdurand:
The third commit simplifies the dispensary "updater".

willdurand:
>  The diff coverage is `95.77%`.

I can probably write some more tests to reach 100%.

package.json
willdurand: done

README.md
willdurand: Let's file a follow-up for this.

willdurand:
@wagnerand @rpl this is now ready for a review. Each commit should be self-contained, which should help understand which changes have been made (the first one being almost 1:1 as mentioned in the PR description).

src/dispensary/updater.js
```diff
@@ -54,7 +42,8 @@ export default class Updater {
         return this.outputHashes(libraries);
       })
       .then((hashes) => {
-        return this[`${this.command}Command`](hashes, _console);
+        _console.log(hashes.join('\n'));
+        return Promise.resolve(hashes);
```
wagnerand, line 46: Shouldn't we have the logic of the former `updateCommand` here? If I understand correctly, this would not update the hashes file?

tests/unit/dispensary/test.updater.js
```diff
@@ -68,135 +68,75 @@ describe(__filename, () => {
...
-        return dispensary.getFiles(libraries);
+        return updater.getFiles(libraries);
       })
       .then((libraries) => {
-        return dispensary.getHashes(libraries);
+        return updater.getHashes(libraries);
       })
       .then((libraries) => {
-        expect(libraries[0].files.length).toBe(32);
+        expect(libraries[0].files.length).toBe(8);
```
wagnerand, line 116: Why did this change from 32 to 8?

wagnerand (commented):
Looks great to me overall, got a few questions here and there.

Great job! 👏 🏆

src/dispensary/updater.js
willdurand: The docs I ported to the README indicates that we have to redirect the output of the command to the txt file, so I kept that. It is useful because that way we kinda have two options: showing the content of the hashes and updating the file (by redirecting the output).

We can add a npm script to abstract this part so that `npm run update-hashes` updates the txt file directly.

tests/unit/dispensary/test.updater.js
willdurand: I think this was a bug. Reading the code, I don't understand how we could have 32 files given that the fixtures file contains 4 versions for `backbone` with two possible files (minified and "normal"). I think, previously, test cases were leaking because skipping some changed these numbers as well.. That's related to the `_files` hack I believe.

src/dispensary/updater.js
wagnerand: Ah! Yes please on having an npm script. Can we add that to this PR or file a follow-up so we don't forget?

willdurand: added in this patch

wagnerand: Sorry I missed that. Thanks!

willdurand:
(rebased)

src/dispensary/updater.js
```diff
@@ -0,0 +1,249 @@
+/* eslint  import/no-extraneous-dependencies: 0 */
```
rpl, line 1: Instead of disabling the `import/no-extraneuous-dependencies` eslint rule on this file, I would slightly prefer it to live outside of the `src/` directory.

From a bundling perspective it does not make a real difference, webpack treeshaking behavior would be omitting this module entirely from the bundle because none of the src modules being bundled does import this one.

Nevertheless, moving this out of `src/` would make it crystal clear that it is not something to import from another addons-linter source.

We could either:
- move it in the form of an ESM (almost as is) into `scripts/lib` and make the script to import it from there.

- Or we could merge it with `scripts/dispensary` itself if we prefer

Personally I'm fine either way.

------

Merging it with `scripts/dispensary` seems also doable, e.g. by making it to not behave like cli script when imported as a module from the unit tests:
```
if (require.main === module) {
  new Updater().run().catch((err) => {
    console.error('Uncaught error', err);
  });
}
```

but most of the changes to merge it into the `script/dispensary` script may likely be required due to the need to adapt it to a commonjs module, which may be more or less:
```
const path = require('path');
const fs = require('fs');

const async = require('async');
const fetch = require('node-fetch');
const naturalCompare = require('natural-compare-lite');

const { default: log } = require('../src/logger');
const { default: createHash } = require('../src/dispensary/hasher');
const { urlFormat } = require('../src/dispensary/utils');

class Updater {
  ...
}

module.exports = { Updater };
...
```

In both cases the path to the files would need to change to be relative to the `scripts/lib` or `scripts` directory (`DEFAULT_PATH_TO_HASHES` and `DEFAULT_LIBRARY_FILE`).

and I guess also one change on the unit test side:
```
import { Updater } from '../../../scripts/dispensary';
```

Everything I think that would still work by staying exactly the same as in this version.

src/dispensary/index.js
```diff
@@ -0,0 +1,39 @@
+import HASHES from 'dispensary/hashes.txt';
```
rpl, line 1: This import means that the hashes.txt data will be now bundled in the addons-linter repo.

At the moment I think this is fine (even if previously it was likely part of the dispensary bundle and not of the addons-linter bundle), and maybe even good (one less file that the linter would need to access).

And so I think we should just add an inline comment right above `import HASHES ...` just to point this out (e.g. something like `// The entire content of hashes.txt will be part of the addons-linter.js bundle`).

The hashes.txt file is currently ~100kB, and [dist/addons-linter.js bundle file on addons-linter v4.0.0 is 712kB](https://unpkg.com/browse/addons-linter@4.0.0/dist/), that means that I expect the `dist/addons-linter.js` bundle to become ~800kB with this changes, which seems still more than reasonable increase to be ok with.

@willdurand how that sounds? (basically "adding a note in the README.md section and an inline comment right above this line" would be the only changes I'm requresting)

rpl (changes requested):
@willdurand Thank you so much for having looked into this, I'm looking forward to the reductions of the node_modules/ size in web-ext too as a side effect of this changes :partying_face:

This looks pretty great to me, I do have only two small changes to request described below (the one related to the list of urls in the README does not really matter that much, the other two are the ones I mainly care about, they should both require only some small tweaks to what is already in this PR).

src/dispensary/index.js
willdurand: Sounds good, useful comments are always good.

I am fine with the increase of the bundle size. Ultimately, the hashes.txt content is loaded from a JS file eventually. The fact that it is going to be from the addons-linter bundle and not the dispensary bundle is an implementation detail. The same `txt` file was embedded in a bundle before.

src/dispensary/updater.js
willdurand: Yeah, I think I am good with that. I suppose we can merge the two files.

src/dispensary/index.js
rpl: @willdurand :+1: yep, I went a bit more into the detail (just for the sake of leaving a record of what we have evaluated in practice), but ultimately that is exactly what I meant.

README.md
willdurand: https://github.com/mozilla/addons-linter/issues/4030

willdurand:
I think I addressed all the comments (correctly). The command continues to work, tests pass, all other checks seem OK too.

r?

rpl (approved):
Lgtm. Thank you!

PR #4013, opened 2021-11-03
Extract and merge locales

willdurand:
~:warning: depends on #4012~

Fixes https://github.com/mozilla/addons-linter/issues/4014 

---

You should probably only look at the first commit. The second one is proving that things work again and we didn't run the extraction script for a while.


wagnerand (approved):
r+ although I admit I'm not too familiar with babel. If this needs a thorough review, we might want to ask a babel expert to have a second look.

willdurand:
> r+ although I admit I'm not too familiar with babel. If this needs a thorough review, we might want to ask a babel expert to have a second look.

Thanks, it's fine. Extraction wouldn't work with a valid babel/webpack config.

Issue #3972, opened 2021-10-18
Validate `install_origins` property


diox:
Per [Intend to Implement message on firefox-dev](https://groups.google.com/g/firefox-dev/c/U7GpHE4R-ZY), we want the linter to support a new property : `install_origins`, which is an array of origin strings.

For the moment, the property should be completely optional, and only validated if present (in the future we may warn if it's absent). We'll need to import the schema from Firefox.

A few restrictions that might or might not be in the schema:
- Each entry in the array should be a valid [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) (scheme included, and no trailing slash)
  - Wildcards are forbidden
  - ~~Origins in the public suffix list are forbidden (i.e. forbid `com`, allow `foo.com`)~~
  - ~~Origins should be in "canonical form" - punycode version, lowercase, not extra `.` at the end.~~
- Limit to 5 entries

diox:
There will be one soon, I had a head start.

AlexandraMoga:
@diox I've uploaded add-ons on -dev with several variations of `install_origins` values and I have the following results:

- Not accepted:
  - https://example.com/
  - http://www.example.com/
  - www.example.com
  - http://*.example.com
  - *://example.com
  - https://*
  - http://example.com/app2/index.html
  - https://example.com/page_one/page_two?query=something
  - http://example.com/app1


- Accepted:
  - https://example.com
  - http://example
  - http://www.example.com
  - https://myapp.example.com
  - http://example.com:8080
  - https://example.page.io
  - http://example-page-io.com
  - http://example_page_1231io.com
  - https://EXAMPLE.IO
  - http://网站.com
  - http://xn--5tzm5g.com


- Other validation results:
	- max number of allowed origins is 5
	- `install_origins` needs to be an array
	- `install_origins` accepts only string values

- No validation messages right now for:
	- `install_origins` key mispelled, i.e. `install_origin`
	- empty array, i.e. `"install_origins": []`

Let me know if something doesn't match what you've expected.

diox:
> http://example - should this be accepted?

Debatable. This can technically be valid on a local network for instance... Can you file a follow-up and I'll ask for more opinions. In the meantime, please raise an issue if this turns out to cause further issues down the chain like 500 errors or inability to block such origins with [the new admin I just landed](https://github.com/mozilla/addons-server/issues/18331).

PR #4018, opened 2021-11-03
Validate install_origins

diox:
Fixes https://github.com/mozilla/addons-linter/issues/3972


src/schema/formats.js
```diff
@@ -103,6 +103,37 @@ export function isSecureUrl(value) {
...
+    // Path, query string and hash shouldn't be included in origins.
+    // URL().pathname will always be '/' if no path is provided, so we have to
+    // check value doesn't end with one either.
+    return false;
+  }
+  // value === url.origin would be enough if it wasn't for IDNs but
+  //  URL().origin returns punycode, so we have to compare against URL().href
+  // minus the last character instead, having checked that there was no path,
+  // query string or hash earlier.
+  return url.href.slice(0, -1) === url.origin;
```
rpl, line 134: We should keep this as much as possible in sync with what we are going to do on the Firefox side, when the patch on the Firefox side will be landed it may be a good idea to put a permalink to the formatter used in Firefox.

In the meantime, here is the related review comment I added on the Firefox patch:
https://phabricator.services.mozilla.com/D129105#inline-715951

Shane and I haven't touched base about that yet, but on the Firefox side we will proceed to use a Schemas formatter like the one I described in that phabricator comment, then in nodejs I think that it may look more or less like the following code snippet:
```js
const ALLOWED_INSTALL_ORIGINS_PROTOCOLS = ["http:", "https:"];

...

function isValidInstallOriginUrl(value) {
  let origin;
  try {
    const url = new URL(value);
    if (!ALLOWED_INSTALL_ORIGINS_PROTOCOLS.includes(url.protocol)) {
      // URL protocol is not one of the allowed ones (e.g. ["http:", "https:"]).
      return false;
    }
    origin = url.origin;
  } catch (err) {
    // Not a valid url.
    return false;
  }

  const { href: baseURL } = new URL("/", origin);
  // Make sure value URL doesn't include any other URL component besides the origin
  // and path set to "/". 
  return baseURL.slice(0, -1) === value;
} 
```

src/schema/imported/manifest.json
```diff
@@ -780,6 +780,16 @@
...
+        },
+        "install_origins": {
+          "type": "array",
+          "optional": true,
+          "minItems": 1,
+          "maxItems": 5,
+          "items": {
+            "type": "string",
+            "format": "origin"
+          }
```
rpl, line 792: As a side note, as part of the review on the Firefox patch, I did mention that `FirefoxSpecificProperties` JSONSchema doesn't allow arbitrary properties that are not explicitly listed in its schema, and so a manifest.json file including `install_origins` inside `browser_specific_settings.gecko` (or `applications.gecko`) would be invalid for older Firefox releases (summarized from [this comment](https://phabricator.services.mozilla.com/D129105#inline-715990)).

I briefly touched base about that with Shane yesterday, and we agreed to discuss about it more today during out team standup.

And so, for now there are no changes requested, but we may have to change this to match what we will agree on doing on the Firefox side (e.g. we can't exclude yet that we may choose to move the property outside of `FirefoxSpecificProperties`, e.g. in a new top level manifest key where the pre-existing JSONSchema does already allow additional properties).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3601,4 +3601,125 @@ describe('ManifestJSONParser', () => {
...
+      testInstallOrigins(['https://xn--fo-9ja.com'], true);
+    });
+
+    it('allows install origins with IDN origin', () => {
+      testInstallOrigins(['https://foo.bar.栃木.jp'], true);
+    });
+
+    it('allows install origins with port', () => {
+      testInstallOrigins(['https://example.com:8888'], true);
+    });
```
rpl, line 3642: Nit, we may use `it.each` Jest helper as we do for some other similar test cases, e.g. this one related to the homepage_url:
https://github.com/mozilla/addons-linter/blob/563ad4ffa2cb877fbc94c29debf5a3759aad18a9/tests/unit/parsers/test.manifestjson.js#L2977-L2982

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3601,4 +3601,125 @@ describe('ManifestJSONParser', () => {
...
+    it('allows install origins with subdomain origin', () => {
+      testInstallOrigins(['https://foo.example.com'], true);
+    });
+
+    it('allows install origins with punnycode origin', () => {
+      testInstallOrigins(['https://xn--fo-9ja.com'], true);
+    });
+
+    it('allows install origins with IDN origin', () => {
+      testInstallOrigins(['https://foo.bar.栃木.jp'], true);
```
rpl, line 3637: Side note: it would be nice to somehow make it easier for us to keep at least some of the test cases that we will have to run on both Firefox and the addons-linter sides in sync between them.

Sharing the entire test case is not really doable, but may be worth moving the list of test case into arrays (in both the Firefox tests and these tests) and just make sure the list is as in sync as it can be (especially in terms of cases being tested).

rpl (commented):
@diox my apologies for letting you wait that long. 

This PR seems to be defintiely going in the right direction.

Follows a more detailed initial round of feedback comments (included a mention to some details that we may still have to discuss and agree on for the Firefox side of this change, which may require changes to this patch, e.g. if we will end up moving the `install_origins` manifest property).

tests/unit/parsers/test.manifestjson.js
diox: Yeah. Kinda similar to your `it.each` suggestion below, the issue with having arrays containing a bunch of invalid/valid values is that when it's invalid, the reason can be different, and we want to check it... Not sure if it's worth doing several arrays of invalid values, one for each kind of error...

src/schema/imported/manifest.json
rpl: > TIL. We want developers to adopt `install_origins` sooner rather than later, if it makes their add-on incompatible with older Firefox versions that sounds like a significant problem...

It is definitely a significant problem (and I'm really glad that we spotted it earlier in the review passes).
That's basically why yesterday I asked Shane to have a quick preliminary chat about that single point from my review pass on the Firefox patch, and we agreed to discuss about it more during the team standup today.

I'm confident we will get to an agreement quickly, either today or in the worst case by the end of this week (you are also welcome to join us if you want to, I can ping you when we are about to discuss about this point, but in any case I will report back here the outcome of that discussion).

rpl: @diox we discussed about this issue during the meeting (and Shane told me that it has been discussed also in another meeting, I think it was the Ops one) and we agreed to move the key to be a top level manifest key (same name and same JSONSchema definition, just moved to be a top level one).

I'm not yet sure if it is meant to be on the `ManifestBase` one, but it seems a reasonable assumption given that it was in the `FirefoxSpecificProperties`, which are part of the `ManifestBase` schema.

tests/unit/parsers/test.manifestjson.js
willdurand: nit: also if we can avoid `testXXX` and have the assertions directly in the test function, that would avoid some ESLint warnings about tests without assertions :)

tests/unit/parsers/test.manifestjson.js
willdurand: Not sure exactly if that's what you're looking for but we can totally do:

```
it.each([
  { origin: 'xyz', expectedError: /some error/ },
  { origin: 'qbc', expectedError: /some other error/ },
])() => {
});
```

diox: Sure, I can do that. Note that if we want to share the test values then Firefox should also have the expected error code in its tests.

src/schema/imported/manifest.json
```diff
@@ -780,6 +780,16 @@
           "default": "*",
           "description": "Maximum version of Gecko to support. Defaults to '*'. (Requires Gecko 45)",
           "pattern": "^[0-9]{1,3}(\\.[a-z0-9*]+)+$"
+        },
+        "install_origins": {
```
rpl, line 784: @diox given that it looks that we now got to an agreement about the pending issue related to this property [^1], and that I'm already reviewing the patches on the Firefox side, I though to run the schema imports for you and push them into my fork (in a [branch named `pre-review/install-origins`](https://github.com/mozilla/addons-linter/compare/master...rpl:pre-review/install-origins)) to make it easier for you to cherry pick what is needed directly from there, the one that may be more directly useful for your work in this pull request are:
- https://github.com/mozilla/addons-linter/commit/b94bba48e4559fdba9deb26293eb359da3d8d859 (this one is just the schema import of the "install_origins" definition from [D129105](https://phabricator.services.mozilla.com/D129105), other schema unrelated to `install_origins' is in separate patches in that same fork, hopefully you will only need this one to try our your changes)
- https://github.com/mozilla/addons-linter/commit/720066a79a03bdb64d13b1274cc5084d27bca927 (this one contains the small tweak that we will need to add to `src/schema/updates/manifest.json`, plus the changes to `src/schema/updates/manifest.json` got by re-importing the schema again on top of the current version of this patch, which means that it does remove this definition and instead adds "maxItems" to the one added by the schema data import from the other patch + a small tweak to the test case to confirm that everything else works as expected)

Let me know how it goes when you will give it a look to cherry picking what's needed from that branch (especially if there is something that isn't really clear or if it isn't working for you).

I'm going to mark the other comment (https://github.com/mozilla/addons-linter/pull/4018/files/0d3278bc1b2fbaad5431df465d19930830757ac3#r745572036) as resolved.

[^1]: in particular: where the new property should be placed (top level manifest property in the `ManifestBase` schema), its manifest property name (`install_origins`) and what should be the format expected to be enforced (`origin` as defined by the specs), I did

rpl:
(as a side note: I added the label "state: blocked by upstream" as a reminder that this may not be merged as is until the schema changes on the Firefox side got to beta even if we have signed it off already in the meantime, we can evaluate if it is possible and we want/need to split out part of this and landing sooner later on, after the PR is ready or almost ready from the reviewers perspective).

src/schema/imported/manifest.json
rpl: @diox here is a patch for the strategy that we have been discussing over Matrix today (I did rerun the import from the current beta on top of the change to the updates/manifest.json file, and tweaked the test case from your PR to confirm that they all pass when the install_origins is moved to the top level):
https://github.com/mozilla/addons-linter/commit/3df53d6fc6802aae252b57266620d9f448119127

rpl:
I'm going to remove the `state: blocked by upstream` label, the approach we discussed this morning ([^1]), and drafted in the patch linked in https://github.com/mozilla/addons-linter/pull/4018#discussion_r751194894, will allow us to don't block merging this once signed-off.

[^1]: "injecting the entire "install_origin" schema entry from src/updates/manifest.json" will allow us to keep the entry in place on schema data imports that we may have to do before the Firefox release with the same schema data will actually get to beta.

src/schema/formats.js
```diff
@@ -103,6 +103,41 @@ export function isSecureUrl(value) {
...
+  try {
+    url = new URL(value);
+  } catch (e) {
+    // It's invalid or not absolute.
+    return false;
+  }
+  if (value.includes('*')) {
+    // Wildcards are not valid in origins.
+    return false;
+  }
```
rpl, line 118: We may not need to explicitly check for "*" given that this is an origin and we are already going to require that the string is strictly equal to the origin (and with that check we should already excluding a number of url components where `new URL(...)` would not be complaining about a "*"), that should be leaving only protocol, hostname and protocol and new URL should be throwing already with a "*" placed there.

@diox do you see any issue in trying to match even more the kind of checks that the patch on the Firefox side is going to be doing? is there any case covered by this method but not by the one in Firefox?

In the current version of that patch, the origin formatted does look like in the following snippet (pasting here because phabricator does not allow easily to link a certain like in a revision):
```
  origin(string, context) {
    let url;
    try {
      url = new URL(string);
    } catch (e) {
      throw new Error(`Invalid origin: ${string}`);
    }
    if (!/^https?:/.test(url.protocol)) {
      throw new Error(`Invalid origin must be http or https for URL ${string}`);
    }
    // url.origin is punycode so a direct check against string wont work.
    // url.href appends a slash even if not in the original string, we we
    // additionally check that string does not end in slash.
    if (string.endsWith("/") || url.href != new URL(url.origin).href) {
      throw new Error(
        `Invalid origin for URL ${string}, replace with origin ${url.origin}`
      );
    }
    if (!context.checkLoadURL(url.origin)) {
      throw new Error(`Access denied for URL ${url}`);
    }
    return url.origin;
  },
``` 

Besides the returned value (and the call to context.checkLoadURL, which it shouldn't matter nor be needed for the addons-linter validation check), it looks we may be able to derive the addons-linter validation function with a almost 1-1 match with the Firefox checks.

wdyt?

rpl (changes requested):
I'm going to look more closely to the rest of the PR, but in the meantime I wanted to bring up a couple of comments I got based on the last review passes to the patch on the Firefox side that I've been doing today.

src/schema/updates/manifest.json
```diff
@@ -106,6 +106,15 @@
           },
           "manifest_version": {
             "maximum": 2
+          },
+          "install_origins": {
+            "type": "array",
+            "minItems": 1,
```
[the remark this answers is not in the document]
willdurand: Isn't the empty array only for MV2?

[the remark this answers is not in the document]
diox: That's correct, we now want the empty array allowed, this needs to be updated. Eventually, when we start enforcing, AMO will add an empty array to any submission that are missing the property (explicitly disallowing their install from any remote origin other than AMO).

src/schema/formats.js
diox: Yes, I haven't had the time to do that yet but I plan to.

diox: Did that in https://github.com/mozilla/addons-linter/pull/4018/commits/86a1a32e5a314e9a7f3c64b3a02a2bd529c36e7d (and https://github.com/mozilla/addons-linter/pull/4018/commits/104f477a3e7d81d19e595aaa9534ac8ed9d3746c to make `eslint` happy).

I kept the logic but had to make a few changes:
- Changed the `throw`s for `return false`
- Returned `true` instead of `url.origin` at the end
- Dropped the `context` variable and the `context.checkLoadURL` call
- Add something for the wildcards (which is more for convenience so that developers don't try to use them only to realize it doesn't work that way when they try to install their add-on later). 
- Renamed `string` to `value` for consistency with the rest of the file

src/schema/updates/manifest.json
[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-linter/pull/4018/commits/e83e4796e0f48aa302dfc0dd0deeabf5d419abd0

rpl (commented):
This looks pretty good to me.

As a side note, I would really like if we could also assign a more specific rule code for invalid install_origins values, so that we can also make sure to show a message that would help the developer to know how to make the manifest key value valid, e.g. either:
- with a link to a doc page (as we do for permissions and optional permissions invalid values)
- or including the expected value in the error message when possible  as we are doing in the error we are going to log on the Firefox side).

To achieve that I think we may use a strategy similar to the one we use for the validation errors related to the `permissions` and `optional_permissions`, where the error got from the JSONSchema validator is then "changed / improved" in the ManifestJSONParser class (from the [erroLookup method here](https://github.com/mozilla/addons-linter/blob/ed8da2e1aedeceefc5f66d8911c111d699781a17/src/parsers/manifestjson.js#L222)).

Given that the changes to `ManifestJSONParser`'s `errorLookup` would require some more work (and potentially bring some more risks with it, depending from the amount and kind of changes we may have to apply to it), personally I would be ok to defer that to a separate follow up and proceed with this part as is in the short run. 

@willdurand wdyt?

src/schema/formats.js
```diff
@@ -103,6 +103,28 @@ export function isSecureUrl(value) {
   return ['https:', 'wss:'].includes(url.protocol);
 }
 
+export function isOrigin(value) {
+  let url;
+  try {
+    url = new URL(value);
+  } catch (e) {
```
willdurand, line 110: I don't know if our code supports it yet but theoretically we could use a catch without an error if we don't use it:

```suggestion
  } catch {
```

src/schema/formats.js
```diff
@@ -103,6 +103,28 @@ export function isSecureUrl(value) {
...
+    return false;
+  }
+  if (!/^https?:/.test(url.protocol)) {
+    return false;
+  }
+  if (value.includes('*')) {
+    return false;
+  }
+  // url.origin is punycode so a direct check against string wont work.
```
willdurand, line 119:
```suggestion
  // url.origin is punycode so a direct check against string won't work.
```

src/schema/formats.js
```diff
@@ -103,6 +103,28 @@ export function isSecureUrl(value) {
...
+  // url.href appends a slash even if not in the original string, we we
+  // additionally check that string does not end in slash.
```
willdurand, line 121:
```suggestion
  // url.href appends a slash even if not in the original string, so we
  // additionally check that the value does not ends with slash.
```

tests/unit/schema/test.formats.js
```diff
@@ -123,18 +124,68 @@ describe('formats', () => {
       expect(isAnyUrl(value)).toEqual(true);
       expect(isAbsoluteUrl(value)).toEqual(true);
       expect(isSecureUrl(value)).toEqual(true);
+      expect(isOrigin(value)).toEqual(true);
 
       expect(isStrictRelativeUrl(value)).toEqual(false);
       expect(imageDataOrStrictRelativeUrl(value)).toEqual(false);
     });
 
+    it('trailing slash', () => {
```
willdurand, line 133: would be good to put actual descriptions/sentences here and in the other _added_ test cases, otherwise the output will look like:

```
it trailing slash
it domain with no path but with query string
```

willdurand (commented):
some minor comments

willdurand:
> Given that the changes to ManifestJSONParser's errorLookup would require some more work (and potentially bring some more risks with it, depending from the amount and kind of changes we may have to apply to it), personally I would be ok to defer that to a separate follow up and proceed with this part as is in the short run.
>
> @willdurand wdyt?

+1, I think we can merge this PR now (after some more cleanups if possible) and update the error message in a follow-up, which @diox would take care of ideally.

tests/unit/schema/test.formats.js
diox: I did this in the other file, but here I was trying to stay consistent on the other URL formats tests, plus those tests make several assertions (some true, some false) so I wasn't sure what to do...

src/schema/formats.js
```diff
@@ -103,6 +103,28 @@ export function isSecureUrl(value) {
...
+    url = new URL(value);
+  } catch (e) {
+    return false;
+  }
+  if (!/^https?:/.test(url.protocol)) {
+    return false;
+  }
+  if (value.includes('*')) {
+    return false;
+  }
```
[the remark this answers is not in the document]
diox: Yeah that surprised me, but tests caught it :)

tests/unit/schema/test.formats.js
willdurand: Yeah, it is quite involved actually so let's forget about this suggestion.

willdurand (approved):

rpl (approved):
lgtm too

I did also confirm that:
- by reimporting the schema data from beta, no changes are applied to the imported schema (in other words the install_origins definition injected from the updates schema is imported as is into the imported schema)
- by importing the schema data from mozilla-central tip, install_origins definition from mozilla-central gets the additional `maxItems: 5` from the schema updates as expected (and all tests added in this PR still pass successfully as expected)

PR #4025, opened 2021-11-05
chore: Fixed typo in the IGNORED_APPLICATIONS_PROPERTY localizable message

### LINKED ISSUE #4024: Fix typo in the IGNORED_APPLICATIONS_PROPERTY localizable message
We have spotted this typo while preparing to localize this message, the last quote character (`"`) is a typo and should be removed:

https://github.com/mozilla/addons-linter/blob/c96d7aebcbcd1dc0cfa5a62cbb4645e3801334e2/src/messages/manifestjson.js#L506

rpl:
Fixes #4024


rpl:
@willdurand I did run the unit tests locally just in case there was a unit test expecting the string with the typo but all tests passed just fine locally.

The localizers will wait for this fix to start localizing this string.

willdurand (approved):
r+, thanks

Please run the extraction script either in this patch or a different PR (and no need to request a review if it's green)

willdurand:
@rpl I think you need to remove the last commit and re-run the l10n extraction

willdurand:
r+ on the first commit, we usually don't review the l10n commits (because of the excessive amount of changes)

PR #4027, opened 2021-11-05
Add script to run the l10n extraction and update the README

### LINKED ISSUE #4026: Document how to run the l10n extraction
This isn't documented yet and we tend to forget about this process. Let's document it to hopefully do it more often.

willdurand:
Fixes #4026

---

I added the script we use in addons-frontend (see [here](https://github.com/mozilla/addons-frontend/blob/4a7bda87683dc52664ad6a9c2cc78c763ec74f5c/bin/run-l10n-extraction)). It allows to run 1 command instead of 2 and also bails out when there are no changes.

### DIFF
--- a/scripts/run-l10n-extraction
+++ b/scripts/run-l10n-extraction
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+# Exit immediately when a command fails.
+set -e
+# Make sure exit code are respected in a pipeline.
+set -o pipefail
+# Treat unset variables as an error and exit immediately.
+set -u
+
+INITIAL_GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+GIT_CHANGES=$(git status --porcelain)
+DIFF_WITH_ONE_LINE_CHANGE="1 file changed, 1 insertion(+), 1 deletion(-)"
+GIT_REPO_NAME=addons-linter
+GIT_REMOTE="${GIT_REMOTE:-git@github.com:mozilla/$GIT_REPO_NAME.git}"
+LOG_FILE="l10n-extraction.log"
+
+info() {
+  local message="$1"
+
+  echo ""
+  echo "INFO: $message"
+  echo ""
+}
+
+error() {
+  local message="$1"
+
+  echo "ERROR: $message"
+  exit 1
+}
+
+run_l10n_extraction() {
+  local branch="$GIT_REPO_NAME-locales-$(date -u --iso-8601=date)"
+
+  info "Starting l10n extraction"
+
+  # Detect local (uncommitted) changes.
+  if [[ ! -z "$GIT_CHANGES" ]]; then
+    error "You have local changes, therefore this script cannot continue."
+  fi
+
+  # Switch to the `master` branch if we are not on it already.
+  if [[ "$INITIAL_GIT_BRANCH" != "master" ]]; then
+    git checkout master
+  fi
+
+  # Make sure the 'master' branch is up-to-date.
+  git pull "$GIT_REMOTE" master
+
+  # Update dependencies
+  npm i >> "$LOG_FILE" 2>&1
+
+  # Ensure the branch to extract the locales is clean.
+  if [[ $(git branch --list "$branch") ]]; then
+    info "Deleting branch '$branch' because it already exists"
+    git branch -D "$branch"
+  fi
+
+  info "Creating and switching to branch '$branch'"
+  git checkout -b "$branch"
+
+  info "Extracting locales (this can take several minutes)"
+  npm run extract-locales >> "$LOG_FILE" 2>&1
+
+  local git_diff_stat
+  git_diff_stat=$(git diff --shortstat)
+
+  if [[ -z "$git_diff_stat" ]] || [[ "$git_diff_stat" == *"$DIFF_WITH_ONE_LINE_CHANGE"* ]]; then
+    info "No locale changes, nothing to update, ending process"
+    git checkout -- .
+    git checkout "$INITIAL_GIT_BRANCH"
+    git branch -d "$branch"
+    return
+  fi
+
+  info "Merging locales"
+  ./scripts/merge-locales >> "$LOG_FILE" 2>&1
+
+  git commit -a -m "Extract and merge locales ($(date -u --iso-8601=date))" --no-verify
+
+  git push "$GIT_REMOTE" "$branch"
+
+  git checkout "$INITIAL_GIT_BRANCH"
+  git branch -D "$branch"
+
+  echo ""
+  echo "----------------------------------------------------------------------"
+  echo ""
+  echo "     You can now open the link below to submit your Pull Request:"
+  echo "     https://github.com/mozilla/$GIT_REPO_NAME/pull/new/$branch"
+  echo ""
+  echo "----------------------------------------------------------------------"
+}
+
+# Clear log file
+echo "" > "$LOG_FILE"
+
+run_l10n_extraction
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 scripts/download-import-tag
 scripts/list-firefox-tags
+scripts/run-l10n-extraction
 scripts/webext-test-functional
 dist
 coverage
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@
 # exclude these files
 scripts/download-import-tag
 scripts/list-firefox-tags
+scripts/run-l10n-extraction
 scripts/webext-test-functional
 package-lock.json
 LICENSE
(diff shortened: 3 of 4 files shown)


scripts/run-l10n-extraction
```diff
@@ -0,0 +1,101 @@
...
+
+  # Detect local (uncommitted) changes.
+  if [[ ! -z "$GIT_CHANGES" ]]; then
+    error "You have local changes, therefore this script cannot continue."
+  fi
+
+  # Switch to the `master` branch if we are not on it already.
+  if [[ "$INITIAL_GIT_BRANCH" != "master" ]]; then
+    git checkout master
+  fi
```
rpl, line 45: I gave this script a try, given that I did have a PR with changes to the locales.

I think that I get why the script assumes that we always want to regenerate locales from the master branch and not arbitrary feature branches, but doing it automatically and automatically create a branch in the official upstream repo feels a bit too auto-magic to me and I'm wondering if that may be surprising only from my perspective. 

As an example, as in this case, as a developer my intention may be to fix a string and to ensure the same PR does also update the locale files accordingly, but that isn't really the case, it is not actually how the script is meant to be used.

Not a blocking issue (especially given that I assume we are using and heppy with this strategy on the addons-frontend side for quite some time), but I'm wondering if:

- when the current branch isn't the master branch, then failing explicitly and printing a guidance message (that explains this is meant to only run on a clean master branch) may make this a bit less surprising
  - or alternatively to ask confirmation first (after printing a message to explain this is only meant to regenerate locales on the master branch

- deleting a branch with `-D` is slightly scary, I see that there shouldn't be too many chances to remove the wrong branch in the end, but asking for confirmation may still be reasonable (in the end this doesn't seem a script we intend to run on CI and I guess we may not need to run it too ofte, I guess ~1-2 times during a cycle and before each release may be enough)

- Pushing to the upstream remote repo is also something I'm wondering if we could ask a confirmation before doing it, maybe even after having printed the `git log --stat` output for the new commits range 

@willdurand wdyt?

scripts/run-l10n-extraction
```diff
@@ -0,0 +1,101 @@
...
+
+  if [[ -z "$git_diff_stat" ]] || [[ "$git_diff_stat" == *"$DIFF_WITH_ONE_LINE_CHANGE"* ]]; then
+    info "No locale changes, nothing to update, ending process"
+    git checkout -- .
+    git checkout "$INITIAL_GIT_BRANCH"
+    git branch -d "$branch"
+    return
+  fi
+
+  git commit -a -m "Extract locales" --no-verify
```
rpl, line 76: Just a secondary detail:
- as you already know I like the commit messages to follow the conventional changelog format (maybe `chore: ...` in this case)
- in this repo we follow that convention as strictly as we do for the web-ext repo and so we may as well opt to don't use that convention for these commit messages neither.

scripts/run-l10n-extraction
willdurand: > I gave this script a try, given that I did have a PR with changes to the locales.
> 
> I think that I get why the script assumes that we always want to regenerate locales from the master branch and not arbitrary feature branches, but doing it automatically and automatically create a branch in the official upstream repo feels a bit too auto-magic to me and I'm wondering if that may be surprising only from my perspective.

We created this script in addons-frontend to remove friction because the old process was manual with too many steps. This script automates all these steps, without any confirmation because that would be equally annoying.

> As an example, as in this case, as a developer my intention may be to fix a string and to ensure the same PR does also update the locale files accordingly, but that isn't really the case, it is not actually how the script is meant to be used.

Yeah, that isn't possible. Except for #4024, we usually add or update new strings but we don't do the l10n extraction as part of the PR. We usually do the l10n extraction from time to time (although with a script, I would probably do more frequently). In addons-frontend, l10n extraction happens every week after the code freeze.

> Not a blocking issue (especially given that I assume we are using and heppy with this strategy on the addons-frontend side for quite some time), but I'm wondering if:
> 
>     * when the current branch isn't the master branch, then failing explicitly and printing a guidance message (that explains this is meant to only run on a clean master branch) may make this a bit less surprising

It is already aborting when the repo is in a dirty state. I don't think we should abort because the process *requires* to do the l10n extraction on the master branch. We could improve the script to by pass this check, though, which would cover the use case mentioned above (but not sure it is worth it).

> or alternatively to ask confirmation first (after printing a message to explain this is only meant to regenerate locales on the master branch

Maybe? We didn't have docs about the l10n extraction process but the addons-frontend instructions already mention that: https://addons-frontend.readthedocs.io/en/latest/i18n/#updating-locales. I think that's why it isn't a surprise for me or the rest of the AMO team.

> deleting a branch with `-D` is slightly scary, I see that there shouldn't be too many chances to remove the wrong branch in the end, but asking for confirmation may still be reasonable

I could rename the conflicting branch or use a longer branch name, potentially including the datetime?

> (in the end this doesn't seem a script we intend to run on CI 

I'd be good to fully automate this process with CI but... we never managed to do that on other AMO project.

> I guess we may not need to run it too ofte, I guess ~1-2 times during a cycle and before each release may be enough)

I think it'd be nice to run this script after every release.

>  Pushing to the upstream remote repo is also something I'm wondering if we could ask a confirmation before doing it, maybe even after having printed the `git log --stat` output for the new commits range

Isn't the summary printed at the end enough? It is going to push a feature branch and only users with WRITE access can do that anyway.

---

To summarize, the "no-confirmation" approach is a feature, not a bug. I can make changes to avoid branch conflicts and maybe some more `info` messages but I am not convinced that we should ask confirmation.

(FWIW the script in addons-frontend has been used by 4 devs on 3 different OSes every week for the last 18 months without much problems. There were some issues with Windows initially but no work has been lost).

willdurand: I suppose the README section should make it crystal-clear that the script runs without any confirmation and explain the different main steps?

rpl (approved):
r+wc (all basically optional).

> ...
> To summarize, the "no-confirmation" approach is a feature, not a bug. I can make changes to avoid branch conflicts and maybe some more info messages but I am not convinced that we should ask confirmation.
>
> (FWIW the script in addons-frontend has been used by 4 devs on 3 different OSes every week for the last 18 months without much problems. There were some issues with Windows initially but no work has been lost).

I wasn't considering it a bug, just a slightly surprising behavior when you are not fully aware, also given that you originally mentioned this same script is being used for the addons-frontend repo, and so all my comments were mainly questions or doubts based on personal preferences.

> It is already aborting when the repo is in a dirty state. I don't think we should abort because the process requires to do the l10n extraction on the master branch. We could improve the script to by pass this check, though, which would cover the use case mentioned above (but not sure it is worth it).

I don't bother adapting the script, but may be worth to make it a bit more clear what is the expected process as part of the README section (e.g. we may just mention that locales are always updated on the master branch, any PR that changes or introduces new localizable strings should be merged on master first and after that the script executed to regenerate the locales)

> Maybe? We didn't have docs about the l10n extraction process but the addons-frontend instructions already mention that: https://addons-frontend.readthedocs.io/en/latest/i18n/#updating-locales. I think that's why it isn't a surprise for me or the rest of the AMO team.

ah, yeah, that makes the process pretty clear upfront, and so the automated behavior wouldn't be that surprising.

Maybe (still optional for this PR) we could mention in the README section that:
- (1) the addons-linter localization process follows the same conventions of the AMO one (and link the AMO docs for reference)
- (2) then mention that this new script does actually automates the manual process described in the AMO docs

> I could rename the conflicting branch or use a longer branch name, potentially including the datetime?

I would actually like if the branch name could include the datetime.
+1 if you decide to do that (still ok if you don't, or defer to a separate PR)

if we do that, I would also remove the `-f` from the `git push` command (there would be less changes to get a conflict one, if we still do then the developer is likely better placed to decide if the existing branch should be force updated or not)

> Isn't the summary printed at the end enough? It is going to push a feature branch and only users with WRITE access can do that anyway.

I think that the script computes the diff stat to decide if there was any update, but I'm not sure I saw it printed (from a quick look to the patch I don't think we log the stats anywhere).

Anyway, given that we have created a branch and that branch it is submitted automatically, I don't think we need to bother to emit that in the script output.

willdurand:
> (1) the addons-linter localization process follows the same conventions of the AMO one (and link the AMO docs for reference)

done

> (2) then mention that this new script does actually automates the manual process described in the AMO docs

done

> I would actually like if the branch name could include the datetime. +1 if you decide to do that (still ok if you don't, or defer to a separate PR)
> 
> if we do that, I would also remove the `-f` from the `git push` command (there would be less changes to get a conflict one, if we still do then the developer is likely better placed to decide if the existing branch should be force updated or not)

done, see:

```
$ ./scripts/run-l10n-extraction

INFO: Starting l10n extraction

From github.com:mozilla/addons-linter
 * branch              master     -> FETCH_HEAD
Current branch master is up to date.

INFO: Creating and switching to branch 'addons-linter-locales-2021-11-08'

Switched to a new branch 'addons-linter-locales-2021-11-08'

INFO: Extracting locales (this can take several minutes)

[...]
```

PR #4066, opened 2021-11-24
Disable variable tracing because of a bug in the eslint plugin

willdurand:
The issue is described in jhttps://github.com/mozilla/eslint-plugin-no-unsanitized/issues/188 and we think it is the underlying problem behind https://github.com/mozilla/addons/issues/1368. The test case is temporary but it shows that the config change fixes the issue.

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -1,10 +1,11 @@
 export const ESLINT_ERROR = 2;
 export const ESLINT_WARNING = 1;
 
-// Disable escapers (Sanitizer.escapeHTML, escapeHTML)
-// and unwrappers (Sanitizer.unwrapSafeHTML, unwrapSafeHTML)
-// which are allowed by default by this plugin.
-const DISABLE_ALLOWED_ESCAPERS = {
+const NO_UNSANITIZED_OPTIONS = {
+  variableTracing: false,
+  // Disable escapers (Sanitizer.escapeHTML, escapeHTML) and unwrappers
+  // (Sanitizer.unwrapSafeHTML, unwrapSafeHTML) which are allowed by default by
+  // this plugin.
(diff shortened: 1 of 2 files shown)


src/const.js
```diff
@@ -5,6 +5,7 @@ export const ESLINT_WARNING = 1;
 // and unwrappers (Sanitizer.unwrapSafeHTML, unwrapSafeHTML)
 // which are allowed by default by this plugin.
 const DISABLE_ALLOWED_ESCAPERS = {
```
rpl, line 7: Nit, we may rename this const into NO_UNSANITIZED_OPTIONS or something like that, given that it isn't only about escapers allowed anymore. Given that this is a temporary measure, I don't consider that as a blocking issue if you don't.

tests/unit/scanners/test.javascript.js
```diff
@@ -419,6 +419,19 @@ describe('JavaScript Scanner', () => {
...
   });
 
+  // See: https://github.com/mozilla/eslint-plugin-no-unsanitized/issues/188
+  it('has variable tracing disabled to avoid a bug', async () => {
+    const code = 'let c; a.innerHTML = `${c}`;';
+
+    const jsScanner = new JavaScriptScanner(code, 'code.js');
+
+    // We are not so much interested in the actual result. This would throw an
+    // error because of thebug in the upstream library.
```
rpl, line 429: :ok_hand: as long as it visible fail when that options is turned on but not yet fixed in the dependency, I'm totally fine with it (on the contrary thanks for taking care of adding a smoke test :heart:).

rpl (approved):

src/const.js
willdurand: Good idea.

PR #4078, opened 2021-12-01
fix: Use ESLint Linter API to more directly control parser and rules module loading

rpl:
Draft of the changes needed to migrate our `JavaScriptScanner` class to the [ESlint Linter API](https://eslint.org/docs/developer-guide/nodejs-api#linter), as one possible option to directly control the parser and rules module loading and make the addons-linter less likely to break because of issues like the ones described in mozilla/web-ext#2245 and mozilla/addons-linter#4074.

Along with moving to the `Linter` class, all the custom eslint rules embedded in the addons-linter itself are now actually loaded from the addons-linter bundle, which means that:
- we don't need to bundle each of the eslint rules separately (which was also duplicating the utility modules used in the eslint rule into the eslint rule bundle itself, which isn't going to be the case with the rules embedded into the addons-linter bundle).
- we don't need the appRoot global anymore (it was used only to point ESLint to the addons-linter dist/rules directory) 

The `Linter` class does never access disk directly, which means that we can also remove some workarounds that were needed to prevent the ESLint class instance from loading eslint ignore lists and configuration from files.


rpl:
@willdurand I need to confirm how this version handles the scenario where there are multiple eslint major versions in the dependency tree, but this now passes all existing tests[^1] included the web-ext functional tests, and so you may start to take a look at it.

As a side note the changes to the `JavaScriptScanner` class are actually smaller that what it may seem from a quick look to the diff, most of the lines changes are only changing nesting level but are unmodified in practice.

The core of the change to `JavaScriptScanner` is the part that calls `linter.defineRules` and `linter.defineParser`.

[^1]: and it also allowed me to remove some workarounds related to the `ESLint` class along with migrating to the `Linter` class, which was a nice addition to the original reasons that motivated me to look into this approach

src/rules/javascript/index.js
```diff
@@ -0,0 +1,25 @@
...
+  'event-listener-fourth': event_listener_fourth,
+  'global-require-arg': global_require_arg,
+  'opendialog-nonlit-uri': opendialog_nonlit_uri,
+  'opendialog-remote-uri': opendialog_remote_uri,
+  'webextension-api': webextension_api,
+  'webextension-api-compat': webextension_api_compat,
+  'webextension-api-compat-android': webextension_api_compat_android,
+  'webextension-deprecated-api': webextension_deprecated_api,
+  'webextension-unsupported-api': webextension_unsupported_api,
+};
```
willdurand, line 25: This looks almost like content we should generate...

Do we have test coverage otherwise? (thinking about preventing missing/forgotten imports and/or new rules not registered here)

tests/unit/rules/javascript/test.eslintRulesObject.js
```diff
@@ -4,15 +4,30 @@ import { readdirSync } from 'fs';
...
+  it.each(files)(
+    '%s rule module should be exported by the index.js module',
+    async (fileName) => {
+      const modName = path.parse(fileName).name;
+      const indexMod = (await import('rules/javascript')).default;
+      const ruleMod = (await import(`rules/javascript/${fileName}`)).default;
+      expect(typeof indexMod[modName]?.create).toBe('function');
+      expect(indexMod[modName]).toBe(ruleMod);
+    }
+  );
```
willdurand, line 32: haha, there it is, thanks!

src/scanners/javascript.js
```diff
@@ -61,145 +58,145 @@ export default class JavaScriptScanner {
...
+      linter.defineRule(key, customEslintRules[key]);
+    }
+
+    // Load plugins rules.
+    const pluginRules = noUnsanitized.rules;
+    for (const key of Object.keys(pluginRules)) {
+      linter.defineRule(`no-unsanitized/${key}`, pluginRules[key]);
+    }
+
+    linter.defineParser('addons-linter-espree', espree);
```
willdurand, line 81: :+1:

willdurand (commented):
Looks good to me, are you happy with the current patch?

rpl:
> Looks good to me, are you happy with the current patch?

yep, I wanted to actually try out the scenario from the issue this should fix, which I did today and it worked like a charm \o/

I was thinking of adding a integration / smoke test to cover the issue, it should be pretty doable (it wouldn't differ too much from the "web-ext functional smoke test" script we do already run in the CI jobs). 

@willdurand wdyt?

In the meantime, here is the STR of the manual test I did:
- created a new dir and a package.json
- added eslint v7 as a dev dependency
- added addons-linter 4.4.0 npm version as a dev dependency
- created a fake extension dir `npx create-webextension fake-ext-src`
- run `npx addons-linter fake-ext-src` and confirmed addons-linter results are broken because of the espree version mismatch
- in the addons-linter sources with its working tree pointed to this pull request: `npm ci && npm build && npm pack`
- in the smoke test project:
  - renamed `node_modules/addons-linter` into `node_modules/addons-linter.npm`
  - replaced `node_modules/addons-linter` with the content of the archive generated by `npm pack`
  - preserved npm transitive dependencies by copying `node_modules/addons-linter.npm/node_modules dir` to `node_modules/addons-linter/node_modules`
- run `npx addons-linter fake-ext-src` and confirmed addons-linter results are NOT broken anymore because of the espree version mismatch

As a side note, these changes seems to have also a positive impact on the package size:
- before these changes: `package size:  1.2 MB / unpacked size: 5.2 MB`
- after these changes: `package size:  361.5 kB / unpacked size: 1.4 MB`

src/rules/javascript/index.js
rpl: yes, initially I was also considering to auto-generate it, but to be honest we didn't changed the list of rules that often and so I didn't bothered for now, and instead I opted to cover it with an additional test case for now (and defer generating it automatically to when it may become useful or necessary in the future).

willdurand:
> I was thinking of adding a integration / smoke test to cover the issue, it should be pretty doable (it wouldn't differ too much from the "web-ext functional smoke test" script we do already run in the CI jobs).

+1

willdurand:
> As a side note, these changes seems to have also a positive impact on the package size:
> 
>     * before these changes: `package size:  1.2 MB / unpacked size: 5.2 MB`
> 
>     * after these changes: `package size:  361.5 kB / unpacked size: 1.4 MB`

Is that "just" because we don't bundle the js rule files anymore? They are still embedded in the bundle (I hope) so I am very surprised but that's great!

rpl:
> > As a side note, these changes seems to have also a positive impact on the package size:
> > ```
> > * before these changes: `package size:  1.2 MB / unpacked size: 5.2 MB`
> > 
> > * after these changes: `package size:  361.5 kB / unpacked size: 1.4 MB`
> > ```
> 
> Is that "just" because we don't bundle the js rule files anymore? They are still embedded in the bundle (I hope) so I am very surprised but that's great!

yes, basically we were creating bundles for each rule and we were basically adding inside each rule bundle a copy of the other addons-linter modules they depends on (which means we had more than one copy of those modules, one of each rule and one more in the main bundle) plus the source maps for all the rules.

I was not surprised the package was going to be smaller, but I was surprised by the amount of the size reduction (until I did notice we had multiple copies of the same modules in each of the rule bundles, that definitely explains the size reduction observed)

rpl:
> > I was thinking of adding a integration / smoke test to cover the issue, it should be pretty doable (it wouldn't differ too much from the "web-ext functional smoke test" script we do already run in the CI jobs).
> 
> +1

@willdurand I added the smoke test in 52490a0:
- locally I first wrote it and verified that it does fail as expected on the current master (without any of the changes in this PR)
- then I double-checked that it was passing locally with the changes in this PR
- finally I added it to the circleci jobs (to all three of them, to be sure to cover the 2 node/npm version we are explicitly testing our package) and pushed into this PR to confirm that all CI jobs pass as expected.

This should now be ready for  your review.

scripts/smoke-test-eslint-version-conflicts
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
willdurand, line 1: I have too many comments on this file (because I have been traumatized by `shellcheck`). Here is a diff to apply:

```diff
diff --git a/scripts/smoke-test-eslint-version-conflicts b/scripts/smoke-test-eslint-version-conflicts
index 5f2ec1b1..2bb78afb 100755
--- a/scripts/smoke-test-eslint-version-conflicts
+++ b/scripts/smoke-test-eslint-version-conflicts
@@ -2,14 +2,14 @@
 
 set -e
 
-LINTER_PATH=`pwd`
-TARGET_TEST_PATH=$LINTER_PATH/tmp/test-project
-FIXTURE_TEST_PATH=$LINTER_PATH/tests/fixtures/eslint-versions-conflicts-package
-LINTER_VERSION=`node -p 'require("./package.json").version'`
+LINTER_PATH=$(pwd)
+TARGET_TEST_PATH="$LINTER_PATH/tmp/test-project"
+FIXTURE_TEST_PATH="$LINTER_PATH/tests/fixtures/eslint-versions-conflicts-package"
+LINTER_VERSION=$(node -p 'require("./package.json").version')
 
 # Fail if the target path already exists.
 [[ -f "$TARGET_TEST_PATH" ]] && (
-  echo "ERROR: Temporary test dir '$TARGET_TEST_PATH' already exist."
+  echo "ERROR: Temporary test dir '$TARGET_TEST_PATH' already exists."
   exit 1
 )
 
@@ -20,15 +20,15 @@ npm ci
 npm run build
 
 # Create target test path directory and cleanup on exit.
-cp -rf $FIXTURE_TEST_PATH $TARGET_TEST_PATH
-trap "rm -rf $TARGET_TEST_PATH" EXIT
+cp -rf "$FIXTURE_TEST_PATH" "$TARGET_TEST_PATH"
+trap 'rm -rf "$TARGET_TEST_PATH"' EXIT
 
 pushd "$TARGET_TEST_PATH"
-npm pack $LINTER_PATH
+npm pack "$LINTER_PATH"
 npm install
-npm install --save-dev ./addons-linter-$LINTER_VERSION.tgz
+npm install --save-dev "./addons-linter-$LINTER_VERSION.tgz"
 
-echo -e "INFO: Run addons-linter on the test fixture project\n" 
+echo -e "INFO: Run addons-linter on the test fixture project\n"
 
 npx addons-linter ./src || (
   echo "ERROR: addons-linter validation failed".
```

src/scanners/javascript.js
```diff
@@ -45,10 +43,9 @@ export default class JavaScriptScanner {
     _ESLint = ESLint,
     _messages = messages,
     _ruleMapping = ESLINT_RULE_MAPPING,
-    // This tells ESLint where to expect the ESLint rules for addons-linter.
-    // Its default value is defined below. This property is mainly used for
-    // testing purposes.
-    _rulePaths = undefined,
+    // This property is used to inject additional custom eslint rules
+    // as part of tests.
+    _ruleModules = undefined,
```
willdurand, line 48: Why having a "modules" prefix and not simply `_rules`? (or `_additionalRules`)

willdurand (approved):
r+wc, thanks

scripts/smoke-test-eslint-version-conflicts
rpl: @willdurand ah, thanks! Actually I should have thought of running shellcheck on the script. Diff applied in the last updated push on this pull request (I rolled it into the second patch patch of this pull request)

src/scanners/javascript.js
rpl: @willdurand No particular reason, I was trying to keep the same prefix as _ruleMapping, but I'm not really attached to `_ruleModules`,  `_rules` would be also totally fine by me.

I opted to _rules and rolled that change into the first patch part of this pull request.

PR #4085, opened 2021-12-06
Make install_origins error messages more user friendly

### LINKED ISSUE #4061: Create specific error codes & messages for invalid `install_origins` values
Follow-up for #3972. In [the pull request, @rpl noted](https://github.com/mozilla/addons-linter/pull/4018#pullrequestreview-812703192):

> As a side note, I would really like if we could also assign a more specific rule code for invalid `install_origins` values, so that we can also make sure to show a message that would help the developer to know how to make the manifest key value valid, e.g. either:
> 
>  * with a link to a doc page (as we do for permissions and optional permissions invalid values)
>  * or including the expected value in the error message when possible  as we are doing in the error we are going to log on the Firefox side).
> 
> 
> To achieve that I think we may use a strategy similar to the one we use for the validation errors related to the `permissions` and `optional_permissions`, where the error got from the `JSONSchema` validator is then "changed / improved" in the `ManifestJSONParser` class (from the [erroLookup method here](https://github.com/mozilla/addons-linter/blob/ed8da2e1aedeceefc5f66d8911c111d699781a17/src/parsers/manifestjson.js#L222)).

We agreed to do that as a follow-up, hence this issue.

diox:
This also changes the generic error message for arrays from `"Unknown ..."` to `"Invalid ..."` to better accommodate install origins

Fixes #4061

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -99,7 +99,7 @@ export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|~$/i;
 export const ALREADY_SIGNED_REGEX = /^META-INF\/manifest\.mf/;
 export const PERMS_DATAPATH_REGEX =
   /^\/(permissions|optional_permissions|host_permissions)\/([\d+])/;
-export const INSTALLORIGIN_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
+export const INSTALL_ORIGINS_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
 
 export const RESERVED_FILENAMES = ['mozilla-recommendation.json'];
 
--- a/src/const.js
+++ b/src/const.js
@@ -97,8 +97,8 @@ export const MAX_FILE_SIZE_TO_PARSE_MB = 4;
 export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|~$/i;
 export const ALREADY_SIGNED_REGEX = /^META-INF\/manifest\.mf/;
-export const PERMS_DATAPATH_REGEX =
-  /^\/(permissions|optional_permissions|host_permissions)\/([\d+])/;
+export const COMPLEX_ARRAYS_DATAPATH_REGEX =
+  /^\/(permissions|optional_permissions|host_permissions|install_origins)\/([\d+])/;
 
 export const RESERVED_FILENAMES = ['mozilla-recommendation.json'];
 
--- a/src/const.js
+++ b/src/const.js
@@ -97,8 +97,9 @@ export const MAX_FILE_SIZE_TO_PARSE_MB = 4;
 export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|~$/i;
 export const ALREADY_SIGNED_REGEX = /^META-INF\/manifest\.mf/;
-export const COMPLEX_ARRAYS_DATAPATH_REGEX =
-  /^\/(permissions|optional_permissions|host_permissions|install_origins)\/([\d+])/;
+export const PERMS_DATAPATH_REGEX =
+  /^\/(permissions|optional_permissions|host_permissions)\/([\d+])/;
+export const INSTALLORIGIN_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
 
 export const RESERVED_FILENAMES = ['mozilla-recommendation.json'];
 
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -1,7 +1,7 @@
 import { oneLine } from 'common-tags';
 
 import { i18n, errorParamsToUnsupportedVersionRange } from 'utils';
-import { MANIFEST_JSON, PERMS_DATAPATH_REGEX } from 'const';
+import { MANIFEST_JSON, COMPLEX_ARRAYS_DATAPATH_REGEX } from 'const';
 
 export const MANIFEST_FIELD_REQUIRED = {
   code: 'MANIFEST_FIELD_REQUIRED',
@@ -49,7 +49,7 @@ export function manifestPermissionUnsupported(permissionName, error) {
   const message = i18n.sprintf(messageTmpl, {
     permissionName,
     versionRange,
-    fieldName: error.dataPath.match(PERMS_DATAPATH_REGEX)[1],
+    fieldName: error.dataPath.match(COMPLEX_ARRAYS_DATAPATH_REGEX)[1],
   });
 
   return {
@@ -109,6 +109,17 @@ export const MANIFEST_HOST_PERMISSIONS = {
   // TODO(https://github.com/mozilla/addons-linter/issues/3893): link host_permissions
   // MDN doc page here once we have created it.
   description: i18n._('Invalid host permission.'),
+  file: MANIFEST_JSON,
+};
+
+export const MANIFEST_INSTALL_ORIGINS = {
+  code: 'MANIFEST_INSTALL_ORIGINS',
+  message: i18n._('Invalid install origin.'),
+  // TODO(https://github.com/mozilla/addons-linter/issues/4084): link install_origins
+  // MDN doc page here once we have created it.
+  description: i18n._(
+    'Invalid install origin (A valid origin has - only - a scheme, hostname and optional port.'
+  ),
   file: MANIFEST_JSON,
 };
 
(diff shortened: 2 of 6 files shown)


src/messages/manifestjson.js
```diff
@@ -112,6 +112,17 @@ export const MANIFEST_HOST_PERMISSIONS = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_INSTALL_ORIGINS = {
+  code: 'MANIFEST_INSTALL_ORIGINS',
+  message: i18n._('Invalid install origin.'),
+  // TODO(https://github.com/mozilla/addons-linter/issues/4084): link install_origins
+  // MDN doc page here once we have created it.
+  description: i18n._(
+    'Invalid install origin (A valid origin has - only - a scheme, hostname and optional port.'
```
willdurand, line 121: unbalanced parens

```suggestion
    'Invalid install origin. A valid origin has - only - a scheme, hostname and optional port.'
```

willdurand (approved):
r+wc

@rpl are you OK with the change from "Unknown permission" to "Invalid permission"? (I don't think it is a problem personally)

src/parsers/manifestjson.js
```diff
@@ -295,7 +295,7 @@ export default class ManifestJSONParser extends JSONParser {
     ) {
       // Choose a different message for permissions unsupported with the
       // add-on manifest_version.
-      if (PERMS_DATAPATH_REGEX.test(error.dataPath)) {
+      if (COMPLEX_ARRAYS_DATAPATH_REGEX.test(error.dataPath)) {
         baseObject = messages.manifestPermissionUnsupported(error.data, error);
```
rpl, line 299: > @rpl are you OK with the change from "Unknown permission" to "Invalid permission"? (I don't think it is a problem personally)

@willdurand @diox I don't really mind changing "Unknown permission" into "Invalid permission", but I'd prefer if we could define a new `ORIGINS_DATAPATH_REGEX` (or `INSTALLORIGIN_DATAPATH_REGEX`) instead of repurposing the original `PERMS_DATAPATH_REGEX`.

As an example by reusing that pref here we are potentially going to be using `messages.manifestPermissionUnsupported` for something that isn't a permission anymore, even if it does not happen in this patch yet because at the moment we enter this branch only for the errors that also involve manifest_version based jsonschema checks (but it may become the case if in the future we may introduce some manifest_version-based changes for the install_origins field schema).

rpl (changes requested):
Looks pretty good, there is just one detail that I'd like to ask changes for (as described in the inline comment below):

personally I'd prefer if we could not reuse PERMS_DATAPATH_REGEX for the install_origins (it does not trigger any issue at the moment, but in the future it could introduce subtle unwanted side-effects because of a new schema data import and we may miss to notice it).

src/parsers/manifestjson.js
diox: @rpl would the following work for you ? :
```diff
diff --git a/src/const.js b/src/const.js
index 7602c4d1..42ff23ae 100644
--- a/src/const.js
+++ b/src/const.js
@@ -97,8 +97,9 @@ export const MAX_FILE_SIZE_TO_PARSE_MB = 4;
 export const HIDDEN_FILE_REGEX = /^__MACOSX\//;
 export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|~$/i;
 export const ALREADY_SIGNED_REGEX = /^META-INF\/manifest\.mf/;
-export const COMPLEX_ARRAYS_DATAPATH_REGEX =
-  /^\/(permissions|optional_permissions|host_permissions|install_origins)\/([\d+])/;
+export const PERMS_DATAPATH_REGEX =
+  /^\/(permissions|optional_permissions|host_permissions)\/([\d+])/;
+export const INSTALLORIGIN_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
 
 export const RESERVED_FILENAMES = ['mozilla-recommendation.json'];
 
diff --git a/src/messages/manifestjson.js b/src/messages/manifestjson.js
index 48e1aeb7..8d89f744 100644
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -1,7 +1,7 @@
 import { oneLine } from 'common-tags';
 
 import { i18n, errorParamsToUnsupportedVersionRange } from 'utils';
-import { MANIFEST_JSON, COMPLEX_ARRAYS_DATAPATH_REGEX } from 'const';
+import { MANIFEST_JSON, PERMS_DATAPATH_REGEX } from 'const';
 
 export const MANIFEST_FIELD_REQUIRED = {
   code: 'MANIFEST_FIELD_REQUIRED',
@@ -49,7 +49,7 @@ export function manifestPermissionUnsupported(permissionName, error) {
   const message = i18n.sprintf(messageTmpl, {
     permissionName,
     versionRange,
-    fieldName: error.dataPath.match(COMPLEX_ARRAYS_DATAPATH_REGEX)[1],
+    fieldName: error.dataPath.match(PERMS_DATAPATH_REGEX)[1],
   });
 
   return {
diff --git a/src/parsers/manifestjson.js b/src/parsers/manifestjson.js
index 260bb68c..371d0f7b 100644
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -25,10 +25,11 @@ import {
   LOCALES_DIRECTORY,
   MESSAGES_JSON,
   FILE_EXTENSIONS_TO_MIME,
+  INSTALLORIGIN_DATAPATH_REGEX,
   STATIC_THEME_IMAGE_MIMES,
   RESTRICTED_HOMEPAGE_URLS,
   RESTRICTED_PERMISSIONS,
-  COMPLEX_ARRAYS_DATAPATH_REGEX,
+  PERMS_DATAPATH_REGEX,
 } from 'const';
 import log from 'logger';
 import * as messages from 'messages';
@@ -295,7 +296,7 @@ export default class ManifestJSONParser extends JSONParser {
     ) {
       // Choose a different message for permissions unsupported with the
       // add-on manifest_version.
-      if (COMPLEX_ARRAYS_DATAPATH_REGEX.test(error.dataPath)) {
+      if (PERMS_DATAPATH_REGEX.test(error.dataPath)) {
         baseObject = messages.manifestPermissionUnsupported(error.data, error);
       } else {
         baseObject = messages.manifestFieldUnsupported(error.dataPath, error);
@@ -333,7 +334,11 @@ export default class ManifestJSONParser extends JSONParser {
     // Arrays can be extremely verbose, this tries to make them a little
     // more sane. Using a regex because there will likely be more as we
     // expand the schema.
-    const match = error.dataPath.match(COMPLEX_ARRAYS_DATAPATH_REGEX);
+    // Note that this works because the 2 regexps use similar patterns. We'll
+    // want to adjust this if they start to differ.
+    const match =
+      error.dataPath.match(PERMS_DATAPATH_REGEX) ||
+      error.dataPath.match(INSTALLORIGIN_DATAPATH_REGEX);
 
     if (
       match &&
```

rpl: @diox yep, that would be ok from my perspective too :+1:

willdurand: > +export const INSTALLORIGIN_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;

I suppose we could add 2 more chars in the variable name to match the regexp: `INSTALLORIGIN_DATAPATH_REGEX` -> `INSTALL_ORIGINS_DATAPATH_REGEX

src/const.js
```diff
@@ -99,6 +99,7 @@ export const FLAGGED_FILE_REGEX = /thumbs\.db$|\.DS_Store$|\.orig$|\.old$|~$/i;
 export const ALREADY_SIGNED_REGEX = /^META-INF\/manifest\.mf/;
 export const PERMS_DATAPATH_REGEX =
   /^\/(permissions|optional_permissions|host_permissions)\/([\d+])/;
+export const INSTALLORIGIN_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
```
willdurand, line 102:
```suggestion
export const INSTALL_ORIGINS_DATAPATH_REGEX = /^\/(install_origins)\/([\d+])/;
```

diox: mmm did I mess up my search & replace...

diox: ee0cf2bd93665dd590b51e5020b4325e7e409d89

src/parsers/manifestjson.js
```diff
@@ -333,7 +334,11 @@ export default class ManifestJSONParser extends JSONParser {
     // Arrays can be extremely verbose, this tries to make them a little
     // more sane. Using a regex because there will likely be more as we
     // expand the schema.
-    const match = error.dataPath.match(PERMS_DATAPATH_REGEX);
+    // Note that this works because the 2 regexps use similar patterns. We'll
+    // want to adjust this if they start to differ.
+    const match =
+      error.dataPath.match(PERMS_DATAPATH_REGEX) ||
+      error.dataPath.match(INSTALL_ORIGINS_DATAPATH_REGEX);
```
rpl, line 341: @diox thanks for leaving a comment about this :heart:

rpl (approved):
lgtm too, thanks a lot @diox!

willdurand:
Merging this to be able to cut a new release.

PR #4120, opened 2022-01-09
fix: Improve detect source type.

### LINKED ISSUE #4020: Syntax error with a top-level await and no import

Files with a _top-level await_ and no _import_: returns a syntax error.

```JavaScript
await browser.storage.local.get();
```

### What happened?

```
Validation Summary:

errors          1              
notices         0              
warnings        0              

ERRORS:

Code              Message        Description               File          Line   Column
JS_SYNTAX_ERROR   JavaScript     There is a JavaScript     testcase.js   1      7     
                  syntax error   syntax error in your                                
                   (Parsing as   code, which might be                                
                  script         related to some                                     
                  error:         experimental JavaScript                             
                  Unexpected     features that aren't an                             
                  token          official part of the                                
                  browser at     language specification                              
                  line: 1 and    and therefore not                                   
                  column: 7)     supported yet. The                                  
                                 validation cannot                                   
                                 continue on this file.
```

No error reported.

I have no error with ESLint.

regseb:
Fixes #4020

I propose a new method to detect the `sourceType` of a file (_script_ or _module_) based only on syntax errors:

- If a file is syntactically correct with only one `sourceType`: analyze with it.
- If a file is syntactically correct with both `sourceType`: use _script_ by default.
- If a file has a syntax error with both `sourceType`: use _script_ by default.

| Syntax error with<br />`{ sourceType: "script" }` | Syntax error with<br />`{ sourceType: "module" }` | sourceType |
| --- | --- | --- |
| no | no | `"script"`
| no | yes | `"script"`
| yes | no | `"module"`
| yes | yes | `"script"`

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -2,7 +2,6 @@ import ESLint from 'eslint';
 import noUnsanitized from 'eslint-plugin-no-unsanitized';
 import { oneLine } from 'common-tags';
 import * as espree from 'espree';
-import * as vk from 'eslint-visitor-keys';
 import { ECMA_VERSION } from 'addons-scanner-utils/dist/const';
 
 import { ESLINT_RULE_MAPPING, ESLINT_TYPES } from 'const';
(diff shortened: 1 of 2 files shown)


rpl (changes requested):
Sorry for not getting to this PR sooner, we actually discussed about this PR and related issue in our last triage and we agreed that we would prefer to just extend the current detection logic to consider a top level await as an additional AST note that would make the linter to prefer sourceType "module" to lint the source file (as @Rob--W also originally suggested in https://github.com/mozilla/addons-linter/issues/4020#issuecomment-979280502).

The approach suggested in Rob's comment is less likely to potentially change the existing behaviors (and besides that with the approach currently drafted in this PR the same source code may have to be parsed 3 times, which is even less ideal than parsing two times as it is in the current approach used for the sourceType detection).

willdurand:
Closing due to inactivity. Feel free to reopen in the future.

PR #4143, opened 2022-01-24
fix: remove redundant and obsolete eslint rules

### LINKED ISSUE #4142: Improve memory usage while scanning bigger js files by removing redundant and obsolete custom eslint rules
As part of reduce OOM issue while scanning bigger js files, I tried the approach we agreed on in our last triage meeting but trying to scan all js files sequentially and run a GC pass after each scanned file was not yet able to prevent the OOM issue.

As an alternative strategy I did run some more tests to bisect which were the eslint rule that are introducing the memory usage issue, especially given that the memory growing behavior it really looked like a leak.

The result of these tests pin-pointed the memory usage issue (at least the one that actually triggers the OOM when included and it doesn't when excluded) on the following two custom eslint rules defined in the addons-linter sources itself:

- [src/rules/javascript/deprecated-entities.js](https://github.com/mozilla/addons-linter/blob/master/src/rules/javascript/deprecated-entities.js):
  - emit warnings on document.write calls, technically overlaps with something that is also covered by rules from [mozilla/eslint-plugin-no-unsanitized/](https://github.com/mozilla/eslint-plugin-no-unsanitized/), but the plugin seems more efficient from a memory usage perspective.
- [src/rules/javascript/event-listener-fourth.js](https://github.com/mozilla/addons-linter/blob/master/src/rules/javascript/event-listener-fourth.js):
  - this rules seems to be emitting warnings on a Firefox-only non-standard parameter of the DOM addEventListener method (`wantsUntrusted`), but it doesn't seem to be relevant anymore for WebExtensions extensions (for WebExtensions documents `wantsUntrusted` should be already defaulting to `true` as it does regular webcontent, while the rule seems to matter only when the extensions were often executed in privileged documents where the `wantsUntrusted` param was defaulting to `false`)

Both the rule were using [a `getNodeReference` helper method defined in src/utils.js](https://github.com/mozilla/addons-linter/blob/255b50f2041bc06a85530d13f393db5a4e488bdb/src/utils.js#L86), and they were the only users of that helper method.

Given that this helper was likely the source of the inefficient memory usage, I'm proposing to remove:
- the two custom eslint rules (which, besides being inefficient from a memory usage perspective, seems to be redundant or obsolete)
- remove alspo the `getNodeReference` helper function along with the only two users, to avoid being tempted to use it in a new rule (if we need something similar, we can look into using the eslint helper that we are using in [mozilla/eslint-plugin-no-unsanitized/](https://github.com/mozilla/eslint-plugin-no-unsanitized/) as part of its "variable tracing" feature).

rpl:
Fix #4142 (see issue description for more details about the motivations and rationale)


Rob--W (approved):
Nice. The `document.write` call is already covered by a different warning, so it's ok to remove the check.

rpl:
@willdurand codecov is complaining about a -0.04% decrease in coverage, but it may be codecov just being "coverage drunk" :-P,  all the 3 files mentioned as impacted have 100% coverage and no new lines missing a hit (just a few unused lines removed) and so we may just have to accept the new baseline and ignore codecov nitpicking-ness ;-)

Let me know wdyt.

wagnerand (approved):

willdurand (approved):

PR #4150, opened 2022-01-25
(Re-)add rule for NO_DOCUMENT_WRITE

### LINKED ISSUE #4148: `document.write()` usage not caught
In https://github.com/mozilla/addons-linter/issues/4142, we removed the `NO_DOCUMENT_WRITE` error (and the ESLint rule) because it was already covered by https://github.com/mozilla/eslint-plugin-no-unsanitized (`UNSAFE_VAR_ASSIGNMENT`). It looks like it isn't working as expected, though.

## STR

Write an add-on with a JS file that contains:

```js
document.write("foo");
```

## Expected

A warning about `document.write()` should be emitted.

## Actual

No error when validating the add-on with the linter v4.7.0 (latest version currently):

```
$ ./bin/addons-linter test
Validation Summary:

errors          0
notices         0
warnings        0
```

This looks like a regression because there is an error when validating the add-on with the linter v4.6.0:

```
$ ./bin/addons-linter test
Validation Summary:

errors          0
notices         0
warnings        1

WARNINGS:

Code                Message        File       Line   Column
NO_DOCUMENT_WRITE   Use of         index.js   1      1
                    document.wr…
                    ite strongly
                    discouraged.
```

willdurand:
Fixes #4148

---

This is a simpler version of what we had before. It won't cover the
exact same things but it should be good enough.

### DIFF
--- a/tests/unit/rules/javascript/test.no_document_write.js
+++ b/tests/unit/rules/javascript/test.no_document_write.js
@@ -1,0 +1,50 @@
+import { VALIDATION_WARNING } from 'const';
+import { NO_DOCUMENT_WRITE } from 'messages';
+import JavaScriptScanner from 'scanners/javascript';
+
+import { runJsScanner } from '../../helpers';
+
+describe(__filename, () => {
+  it('should warn about using document.write()', async () => {
+    const code = 'document.write();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    linterMessages.sort();
+    expect(linterMessages.length).toEqual(1);
+    expect(linterMessages[0].code).toEqual(NO_DOCUMENT_WRITE.code);
+    expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
+  });
+
+  it('should not warn about using other member functions', async () => {
+    const code = 'document.doNothing();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    expect(linterMessages.length).toEqual(0);
+  });
+
+  it('should not warn about calling write() in general', async () => {
+    const code = 'foo.write();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    expect(linterMessages.length).toEqual(0);
+  });
+
+  it('should not warn about calling write() to a non-global document', async () => {
+    const code = 'foo.document.write();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    expect(linterMessages.length).toEqual(0);
+  });
+
+  it('should not warn about calling write() to non member of document', async () => {
+    const code = 'document.foo.write();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    expect(linterMessages.length).toEqual(0);
+  });
+});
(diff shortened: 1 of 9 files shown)


tests/unit/rules/javascript/test.no_document_write.js
```diff
@@ -0,0 +1,50 @@
...
+  it('should not warn about calling write() in general', async () => {
+    const code = 'foo.write();';
+    const jsScanner = new JavaScriptScanner(code, 'badcode.js');
+
+    const { linterMessages } = await runJsScanner(jsScanner);
+    expect(linterMessages.length).toEqual(0);
+  });
+
+  it('should not warn about calling write() to a non-global document', async () => {
+    const code = 'foo.document.write();';
```
rpl, line 36: @willdurand question: do we care about this warning being collected also for expressions like 'window.document.write();` or `globalThis.document.write()`?

willdurand: Good question, the test suite didn't cover that before I think. I'd say "no" because we want a simple rule and I don't think these two expressions are common (especially the second one) but I might be wrong :)

rpl: if we don't care about that kind of calls and other corner cases, there a bunch other that we can see from [eslint-plugin-no-unsanitized test cases](https://github.com/mozilla/eslint-plugin-no-unsanitized/blob/master/tests/rules/method.js), e.g. the following two may no be caught by the current version of the rule:

- document.writeln ;-)
- document.write`something`


I think this version may already be good to go.
If on the contrary we want more cases to be detected, it may be good to use an approach similar to the one that eslint-plugin-no-unsanitized uses (that logic in splitted between https://github.com/mozilla/eslint-plugin-no-unsanitized/blob/master/lib/rules/method.js and https://github.com/mozilla/eslint-plugin-no-unsanitized/blob/master/lib/ruleHelper.js)

Let's confirm what the others people involved thinks about those cases, just to confirm if there is more to do (either in this PR or in follow up) of if what this PR already does would be enough.

wagnerand: I am not sure what other cases we'd want to be detected, as far as I understand this patch, it is currently detecting pretty much anything. For me the question is: Do we want to exclude some obvious non-issues.

I acknowledge this might be solving for something we used to live with with the previous rule as well, so we might want to do this in a follow up if we want to minimize short-term effort:

* Don't match `document.write();` - Writing nothing is useless but also not dangerous.
* Consider evaluating the type of the argument passed. We might want to ignore static strings* and focus on variables, or a combination of strings and variables...anything dynamic

*`document.write("<script src='http...'>");` is obviously something that we'd want to flag even though it's a static string. Rob suggested we could look for `<`. The question is, how far do we want to go.

willdurand: My understanding was that this rule was clearly not a security rule (like most linter rules anyway, i.e. this isn't a security scanner) and therefore we wanted to re-add a simple rule targeting developers.

Since it is a warning, it isn't so much a problem if it matches `document.write("whatever static string")` because the general idea is to discourage developers to use it, even if there is nothing dangerous (because of what I said above).

`eslint-plugin-no-unsanitized` already provides "security-oriented" rules so I don't think we need to duplicate that.

rpl (approved):

diox (approved):

tests/unit/rules/javascript/test.no_document_write.js
wagnerand: Some rules are targeted towards both developers and reviewers, and this is one of them.
Yes, we want to discourage developers from using that, but we shouldn't warn them unnecessarily. We do get complaints from users why we raise warnings for bad, but safe code.

If we have strong opinions on discouraging, we should consider throwing an error. I don't see a good reason to use `document.write('...static...')`, and we clearly do not want developers to pass something dynamic instead.

The reason this rule is reintroduced is because `eslint-plugin-no-unsanitized` does not cover it, for some reason.

wagnerand (commented):
r+wc. Maybe we can optimize it in a follow-up.

wagnerand (approved):

PR #4151, opened 2022-01-25
feat: Imported Firefox 97.0b8 API schema

### LINKED ISSUE #4149: Import Firefox 97 schema
This issue is about importing the Firefox 97 schema. We should make sure that validation for `sitepermission` add-ons actually works, and potentially have a new `validate` function for that specific add-on type.

rpl:
Fixes #4149

-  removed "threads" from the list `geckoProfiler` API features, [Bug 1729815](https://bugzilla.mozilla.org/1729815)
-  added "cookieStoreId" to `tabs.query` API options, [Bug 1730931](https://bugzilla.mozilla.org/1730931)
-  added "cookieStoreId" to `contentScripts.register` API options, [Bug 1470651](https://bugzilla.mozilla.org/1470651)
- new manifest format "WebExtensionSitePermissionsManifest"

The "WebExtensionSitePermissionsManifest" JSONSchema also required some additional changes to the schema validator and to the manifest parser, and additional tests to cover the sitepermission manifest validation.

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -16,6 +16,7 @@ import {
   validManifestJSON,
   validDictionaryManifestJSON,
   validLangpackManifestJSON,
+  validSitePermissionManifestJSON,
   validStaticThemeManifestJSON,
   getStreamableIO,
   EMPTY_PNG,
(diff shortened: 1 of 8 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -2953,6 +2954,52 @@ describe('ManifestJSONParser', () => {
...
+      const json = validSitePermissionManifestJSON({
+        site_permissions: ['midi', 'not_a_valid_webapi_name'],
+      });
+      const manifestJSONParser = new ManifestJSONParser(json, linter.collector);
+      expect(linter.collector.errors).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({
+            code: messages.JSON_INVALID.code,
+            message:
+              '"/site_permissions/1" is not a valid key or has invalid extra properties',
```
rpl, line 2994: @willdurand @Rob--W 

To be fair this isn't exactly the same kind of validation error message we use for invalid permissions, for those we tweak the message as collected by ajv from the manifest parser class. 

We could do the same for the site_permissions but I'm not sure it would be really valuable given that this addon type submissions should be actually go through the usual submission flow used for the "real" webextensions (and so I opted to not do that, at least not until we decided that there would be value in doing that).

Issue #4169, opened 2022-02-04
this.parsedJSON.background.scripts.forEach is not a function


diox:
Someone hit this when validating their add-on in production:

https://sentry.prod.mozaws.net/operations/olympia-prod/issues/15978750/?referrer=github_plugin

The stacktrace was cut, here is an edited version of it:

```
this.parsedJSON.background.scripts.forEach is not a function

UnhandledPromiseRejectionWarning: TypeError: this.parsedJSON.background.scripts.forEach is not a function
    at ManifestJSONParser._validate (/addons-linter/src/parsers/manifestjson.js:448:44)
    at new ManifestJSONParser (/addons-linter/src/parsers/manifestjson.js:127:12)
```

AlexandraMoga:
I've verified this issue with the latest linter version (v4.9.0) using an add-on with incorrect `background.scripts` syntax in the manifest:
```
"background": {
		"scripts": "background_script.js"		
	}
```

These are the validation errors received:

[CLI]
```
ERRORS:

Code                      Message                            Description                                                   File            Line   Column
JSON_INVALID              "/background/scripts" is an        Your JSON file could not be parsed.
                          invalid additional property
MANIFEST_FIELD_REQUIRED   "/background/page" is a required   See https://mzl.la/1ZOhoEN (MDN Docs) for more information.   manifest.json
                          property
MANIFEST_FIELD_INVALID    "/background/scripts" should be    See https://mzl.la/1ZOhoEN (MDN Docs) for more information.   manifest.json
                          array
JSON_INVALID              "/background" is not a valid key   Your JSON file could not be parsed.
                          or has invalid extra properties
```

[AMO -dev]
![image](https://user-images.githubusercontent.com/31961530/153021095-fda36aca-5832-4b57-85d3-2064177c8835.png)

@willdurand besides the expected `"/background/scripts" should be array` error, there are some additional error messages related to the background field but, if I remember correctly, these are expected?

PR #4174, opened 2022-02-07
Check that background.scripts is an array before attempting to iterate over it

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4169


rpl (approved):
@willdurand ty! :heart:

PR #4175, opened 2022-02-07
Remove doctoc dependency

willdurand:
Here is a proposal to get rid of the `doctoc` dependency because (1) we
are not using it much and (2) some of its transitive dependencies have
"security" issues that we don't necessarily need.

(and (3) using this dependency seems overkill)


rpl (approved):
sgtm

PR #4181, opened 2022-02-08
Use `.test()` instead of `.match()`

### LINKED ISSUE #4179: Performance: match vs test
When using `match() as `if` condition where `true/false` is required, `test()` would be at least twice faster.

https://github.com/mozilla/addons-linter/blob/872132e9661ea02423b1226a8926d1c9ce890810/src/scanners/filename.js#L16-L28

If it is acceptable, I can make a PR.

erosman:
Fixes #4179


willdurand (approved):

willdurand:
Thanks @erosman

PR #4189, opened 2022-02-14
Prevent TypeError in opendialog-remote-uri rule

### LINKED ISSUE #4188: openDialog triggers addon-linter

I had a function called openDialog(null) in one of my React components. 

### What happened?

The addon-linter threw an error because it misunderstood it as window.openDialog().

I expect addon-linter to not throw an error in this situation.

When submitting a new addon (https://addons.mozilla.org/de/developers/addon/submit/distribution) with this problem on the web, just a generic error message is displayed. 

I ran the linter on my local machine to find out what the error was:
image

willdurand:
Fixes #4188

---

We don't check which object calls `openDialog()` so we may find a
completely unrelated method called `openDialog()` and that one could
take `null` as first argument (which isn't allowed in `window.openDialog()`).

In this case, we currently throw a `TypeError` but we shouldn't, which
is what this patch solves. I am not too sure whether we should report
a warning here, though.

The current patch does report a warning because it is _just_ a warning
but maybe we shouldn't?

### DIFF
--- a/src/rules/javascript/opendialog-remote-uri.js
+++ b/src/rules/javascript/opendialog-remote-uri.js
@@ -13,7 +13,12 @@ const rule = {
         ) {
           if (node.arguments.length) {
             const uri = node.arguments[0];
-            if (uri.type === 'Literal' && isLocalUrl(uri.value) === false) {
+            if (
+              uri.type === 'Literal' &&
+              // The first argument should be of type `string` but, if it is
+              // not for some reasons, report a warning to be extra-safe.
+              (typeof uri.value !== 'string' || isLocalUrl(uri.value) === false)
+            ) {
               return context.report(node, OPENDIALOG_REMOTE_URI.code);
             }
           }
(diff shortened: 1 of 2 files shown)


rpl (approved):
> The current patch does report a warning because it is _just_ a warning but maybe we shouldn't?

I agree with you that we should double-check if we really need a warning, and actually if we really still need this rule at all:

I think I may have also mentioned it in some other issue as part of which we investigated what the custom eslint rules are covering and if they are actually still useful, and this rule looked like one that we may actually not need anymore:

- in [Window.webidl the openDialog method that this rule is meant to cover is actually marked as ChromeOnly](https://searchfox.org/mozilla-central/rev/e66593593f3b356901011ea0fcdf9979728e9ae8/dom/webidl/Window.webidl#512-515) and so it shouldn't be actually accessible to the code running in an extensions page as it isn't for a regular webpage.

Anyway, in the meantime this short run fix looks good to me. 
 
(a privileged extension may still be able to reach it, by still not from the extension pages but only from the privileged JS code of an experimental API embedded in a privileged extensions, but it seems like for all non-privileged extensions this rule isn't actually going to cover anything).

PR #4263, opened 2022-04-11
fix: Migrate to addons-linter JSONSchema validator and Firefox JSONSchema import to ajv v8

rpl:
- changes migrate to draft07 (this mainly required to rename top level `id` property into `$id`)
- changes needed to account for `error.dataPath` renamed into `error.instancePath` in ajv v8
- changes needed to account for `should` changed into `must` in all schema validator generated messages (most of the changes related to this have been on the test side, but at least one change related to this was on the implementation side, where an ajv error was meant to be replaced with a custom error message)
- changes to tests assumptions due to `error.instancePath` on errors related to required properties missing changing from the missing property path to the parent property path 

Details pending a discussion:
- validation errors related to cryptominers missing an instancePath property (I haven't double-checked yet why that is the case, the expected validation error is still collected but it doesn't have the expected instancePath property).
- some more of the old options passed to the Ajv constructor have not been replaced with new options supported ajv v8


rpl:
@willdurand in the end I decided to keep the ajv-merge-patch dependency for now, there is already an amount of changes I had to apply to deal with porting the linter to ajv v8, and thre are some backward incompatible changes and so keeping as much or the surrounding code unmodified (e.g. the js code for the schema normalization and import, and the schema data previously imported as well) may be helpful to make it easier to identify and fix other potential changes in behavior and regressions that our test suite may have not caught already.

As a side note: I also tried on this branch to import the updated schema data from Firefox 100 beta, to confirm that the (even if minimal) changes to the code that imports the data are working correctly with the changes needed on the JSONSchema validator  module.

src/schema/validator.js
```diff
@@ -205,17 +203,23 @@ export class SchemaValidator {
...
-      errorDataPath: 'property',
-      jsonPointers: true,
+      // include schema and data properties in error objects.
       verbose: true,
-      schemas: this.schemas,
-      schemaId: 'auto',
+      // TODO: do the following options have new replacements in ajv v8?
+      // what was the reason for setting these options in older ajv version?
+      //
+      // errorDataPath: 'property',
```
willdurand, line 214: [This commit](https://github.com/ajv-validator/ajv/commit/ef40fbbda6f165d95a82dddb9103d6475b0fd130) says:

>  - _errorDataPath_ (deprecated): set `dataPath` to point to 'object' (default) or to 'property' when validating keywords `required`, `additionalProperties` and `dependencies`.

But that has been deprecated a while ago and removed in v8 so I am not sure it is super useful these days. If that does not break any test, we should be good?

src/schema/validator.js
```diff
@@ -205,17 +203,23 @@ export class SchemaValidator {
...
+      // jsonPointers: true,
```
willdurand, line 215: This configures the format of the `dataPath` (now `instancePath`). This is now the default according to https://github.com/ajv-validator/ajv/blob/0c40ebfa937d1d9cc01b4c5f06e7edfc4cfd5b00/docs/v6-to-v8-migration.md#removed-options

> jsonPointers: JSONPointers are used to report errors by default, jsPropertySyntax: true (deprecated) can be used if old format is needed.

tests/unit/test.linter.js
```diff
@@ -1501,15 +1501,18 @@ describe('Linter.extractMetadata()', () => {
       code: 'COINMINER_USAGE_DETECTED',
       column: 284,
       line: 2,
-      dataPath: 'CoinHive.CONFIG',
+      // TODO: determine why with ajv 8 dataPath or instancePath
+      // are not set to these values, and if we actually care
+      // (given that the validation warning is added anyway).
+      // instancePath: 'CoinHive.CONFIG',
```
willdurand, line 1507: This is probably caused by a missing change in `src/linter.js`. We need to update `dataPath` to `instancePath`:

```diff
diff --git a/src/linter.js b/src/linter.js
index 804c663f..9f6df5a5 100644
--- a/src/linter.js
+++ b/src/linter.js
@@ -758,9 +758,9 @@ export default class Linter {
             file: filename,
             line: matchedLine,
             column: matchedColumn,
-            // use dataPath for our actual match to avoid any obvious
+            // use instancePath for our actual match to avoid any obvious
             // duplicates
-            dataPath: match[0],
+            instancePath: match[0],
           });
         }
       });
```

willdurand (changes requested):
Thanks. That looks good.

Should we update `dataPath` in `src/parsers/locale-messagesjson.js` too?

src/parsers/manifestjson.js
```diff
@@ -261,15 +261,15 @@ export default class ManifestJSONParser extends JSONParser {
...
+      if (lowerCaseMessage === 'must match a schema in anyof') {
         // eslint-disable-next-line no-param-reassign
         error.message = 'is not a valid key or has invalid extra properties';
       }
     }
 
     const overrides = {
-      message: `"${error.dataPath}" ${error.message}`,
-      dataPath: error.dataPath,
+      message: `"${error.instancePath || '/'}" ${error.message}`,
```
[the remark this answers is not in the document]
rpl: I couldn''t find any explicit callout for this change in their v6-to-v8 migration doc (or v8 release notes) but it seems that for the error related to the nested property the error used to be associated to the nested property while now it is instead associated to the parent level.

And so when the error is associated to the root level, the instance path seems to undefined instead of '/'.

src/schema/validator.js
rpl: The instancePath is included and so we should be good on that side, but now that I re-read that point I think that it may be actually the change that is triggering the different behavior with what the instancePath to the parent object (what I've been describing in https://github.com/mozilla/addons-linter/pull/4263#discussion_r860237509).

src/schema/validator.js
rpl: yeah, that is how I was reading that, but wanted to double-check it. I'm going to rip that off.

tests/unit/test.linter.js
rpl: thank you! I looked to linter but at that point my brain got tired of looking for dataPath and filtered this out without telling me :-)

I'm going to apply this change and confirm that the related test case does pass as expected (and remove the TODO on the test side if that is the case as it seems it should be).

src/collector.js
```diff
@@ -51,26 +51,26 @@ export default class Collector {
     return this[`${type}s`];
   }
 
-  messagesAtDataPath(dataPath) {
-    if (dataPath === undefined) {
-      throw new Error('dataPath is required');
+  messagesAtDataPath(instancePath) {
```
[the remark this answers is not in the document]
rpl: I was really tempted to rename the function as well, but I was afraid that this PR was getting already annoying enough to review and make sure that we are not introducing any regression.

Maybe we can do that in a separate PR to merge this one (and the schema data import built on top of this PR, which I have in another branch that I can force push into #4276) sooner and deal with this additional cleanup separately. wdyt?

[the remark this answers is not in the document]
rpl: well, actually it seems to be used in just 3 places (1 actually on the test side), I'm going to rename this function in this PR.

src/schema/validator.js
rpl: I just re-tried what happens if I try to pass it, it throws and so I have removed the comment, it is likely related to the change in behavior that I did mention in the previous comment, but I guess that we have to accept this change in behavior.

rpl:
@willdurand this should now be ready for another review pass

willdurand:
> Should we update `dataPath` in `src/parsers/locale-messagesjson.js` too?

@rpl thoughts on this question?

rpl:
> > Should we update `dataPath` in `src/parsers/locale-messagesjson.js` too?
> 
> @rpl thoughts on this question?

@willdurand eh I've actually applied the additional changes related to that question but forgot to answer to the question, sorry about that.

The answer is:
- yes, those should also be updates
- ... which means that we may not have coverage for those bits because the tests didn't caught it

and so I applied the additional changes along with improving the test coverage of that part accordingly:
 -https://github.com/mozilla/addons-linter/pull/4263/commits/ac24af3fd61d93e19467f5abe90a6f0eb39ade01

Thanks for spotting that!

willdurand (approved):
r+, thanks. 🚢

Issue #4306, opened 2022-05-07
Detect and report permissions and properties requiring a privileged signature


rpl:
The changes to the Firefox JSONSchema data in work as part of [Bug 1739114 - Gate each privileged extension feature by a webextension permission](https://bugzilla.mozilla.org/show_bug.cgi?id=1739114) will allow the addons-linter ManifestJSONParser to detect if an extension is including a permission or manifest property that requires a privileged signature.

To make it easier to detect when an addon is being submitted to be signed by AMO while it should have been signed with a privileged signature (and then submitted to AMO as already signed to be discoverable and listed on AMO) we plan to:

- make use of the updates to the schema to allow src/schema/validator to detect and collect linting results for the permissions and properties that are marked as privileged in the schema data
- report the linting results related to privileged permissions and properties as:
  - blocking linting errors if the addon isn't already signed
  - non-blocking linting warning if the addon is already signed

(detecting the "already signed" status by checking for the presence of the signatures files in the add-on assets, similarly to how the `ALREADY_SIGNED` rule detects it).

wagnerand:
See also https://github.com/mozilla/addons-server/issues/18254

willdurand:
> To make it easier to detect when an addon is being submitted to be signed by AMO while it should have been signed with a privileged signature (and then submitted to AMO as already signed to be discoverable and listed on AMO) we plan to:

Note that this is also a requirement for Shipit as per https://github.com/mozilla-extensions/xpi-manifest/issues/136

ioanarusiczki:
@willdurand

I tested locally with linter 5.5.0, I added permissions from https://bugzilla.mozilla.org/show_bug.cgi?id=1739114#c5 into a standard add-on (with no privileged signature), and these are the results:

- for activityLog, mozillaAddons, networkStatus, normandyAddonStudy, telemetry, urlbar ->

![screen1](https://user-images.githubusercontent.com/33448286/169833496-ebbb9da9-42ad-4161-ac8b-203c24ec2beb.png)


-adding the permissions above, validation will fail with a red warning when `--privileged` is passed into the CLI

![needsmozilladdons](https://user-images.githubusercontent.com/33448286/169834139-4c45f369-84f4-4e14-a78f-a07ce22f9a94.png)

- validation will pass with no red warnings if mozillaAddons is also added together with the other permissions with `--privileged`

[...]

PR #4307, opened 2022-05-07
feat: Detect and report permissions and properties requiring a privileged signature

rpl:
This PR contains a draft for the changes related to fixing #4306 

- [x] consider command line flag / addons-linter Linter constructor optional parameter to adjust the behavior of the new rules during development or when executed by a non-AMO signing pipeline
- [x] move keyword and postprocess strings into src/const.js
- [x] introduce new rule code for errors reported only for privileged addons
- [x] remove import of Firefox 101 and 102 schema data used to prototype the rest of the changes 
- [x] look into how we may decouple the required changes to the addons-linter internals from the data being imported from Firefox Nightly 102, as a temporary measure to use until Firefox 102 got to the beta channel
- [x] keep the schema data temporarily imported from Firefox 101 and 102 updated with [Bug 1739114](https://bugzilla.mozilla.org/show_bug.cgi?id=1739114) schema changes


docs/index.html
```diff
@@ -396,6 +396,16 @@ <h2 id="web-extensions-%2F-manifest.json" tabindex="-1">Web Extensions / manifes
 <td>A manifest permission is not supported for the extension manifest_version</td>
 </tr>
 <tr>
+<td><code>MANIFEST_PERMISSIONS_PRIVILEGED</code></td>
```
willdurand, line 399:
```suggestion
<td><code>MANIFEST_PERMISSION_PRIVILEGED</code></td>
```

for consistency with the other names

src/messages/manifestjson.js
```diff
@@ -107,6 +124,27 @@ export const MANIFEST_PERMISSIONS = {
...
+  'MANIFEST_PERMISSIONS_PRIVILEGED';
+export function manifestPermissionsPrivileged(error) {
+  const messageTmpl =
+    i18n._(oneLine`%(instancePath)s: privileged permissions (%(privilegedPermissions)s)
+                     are only allowed in add-ons signed as privileged.`);
+  const message = i18n.sprintf(messageTmpl, {
+    instancePath: error.instancePath,
+    privilegedPermissions: error.params.privilegedPermissions
+      .map((p) => `"${p}"`)
+      .join(', '),
```
willdurand, line 137: nit: I think we should add this list at the end and use `JSON.stringify(error.params.privilegedPermissions)` so that it outputs something like this:

```
/instancePath: the following privileged permissions are only allowed in add-ons signed as privileged: ["foo", "bar"]
```

src/parsers/manifestjson.js
```diff
@@ -387,14 +404,25 @@ export default class ManifestJSONParser extends JSONParser {
...
       messages.MANIFEST_HOST_PERMISSIONS.code,
       messages.MANIFEST_PERMISSION_UNSUPPORTED,
       messages.MANIFEST_FIELD_UNSUPPORTED,
     ];
 
+    // If the extension is already signed, just warn about privileged permissions
+    // and privileged properties (add-ons signed as privileged are expected to be
+    // submitted to AMO as already signed if they need to be listed).
+    if (this.isAlreadySigned) {
+      warnings = [
```
willdurand, line 419: Maybe use `push(messages.MANIFEST_PERMISSIONS_PRIVILEGED, messages.MANIFEST_FIELD_PRIVILEGED)` and keep the declaration above with `const`.

src/linter.js
```diff
@@ -306,6 +306,10 @@ export default class Linter {
       const json = await this.io.getFileAsString(constants.MANIFEST_JSON);
       const manifestParser = new ManifestJSONParser(json, this.collector, {
         io: this.io,
+        // Account for unit tests where io.files isn't set.
```
willdurand, line 309: I don't understand why this is needed.

src/parsers/manifestjson.js
```diff
@@ -387,14 +404,25 @@ export default class ManifestJSONParser extends JSONParser {
...
       messages.MANIFEST_PERMISSIONS.code,
       messages.MANIFEST_OPTIONAL_PERMISSIONS.code,
       messages.MANIFEST_HOST_PERMISSIONS.code,
       messages.MANIFEST_PERMISSION_UNSUPPORTED,
       messages.MANIFEST_FIELD_UNSUPPORTED,
     ];
 
+    // If the extension is already signed, just warn about privileged permissions
+    // and privileged properties (add-ons signed as privileged are expected to be
+    // submitted to AMO as already signed if they need to be listed).
```
willdurand, line 417: Do we have something like this already? I am wondering if this isn't something AMO already takes into account (e.g., maybe some errors are skipped).

willdurand:
> keep the schema data temporarily imported from Firefox 101 and 102 updated with [Bug 1739114](https://bugzilla.mozilla.org/show_bug.cgi?id=1739114) schema changes
> remove import of Firefox 101 and 102 schema data used to prototype the rest of the changes

Maybe only commit the changes related to `PermissionPrivileged` so that there is no need for that.

src/schema/validator.js
```diff
@@ -542,6 +542,67 @@ export class SchemaValidator {
...
     });
+
+    const getPrivilegedPermissionsSet =
+      this.getPrivilegedPermissionsSet.bind(this);
+
+    validator.addKeyword({
+      keyword: 'postprocess',
+      validate: function validate(keywordSchemaValue, propValue) {
+        if (keywordSchemaValue === 'checkPrivilegePermission') {
+          const privilegedPermissions = getPrivilegedPermissionsSet(validator);
```
willdurand, line 553: Is this needed to have the right `this` in the validate function below? Maybe try arrow function for the `validate`.

src/schema/validator.js
```diff
@@ -542,6 +542,67 @@ export class SchemaValidator {
...
+          for (const permission of propValue) {
+            if (privilegedPermissions.has(permission)) {
+              found.add(permission);
+            }
+          }
+
+          if (found.size > 0) {
+            validate.errors = [
+              {
+                keyword: 'checkPrivilegedPermission',
```
willdurand, line 565: Probably a good idea to have a constant for this string value.

src/schema/validator.js
```diff
@@ -542,6 +542,67 @@ export class SchemaValidator {
...
+        const schemaNamespace = validator.getSchema(namespace).schema;
+        schemaNamespace.permissions.forEach((str) => results.add(str));
+      }
+    }
+    if (results.size === 0) {
+      throw new Error(
+        'Unable to retrieve the Privileged Permissions set from the schema data'
+      );
+    }
+    return results;
```
willdurand, line 605: This looks a bit packed but that also looks good to me.

docs/index.html
rpl: I'm ok with changing this name for consistency, maybe it may be even more consistent renamed as `MANIFEST_PRIVILEGED_PERMISSION` but I wanted to mention why I named is as `*_PERMISSIONS_*`:

- most of the other permissions rule named as `*_PERMISSION` are reporting issues related to a single permission
- whereas this one has to report one error for all privileged permissions being found, because the custom keyword is set on the property and gets the entire set of permissions and reports one linting issue for all privileged permissions found in that array

I'm not against renaming it form a plural into a singular form, but I wanted to point that out in case that may affect our decision about it.

src/linter.js
rpl: This was something I was meant to take another look into, basically it looked like not all unit tests were mocking io.files and so some unrelated test was failing because of this additional check.

But I took another look as I was planning to, and it turns out that `io.getFiles()` is actually mocked in all there where it should and it would be a better way to retrieve the map of the assets packaged with the extensions, I changed this part and now all tests are passing without any need to relax the expectations on the implementation side. 

I'll submit that part of the chages in this PR soon.

src/parsers/manifestjson.js
willdurand: @diox

rpl:
> Maybe only commit the changes related to PermissionPrivileged so that there is no need for that (git add -p)

@willdurand for the schema data I was planning to:
- create a separate PR for importing Firefox 101 data (tomorrow we will be already on Firefox 101.0b6 in the beta channel)
- then refresh this PR and look what is left in the diff and which is the actual final form of the schema changes from the Phabricator revision we will be landing in Nightly 102, based on that either: 
  - import the pieces be need by cherrypicking them from the diff
  - or use the overrides we apply from the data we put into "src/schema/updates" if that is feasible based on the form that the schema changes will have

src/schema/validator.js
rpl: yeah, it is

Ideally this would be much simpler if the schema imported becomes simpler ("simpler" as in "using less indirections", due to having to resolve $extend, $ref etc) or if we would be simplify the schema with those normalizations that I was looking into and mentioned to you some times ago.

Unfortunately I didn't got time to complete that and it was risky enough to do that along with updating to ajv v8, I feel that changing the format during the import is something that may be safer if done in its own separate PR (separately from other changes introducing new rules, or updating a complex-enough dependency like ajv).

src/schema/validator.js
rpl: yeah, I was waiting for the schema in the Phabricator revision to be finalized, before moving these into consts, I'll leave a TODO comment in this PR description comment for now.

rpl: Well, in the end I decided to anticipate this, and just update it if it changes on the phabricator patch side, this is now part of this PR, changes introduced in [c3ba1b0](https://github.com/mozilla/addons-linter/pull/4307/commits/c3ba1b0792303b4b9500059277c79ffba5838515).

rpl:
@willdurand given that you were already giving this some review passes, I decided to open it up to double-check if marking it ready for review would also convince CircleCI to run a job for it :-)

willdurand:
> [rpl](/rpl) (Luca Greco) requested a review from [willdurand](/willdurand) [ 12 hours ago](#event-6600987517)

@rpl I thought you wanted to land #4318 first and then rebase this PR?

tests/unit/schema/test.privileged.js
```diff
@@ -0,0 +1,71 @@
...
+
+describe('/permissions', () => {
+  const privilegedPermissions = ['mozillaAddons', 'telemetry', 'networkStatus'];
+
+  it.each(privilegedPermissions)(
+    'should report "%s" as a privileged permission',
+    (privilegedPerm) => {
+      const manifest = cloneDeep(validManifest);
+      manifest.permissions = ['fakeUnknownPerm', privilegedPerm, 'tabs'];
+      validateAddon(manifest);
```
willdurand, line 57: On all these test cases, there would be some value in having an empty line before and after `validateAddon(manifest);`.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -474,6 +474,106 @@ describe('ManifestJSONParser', () => {
...
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector,
+        {
+          // The following file is making sure that from a linter perspective
+          // the add-on being validated is including a signature file in the
+          // assets packaged with the extension.
+          //
+          // A separate Linter unit test is making sure this is set to true based
+          // on files included in the asserts packages with the extension.
```
willdurand, line 553: Not sure I understand this sentence.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -474,6 +474,106 @@ describe('ManifestJSONParser', () => {
...
+              )
+            ),
+          }),
+        ])
+      );
+      expect(warnings).toEqual([]);
+    });
+
+    // This test case confirms that we don't block the submittions if
+    // an addon already signed is including a privileged permission
```
willdurand, line 534:
```suggestion
    // an addon already signed includes a privileged permission
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -474,6 +474,106 @@ describe('ManifestJSONParser', () => {
...
+    // (as it is expected for privileged extensions that wants to be listed
```
willdurand, line 535:
```suggestion
    // (as it is expected for privileged extensions that want to be listed
```

src/yargs-options.js
```diff
@@ -37,6 +37,12 @@ const options = {
     type: 'boolean',
     default: false,
   },
+  'privileged-addon': {
```
willdurand, line 40: I'd name this flag `privileged` only, it is shorter and the linter CLI is already called "addons-linter".

src/parsers/manifestjson.js
```diff
@@ -395,6 +413,33 @@ export default class ManifestJSONParser extends JSONParser {
       messages.MANIFEST_FIELD_UNSUPPORTED,
     ];
 
+    // Message with the following codes will be:
+    //
+    // - omitted if the add-on is being explicitly validated as privileged
+    //   (when --privileged-addon cli option was passed or privilegedAddon
+    //   linter config is being set to true, e.g. locally as part of a CI
+    //   job or when validated as part of submitting the addon to the service
+    //   that will sign it as privileged).
```
willdurand, line 422: Where is this done? The code below only considers `isAlreadySigned` as far as I can see.

willdurand (commented):
I didn't finish the review yet. Here is a small list of comments.

tests/unit/test.linter.js
```diff
@@ -773,6 +773,28 @@ describe('Linter.getAddonMetadata()', () => {
...
+      },
+    };
+
+    const FakeManifestParser = sinon.spy(ManifestJSONParser);
+    await addonLinter.getAddonMetadata({
+      ManifestJSONParser: FakeManifestParser,
+    });
+    sinon.assert.calledOnce(FakeManifestParser);
+    expect(FakeManifestParser.firstCall.args[2].isAlreadySigned).toEqual(true);
+  });
```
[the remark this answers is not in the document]
rpl: :ok_hand: I'll check that we do have that or I'll add one in the next right of changes.

src/yargs-options.js
rpl: I'm going to rename the option in the next round of changes as we agreed.

src/yargs-options.js
```diff
@@ -37,6 +37,12 @@ const options = {
...
+    describe:
+      'Lint the addon as privileged signature (disable errors on privileged manifest properties and permissions).',
```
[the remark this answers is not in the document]
rpl: I guess that I may actually remove this part of the change in this PR if the description may end up in the other PR where we are going to land the dummy option upfront this additional set of changes. I'll keep the comment as unresolved until I'll rebase the PR and confirm this part is going to be already covered in the rebased change.

src/schema/imported/privacy.json
```diff
@@ -297,7 +297,8 @@
         "nonPersistentCookies": {
           "type": "boolean",
           "default": false,
-          "description": "Whether to create all cookies as nonPersistent (i.e., session) cookies."
+          "description": "Whether to create all cookies as nonPersistent (i.e., session) cookies.",
+          "deprecated": "This property has no effect anymore and its value is always <code>false<code>."
```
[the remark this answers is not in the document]
rpl: None, this is part of the schema changes that I'll rip off for sure.
I'm just not sure yet if I'll keep the diff as is or try to apply changes from a file in src/schema/updates, but I'll look into that asap (likely as part of the next round of changes).

rpl:
@willdurand I've added the piece that we discussed today, the validation  rule meant to be executed only when --privileged is passed. 

In case it may be helpful to be able to look only the new additional changes, all those changes are built on top of the rest of the previous state of this PR in [9ef2cad](https://github.com/mozilla/addons-linter/pull/4307/commits/9ef2cadaaa2b412d1b0ac865fd95032028f96209).

tests/unit/test.linter.js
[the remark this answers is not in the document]
rpl: This should now be covered by the changes to the test case applied in [4c6b5dd](https://github.com/mozilla/addons-linter/pull/4307/commits/4c6b5dddb041d0b33811268f62aec2d3ba260e3c).

src/yargs-options.js
rpl: Option renamed by rebasing this PR on top of #4326.

src/yargs-options.js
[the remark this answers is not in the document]
rpl: this should also be covered by the rebase of this PR on top of #4326.

src/schema/imported/privacy.json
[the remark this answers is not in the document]
rpl: I've reverted the changes to the schema data unrelated to privileged permissions and manifest properties in the last push in this PR.

Unfortunately as I was mentioning yesterday, the changes to the schema are a bit too spread for what the schema updates files can easily provide (e.g. there are types renamed along with some new property added, and I don't think we do support renaming a type from schema updates at the moment).

We may end up landing the changes to the schema and them reimport the rest of Firefox 102 data once the current Nightly gets to beta.

rpl:
@willdurand this PR should now be ready for another review pass from your perspective.

docs/rules.md
```diff
@@ -99,6 +99,10 @@ Rules are sorted by severity.
 | `MANIFEST_FIELD_PRIVILEGEDONLY`                         | warning  | A manifest field ignored on non-privileged add-ons                                              |
 | `MANIFEST_FIELD_UNSUPPORTED`                            | warning  | A manifest field is not supported (or not supported for the extension manifest_version)         |
 | `MANIFEST_PERMISSION_UNSUPPORTED`                       | warning  | A manifest permission is not supported for the extension manifest_version                       |
+| `MANIFEST_PERMISSIONS_PRIVILEGED`                       | error    | A manifest permission only allowed in add-ons signed as privileged                              |
+| `MOZILLA_ADDONS_PERMISSION_REQUIRED`                    | error    | "mozillaAddons" permission is a mandatory requirement for signing add-ons as privileged         |
+| `PRIVILEGED_FEATURES_REQUIRED`                          | error    | Privileged features are mandatory for privileged add-ons, but none has been detected            |
+| `MANIFEST_FIELD_PRIVILEGED`                             | error    | A manifest field only allowed in add-ons signed as privileged                                   |
```
willdurand, line 105:
```suggestion
| `MANIFEST_FIELD_PRIVILEGED`                             | error    | A manifest field is only allowed in privileged extensions                                       |
```

src/messages/manifestjson.js
```diff
@@ -51,6 +51,23 @@ export function manifestFieldUnsupported(fieldName, error) {
   };
 }
 
+export const MANIFEST_FIELD_PRIVILEGED = 'MANIFEST_FIELD_PRIVILEGED';
+export function manifestFieldPrivileged(error) {
+  const messageTmpl =
+    i18n._(oneLine`%(instancePath)s: privileged manifest fields
+                     are only allowed in add-ons signed as privileged.`);
```
willdurand, line 58:
```suggestion
                     are only allowed in privileged extensions.`);
```

src/messages/manifestjson.js
```diff
@@ -107,6 +124,26 @@ export const MANIFEST_PERMISSIONS = {
   file: MANIFEST_JSON,
 };
 
+export const MANIFEST_PERMISSIONS_PRIVILEGED =
+  'MANIFEST_PERMISSIONS_PRIVILEGED';
+export function manifestPermissionsPrivileged(error) {
+  const messageTmpl =
+    i18n._(oneLine`%(instancePath)s: the following privileged permissions
+                     are only allowed in add-ons signed as privileged:
```
willdurand, line 132:
```suggestion
                     are only allowed in privileged extensions:
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
   ),
   file: MANIFEST_JSON,
 };
+
+export const PRIVILEGED_FEATURES_REQUIRED = 'PRIVILEGED_FEATURES_REQUIRED';
+export function privilegedFeaturesRequired(error) {
+  const messageTmpl = i18n._(
+    oneLine`%(instancePath)s: Requiring a privileged permission is a mandatory requirement for signing an add-on as privileged.
```
willdurand, line 720:
```suggestion
    oneLine`%(instancePath)s: Privileged extensions should declare privileged permissions.
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
...
+
+  const message = i18n.sprintf(messageTmpl, {
+    instancePath: error.instancePath,
+  });
+
+  return {
+    code: PRIVILEGED_FEATURES_REQUIRED,
+    message,
+    description: i18n._(oneLine`
+      This submission does not declare any privileged permission. It does not need to be signed with the privileged certificate.
```
willdurand, line 732:
```suggestion
      This extension does not declare any privileged permission. It does not need to be signed with the privileged certificate.
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
...
+      Please upload it directly to addons.mozilla.org.
```
willdurand, line 733:
```suggestion
      Please upload it directly to https://addons.mozilla.org/.
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
...
+  };
+}
+
+export const MOZILLA_ADDONS_PERMISSION_REQUIRED =
+  'MOZILLA_ADDONS_PERMISSION_REQUIRED';
+export function mozillaAddonsPermissionRequired(error) {
+  const messageTmpl =
+    error.instancePath === '/permissions'
+      ? i18n._(
+          oneLine`%(instancePath)s: "mozillaAddons" permission is a mandatory requirement for signing an add-on as privileged.`
```
willdurand, line 745:
```suggestion
          oneLine`%(instancePath)s: The "mozillaAddons" permission is required for privileged extensions.`
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
...
+        )
+      : i18n._(
+          oneLine`%(instancePath)s: "mozillaAddons" permission is a mandatory requirement for add-ons including privileged manifest fields.`
```
willdurand, line 748:
```suggestion
          oneLine`%(instancePath)s: The "mozillaAddons" permission is required for extensions that include privileged manifest fields.`
```

src/messages/manifestjson.js
```diff
@@ -676,3 +713,54 @@ export const EXTENSION_ID_REQUIRED = {
...
+    instancePath: error.instancePath,
+  });
+
+  return {
+    code: MOZILLA_ADDONS_PERMISSION_REQUIRED,
+    message,
+    description:
+      error.instancePath === '/permissions'
+        ? i18n._(
+            oneLine`This submission does not require the "mozillaAddons" permission, which is mandatory for add-ons expected to be signed as privileged.`
```
willdurand, line 761:
```suggestion
            oneLine`This submission does not require the "mozillaAddons" permission, which is required for privileged extensions.`
```

src/messages/manifestjson.js
```diff
@@ -107,6 +124,26 @@ export const MANIFEST_PERMISSIONS = {
...
+                     %(privilegedPermissions)s.`);
+  const message = i18n.sprintf(messageTmpl, {
+    instancePath: error.instancePath,
+    privilegedPermissions: JSON.stringify(error.params.privilegedPermissions),
+  });
+
+  return {
+    code: MANIFEST_PERMISSIONS_PRIVILEGED,
+    message,
+    description: message,
```
willdurand, line 142: I think the `description` should include a bit more information so I'd reuse `message` + the following:

```
Please refer to https://github.com/mozilla-extensions/xpi-manifest to learn more about privileged extensions and signing.
```

src/messages/manifestjson.js
```diff
@@ -51,6 +51,23 @@ export function manifestFieldUnsupported(fieldName, error) {
...
+    i18n._(oneLine`%(instancePath)s: privileged manifest fields
+                     are only allowed in add-ons signed as privileged.`);
+  const message = i18n.sprintf(messageTmpl, {
+    instancePath: error.instancePath,
+  });
+
+  return {
+    code: MANIFEST_FIELD_PRIVILEGED,
+    message,
+    description: message,
```
willdurand, line 66: I think the `description` should include a bit more information so I'd reuse `message` + the following:

```
Please refer to https://github.com/mozilla-extensions/xpi-manifest to learn more about privileged extensions and signing.
```

src/parsers/manifestjson.js
```diff
@@ -395,6 +421,33 @@ export default class ManifestJSONParser extends JSONParser {
       messages.MANIFEST_FIELD_UNSUPPORTED,
     ];
 
+    // Message with the following codes will be:
+    //
+    // - omitted if the add-on is being explicitly validated as privileged
+    //   (when --privileged-addon cli option was passed or `privileged` set
+    //   to true in linter config is being set to true, e.g. locally as part
+    //   of a CI job or when validated as part of submitting the addon to the
+    //   service that will sign it as privileged).
```
willdurand, line 430:
```suggestion
    // - omitted if the add-on is being explicitly validated as privileged
    //   when the `--privileged` cli option was passed or `privileged` is
    //   set to true in the linter config. This is the case for privileged
    //   extensions in the https://github.com/mozilla-extensions/ org.
```

src/parsers/manifestjson.js
```diff
@@ -395,6 +421,33 @@ export default class ManifestJSONParser extends JSONParser {
...
+    // - reported as warnings if the add-on is already signed
+    //   (because it is expected for addons signed as privileged to be
+    //   submitted to AMO to become listed, and so the warning is meant to
+    //   be just informative and to let extension developers and reviewers
+    //   to know that the extension is expected to be signed as privileged
+    //   or it wouldn't work).
+    //
+    // - reported as errors if the add-on isn't signed
+    //   (which prevents to submit a privileged addon on AMO and have it
+    //   signed with a non privileged keys by mistake).
```
willdurand, line 441:
```suggestion
    // - reported as errors if the add-on isn't signed, which should
    //   reject the submission of a privileged extension on AMO (and
    //  have it signed with a non privileged certificate by mistake).
```

willdurand (commented):
Some more comments, it is looking good overall.

src/const.js
```diff
@@ -233,3 +233,14 @@ export const RESTRICTED_PERMISSIONS = new Map([
...
+
+export const SCHEMA_KEYWORDS = {
+  ANY_OF: 'anyOf',
+  DEPRECATED: 'deprecated',
+  MIN_MANIFEST_VERSION: 'min_manifest_version',
+  MAX_MANIFEST_VERSION: 'max_manifest_version',
+  PRIVILEGED: 'privileged',
+  REQUIRED: 'required',
+  TYPE: 'type',
+  VALIDATE_PRIVILEGED_PERMISSIONS: 'validatePrivilegedPermissions',
```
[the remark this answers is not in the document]
rpl: sure, I also splitted the custom keywords not part of the JSONSchema specs into a separate const, and then merged that group into the one listing all keywords we refer to in the manifest parser internals through these consts.

src/messages/manifestjson.js
rpl: Composing the string would be technically possible by using i18n.sprintf, but Instead of repeating the message in a description defined as a composed string I oped to set the description to the string you suggested (and the message as is).

Same for the one that follows.

(As a side note, this should be the only deviation from the suggested edits, everything else should be applied exactly as in the related suggested change).

src/parsers/manifestjson.js
```diff
@@ -342,8 +362,14 @@ export default class ManifestJSONParser extends JSONParser {
     ) {
       baseObject = messages.MANIFEST_BAD_HOST_PERMISSION;
       overrides.message = `Permissions ${error.message}.`;
-    } else if (error.keyword === 'type') {
+    } else if (error.keyword === SCHEMA_KEYWORDS.TYPE) {
       baseObject = messages.MANIFEST_FIELD_INVALID;
+    } else if (error.keyword === SCHEMA_KEYWORDS.PRIVILEGED) {
+      baseObject = error.params?.isPrivilegedAddon
```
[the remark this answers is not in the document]
rpl: In a previous version (I think one that only existed locally on my machine) the error reported on non privileged addons did not have any params at all, but then I changed it to always include the params (mainly because it seems more consistent to make sure that the error collected by a certain custom keyword has always the same format).

And so in practice error.params should always be defined with a keyword set to SCHEMA_KEYWORDS.PRIVILEGED, and so I'll just be removing the unnecessary optional chaining from this one.

I think that for L336 I may have not being passing through such an intermediate version (both the errors did have the exact same privilegedPermissions property set in the error params, and that is likely the only reason why the optional chaining was left only in this one.

src/schema/validator.js
```diff
@@ -525,23 +530,152 @@ export class SchemaValidator {
...
+      //   permissions already required).
+      if (this.privileged) {
+        if (found.size === 0 || !hasMozillaAddonsPermission) {
+          checkForPrivilegedPermissionsValidate.errors = [
+            {
+              keyword: SCHEMA_KEYWORDS.VALIDATE_PRIVILEGED_PERMISSIONS,
+              params: {
+                postprocess: keywordSchemaValue,
+                privilegedPermissions: Array.from(found),
+                isPrivilegedAddon: this.privileged,
```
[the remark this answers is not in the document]
rpl: I preferred to leave this to isPrivilegedAddon because there is also a privilegedPermissions and so I think that the current name is more immediately clear ("clear that it refers to the <<addon>> being privileged") then if it would be called more generically`isPrivileged`.

Given that it is an internal implementation detail it doesn't sound like a big deal that it doesn't match exactly how the cli option is named.

src/schema/validator.js
```diff
@@ -525,23 +530,152 @@ export class SchemaValidator {
...
+      keyword: SCHEMA_KEYWORDS.MIN_MANIFEST_VERSION,
       validate: createManifestVersionValidateFn(
-        'min_manifest_version',
+        SCHEMA_KEYWORDS.MIN_MANIFEST_VERSION,
         (minMV, manifestVersion) => minMV <= manifestVersion
       ),
       errors: true,
     });
+
+    const checkForPrivilegedPermissionsValidate = (
```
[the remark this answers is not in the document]
rpl: I did the same also for the other validate function introduced in this PR.

src/messages/manifestjson.js
```diff
@@ -676,3 +717,53 @@ export const EXTENSION_ID_REQUIRED = {
...
+  const message = i18n.sprintf(messageTmpl, {
+    instancePath: error.instancePath,
+  });
+
+  return {
+    code: PRIVILEGED_FEATURES_REQUIRED,
+    message,
+    description: i18n._(oneLine`
+      This extension does not declare any privileged permission. It does not need to be signed with the privileged certificate.
+      Please upload it directly to "https://addons.mozilla.org/".
```
rpl, line 736: @willdurand while I was going through the last round of suggested edits I looked back to this description and I'm wondering if this may sounds like something the linter is 100% sure of and then pushing privileged extensions developers to don't even double-check if the addon was actually expected to be privileged.

Besides the couple of corner case that the linter can't detect at the moment, there is a chance that the developer may have missed to specify a permission or mistyped a property which would prevent the addon to be detected as requiring a privileged signature (but may be only breaking a secondary feature).

I'm also thinking that we could also detect if the manifest is including a restricted match pattern (which require also the "mozillaAddons" permission to work as expected) and use also that as an hint that the add-on is expected to be privileged. 

wdyt? 

do we want to leave that as a separate followup to this PR?

rpl:
@willdurand this should be ready for another review pass, I left as not resolved previous review comments that may still be worth touching base about, but I'm also linking then below to make it easier to find them in the PR:

- https://github.com/mozilla/addons-linter/pull/4307#discussion_r875117948
- https://github.com/mozilla/addons-linter/pull/4307#discussion_r875118164
- https://github.com/mozilla/addons-linter/pull/4307#discussion_r875133503
- https://github.com/mozilla/addons-linter/pull/4307#discussion_r875135631

And I added the following discussion point(s):
- https://github.com/mozilla/addons-linter/pull/4307#discussion_r875151550

src/messages/manifestjson.js
willdurand: ah ok yeah, that's probably fine. I don't know if this is consistent with what we're doing already but it isn't wrong so thank you for that

src/schema/validator.js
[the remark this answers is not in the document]
willdurand: yep, no problem.

src/messages/manifestjson.js
willdurand: Yeah, I think that's why we'll need to enforce `"mozillaAddons"` for privileged extensions in Firefox in the future. That will be the best way for developers to think about that and know whether their extension is privileged or not.

In the meantime, we can probably detect restricted match patterns indeed. I would think that a separate PR could be easier to review but up to you really.

src/const.js
```diff
@@ -233,3 +233,24 @@ export const RESTRICTED_PERMISSIONS = new Map([
...
 ]);
+
+export const SCHEMA_KEYWORDS_CUSTOM = {
+  MIN_MANIFEST_VERSION: 'min_manifest_version',
+  MAX_MANIFEST_VERSION: 'max_manifest_version',
+  PRIVILEGED: 'privileged',
+  // This custom keyword doesn't exist on the Firefox side, but it is injected into the
+  // schema data as part of the schema data inmport and used by the linter to hook up
+  // custom validation logic for privileged permissions listed in the privileged extensions
+  // manifest permissions property.
```
willdurand, line 244:
```suggestion
  // custom validation logic for privileged extensions.
```

src/messages/manifestjson.js
willdurand, line 736: Do we need the quotes around `https://addons.mozilla.org/`?

willdurand (approved):
Thank you! r+wc

src/parsers/manifestjson.js
```diff
@@ -92,6 +93,7 @@ export default class ManifestJSONParser extends JSONParser {
       selfHosted = getDefaultConfigValue('self-hosted'),
```
willdurand, line 93: Here we pass the value of the `self-hosted` flag, I am wondering if we could/should do the same for `privileged`. I think it's going to be needed for #4333. WDYT?

src/messages/manifestjson.js
rpl: Not necessarily, I always quote urls to make it clear where they start and end, and so I have done it without realizing it wasn't quoted in your suggested version.

I removed the quote, given that the url is known (and not an arbitrary one interpolated into the message) I don't think we really need to quote it.

src/parsers/manifestjson.js
rpl: @willdurand sure, makes sense. I applied this diff as if and included it in the last updated on this PR (btw, thanks for the diff :heart:)

src/messages/manifestjson.js
rpl: k, I'll create a separate PR for the restricted match pattern detection.

For the Firefox side, we could consider to introduce sooner a warning logged on missing "mozillaAddons" permission in extensions detected as privileged, that wouldn't break backward compatibility and may provide more visible hints to the extension developers hitting it.

willdurand (approved):

Issue #4303, opened 2022-05-06
Require the addon id to be included in the manifest for manifest_version 3 extensions


rpl:
As part of [Bug 1755872 - Make add-on ID mandatory in the manifest for MV3](https://bugzilla.mozilla.org/show_bug.cgi?id=1755872) triage we agreed to look into enforcing the manifest of manifest_version 3 extension to explicitly include the addon id in the manifest at submission time before being signed.

After we have enforced it for all signed extension we may consider what other issues we still have and if additional changes on the Firefox side would help with fill those gaps.

ioanarusiczki:
@willdurand 
I've tested today on addons-linter 5.3.0 with addon samples (from prod) and a dictionary. 
- I added into manifest.json "manifest_version": 3
- used "id" under "applications" or  "browser_specific_settings" 

With 'addons-linter sideberry.xpi  --max-manifest-version=3':

This is the error message when there's no id in the manifest:

![error message](https://user-images.githubusercontent.com/33448286/168610337-4bbf2692-2c15-43fd-a413-91350d4147f0.png)

When an invalid id format is used I get the same error messages as before:
![invalid id](https://user-images.githubusercontent.com/33448286/168611853-cfd3de3e-ea41-4b91-8daa-02a8a2a37530.png)

+2 erros that have been detailed into https://github.com/mozilla/addons-linter/issues/3843#issuecomment-892687526 when there's MV3 with MV2 specific keys.

I'm going to complete all my testing tomorrow, check themes and lang packs too. So far it looks good to me.

ioanarusiczki:
@willdurand 

- All addon types require an id when manifest version is 3. 

- I've noticed on AMO dev an error that it not thrown by the linter locally, so it must be from server:

![error from server](https://user-images.githubusercontent.com/33448286/168816258-7c649e6d-f494-4ae3-ba29-3ca6bb5c250e.png)

Using the CLI `addons-linter langpack.xpi  --max-manifest-version=4` I will not get the error above, I hope that's alright.

willdurand:
> * I've noticed on AMO dev an error that it not thrown by the linter locally when the MV > 3, so it must be from server:

@ioanarusiczki so you essentially have a `manifest.json` with `"manifest_version": 4` in it?

> Using the CLI `addons-linter langpack.xpi --max-manifest-version=4` I will not get the error above, I hope that's alright.

[...]

PR #4315, opened 2022-05-11
Require extension ID for MV3+

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4303

---

I couldn't use the schema because it is MV3+ only and it was easier to introduce another method.

The link to the EW page is also mentioned in https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings so I thought it'd make sense to use that as well. We'll probably want to add a notice for this new MV3+ requirement, see: https://github.com/mozilla/extension-workshop/issues/1310.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -570,6 +570,11 @@
 <td>error</td>
 <td>A permission requires &quot;strict_min_version&quot; to be set to a specific Firefox version</td>
 </tr>
+<tr>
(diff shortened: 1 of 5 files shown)


willdurand:
hum, not sure why there is no Circle CI...

PR #4326, opened 2022-05-16
Add a new --privileged CLI option

### LINKED ISSUE #4324: Add a `--privileged` CLI flag
This issue is related to #4306. We want to add a new CLI flag named `--privileged` (off/disabled by default), which will configure the linter to handle privileged extensions. If we can add it ahead of time, we can start adding it to xpi-template and xpi-manifest, even if there is no rules for privileged extensions yet. That will avoid potential issues introduced by #4306.

willdurand:
Fixes #4324

---

This PR adds a new CLI flag that does not do anything (so far, see: #https://github.com/mozilla/addons-linter/issues/4306).  (I also fixed typos)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,13 @@ The addons-linter will check your add-on and show you errors, warnings, and frie
 addons-linter --help
 ```
 
+#### Privileged extensions
+
+The addons-linter can lint privileged extensions **only** when the `--privileged` option is passed to it. This option changes the behavior of the linter to:
+
+1. emit errors when the input file (or directory) is a regular extension (i.e. the extension does not use privileged features)
+2. hide messages related to privileged features (e.g., permissions and properties) when the input file (or directory) is a privileged extension
+
 ### Linter API Usage
 
 You can use the linter directly as a library to integrate it better into your development process.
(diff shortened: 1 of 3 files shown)


willdurand:
ah wait, I forgot to update the docs...

README.md
```diff
@@ -40,6 +40,13 @@ The addons-linter will check your add-on and show you errors, warnings, and frie
 addons-linter --help
 ```
 
+#### Privileged extensions
+
+The addons-linter can lint **privileged extensions** using the `--privileged` CLI flag. This option changes the behavior of the linter to:
```
rpl, line 45: Not a big dea, but I wonder if this should be actually stating the opposite (something like: "For linting privileged extensions addons-linter should be executed by including the `--privileged` CLI flag").

README.md
```diff
@@ -40,6 +40,13 @@ The addons-linter will check your add-on and show you errors, warnings, and frie
...
+
+1. emit errors when the input file (or directory) is a regular extension (i.e. the extension does not use privileged features)
```
rpl, line 47: Nit, Maybe "the extension does not require any privileged feature through its manifest.json file"? just to be clear that it is what is in declared are required in the manifest to make a difference here).

rpl (approved):
@willdurand looks good, the following 2 review comments inline are just thoughts / optional nits that we may want to consider but not really blocking for this PR.

Issue #4333, opened 2022-05-17
Mute error about `update_url` when `--privileged` is specified


willdurand:
I just noticed that the linter emits an error when `update_url` is used in a privileged extension (see: [TaskCluster `live.log` for the remote-settings-devtools extension](https://firefox-ci-tc.services.mozilla.com/tasks/QORvbwUaQwi0TGYFNcZPng/runs/0/logs/public/logs/live.log)).

I believe this is OK because we cannot determine whether the extension will be listed or unlisted so we should disable this message when `--privileged` is passed.

ioanarusiczki:
@willdurand 

1. I added the update_url into a manifest for a standard extension plus ""mozillaAddons" permission:
- without --privileged flag, 2 red errors are triggered CODE MANIFEST_PERMISSIONS_PRIVILEGED and MANIFEST_UPDATE_URL
- with --privileged flag, only one yellow warning is triggered ->  MANIFEST_UPDATE_URL (which used to be red without the flag)

2. A Mozilla signed extension:
- without --privileged flag, the update_url throws an error.
- with  --privileged flag, the update_url becomes a yellow warning. 

I believe this is the expected result.

PR #4334, opened 2022-05-18
Do not emit an error for update_url when --privileged is used

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4333
~~:warning: Depends on #4307~~

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -1609,6 +1609,60 @@ describe('ManifestJSONParser', () => {
       expect(errors[0].message).toContain('update_url');
     });
 
+    it('emits a warning for privileged extensions', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        applications: {
+          gecko: {
+            update_url: 'https://foo.com/bar',
+          },
+        },
+      });
+
+      const manifestJSONParser = new ManifestJSONParser(
+        json,
+        addonLinter.collector,
+        {
+          selfHosted: false,
+          schemaValidatorOptions: {
+            privileged: true,
+          },
+        }
+      );
+
+      expect(manifestJSONParser.isValid).toEqual(true);
+      expect(addonLinter.collector.errors.length).toBe(0);
+      const { warnings } = addonLinter.collector;
+      expect(warnings[0].code).toEqual(messages.MANIFEST_UPDATE_URL.code);
+      expect(warnings[0].message).toContain('update_url');
+    });
+
+    it('is not an issue if self-hosted and privileged', () => {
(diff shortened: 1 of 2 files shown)


willdurand:
need a rebase obviously but the last commit should be good for a review (and the previous one is what I mentioned in #4307)

rpl:
@willdurand [006bc49](https://github.com/mozilla/addons-linter/pull/4334/commits/006bc4951b7f9bf7a346e792ee6f852715a02ad2) looks good to me, thanks for catching it and looking into how to handle that additional corner case.

I'm wondering if we should be still stll add a message but as a warning, to let the developer to know that if they intend to submit the resulting signed xpi to be listed on AMO then they should remove the update_url otherwise they wouldn't be able to submit that version as it would be blocked by a linting error when submitted to AMO as not-selfhosted.

wdyt?

tests/unit/parsers/test.manifestjson.js
```diff
@@ -1609,6 +1609,60 @@ describe('ManifestJSONParser', () => {
...
+            privileged: true,
+          },
+        }
+      );
+
+      expect(manifestJSONParser.isValid).toEqual(true);
+      expect(addonLinter.collector.errors.length).toBe(0);
+      const { warnings } = addonLinter.collector;
+      expect(warnings[0].code).toEqual(messages.MANIFEST_UPDATE_URL.code);
+      expect(warnings[0].message).toContain('update_url');
```
rpl, line 1637: given that the rest of the file is using this pattern I think that it is ok to use it here as well for consistency.

In more recent test case I opted to use jest's expect.arrayContains/expect.objectContains to make the test to check just what it cares about and don't fail if there are other warnings logged that the test case doesn't care about, but I think it would be better doing it consistently for the entire test file instead of using a different pattern only on this one.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -1609,6 +1609,60 @@ describe('ManifestJSONParser', () => {
...
+          selfHosted: true,
+          schemaValidatorOptions: {
+            privileged: true,
+          },
+        }
+      );
+
+      expect(manifestJSONParser.isValid).toEqual(true);
+      expect(addonLinter.collector.errors.length).toBe(0);
+      expect(addonLinter.collector.warnings.length).toBe(0);
```
rpl, line 1663: Nit, I prefer the pattern: `expect(addonLinter.collector.errors).toEqual([])` because it produces a more detailed failure message when it fails and it can more easily figured out directly from the failure we may get from the CI.

rpl (approved):
Looks good, just one small optional nit on the test side (asserting no error or warnings using `expect(someArray).toEqual([])` instead of `expect(someArray.length).toBe(0)`).

PR #4345, opened 2022-05-25
feat: allow 'wasm-unsafe-eval' keyword

### LINKED ISSUE #4343: Accept 'wasm-unsafe-eval' in content_security_policy
Firefox has introduced support for the `'wasm-unsafe-eval'` source in a Content Security Policy: https://bugzilla.mozilla.org/show_bug.cgi?id=1740263.

This source is also permitted in extensions, as explained at https://bugzilla.mozilla.org/show_bug.cgi?id=1766027#c2.

This value is currently rejected by addons-linter by the following logic: 
https://github.com/mozilla/addons-linter/blob/ba3b667a13bde65250bcb1ebd814c9d57ccfa80d/src/const.js#L205
https://github.com/mozilla/addons-linter/blob/12b64da87e223f042f7f5fd2ab2a6832435243f1/src/parsers/manifestjson.js#L1051

We need to update the logic, e.g. by adding `wams-unsafe-eval` to the regexp, and add a regression test around
https://github.com/mozilla/addons-linter/blob/eee957380ed69574aaafef50ab948951e1b29142/tests/unit/parsers/test.manifestjson.js#L1509

Rob--W:
Fixes #4343

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -1456,12 +1456,22 @@ describe('ManifestJSONParser', () => {
         // forseeable future. See http://bit.ly/2wG6LP0 for more details-
         "script-src 'self' 'unsafe-inline';",
 
+        // 'wasm-unsafe-eval' is permitted, despite the unsafe-eval substring.
+        "script-src 'self' 'wasm-unsafe-eval'",
+
         // 'default-src' is insecure, but the limiting 'script-src' prevents
         // remote script injection
         "default-src *; script-src 'self'",
         "default-src https:; script-src 'self'",
         "default-src example.com; script-src 'self'",
         "default-src http://remote.com/; script-src 'self'",
+
+        // In theory, script-src should override default-src, and the insecure
+        // 'unsafe-eval' directive should be ignored. But the implementation
+        // rejects 'unsafe-eval' unconditionally, despite the script-src
+        // override. So this test case fails with MANIFEST_CSP_UNSAFE_EVAL.
+        // See https://github.com/mozilla/addons-linter/pull/4345#discussion_r881860151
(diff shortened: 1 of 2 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -1456,12 +1456,19 @@ describe('ManifestJSONParser', () => {
...
         // 'default-src' is insecure, but the limiting 'script-src' prevents
         // remote script injection
         "default-src *; script-src 'self'",
         "default-src https:; script-src 'self'",
         "default-src example.com; script-src 'self'",
         "default-src http://remote.com/; script-src 'self'",
+        // This should be accepted because script-src overrides default-src,
+        // but because unsafe-eval is unconditionally rejected despite the
+        // script-src override, the linter complains anyway:
+        // "script-src 'wasm-unsafe-eval'; default-src 'unsafe-eval'",
```
willdurand, line 1471: Should we fix that somehow?

Rob--W: I have considered doing so, but decided not to since it's not really a big deal to block its use from `default-src`.

`'unsafe-eval'` is only meaningful in `script-src` and `style-src`:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_style_expressions

... and `style-src` is not validated here. Extensions wanting to use the `unsafe-eval` part of the CSP should just use `style-src` instead of `default-src`, for clarity.

willdurand: ok, can you maybe rework the comment a bit?

willdurand (approved):
thanks

Issue #4351, opened 2022-06-01
Should we raise `UNSUPPORTED_API` if priviledged addon?


leplatrem:
### Describe the problem and steps to reproduce it:

Our Remote Settings DevTools extension defines an Experiments API in the `manifest.json`:
https://github.com/mozilla-extensions/remote-settings-devtools/blob/3e6463c56ff5b2db401794dcee47221c0a8193cb/extension/manifest.json#L34-L43

But the linter still warns about an `UNSUPPORTED_API` in `content/script.js`.

```
[task 2022-05-24T16:52:12.500Z] UNSUPPORTED_API     experiment…   This API has not been   content/sc…   7      24    
[task 2022-05-24T16:52:12.500Z]                     s.remotese…   implemented by          ript.js                    
[task 2022-05-24T16:52:12.500Z]                     ttings is     Firefox.                                           
[task 2022-05-24T16:52:12.500Z]                     not                                                              
[task 2022-05-24T16:52:12.501Z]                     supported           
```

### What happened?

The ability to define new experiments API is reserved to priviledged addons. And the linter does not seem aware of this. (similar to #4334 ?)

### What did you expect to happen?

The linter should not warn about an unsupported API if a priviledged addon is built and the experiment API is defined in the manifest.

willdurand:
Thanks for filing this issue Mathieu. That's also something I was wondering and I think we could silent this warning because it does not make much sense when `--privileged` is passed anyway. We don't have extra validation.

ioanarusiczki:
@willdurand I ran locally linter 5.8.0 for the remote-settings-devtools addon and when I add --privileged flag, the yellow warning about the API is no longer displayed.

Validation summary: 

![validation results](https://user-images.githubusercontent.com/33448286/173546966-9f61ea13-33c5-40ed-8b35-fc6b7ea17da3.png)

PR #4354, opened 2022-06-03
Detect (some) privileged APIs defined in the manifest.json

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4351 (maybe)

---

This is currently a draft to avoid raising `UNSUPPORTED_API` warnings when privileged extensions use experimental APIs defined in the `manifest.json` file.


alht962:
در تاریخ جمعه ۳ ژوئن ۲۰۲۲،‏ ۲۱:۰۷ William Durand ***@***.***>
نوشت:

willdurand:
@rpl WDYT (before I write tests)

src/parsers/manifestjson.js
```diff
@@ -1088,6 +1088,44 @@ export default class ManifestJSONParser extends JSONParser {
...
+    // API. The data in the `manifest.json` would look like this:
+    //
+    // "experiment_apis": {
+    //   "some-name": {
+    //     "schema": "experiments/some-name/schema.json",
+    //     "parent": {
+    //       "scopes": ["addon_parent"],
+    //       "script": "experiments/some-name/api.js",
+    //       "paths": [["some", "name"]]
+    //     }
```
rpl, line 1108: @willdurand technically the "experiment_apis" can also have a "child" property for API namespace that are meant to be implemented in the child process, I'm not sure if there is any out-of-the-tree privileged extensions that is actually using that, but we do support it and there are at least a couple of privileged extensions in tree using that:
- https://searchfox.org/mozilla-central/source/browser/extensions/pictureinpicture/manifest.json#23-30
- https://searchfox.org/mozilla-central/source/browser/extensions/webcompat/manifest.json#38-45

and so it seems reasonable to silent the error also for API namespace declared as "child".

rpl (commented):
> @rpl WDYT? (before I write tests)

@willdurand :+1: the approach seems pretty reasonable to me, I haven't looked closely all the bits but from a quick look it seems to be doing exactly what I was thinking of when we briefly discussed about it last week.

I added a comment below related to the "extension_apis" declared with a "child process"-side implementation, it is less common but it seems reasonable to silence errors related to those as well if the extension is being validated as privileged.

src/parsers/manifestjson.js
willdurand: Ah yeah, thanks for letting me know. That's something I wanted to double-check tomorrow :P

tests/unit/rules/javascript/test.unsupported_browser_api.js
```diff
@@ -121,4 +121,67 @@ describe('unsupported browser APIs', () => {
...
+      const code = 'browser.some.experiment.foo()';
+      const jsScanner = new JavaScriptScanner(code, 'api.js', {
+        addonMetadata: {
+          experimentApiPaths: ['some.experiment'],
+        },
+        privileged,
+      });
+
+      const { linterMessages } = await runJsScanner(jsScanner);
+      expect(linterMessages.length).toEqual(expectedMessages);
```
rpl, line 147: could we also assert that we got the expected reported linting issue along with checking we got the expected number of linting issues reported?

rpl (changes requested):
Follows a new round of review comments, mostly nits and a tweak to the schema data to cover a corner case (empty extension_apis object) that would be nice to be raising a blocking linting error (because it is currently leaving the extension installed and enabled but in a broken state when installed in Firefox).

Let me know wdyt.

src/parsers/manifestjson.js
```diff
@@ -1088,6 +1088,49 @@ export default class ManifestJSONParser extends JSONParser {
...
+    // "paths" like:
+    //
+    // ['some.name']
+    //
+    // We could have either a "parent" or "child" property for each API.
+    //
+    const apiPaths = [];
+    for (const key of Object.keys(experiment_apis)) {
+      const { child, parent } = experiment_apis[key];
+      const paths = child?.paths || parent?.paths;
```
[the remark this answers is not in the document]
willdurand: mm right

tests/unit/parsers/test.manifestjson.js
```diff
@@ -4203,4 +4204,142 @@ describe('ManifestJSONParser', () => {
...
+          schema: 'path/to/schema.json',
+          ...props,
+        },
+      };
+    };
+
+    it.each([
+      { title: 'undefined apis', experiment_apis: undefined, expected: [] },
+      { title: 'no apis', experiment_apis: null, expected: [] },
+      { title: 'empty apis', experiment_apis: {}, expected: [] },
```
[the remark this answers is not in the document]
willdurand: I can do that but in a different PR.

tests/unit/rules/javascript/test.unsupported_browser_api.js
willdurand: I think so

src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -28,6 +28,16 @@ const rule = {
...
 
+          if (
+            privileged &&
+            addonMetadata.experimentApiPaths &&
+            addonMetadata.experimentApiPaths.some((apiPath) =>
+              api.startsWith(apiPath)
+            )
+          ) {
+            return;
+          }
```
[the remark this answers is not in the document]
willdurand: seems a bit excessive to me, with the optional chaining requested above, it should be enough.

tests/unit/parsers/test.manifestjson.js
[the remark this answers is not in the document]
rpl: sure, if you prefer let's just do it in a separate PR.

I filed a bug for the Firefox side of this  as [Bug 1773301](https://bugzilla.mozilla.org/show_bug.cgi?id=1773301).

I also forgot to link the section related to `minProperties` from the JSONSchema specs: https://json-schema.org/understanding-json-schema/reference/object.html#size

src/parsers/manifestjson.js
```diff
@@ -1088,6 +1088,49 @@ export default class ManifestJSONParser extends JSONParser {
...
+      const { child, parent } = experiment_apis[key];
+      const parentPaths = parent?.paths ?? [];
+      const childPaths = child?.paths ?? [];
+
+      [...parentPaths, ...childPaths]
+        .filter((p) => Array.isArray(p) && p.length)
+        .forEach((p) => apiPaths.add(p.join('.')));
+    }
+
+    return Array.from(apiPaths.values());
```
rpl, line 1131: Nit, `return Array.from(apiPaths);` would also get us the same result, or maybe we may even just leave it as a Set?

src/rules/javascript/webextension-unsupported-api.js
```diff
@@ -28,6 +28,15 @@ const rule = {
             return;
           }
 
+          if (
+            privileged &&
+            addonMetadata?.experimentApiPaths?.some((apiPath) =>
+              api.startsWith(apiPath)
```
rpl, line 34: given that we only care about the namespace, could we change this check to `apiPath === namespace`?
(and if we could do that, maybe we may opt to leave `experimentaApiPaths` to be of type Set and just do `addonMetadata?.experimentApiPaths?.has(module)`). 

I'm wondering also if by checking for `api.startsWith` we may be actually ignoring an API call that isn't technically defined by "experiment_apis"  but may still share a prefix with one that is defined and pass this check, but I haven't checked if that is actually technically possible.

Not a super big deal given that this is mainly silencing warnings and so even if there was some reported as a warning it wouldn't be blocking signing the addon.

tests/unit/rules/javascript/test.unsupported_browser_api.js
```diff
@@ -121,4 +121,96 @@ describe('unsupported browser APIs', () => {
     const { linterMessages } = await runJsScanner(jsScanner);
     expect(linterMessages.length).toEqual(0);
   });
+
+  describe('Experiement APIs', () => {
```
rpl, line 125: Nit, typo: `Experiement` => `Experiment`

tests/unit/rules/javascript/test.unsupported_browser_api.js
```diff
@@ -121,4 +121,96 @@ describe('unsupported browser APIs', () => {
...
+    it.each([
+      {
+        title: 'allows known privileged APIs when the extension is privileged',
+        privileged: true,
+        expectedMessage: null,
+      },
+      {
+        title: 'emits an error when the extension is not privileged',
```
rpl, line 133: Not a big deal because the categorization between errors and warnings does not happen at the JavascriptScanner level, but given that technically [UNSUPPORTED_API is categorized as a warning](https://mozilla.github.io/addons-linter/#web-extensions-%2F-manifest.json) we may change `emits an error` into  `emits a warning` if we prefer to reflect that in the tests descriptions.

rpl (approved):
Thanks Will, r+wc

src/rules/javascript/webextension-unsupported-api.js
willdurand: I don't think `apiPath === namespace` would work for compound api paths, e.g., `some.experiment`.

willdurand: Here is an example: https://github.com/mozilla-extensions/searchengine-devtools/blob/2726d3e07a6f9723eda370f8635f4fcc5df976c4/extension/manifest.json#L41

The namespace should be `experiments.searchengines` I think. Using `api` in the check works currently but that wouldn't work for `foo.bar.baz.api()`. We should fix how we compute the `namespace` but that does not seem trivial.

tests/unit/parsers/test.manifestjson.js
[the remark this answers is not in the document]
willdurand: WIP at: https://github.com/mozilla/addons-linter/pull/4363

src/rules/javascript/webextension-unsupported-api.js
willdurand: See: https://github.com/mozilla/addons-linter/issues/4364

PR #4397, opened 2022-07-05
fix: Prevent mv3 permissions errors to be wrongly reported in mv2 addon validation

### LINKED ISSUE #4367: Host permissions regarded as invalid in `permissions` if other invalid permissions are present

Load an extension with the following manifest file:
```
{
  "manifest_version": 2,
  "version": "25.8.0",
  "name": "Example",
  "permissions": [
    "SOME_INVALID_PERMISSION",
    "https://www.example.com/"
  ]
}
```

### What happened?

`https://www.example.com/` is seen as an invalid permission. 

`https://www.example.com/` should be seen as valid.

This only happens if `SOME_INVALID_PERMISSION` is present.

### LINKED ISSUE #4521: Possible typo in src/schema/validator.js (mix_manifest_version)
https://github.com/mozilla/addons-linter/blob/300ad70f9caf8cee628778af2c5d28e2302167e1/src/schema/validator.js#L90

Should that be `max_manifest_version`?

rpl:
Fixes #4367
(also fixes #4521, including the fix for the typo in `isRelevantError` helper function and additional test coverage added to test.schema.js to cover the fixed typo with an explicit test case).  

The unexpected validation warning reported for a valid host permission (filed as addons-linter issue #4367)
seems to be due to the fact that both manifest_version 2 and manifest_version 3 definitions for the
permissions manifest field are being processed.

The current implementation is trying to make sure to filter out the errors not relevant for the current
add-on based on the error.parentSchema property (See isRelevantError helper function defined in
src/schema/validator.js).

Unfortunately this approach is unable to filter out all irrelevant errors, in particular for JSONSchema
types that are referenced in a min/max_manifest_version conditioned entry with more then 1 level deep nesting,
the error.parentSchema will not be the one explicitly including the min/max_manifest_version property and
this is what makes the error triggered by the min_manifest_version 3 version of the "permissions" to be
wrongly reported as part of a manifest_version 2 add-on validation.

This patch is currently filtering the schema data at "JSONSchema data compile"-time and omitting from the
schema data all entries with min/max_manifest_version that do not match the minimum and maximum manifest_version
set per schema validator config or if they don't apply to the actual add-on manifest_version.

NOTE: Ideally a better (and less expesive at runtime) approach may be to this kind of JSONSchema data filtering
at import time (by creating a different set of JSONSchema data per manifest_version).

### DIFF
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -39,20 +39,20 @@ function isRelevantError({
 
   const { minimum, maximum } = allowedManifestVersionsRange;
 
-  const errorMinManifestVersion =
+  let errorMinManifestVersion =
     error.params?.min_manifest_version ??
     error.parentSchema?.min_manifest_version ??
     minimum;
 
-  let errorMaxManifestVersion =
+  const errorMaxManifestVersion =
     error.params?.max_manifest_version ??
     error.parentSchema?.max_manifest_version ??
     maximum;
 
   // Make sure the computed error max version is always >= to the computed min version.
-  errorMaxManifestVersion = Math.max(
-    errorMaxManifestVersion,
-    errorMinManifestVersion
+  errorMinManifestVersion = Math.min(
+    errorMinManifestVersion,
+    errorMaxManifestVersion
   );
 
   const isTopLevelManifestKey =
@@ -140,7 +140,8 @@ function filterErrors(
 }
 
 function getManifestVersionsRange(validatorOptions) {
-  const { minManifestVersion, maxManifestVersion } = validatorOptions;
+  const { minManifestVersion, maxManifestVersion, addonManifestVersion } =
+    validatorOptions;
 
   const minimum =
     minManifestVersion == null
@@ -151,6 +152,8 @@ function getManifestVersionsRange(validatorOptions) {
     maxManifestVersion == null
       ? getDefaultConfigValue('max-manifest-version')
       : maxManifestVersion;
+
+  const addon = addonManifestVersion == null ? minimum : addonManifestVersion;
 
   // Make sure the version range is valid, if it is not:
   // raise an explicit error.
@@ -163,7 +166,7 @@ function getManifestVersionsRange(validatorOptions) {
     );
   }
 
-  return { minimum, maximum };
+  return { minimum, maximum, addon };
 }
 
 export class SchemaValidator {
@@ -409,9 +412,55 @@ export class SchemaValidator {
   }
 
   _compileAddonValidator(validator) {
-    const { minimum, maximum } = this.allowedManifestVersionsRange;
+    const { minimum, addon, maximum } = this.allowedManifestVersionsRange;
 
-    const schemaData = deepPatch(this.schemaObject, {
+    const replacer = (key, value) => {
+      if (Array.isArray(value)) {
+        const patchedValue = value.filter((item) => {
+          let includeItem = true;
+          if (
+            item?.min_manifest_version &&
+            minimum < item.min_manifest_version
+          ) {
+            includeItem =
+              item.min_manifest_version >= minimum &&
+              item.min_manifest_version <= maximum &&
+              item.min_manifest_version <= addon;
+          }
+          if (
+            item?.max_manifest_version &&
+            maximum > item.max_manifest_version
+          ) {
+            includeItem =
+              item.max_manifest_version >= minimum &&
+              item.max_manifest_version <= maximum &&
+              item.max_manifest_version >= addon;
+          }
+
+          return includeItem;
+        });
+        return patchedValue;
+      }
+      return value;
+    };
+
+    // Omit from the manifest schema data all entries that includes a
+    // min/max_manifest_version which is outside of the minimum
+    // and maximum manifest_version currently allowed per validator
+    // config and if they do not apply to the addon manifest_version.
+    //
+    // NOTE: the rest of the schema data isn't filtered based on the
+    // current manifest version.
+    //
+    // TODO(https://github.com/mozilla/addons-linter/issues/4512):
+    // this shouldn't be necessary anymore if we do generate two sets
+    // of schema data (one for MV2 and one for MV3) as part of
+    // importing and normalizing the JSONSchema data from Firefox.
+    const patchedSchemaObject = JSON.parse(
+      JSON.stringify(this.schemaObject, replacer, 2)
+    );
+
+    const schemaData = deepPatch(patchedSchemaObject, {
       types: {
         ManifestBase: {
           properties: {
@@ -518,6 +567,12 @@ export class SchemaValidator {
         }
 
         if (!res) {
+          // If the addon manifest is out of an enum values min/max manifest version range,
+          // don't report an additional validation error for the mix/max_manifest_version
+          // keyword validation function.
+          if (schema.enum) {
+            return true;
+          }
           validate.errors = [
             {
               keyword,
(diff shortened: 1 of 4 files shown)


src/schema/validator.js
```diff
@@ -409,9 +412,47 @@ export class SchemaValidator {
...
+    };
+
+    // Omit from the schema data all entries that includes a
+    // min/max_manifest_version which is outside of the minimum
+    // and maximum manifest_version currently allowed per validator
+    // config and if they do not apply to the addon manifest_version.
+    //
+    const schemaObject = JSON.parse(
+      JSON.stringify(this.schemaObject, replacer, 2)
+    );
```
rpl, line 453: @willdurand this is currently traversing the JSONSchema data by using JSON.stringify, which is likely not the best way to traverse the json objects given that it is stringifying all data and then parsing it again, I used it mainly as a quick way to give the approach a try and confirm if that would be a possible way to avoid these unexpected errors.

Besides the JSON.stringify trick, as I also mentioned in the PR description comment a better (and less expesive at runtime) approach may be to this kind of JSONSchema data filtering at import time (by creating a different set of JSONSchema data per manifest_version).

We may also consider looking into that approach now and see if we can get an alternative patch without a too big of an effort.

Let me know wdyt.

(in the PR description I tried to explain in a bit more detail what is going on in practice, but I'm happy to have a chat if that description didn't yet help you to get a good picture of the underlying issue).

willdurand:
> Let's touch base about this points in the next few days, e.g. after you got a chance to take a look to this version.

Yeah, that's actually something I forgot to let you know earlier today.. I read this new patch and your comments/concerns. We should discuss some time next week!

willdurand:
As per our last discussion, the idea was to:

1. move these changes to a new PR with a different goal: we eventually want to simplify the JSON schema in the linter so that will be a first step (having two sets of schema for mv2 and mv3). Later, we'll merge all files to have a JSON file per manifest version. After that we'll try to simplify the structure of the JSON schema
2. switch back to the old patch (with the replacer function) to fix https://github.com/mozilla/addons-linter/issues/4367. The unproven fear was that it would impact the perfs negatively but we came to the conclusion that this was unproven and rather unlikely

willdurand:
> [...] Later, we'll merge all files to have a JSON file per manifest version [...]

I filed #4511 for that, let's file issues for:

> [...] we eventually want to simplify the JSON schema in the linter so that will be a first step (having two sets of schema for mv2 and mv3) [...]

and

> [...]  After that we'll try to simplify the structure of the JSON schema [...]

willdurand:
I filed https://github.com/mozilla/addons-linter/issues/4512 and https://github.com/mozilla/addons-linter/issues/4513

rpl:
@willdurand I have restored the initial version of this PR as we agreed on, then added the test case introduced along with the other approach to provide explicit test coverage around the issue this is meant to fix.

willdurand:
> (also fixes #4523, including the fix for the typo in `isRelevantError` helper function and additional test coverage added to test.schema.js to cover the fixed typo with an explicit test case).

That fixes https://github.com/mozilla/addons-linter/issues/4521, not the PR.

rpl:
> > (also fixes #4523, including the fix for the typo in `isRelevantError` helper function and additional test coverage added to test.schema.js to cover the fixed typo with an explicit test case).
> 
> That fixes #4521, not the PR.

right, reference to the right issue number fixed in the PR description.

src/schema/validator.js
```diff
@@ -409,9 +412,55 @@ export class SchemaValidator {
...
+          }
+
+          return includeItem;
+        });
+        return patchedValue;
+      }
+      return value;
+    };
+
+    // Omit from the manifest schema data all entries that includes a
```
willdurand, line 447:
```suggestion
    // Omit from the manifest schema data all entries that include a
```

src/schema/validator.js
```diff
@@ -518,6 +567,12 @@ export class SchemaValidator {
         }
 
         if (!res) {
+          // If the addon manifest is out of an enum values min/max manifest version range,
+          // don't report an additional validation error for the mix/max_manifest_version
```
willdurand, line 571:
```suggestion
          // don't report an additional validation error for the min/max_manifest_version
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -528,6 +528,89 @@ describe('ManifestJSONParser', () => {
...
+          }),
+        ])
+      );
+      expect(manifestJSONParser.isValid).toEqual(false);
+
+      // Confirm that the unexpected origin is reported as a validation
+      // error if the extension is a manifest version 3 extension.
+      const jsonv3 = validManifestJSON({
+        manifest_version: 3,
+        applications: {
```
willdurand, line 573:
```suggestion
        browser_specific_settings: {
```

tests/unit/schema/test.schema.js
```diff
@@ -146,33 +200,138 @@ describe('Schema JSON', () => {
...
+          expect.objectContaining({
+            instancePath: '/browser_action/popup',
+            keyword: 'type',
+            message: 'must be string',
+          }),
+        ])
+      );
 
-      expect(validator.validateAddon.errors).toEqual(
+      // Errors for values in the fantary_api enum restricted to MV1 and MV3 should not be reported.
```
willdurand, line 229:
```suggestion
      // Errors for values in the fantasy_api enum restricted to MV1 and MV3 should not be reported.
```

tests/unit/schema/test.schema.js
```diff
@@ -146,33 +200,138 @@ describe('Schema JSON', () => {
...
+      );
+      expect(validateAddon.errors).toEqual(
+        expect.not.arrayContaining([
+          expect.objectContaining({
+            instancePath: '/fantasy_api/fantasy_key_obsolete',
+          }),
+        ])
+      );
+
+      // Errors for values in the fantary_api enum supported by MV2 should be reported.
```
willdurand, line 256:
```suggestion
      // Errors for values in the fantasy_api enum supported by MV2 should be reported.
```

tests/unit/schema/test.schema.js
```diff
@@ -146,33 +200,138 @@ describe('Schema JSON', () => {
...
+        expect.not.arrayContaining([
+          expect.objectContaining({
+            instancePath: '/browser_action/popup',
+            keyword: 'type',
+            message: 'must be string',
+          }),
+        ])
+      );
+
+      // Errors for values in the fantary_api enum restricted to MV1 and MV2 should not be reported.
```
willdurand, line 297:
```suggestion
      // Errors for values in the fantasy_api enum restricted to MV1 and MV2 should not be reported.
```

tests/unit/schema/test.schema.js
```diff
@@ -146,33 +200,138 @@ describe('Schema JSON', () => {
...
+      );
+      expect(validateAddon.errors).toEqual(
+        expect.not.arrayContaining([
+          expect.objectContaining({
+            instancePath: '/fantasy_api/fantasy_key_obsolete',
+          }),
+        ])
+      );
+
+      // Errors for values in the fantary_api enum supported by MV3 should be reported.
```
willdurand, line 324:
```suggestion
      // Errors for values in the fantasy_api enum supported by MV3 should be reported.
```

willdurand (changes requested):
Looking good, a few questions and typos.

src/schema/validator.js
```diff
@@ -163,7 +166,7 @@ function getManifestVersionsRange(validatorOptions) {
     );
   }
 
-  return { minimum, maximum };
+  return { minimum, maximum, addon };
```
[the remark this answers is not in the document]
rpl: I renamed it to currentAddon (just to make it clear it is the version of the current addon being validated, and not a current manifest version option set on the linter, like minimum and maximum are).

Let me know if that is a reasonable name from this property also from your perspective.

rpl:
@willdurand I've just updated this PR based on your last round of review comments (https://github.com/mozilla/addons-linter/pull/4397/commits/a14d5017569935f2d6f52251109091f2e5f6b052), this should be ready for a final review pass.

src/schema/validator.js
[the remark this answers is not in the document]
willdurand: Ok so that's my fault actually. I should have suggested something else: I don't like that we return the addon manifest version in something called "range". This is no longer a range anymore since it has more than min/max.

I tried to move the logic for `currentAddon` where it is actually used and the test suite still passes:

```diff
diff --git a/src/schema/validator.js b/src/schema/validator.js
index 6c700742..57f0ef25 100644
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -140,8 +140,7 @@ function filterErrors(
 }
 
 function getManifestVersionsRange(validatorOptions) {
-  const { minManifestVersion, maxManifestVersion, addonManifestVersion } =
-    validatorOptions;
+  const { minManifestVersion, maxManifestVersion } = validatorOptions;
 
   const minimum =
     minManifestVersion == null
@@ -164,10 +163,7 @@ function getManifestVersionsRange(validatorOptions) {
     );
   }
 
-  const currentAddon =
-    addonManifestVersion == null ? minimum : addonManifestVersion;
-
-  return { minimum, maximum, currentAddon };
+  return { minimum, maximum };
 }
 
 export class SchemaValidator {
@@ -413,8 +409,11 @@ export class SchemaValidator {
   }
 
   _compileAddonValidator(validator) {
-    const { minimum, currentAddon, maximum } =
-      this.allowedManifestVersionsRange;
+    const { minimum, maximum } = this.allowedManifestVersionsRange;
+    const { addonManifestVersion } = this._options;
+
+    const manifestVersion =
+      addonManifestVersion == null ? minimum : addonManifestVersion;
 
     const replacer = (key, value) => {
       if (Array.isArray(value)) {
@@ -427,7 +426,7 @@ export class SchemaValidator {
             includeItem =
               item.min_manifest_version >= minimum &&
               item.min_manifest_version <= maximum &&
-              item.min_manifest_version <= currentAddon;
+              item.min_manifest_version <= manifestVersion;
           }
           if (
             item?.max_manifest_version &&
@@ -436,7 +435,7 @@ export class SchemaValidator {
             includeItem =
               item.max_manifest_version >= minimum &&
               item.max_manifest_version <= maximum &&
-              item.max_manifest_version >= currentAddon;
+              item.max_manifest_version >= manifestVersion;
           }
 
           return includeItem;
```

WDYT?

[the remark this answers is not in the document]
rpl: @willdurand ah, now I see what you meant, I applied your diff and made some more small tweaks on top of it (basically moved it into a Validator class getter, see https://github.com/mozilla/addons-linter/pull/4397/commits/3c2a0a7c72e0d92f29da3c110b9e05f5f64a77f9), how that looks to you?

willdurand (approved):

PR #4363, opened 2022-06-08
Add minProperties to experiment_apis in schema

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4365

---

With privileged flag:

```
$ ./bin/addons-linter ../test-ext --privileged
Validation Summary:

errors          4
notices         0
warnings        0

ERRORS:

Code                                 Message                          Description                                                     File            Line   Column
JSON_INVALID                         "/manifest_version" must be <=   Your JSON file could not be parsed.
                                     2
PRIVILEGED_FEATURES_REQUIRED         /permissions: Privileged         This extension does not declare any privileged permission. It   manifest.json
                                     extensions should declare        does not need to be signed with the privileged certificate.
                                     privileged permissions.          Please upload it directly to https://addons.mozilla.org/.
MOZILLA_ADDONS_PERMISSION_REQUIRED   /experiment_apis: The            /experiment_apis: The "mozillaAddons" permission is required    manifest.json
                                     "mozillaAddons" permission is    for extensions that include privileged manifest fields.
                                     required for extensions that
                                     include privileged manifest
                                     fields.
JSON_INVALID                         "/experiment_apis" must NOT      Your JSON file could not be parsed.
                                     have fewer than 1 properties
```

Without privileged flag:

```
$ ./bin/addons-linter ../test-ext
Validation Summary:

errors          3
notices         0
warnings        0

ERRORS:

Code                        Message                          Description                                                   File            Line   Column
JSON_INVALID                "/manifest_version" must be <=   Your JSON file could not be parsed.
                            2
MANIFEST_FIELD_PRIVILEGED   /experiment_apis: privileged     Please refer to                                               manifest.json
                            manifest fields are only         https://github.com/mozilla-extensions/xpi-manifest to learn
                            allowed in privileged            more about privileged extensions and signing.
                            extensions.
JSON_INVALID                "/experiment_apis" must NOT      Your JSON file could not be parsed.
                            have fewer than 1 properties

```

PR #4407, opened 2022-07-14
feat: Import Firefox 103 schema data

rpl:
This PR includes a refresh of the Firefox imported JSONSchema data.

There is no change to the actual JSONSchema, besides a tweak to the experiments.json file which is actually equivalent to the one previously set in #4363 

In addition to that this PR include another commit to reimport a small set of changes to the import script to make sure the updates for experiments.json are actually applied (which used to be in #4363 but got lost in the final updates of the PR and didn't land in master as it should have).

NOTE: no change of the previous addons-linter behaviors is expected to be introduced by these changes.

### DIFF
--- a/src/schema/firefox-schemas-import.js
+++ b/src/schema/firefox-schemas-import.js
@@ -703,17 +703,14 @@ function loadSchemasFromFile(basePath) {
 
 export function importSchemas(firefoxPath, ourPath, importedPath) {
   const rawSchemas = loadSchemasFromFile(firefoxPath);
-  const ourSchemas = {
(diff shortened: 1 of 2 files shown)


willdurand:
> In addition to that this PR include another commit to reimport a small set of changes to the import script to make sure the updates for experiments.json are actually applied (which used to be in #4363 but got lost in the final updates of the PR and didn't land in master as it should have).

Whaaaaat?  I was trying to find where this code could have been lost but losing changes in the PR right before merging it must be a first! 😱

willdurand (approved):

Issue #4432, opened 2022-08-08
Make hidden and action mutually exclusive


willdurand:
Hidden extensions should ideally not have any browser actions because it is a bit odd to not see the extension anywhere (e.g., not in about:addons) but still see a button in the toolbar and no way to remove the extension for instance. FTR, extensions can only be "hidden" if they are _privileged_ extensions. That isn't possible for "regular" extensions.

We currently don't have any privileged extensions with browser actions but there is no restriction. We are going to [make that happen in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1781998) but it would be good to emit an error too.

ioanarusiczki:
@rpl @willdurand 

I tried with linter version 5.14.0 using a mozilla-signed extension to add browswer_action and page_action into manifest.json.

 I didn't get an error with my testing sample. I'm attaching it, maybe I'm doing something wrong? 

[test2.zip](https://github.com/mozilla/addons-linter/files/9351134/test2.zip)

willdurand:
This is only for privileged extensions with `hidden: true` in the `manifest.json` file and one (or all) of `browser_action`, `page_action` and `action`.

ioanarusiczki:
@willdurand 

I get the error message locally while testing it with --privileged flag and "hidden":true. Here's action property used in MV3:

![actionmv3](https://user-images.githubusercontent.com/33448286/185057605-484e6893-b961-4f41-ade3-b11c33304943.png)

I've also noticed that on the testing servers (AMO dev) this error is not available. Is this expected? Maybe these addons are uploaded with web-ext (It's just an assumption).

willdurand:
> I've also noticed that on the testing servers (AMO dev) this error is not available. Is this expected? Maybe these addons are uploaded with web-ext (It's just an assumption).

Yes, it is expected I think. The linter shouldn't be executed with the `--privileged` flag on AMO

willdurand:
> I get the error message locally while testing it with --privileged flag and "hidden":true. Here's action property used in MV3:

That looks good to me.

PR #4433, opened 2022-08-08
Make hidden and action/page_action/browser_action mutually exclusive (for privileged add-ons)

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4432

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -122,7 +122,7 @@ Rules are sorted by severity.
 | `NO_PLACEHOLDER_CONTENT`                                | error    | Placeholder is missing the content                                                              |
 | `JSON_INVALID`                                          | error    | JSON is not well formed                                                                         |
 | `JSON_DUPLICATE_KEY`                                    | error    | Duplicate key in JSON                                                                           |
-| `MANIFEST_VERSION_INVALID`                              | error    | manifest_version in manifest.json is not valid.                                                 |
+| `MANIFEST_VERSION_INVALID`                              | error    | manifest_version in manifest.json is not valid                                                  |
 | `PROP_NAME_MISSING`                                     | error    | name property missing from manifest.json                                                        |
 | `PROP_NAME_INVALID`                                     | error    | name property is invalid in manifest.json                                                       |
 | `PROP_VERSION_MISSING`                                  | error    | version property missing from manifest.json                                                     |
@@ -138,7 +138,8 @@ Rules are sorted by severity.
 | `IGNORED_APPLICATIONS_PROPERTY`                         | warning  | Usage of both `applications` and `browser_specific_settings` properties                         |
 | `RESTRICTED_HOMEPAGE_URL`                               | error    | Linking to addons.mozilla.org in `homepage_url` or `developer.url` is not allowed               |
 | `RESTRICTED_PERMISSION`                                 | error    | A permission requires "strict_min_version" to be set to a specific Firefox version              |
-| `EXTENSION_ID_REQUIRED`                                 | error    | The extension ID is mandatory for Manifest Version 3 (and above) extensions.                    |
+| `EXTENSION_ID_REQUIRED`                                 | error    | The extension ID is mandatory for Manifest Version 3 (and above) extensions                     |
+| `HIDDEN_NO_ACTION`                                      | error    | The `hidden` and `browser_action`/`page_action` (or `action`) properties are mutually exclusive |
 
 ### Static Theme / manifest.json
 
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -730,6 +730,26 @@ export default class ManifestJSONParser extends JSONParser {
 
     this.validateRestrictedPermissions();
     this.validateExtensionID();
+    this.validateHiddenAddon();
+  }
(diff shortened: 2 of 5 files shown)


docs/rules.md
```diff
@@ -139,6 +139,7 @@ Rules are sorted by severity.
 | `RESTRICTED_HOMEPAGE_URL`                               | error    | Linking to addons.mozilla.org in `homepage_url` or `developer.url` is not allowed               |
 | `RESTRICTED_PERMISSION`                                 | error    | A permission requires "strict_min_version" to be set to a specific Firefox version              |
 | `EXTENSION_ID_REQUIRED`                                 | error    | The extension ID is mandatory for Manifest Version 3 (and above) extensions.                    |
+| `HIDDEN_NO_BROWSER_ACTION`                              | error    | The `hidden` and `browser_action` (or `action`) properties are mutually exclusive.              |
```
rpl, line 142: @willdurand I'm wondering if we should also do the same for the page_action, in the end we mainly focused our interest on browserAction because it was directly part of details discussed for the other ongoing work, but with those motivations having a page_action wouldn't be that different in the end.

Another one that may not make sense to allow when the addon is hidden is options_ui (at least when there isn't also a open_in_tab set to true).

We may opt to consider these but defer covering some if we think they will need some more discussion, but I thought to mention it in this PR in case we may want to tweak the rule code so that the other ones would also fit in the same rule code.

e.g. we may opt for something like `HIDDEN_NO_ACTION` or `HIDDEN_UI_ACTION` as a rule id if we only intend to use it for browserAction and pageAction

or maybe something like `NO_HIDDEN_WITH_UI`  may be one that we could also use for the options_ui case (if we agree that would also worth).

wdyt?

rpl (commented):
@willdurand this looks pretty good and straightforward, there is just one detail around the rule code and if we should also check for page_action and options_ui (either in this PR or as a follow up).

Let me know wdyt.

docs/rules.md
willdurand: PR updated to restrict `page_action`. I'll file an issue for `options_ui`.

src/parsers/manifestjson.js
```diff
@@ -730,6 +730,26 @@ export default class ManifestJSONParser extends JSONParser {
 
     this.validateRestrictedPermissions();
     this.validateExtensionID();
+    this.ensureHiddenAddonsWithoutBrowserActions();
+  }
+
+  ensureHiddenAddonsWithoutBrowserActions() {
```
rpl, line 736: Nit (totally optional), maybe we can call this `validateHiddenAddon` or `validateHiddenAddonRestrictions`?

src/parsers/manifestjson.js
```diff
@@ -730,6 +730,26 @@ export default class ManifestJSONParser extends JSONParser {
...
+    if (!this.isPrivilegedAddon) {
+      return;
+    }
+
+    if (
+      this.parsedJSON.hidden &&
+      ('action' in this.parsedJSON ||
+        'browser_action' in this.parsedJSON ||
+        // Note: When this was introduced, it was stricter than the Firefox
+        // side because Firefox didn't restrict `page_action` in Bug 1781998.
```
rpl, line 747: Let's file a followup on bugzilla for aligning these restrictions on the Firefox side.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -102,6 +102,125 @@ describe('ManifestJSONParser', () => {
...
+      title: 'hidden and page_action',
+      manifest: {
+        hidden: true,
+        page_action: {},
+      },
+      privileged: true,
+      expectError: true,
+    },
+    {
+      title: 'hidden, browser_action and page_action',
```
rpl, line 134: Nit, technically in mv3 there will be a similar but slightly different combination, 'hidden, action and page_action' at least at the moment, given that adding one more case is pretty cheap thanks to how you wrote this test cases, I'd say let's just add the additional case for now.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -102,6 +102,125 @@ describe('ManifestJSONParser', () => {
...
+        page_action: {},
+      },
+      privileged: true,
+      expectError: false,
+    },
+  ])(
+    'should report an error when hidden and actions are defined - $title',
+    ({ manifest, privileged, expectError }) => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON(manifest);
```
rpl, line 199: I haven't checked if this is affecting somehow the test coverage as a not fully visible side-effect, it is likely not affecting it because the checks for these restrictions are completely independent from the schema validations, but to prevent our future selves to ask themselves the same question, I'm wondering if we should add into the manifest properties for the entries using `action` instead of `browser_action` an explicit `manifest_version: 3`, wdyt?

rpl (approved):
lgtm, just a couple of nits in the inline review comments below.

src/parsers/manifestjson.js
willdurand: https://bugzilla.mozilla.org/show_bug.cgi?id=1784254

Issue #4507, opened 2022-10-12
Treat version validation failures as hard errors


Rob--W:
Moved from https://bugzilla.mozilla.org/show_bug.cgi?id=1593316

In https://bugzilla.mozilla.org/show_bug.cgi?id=1793925, we introduce validation of the version format to the schema, but only warn. We should enforce the format in the linter, by not warning but rejecting submissions.

At the least for MV3. The format is chosen such that MV2 extensions can migrate to it, but since this could result in the rejection of long-established formats such as `1.0pre1`, it may be reasonable to communicate to extension developers before enforcing it in MV2 too.

Rob--W:
More details in https://bugzilla.mozilla.org/show_bug.cgi?id=1793925#c3

In short, the requirement for the version format is:
- At least 1, at most 4 parts separated by dots (as before).
- 9 digits per part.
- Trim leading zeroes.

ioanarusiczki:
@willdurand I added the next milestone.

ioanarusiczki:
@willdurand @Rob--W 

I tested for MV2 and MV3 on linter 5.20.0. -> I believe it should work as expected.

Some examples of valid versions:
"0.0.0.0"
"0.0.0.9"
"1.0.0.0"
"0.1.2.3"
"123456789.214785236.355554444.444444444"
"999999999.99999999.99999999.99999999" 

Invalid versions, when the error is thrown:
"1.2.3.04"
"0.0.0.00"
"0.0.0.w"
"10.01.0.0"
"1.2.3.4.5"
"1.214.44.webext"
"123456789.999999999.00004"
" 123456789.999999999.4" or "123456789.99999 9999.4" or "123456789.99999 9999.4 " -> spaces left
".123456789.999999999.4" or "123456789.999999999.4." -> dots used at the beginning or at the end
"000000000.000000000.000000000.000000000"
"999999999.99999999.99999999.99999999.9" 

This is the error : 

![the error for invalid version](https://user-images.githubusercontent.com/33448286/199027212-d5bdb5b2-f7e0-4e02-b448-c0b10ee7c905.png)

PR #4527, opened 2022-10-18
Implement new version string check

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4507

---

In addition to adding the new logic, I removed a notice that is no longer useful (since there is now a warning for pretty much the same thing) and I also removed an unused code (about the `version` string).


src/utils.js
```diff
@@ -12,6 +12,43 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
 /* global nodeRequire, localesRoot */
 
 const SOURCE_MAP_RE = /\/\/[#@]\s(source(?:Mapping)?URL)=\s*(\S+)/;
+// Firefox's version format is laxer than Chrome's, it accepts:
```
rpl, line 15: Nit, we may mention this is valid only for MV2 addons (e.g. `Firefox's version format for MV2 extensions is laxer than chrome's...`)

src/utils.js
```diff
@@ -12,6 +12,43 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
...
+  }
+
+  const parts = version.split('.');
+
+  if (parts.length > 4) {
+    return false;
+  }
+
+  // Non-zero values cannot start with 0 and we allow numbers up to 9 digits.
+  return !parts.some((part) => !/^(0|[1-9][0-9]{0,8})$/.test(part));
```
rpl, line 50: Nit, move this regex into a const? (e.g. one named `VALID_SIMPLE_VERSION_PART` or something like that)

src/utils.js
```diff
@@ -12,6 +12,43 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
 /* global nodeRequire, localesRoot */
 
 const SOURCE_MAP_RE = /\/\/[#@]\s(source(?:Mapping)?URL)=\s*(\S+)/;
+// Firefox's version format is laxer than Chrome's, it accepts:
+// https://developer.mozilla.org/en-US/docs/Toolkit_version_format
```
rpl, line 16: This page has been moved and the current url is https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version/format

src/utils.js
```diff
@@ -12,6 +12,43 @@ import { PACKAGE_TYPES, LOCAL_PROTOCOLS } from 'const';
...
+// We choose a slightly restricted version of that format (but still more
+// permissive than Chrome) to allow Beta addons, per:
+// https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance
```
rpl, line 19: This one also doesn't exist anymore, I think we may have moved this content to extensionworkshop, but we would need to look into what the old MDN page used to contain to be sure of which one of the extensionworkshop pages may have replaced it

I'd guess it may be https://extensionworkshop.com/documentation/publish/distribute-pre-release-versions/?

tests/unit/schema/test.version.js
```diff
@@ -1,31 +0,0 @@
...
-  });
-
-  it('should be invalid due to missing version', () => {
-    const manifest = cloneDeep(validManifest);
-    manifest.version = undefined;
-    validateAddon(manifest);
-    expect(validateAddon.errors.length).toEqual(1);
-    expect(validateAddon.errors[0].instancePath).toEqual('');
-    expect(validateAddon.errors[0].params.missingProperty).toEqual('version');
-  });
```
rpl, line 23: was this test for the missing version property redundant?

willdurand: moved to `tests/unit/test.utils.js`

src/utils.js
willdurand: I changed the link to EW, at least the page loads.

src/utils.js
willdurand: updated

src/utils.js
willdurand: done

src/messages/manifestjson.js
```diff
@@ -793,3 +774,12 @@ export const APPLICATIONS_INVALID = {
     "browser_specific_settings" instead.`),
   file: MANIFEST_JSON,
 };
+
+export const SIMPLE_VERSION_STRING = {
+  code: 'SIMPLE_VERSION_STRING',
+  message: i18n._('The version string should be simplified.'),
+  description: i18n._(oneLine`The version should be a string with 1 to 4
+    numbers separated with dots. Each number should have up to 9 digits and
+    leading zeros are not allowed. Letters are no longer allowed.`),
```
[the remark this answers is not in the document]
willdurand: See for MDN: https://github.com/mdn/content/pull/21699

It is already done in EW: https://github.com/mozilla/extension-workshop/issues/1427

docs/rules.md
```diff
@@ -78,70 +78,73 @@ Rules are sorted by severity.
...
+| `EXTENSION_ID_REQUIRED`                                 | error       | The extension ID is mandatory for Manifest Version 3 (and above) extensions                                                                                        |
+| `HIDDEN_NO_ACTION`                                      | error       | The `hidden` and `browser_action`/`page_action` (or `action`) properties are mutually exclusive                                                                    |
+| `APPLICATIONS_DEPRECATED`                               | warning     | The `applications` property in the manifest is deprecated and will no longer be accepted in Manifest Version 3 and above. Use `browser_specific_settings` instead. |
+| `APPLICATIONS_INVALID`                                  | error       | The `applications` property is no longer accepted in Manifest Version 3 and above.                                                                                 |
+| `SIMPLE_VERSION_STRING`                                 | warning[^1] | The version string should be simplified.                                                                                                                           |
```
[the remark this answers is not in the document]
willdurand: As per our slack discussion, we'll indeed introduce two separate codes for mv2/warning and mv3/error, and I'll get rid of the unused codes in the documentation.

PR #4532, opened 2022-10-21
Encourage the use of 'browser_specific_settings' over 'applications' in MV2, forbid the latter in MV3+

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4533

This PR adds two new messages:

1. `APPLICATIONS_DEPRECATED` for MV2 extensions to let developers know that they should be using `browser_specific_settings` instead of `applications`
2. `APPLICTIONS_INVALID` for MV3 extensions, for which we want to prevent the use of `applications` in the manifest

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -600,6 +600,16 @@
 <td>error</td>
 <td>The <code>hidden</code> and <code>browser_action</code>/<code>page_action</code> (or <code>action</code>) properties are mutually exclusive</td>
 </tr>
+<tr>
+<td><code>APPLICATIONS_DEPRECATED</code></td>
+<td>warning</td>
+<td>The <code>applications</code> property in the manifest is deprecated and will no longer be accepted in Manifest Version 3 and above. Use <code>browser_specific_settings</code> instead.</td>
+</tr>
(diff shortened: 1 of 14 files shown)


Rob--W:
Is there a Firefox-side bug to mark applications as deprecated in the schema? If yes, could it be marked as a dependency of https://bugzilla.mozilla.org/show_bug.cgi?id=1739641 ?

willdurand:
> Is there a Firefox-side bug to mark applications as deprecated in the schema? If yes, could it be marked as a dependency of [bugzilla.mozilla.org/show_bug.cgi?id=1739641](https://bugzilla.mozilla.org/show_bug.cgi?id=1739641) ?

Not I was planning on filing the bug next week, thanks for the link to this bug! I'll make sure to link both.

willdurand:
> No, I was planning on filing the bug next week, thanks for the link to this bug! I'll make sure to link both.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1797050

Also, this patch is ready for review.

diox (approved):

PR #4549, opened 2022-10-31
feat: Imported Firefox 107.0b8 API schema

### LINKED ISSUE #4548: Import Firefox 107 schema
This issue is about import updated WebExtensions API JSONSchema files from Firefox 107.

rpl:
- [Bug 1740608 - Implement proper error handling for `scripting.executeScript()`](https://bugzilla.mozilla.org/show_bug.cgi?id=1740608).

The change is restricted to the `InjectResult` type, which is only used for the type of an API method resolved value, and so we don't expect any particular change in behavior on the addons-linter side.

Fixes #4548

### DIFF
--- a/src/schema/imported/scripting.json
+++ b/src/schema/imported/scripting.json
@@ -280,17 +280,7 @@
           "description": "The result of the script execution."
         },
         "error": {
-          "type": "object",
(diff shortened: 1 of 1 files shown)


willdurand (approved):
Thanks!

PR #4573, opened 2022-11-08
fix: CSP should ignore repeated directives

### LINKED ISSUE #4575: Incorrect CSP directive used when repeated

The implementation of the MANIFEST_CSP and MANIFEST_CSP_UNSAFE_EVAL rules rely on a CSP validator.
A CSP source list may potentially contain multiple sources, and in that case the first one should be used, and any duplicates be ignored.

Test case, create `manifest.json` and lint the directory containing the following:
```json
{
  "name": "x",
  "version": "1",
  "manifest_version": 2,
  "content_security_policy": "script-src 'unsafe-eval'; script-src 'self'"
}
```

### What happened?

The addons-linter currently uses the last one instead of the first one. No errors are emitted.

The addons-linter should use the first one for validation, and emit a warning for the presence of 'unsafe-eval'.

I have a patch up at #4573

Rob--W:
Fixes #4575.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -267,6 +267,8 @@ export function parseCspPolicy(policy) {
     return {};
   }
 
+  // See https://www.w3.org/TR/CSP3/#parse-serialized-policy
+
   // eslint-disable-next-line no-param-reassign
   policy = policy.toLowerCase();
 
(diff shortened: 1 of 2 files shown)


Rob--W:
This is the first patch out of 6. The full patchset is at https://github.com/mozilla/addons-linter/compare/master...Rob--W:csp-fix-6?expand=1, but I'll push each one by one untl this review has completed. I'm doing it in this way because rebase-and-merge is not enabled for this repo, and I'd like each commit to land separately (since they address different issues).

For some reason CI fails here. Not sure why.

willdurand (approved):

PR #4580, opened 2022-11-08
chore: simplify CSP validation logic

Rob--W:
The original CSP validation implementation was unnecessarily complex. This part removes redundant logic that's already covered elsewhere.

This is a refactor that does not change the behavior that we care about. This is evidenced by the unit tests.

I added even two more unit tests:
- Verification that the presence of extra space around the directives is still correctly parsed. This works because consecutive spaces are already erased by the [parseCspPolicy helper](https://github.com/mozilla/addons-linter/blob/d2ef1c846d5074a9566594d39fdf42a0b91a2171/src/utils.js#L280-L281).
- Verification that `script-src self` raises a warning. The previous implementation stripped `'`, so it could not distinguish between `'self'` and `self` (or `self'` or `'self`). Now it does.

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -202,10 +202,11 @@ export const TEMPORARY_APIS = [
 // See https://mzl.la/2vwqbGU for more details and allowed options.
 export const CSP_KEYWORD_RE = new RegExp(
   [
+    "^'(", // <-- keywords always start and end with '.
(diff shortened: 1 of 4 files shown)


willdurand (approved):
That seems fine, thanks

PR #4582, opened 2022-11-09
fix: array pattern with omitted element should not cause the file to be interpreted as script

### LINKED ISSUE #4581: Array pattern with omitted element would cause the file to be interpreted as script

```js
const [, x] = y
export {}
```

### What happened?

```
Code              Message                        Description                                              File                        Line   Column
JS_SYNTAX_ERROR   JavaScript syntax error        There is a JavaScript syntax error in your code, which   ....../................js   ..     ....  
                  (Parsing as module error:      might be related to some experimental JavaScript                                                  
                  Cannot read properties of      features that aren't an official part of the language                                             
                  null (reading 'type') at       specification and therefore not supported yet. The                                                
                  line: (unknown) and column:    validation cannot continue on this file.                                                          
                  (unknown). This looks like a                                                                                                     
                  bug in addons-linter, please                                                                                                     
                  open a new issue:                                                                                                                
                  https://github.com/mozilla/…                                                                                                     
                  addons-linter/issues)                                                                                                            
                  (Parsing as script error:                                                                                                        
                  'import' and 'export' may                                                                                                        
                  appear only with                                                                                                                 
                  'sourceType: module' at                                                                                                          
                  line: .. and column: ....)        
```

No syntax error.

I'll submit a PR to address this issue.

ikatyang:
`ArrayPattern.elements[i]` may be `null` if the element is omitted, which would cause js error and fallback to script mode regardless of the actual source type

Fixes #4581

### DIFF
--- a/src/scanners/javascript.js
+++ b/src/scanners/javascript.js
@@ -231,7 +231,7 @@ export default class JavaScriptScanner {
 
         if (Array.isArray(child)) {
           for (let j = 0; j < child.length; ++j) {
-            if (this._getSourceType(child[j]) === 'module') {
+            if (child[j] && this._getSourceType(child[j]) === 'module') {
               return 'module';
             }
           }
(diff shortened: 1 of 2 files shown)


willdurand (approved):

willdurand:
Thanks!

Issue #3973, opened 2021-10-18
Raise error if `install_origins` if absent from manifest for `self-hosted` and MV3


diox:
MV3 self-hosted add-ons will need to specify in their manifest what origins to allow installs from.

Therefore, in the linter, when `self-hosted` parameter is `true` and `max-manifest-version` is `3` or higher, if `install_origins` is missing from the manifest, an error should be issued. 

Note that an empty `install_origins` value (`[]`) should be allowed for local installation of those add-ons without a web server - so we should only return an error if the key is absent entirely, and only for self-hosted if MV3 are explicitly allowed.

willdurand:
@diox is this issue still relevant?

diox:
No, let's close it and re-assess later what will need to happen in the linter for `install_origins`

PR #4587, opened 2022-11-14
Raise error when `install_origins` is absent from manifest for self-hosted and MV3

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/3973

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -141,6 +141,7 @@ Rules are sorted by severity.
 | `APPLICATIONS_INVALID`                                  | error    | The `applications` property is no longer accepted in Manifest Version 3 and above.                                                                                 |
 | `VERSION_FORMAT_DEPRECATED`                             | warning  | The version string should be simplified.                                                                                                                           |
 | `VERSION_FORMAT_INVALID`                                | error    | The version string is not valid because its format is too complex.                                                                                                 |
+| `INSTALL_ORIGINS_REQUIRED`                              | error    | The `install_origins` property is required in the manifest for self-hosted versions in Manifest Version 3 and above.                                               |
 
 ### Static Theme / manifest.json
 
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -795,3 +795,14 @@ export const VERSION_FORMAT_INVALID = {
     https://mzl.la/3h3mCRu (MDN Docs) for more information.`),
   file: MANIFEST_JSON,
 };
+
+export const INSTALL_ORIGINS_REQUIRED = {
+  code: 'INSTALL_ORIGINS_REQUIRED',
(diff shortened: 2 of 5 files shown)


docs/rules.md
```diff
@@ -141,6 +141,7 @@ Rules are sorted by severity.
 | `APPLICATIONS_INVALID`                                  | error    | The `applications` property is no longer accepted in Manifest Version 3 and above.                                                                                 |
 | `VERSION_FORMAT_DEPRECATED`                             | warning  | The version string should be simplified.                                                                                                                           |
 | `VERSION_FORMAT_INVALID`                                | error    | The version string is not valid because its format is too complex.                                                                                                 |
+| `INSTALL_ORIGINS_REQUIRED`                              | error    | The `install_origins` property is required in the manifest for self-hosted versions (in Manifest Version 3 and above).                                             |
```
diox, line 144: nit: we're not consistent between this description and the one raised by the rule (and, if we want to be pedantic, we have at least 3 different ways to write `for/in Manifest version 3 and above`)

src/messages/manifestjson.js
```diff
@@ -795,3 +795,13 @@ export const VERSION_FORMAT_INVALID = {
...
   file: MANIFEST_JSON,
 };
+
+export const INSTALL_ORIGINS_REQUIRED = {
+  code: 'INSTALL_ORIGINS_REQUIRED',
+  message: i18n._(
+    'The "install_origins" property is required in the manifest.'
+  ),
+  description: i18n._(oneLine`The "install_origins" property is required in the
+    manifest for self-hosted versions in Manifest Version 3 and above.`),
```
diox, line 805: Should we wait until we have MDN docs live to include the relevant URL, so that we don't have to change the string and redo translations ?

src/parsers/manifestjson.js
```diff
@@ -719,6 +719,15 @@ export default class ManifestJSONParser extends JSONParser {
       this.validateHomePageURL(this.parsedJSON.homepage_url);
     }
 
+    if (
+      this.selfHosted &&
+      this.parsedJSON.manifest_version > 2 &&
+      !this.parsedJSON.install_origins
```
diox, line 725: Where did we land on authorizing an empty array for self-distributed without a host at all ? IIRC Firefox was supposed to let you do `install_origins: []` and we'd let that through, maybe with a warning, for people who want to just drag & drop the xpi locally and such.

src/messages/manifestjson.js
willdurand: Sure

src/parsers/manifestjson.js
diox: Eh, past me is smarter than current me

PR #4594, opened 2022-11-15
fix: Correctly detect insecure CSP when default-src is missing.

### LINKED ISSUE #4578: CSP without default-src is incorrectly detected as secure

The CSP validator's primary purpose is to reject CSP policies that permit remote code.
`script-src` is the primary directive to control that, and if that is not specified, `default-src`.
The validator correctly handles these two cases. But if `default-src` is missing, the default fallback is to be very permissive. This should be flagged as insecure CSP.

Test case, create `manifest.json` and lint the directory containing the following:
```json
{
  "name": "x",
  "version": "1",
  "manifest_version": 2,
  "content_security_policy": "img-src 'none'"
}
```

### What happened?
MANIFEST_CSP warning should be emitted, because this policy allows remote code.

No warnings.

Note: Firefox has a different CSP parser. Its current behavior given the above CSP is to ignore it and use the default CSP that rejects remote scripts. While that matches our intent, the rejection of the CSP means that the extension developer's intent to reject images (`img-src 'none'`) is not respected.

This issue happens because [the implementation](https://github.com/mozilla/addons-linter/blob/a2fbda70d90106900d05be18410285afe22d13b9/src/parsers/manifestjson.js#L1045
) assumes that the CSP is secure by default, and rejects when an insecure directive is encountered. An empty CSP is not secure by default, which results in this bug.

Rob--W:
Fixes #4578.

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -1622,22 +1622,11 @@ describe('ManifestJSONParser', () => {
   });
 
   describe('content security policy', () => {
-    it('should warn on rules allowing remote code execution', () => {
-      const addonLinter = new Linter({ _: ['bar'] });
-      const json = validManifestJSON({
-        content_security_policy: 'foo',
-      });
-      const manifestJSONParser = new ManifestJSONParser(
-        json,
-        addonLinter.collector
-      );
-
-      expect(manifestJSONParser.isValid).toEqual(true);
-      expect(addonLinter.collector.warnings.length).toEqual(0);
-    });
-
     describe('should warn on invalid values according to Add-On Policies', () => {
       const invalidValues = [
+        '', // Empty CSP does not clear the CSP, so it should warn.
+        'foo', // invalid directive, equivalent to empty CSP
+
         'default-src *',
         'default-src moz-extension: *', // mixed with * invalid
         'default-src ws:',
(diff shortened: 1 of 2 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -1922,12 +1919,13 @@ describe('ManifestJSONParser', () => {
 
         const keys = Object.keys(contentSecurityPolicy);
         for (let i = 0; i < keys.length; i++) {
-          expect(warningsV3[i].code).toEqual(messages.MANIFEST_CSP_UNSAFE_EVAL);
-          expect(warningsV3[i].message).toContain(
+          expect(warningsV3[i * 2].code).toEqual(messages.MANIFEST_CSP_UNSAFE_EVAL);
```
willdurand, line 1922: maybe mention that we expect two codes for each key, and so first warning is `MANIFEST_CSP_UNSAFE_EVAL` and second warning is `MANIFEST_CSP`.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -1888,6 +1883,8 @@ describe('ManifestJSONParser', () => {
         expect(warnings[0].message).toEqual(
           messages.manifestCspUnsafeEval('content_security_policy').message
         );
+        expect(warnings[1].code).toEqual(messages.MANIFEST_CSP);
+        expect(warnings.length).toBe(2);
```
willdurand, line 1887:
```suggestion
        expect(warnings.length).toEqual(2);
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -1922,12 +1919,13 @@ describe('ManifestJSONParser', () => {
...
-          expect(warningsV3[i].code).toEqual(messages.MANIFEST_CSP_UNSAFE_EVAL);
-          expect(warningsV3[i].message).toContain(
+          expect(warningsV3[i * 2].code).toEqual(messages.MANIFEST_CSP_UNSAFE_EVAL);
+          expect(warningsV3[i * 2].message).toContain(
             `content_security_policy.${keys[i]}`
           );
+          expect(warningsV3[i * 2 + 1].code).toEqual(messages.MANIFEST_CSP);
         }
-        expect(warningsV3.length).toBe(3);
+        expect(warningsV3.length).toBe(6);
```
willdurand, line 1928:
```suggestion
        expect(warningsV3.length).toEqual(6);
```

willdurand (approved):
r+wc

PR #4602, opened 2022-11-22
feat: Import firefox 108.0b5 schema data

### LINKED ISSUE #4603: Import Firefox 108 schema
This issue is about import updated WebExtensions API JSONSchema files from Firefox 108.

### LINKED ISSUE #4604: manifest_version 3 extensions manifest.json with event pages should not be reported as invalid
On linting a manifest.json using a background event page:

```
{
	"manifest_version": 3,
	"name": "MV3 extension name",
	"version": "1.0.0",
	"background": {
		"page": "index.html"
	},
		"gecko": {
			"id": "an-mv3-extension@test"
		}
	}
}
```

The addons-linter is currently reporting the manifest as invalid and mentioning that the background.service_worker property is required:
```
../addons-linter/bin/addons-linter --max-manifest-version=3 .
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code                      Message                                     Description                                                   File            Line   Column
MANIFEST_FIELD_REQUIRED   "/background" must have required property   See https://mzl.la/1ZOhoEN (MDN Docs) for more information.   manifest.json
                          'service_worker'
```

But:
- Firefox does support event pages in manifest_version 3 extensions, and so the manifest should be valid
- Firefox does not yet support `background.service_worker` and it is only enabled when its separate about:config pref is set to true (tracked by #4605)

This issue is tracking fixing the addons-linter JSONSchema to make sure event pages used in `manifest_version: 3` extension manifests isn't reported as a linting error.

### LINKED ISSUE #4605: Gate background.service_worker support behind an expicit addons-linter feature flag
background.service_worker field was currently disallowed in extensions submitted to AMO due to the `min_manifest_version: 3` part of its JSONSchema definition.

Once we enable `manifest_version: 3` support by default on AMO, `background.service_worker` would be currently allowed in MV3 extensions submitted to AMO, but Firefox support for `background.service_worker` is only enabled when its separate about:config pref is set to true and so AMO should not yet accept extensions using `background.service_worker`.

This issue is tracking the addition of a new separate feature flag to the addons-linter and the new feature flag to keep `background-service_worker` support disabled by default.

rpl:
- lifted `max_manifest_version: 2` restriction applied on background.page and background.scripts JSONSchema definitions (applied as part of the addons-linter overrides on every new Firefox JOSNSchema data imports)
- [Bug 1755763](https://bugzilla.mozilla.org/show_bug.cgi?id=1755763) - additions to action and browserAction.openPopup parameters (not used by the addons-linter in practice)
- [Bug 1745758](https://bugzilla.mozilla.org/show_bug.cgi?id=1745758) - updates on the declarative_net_request schema data (these changes are also currently unused by the addons-linter in practice)

In addition to the JSONSchema changes, this PR also adds a new addons-linter feature flag `--mv3-background-service-worker` to explicitly gate support for the `background.service_worker` manifest field (given that it was currently gated by just the `min_manifest_version: 3` keyword in the JSONSchema data but its support is locked behind an explicit about:config pref in Firefox and currently disabled by default on all channels).

Fixes #4603
Fixes #4604
Fixes #4605

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -314,6 +314,8 @@ export default class Linter {
           privileged: this.config.privileged,
           minManifestVersion: this.config.minManifestVersion,
           maxManifestVersion: this.config.maxManifestVersion,
+          enableBackgroundServiceWorker:
+            this.config.enableBackgroundServiceWorker,
         },
       });
       await manifestParser.validateIcons();
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -513,7 +513,15 @@ export default class ManifestJSONParser extends JSONParser {
         );
       }
       if (this.parsedJSON.background.service_worker) {
-        if (this.parsedJSON.manifest_version >= 3) {
+        if (!this.schemaValidatorOptions?.enableBackgroundServiceWorker) {
+          // Report an error and mark the manifest as invalid if
+          // background service worker support isn't enabled by
+          // the addons-linter feature flag.
+          this.collector.addError(
+            messages.manifestFieldUnsupported('/background')
+          );
+          this.isValid = false;
+        } else if (this.parsedJSON.manifest_version >= 3) {
           this.validateFileExistsInPackage(
             this.parsedJSON.background.service_worker,
             'script'
--- a/src/yargs-options.js
+++ b/src/yargs-options.js
@@ -47,6 +47,11 @@ const options = {
     type: 'boolean',
     default: false,
   },
+  'enable-background-service-worker': {
+    describe: 'Enable MV3 background service worker support',
+    type: 'boolean',
+    default: false,
+  },
   'min-manifest-version': {
     describe:
       'Set a custom minimum allowed value for the manifest_version property',
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -2858,7 +2858,7 @@ describe('ManifestJSONParser', () => {
       );
     });
 
-    it('does not error if background.service_worker is used with manifest_version: 3', () => {
+    it('does not allow background.service_worker unless enabled by feature flag', () => {
       const linter = new Linter({ _: ['bar'] });
       const json = validManifestJSON({
         manifest_version: 3,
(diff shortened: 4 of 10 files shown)


src/linter.js
```diff
@@ -314,6 +314,7 @@ export default class Linter {
           privileged: this.config.privileged,
           minManifestVersion: this.config.minManifestVersion,
           maxManifestVersion: this.config.maxManifestVersion,
+          mv3BackgroundServiceWorker: this.config.mv3BackgroundServiceWorker,
```
willdurand, line 317: I would drop the `mv3` prefix, service worker isn't a thing in mv2 anyway.

src/parsers/manifestjson.js
```diff
@@ -513,7 +513,18 @@ export default class ManifestJSONParser extends JSONParser {
...
       if (this.parsedJSON.background.service_worker) {
-        if (this.parsedJSON.manifest_version >= 3) {
+        if (!this.schemaValidatorOptions?.mv3BackgroundServiceWorker) {
+          // Report an error and mark the manifest as invalid if
+          // background service worker support isn't enabled by
+          // the addons-linter feature flags.
+          this.collector.addError(
+            messages.manifestFieldUnsupported(
+              '/background',
+              {} // mock schema error object.
```
willdurand, line 523: can you describe why we are doing this?

src/yargs-options.js
```diff
@@ -47,6 +47,11 @@ const options = {
     type: 'boolean',
     default: false,
   },
+  'mv3-background-service-worker': {
```
willdurand, line 50:
```suggestion
  'enable-background-service-worker': {
```

or something like that maybe?

src/yargs-options.js
```diff
@@ -47,6 +47,11 @@ const options = {
...
+    describe: 'Enable/Disable MV3 background service worker support',
```
willdurand, line 51:
```suggestion
    describe: 'Enable background service worker support',
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2858,6 +2858,37 @@ describe('ManifestJSONParser', () => {
...
+            message: expect.stringMatching(
+              /"\/background" is in an unsupported format/
+            ),
+            file: 'manifest.json',
+          }),
+        ])
+      );
+    });
+
     it('does not error if background.service_worker is used with manifest_version: 3', () => {
```
willdurand, line 2892:
```suggestion
    it('does not error if background.service_worker is used with manifest_version: 3 and support is enabled', () => {
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2869,7 +2900,50 @@ describe('ManifestJSONParser', () => {
...
+      );
+
+      expect(manifestJSONParser.isValid).toBeTruthy();
+    });
+
+    it('does not error if background.page is used with manifest_version: 3', () => {
+      const linter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({
+        manifest_version: 3,
+        background: { page: 'background_page.html' },
```
willdurand, line 2917: What if we add extra properties?

willdurand (approved):
r+wc, that looks good overall, though I have a number of comments.

src/parsers/manifestjson.js
rpl: it didn't seem worth adding a new specific linting message and so I used the pre-existing `manifestFieldUnsupported` one, which usually expects the ajv error to be passed as the second parameter, alternatively I could just tweak `manifestFieldUnsupported` implementation to account for the error object to not be always passed.

tests/unit/parsers/test.manifestjson.js
rpl: from what I recall extra property wouldn't be allowed by the manifest schema, both unknown ones and persistent which was turned off by default, if you think it would be worth it I can also add test coverage for that if some isn't already provided by some other tests (I haven't checked to be honest).

src/parsers/manifestjson.js
willdurand: > to account for the error object to not be always passed.

ha, if we can turn that parameter into an optional one, that'd be nice yeah.

diox (commented):
Looks good to me, tested that it works.

r+ w/ @willdurand's comments

willdurand (approved):

willdurand:
ha well, nevermind. We'll fix that later, maybe.

PR #4667, opened 2023-01-09
chore: Added unit tests for the dnr manifest fields validation

rpl:
This PR introduces a couple more schema validation tests related to the declarative_net_request manifest fields, as a follow up to #4655


Rob--W (approved):
LGTM. How about a test for a non-existing prop, that it won't cause hard validation failures?

rpl:
> LGTM. How about a test for a non-existing prop, that it won't cause hard validation failures?

sure thing, added in [5856e55](https://github.com/mozilla/addons-linter/pull/4667/commits/5856e5509c71b6b4cde0e26f795d64c123b9ee42)

PR #4678, opened 2023-01-16
Use vendored `ajv-merge-patch` library

willdurand:
This fixes a security issue caused by an indirect dependency.


willdurand:
This isn't exactly what I had in mind when we discussed yesterday but.. this is a working solution.

The problem with vendoring the whole library is that we'd need to compile the linter's dependencies with webpack. Currently, we are only compiling/bundling the source files. Because of that, the vendored library isn't available in the package and the install fails (this is when we specify the vendored library into the top-level `package.json`). We could have copied the `vendor/*` files to the `dist/` file in order to ship the vendor libraries in the npm package but that felt wrong.

So I use the "golang way", which consists in keeping the actual source files only. This is similar to having the implementation in our `src/` directory except that we retain the README and LICENSE in a separate folder (in a `vendor` directory). That makes it obvious that we are using a third-party library and that this isn't "our" code.

I removed the test files. This was a bit annoying to set up because of.. babel? Not sure why to be honest. If you want me to re-add the test files, I'll need to debug that but that didn't work out of the box :/ I trust our own test suite to cover this library a bit..

PR #4739, opened 2023-03-02
fix: Set a custom no-unsanitized config for dynamic import calls

### LINKED ISSUE #4686: await import chrome.runtime.getURL causes UNSAFE_VAR_ASSIGNMENT

I'm trying to import `script.js` from a content script (`test.js`). `script.js` is declared as `web_accessible_resources` in my manifest, so this works perfectly in runtime.

```
// test.js
async function test () {
    await import (chrome.runtime.getURL('script.js'))
}
```

### What happened?

During running `web-ext lint`, I'm getting:

```
UNSAFE_VAR_ASSIGNMENT      Unsafe call to import for    Due to both security and performance concerns, this    test.js         2      11    
                           argument 0                   may not be set using dynamic values which have not                                  
                                                        been adequately sanitized. This can lead to security                                
                                                        issues or fairly serious performance degradation.      
```

In this case the script is provided by the extension itself (`chrome.runtime.getURL('script.js')`), so this feels like a false positive?

In case I'm wrong about it being a false positive:

- what's wrong with it? Perhaps the error message could be improved in this case
- what's the right way to achieve the same effect? I could use `browser.tabs.executeScript({file: 'script.js'})` before injecting `test.js`, but that way the logic is kinda scattered away from the call site. 

Perhaps the closest relevant issue is https://github.com/mozilla/addons-linter/issues/2498, but it's different -- in that case it was a syntax error.

rpl:
This PR includes a new custom config for the no-unsanitized eslint plugin to customize the behavior on validating dynamic import calls.

Fixes #4686

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -9,12 +9,29 @@ const NO_UNSANITIZED_OPTIONS = {
   escape: { taggedTemplates: [], methods: [] },
 };
 
+const NO_UNSANITIZED_METHOD_CUSTOMIZATIONS = {
+  import: {
+    escape: {
+      methods: ['chrome.runtime.getURL', 'browser.runtime.getURL'],
+    },
+    // NOTE: Alternatively using the following option would instead
+    // configure the plugin to consider any method call as allowed
+    // on dynamic import calls:
+    //
+    // objectMatches: [],
+  },
+};
+
 // 3rd party / eslint-internal rules
 export const EXTERNAL_RULE_MAPPING = {
   'no-eval': [ESLINT_WARNING, { allowIndirect: false }],
   'no-implied-eval': ESLINT_WARNING,
   'no-new-func': ESLINT_WARNING,
-  'no-unsanitized/method': [ESLINT_WARNING, NO_UNSANITIZED_OPTIONS],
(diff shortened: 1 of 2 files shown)


src/const.js
```diff
@@ -9,12 +9,29 @@ const NO_UNSANITIZED_OPTIONS = {
...
+  import: {
+    escape: {
+      methods: ['chrome.runtime.getURL', 'browser.runtime.getURL'],
+    },
+    // NOTE: Alternatively using the following option would instead
+    // configure the plugin to consider any method call as allowed
+    // on dynamic import calls:
+    //
+    // objectMatches: [],
+  },
```
rpl, line 22: @willdurand in this inline comment I'm highlighting an alternative configuration that would allow any method call to silence the no-unsanitized plugin on dynamic import calls, in the current version I have currently only added chrome.runtime.getURL and browser.runtime.getURL as valid escapers, but based on the comments on the issue side it seems we may want to also consider the alternative described in this comment (and tweak the test case accordingly).

willdurand: let's be conservative for now

willdurand (approved):

PR #4764, opened 2023-03-18
Bump the value of 'max-manifest-version' to 3

### LINKED ISSUE #4762: `JSON_INVALID` error is logged if you set `manifest_version` to `3`

`JSON_INVALID` error is logged if you set `manifest_version` to `3` in `manifest.json`.

addons-linter version: 5.31.0

If `manifest_version` is `2`, no error logged.

manifest.json:
```json
{
  "name": "MV3 test",
  "manifest_version": 2,
  "background": {
    "page": "/html/background.html"
  },
    "gecko": {
      "id": "test@asamuzak.jp",
      "strict_min_version": "110.0"
    }
  },
  "version": "0.0.1"
}
```

result:
```console
Validation Summary:

errors          0
notices         0
warnings        0
```

But if `manifest_version` is set to `3`, error logged.
Note that the only difference is the value of the `manifest_version`.

manifest.json:
```json
{
  "name": "MV3 test",
  "manifest_version": 3,
  "background": {
    "page": "/html/background.html"
  },
    "gecko": {
      "id": "test@asamuzak.jp",
      "strict_min_version": "110.0"
    }
  },
  "version": "0.0.1"
}
```

result:
```console
Validation Summary:

errors          1
notices         0
warnings        0

ERRORS:

Code           Message               Description                           File   Line   Column
JSON_INVALID   "/manifest_version"   Your JSON file could not be parsed.
               must be <= 2
```

### What happened?

`JSON_INVALID` error is logged.

No error logged.

asamuzaK:
Fixes #4762

* Change the default value of 'max-manifest-version' from 2 to 3.
* Fix error code and message.

- [x] Add tests to cover the changes added in this PR.


tests/test.yargs-options.js
```diff
@@ -4,6 +4,7 @@ describe('getDefaultConfigValue()', () => {
   it('should return the default value', () => {
     expect(getDefaultConfigValue('self-hosted')).toEqual(false);
     expect(getDefaultConfigValue('privileged')).toEqual(false);
+    expect(getDefaultConfigValue('max-manifest-version')).toEqual(3);
```
asamuzaK, line 7: There were only 'self-hosted' and 'privileged' tests.
Should we add more tests like 'min_manivest_version' and so on?

src/parsers/manifestjson.js
```diff
@@ -284,6 +284,13 @@ export default class ManifestJSONParser extends JSONParser {
       // properties are handled properly (e.g. we should also detect when the deprecated
       // keyword is actually used to warn the developer of additional properties not
       // explicitly defined in the schemas).
+    } else if (
+      error.instancePath === '/manifest_version' &&
+      error.keyword === 'maximum'
+    ) {
+      baseObject = messages.MANIFEST_VERSION_INVALID;
+      overrides.message = baseObject.message;
```
asamuzaK, line 292: Here we replace overrides.message with baseObject.message to match the rest of this function.
However, the original overrides.message is `/manifest_json must be <=3`, which gives a clearer indication of what caused the error than the baseObject.message `"manifest_version" in the manifest.json is not a valid value`.
I would suggest to keep the original overrides.message instead of replacing it with baseObject.message.
Thoughts?

willdurand (changes requested):
Could you please split your patch into two distinct PRs? I am fine with bumping the `max-manifest-version` value to `3` but the other changes about the error message are unrelated.

asamuzaK:
It is related.
#4762 contains two bugs:
* Does not support other than manifest version 2
* Reporting a JSON_INVALID error when an unsupported manifest version is specified, even though the JSON is **valid**.

asamuzaK:
For the time being, I removed the fixes for error code and message as requested.

src/schema/updates/manifest.json
```diff
@@ -106,7 +106,7 @@
             "format": "versionString"
           },
           "manifest_version": {
-            "maximum": 2
+            "maximum": 3
```
willdurand, line 109: Could you fully remove the `manifest_version` object instead?

willdurand (changes requested):

willdurand (approved):
Thanks!

src/schema/updates/manifest.json
asamuzaK: Done

PR #4776, opened 2023-03-27
Fix manifest schema

willdurand:
See: https://github.com/mozilla/addons-linter/issues/4762#issuecomment-1484824364


willdurand:
This is a trivial change so I am going to r+ myself.. (I should have caught that in #4764)

PR #4795, opened 2023-04-06
fix: Accept top level await

### LINKED ISSUE #4020: Syntax error with a top-level await and no import

regseb:
Fixes #4020

In the detection of the type of a JavaScript file, if the file contains an `await` in the top level: consider that it's a module.


src/scanners/javascript.js
```diff
@@ -218,10 +218,20 @@ export default class JavaScriptScanner {
...
+      'FunctionDeclaration',
+      'FunctionExpression',
+    ];
 
     if (possibleImportExportTypes.includes(node.type)) {
       return 'module';
     }
+    if (topLevel && node.type === 'AwaitExpression') {
+      return 'module';
+    }
```
willdurand, line 232: nit: could we merge both `if` blocks?

willdurand (changes requested):
This looks _fine_. I have a nit then we can probably merge this PR. Thanks!

willdurand (approved):

Issue #4780, opened 2023-03-28
Add preliminary support for the `bss.gecko_android` manifest key


willdurand:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1824237

---

This is about the schema import/update, mainly.

ioanarusiczki:
@willdurand 

On linter 6.2.0 I tested with a manifest.json that would have 	`"browser_specific_settings": {
    "gecko_android": {
      "strict_min_version": "109.0",
      "strict_max_version": "112.0"
    }
  }`
With MV2., validation passes for gecko_android with strict_min or/and strict_max defined.  

It also passed with nothing defined, I tried  `"browser_specific_settings": {"gecko_android": { } }` (can't tell if that's expected but it's the same on linter 6.1.0)

It throws and error when the min is "strict_min_version": "109.*" (this I know it's expected)

On MV3 I get an error ` EXTENSION_ID_REQUIRED   The extension ID is required in Manifest   See https://mzl.la/3PLZYdo for more information.   manifest.json
                        Version 3 and above.` 

The [docs ](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings) mention that for gecko_android the `id` is not a supported key -> mv3 is not supported yet on Android

willdurand:
I am guessing this is because `gecko.id` must exist?

ioanarusiczki:
@willdurand With or without an `id` specified in the manifest.json on MV3 I get the same error.

willdurand:
can you paste the entire `manifest.json`?

ioanarusiczki:
@willdurand 

`{
    "manifest_version": 3,
    "version": "3.0",
    "description": "A test extension from webext-generator.",
    "name": "gecko android key",
	"browser_specific_settings": {
    "gecko_android": {
      "id": "testdummy6@test",
      "strict_min_version": "109.0"
    }
  }
}`

and 
`{
    "manifest_version": 3,
    "version": "3.0",
    "description": "A test extension from webext-generator.",
    "name": "gecko android key",
	"browser_specific_settings": {
    "gecko_android": {
      "strict_min_version": "109.0",
      "strict_max_version": "112.0"
    }
  }
}`

willdurand:
> ```
> {
>     "manifest_version": 3,
>     "version": "3.0",
>     "description": "A test extension from webext-generator.",
>     "name": "gecko android key",
[...]

PR #4803, opened 2023-04-13
Add preliminary support for the bss.gecko_android manifest key

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4780

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -468,9 +468,16 @@ export default class ManifestJSONParser extends JSONParser {
       });
     }
 
+    if (this.parsedJSON.applications?.gecko_android) {
+      this.collector.addError(
+        messages.manifestFieldUnsupported('/applications/gecko_android')
+      );
+      this.isValid = false;
+    }
+
     if (this.parsedJSON.manifest_version < 3) {
       if (
-        this.parsedJSON.browser_specific_settings &&
+        this.parsedJSON.browser_specific_settings?.gecko &&
         this.parsedJSON.applications
       ) {
         this.collector.addWarning(messages.IGNORED_APPLICATIONS_PROPERTY);
@@ -514,11 +521,11 @@ export default class ManifestJSONParser extends JSONParser {
       }
       if (this.parsedJSON.background.service_worker) {
         if (!this.schemaValidatorOptions?.enableBackgroundServiceWorker) {
-          // Report an error and mark the manifest as invalid if
-          // background service worker support isn't enabled by
-          // the addons-linter feature flag.
+          // Report an error and mark the manifest as invalid if background
+          // service worker support isn't enabled by the addons-linter feature
+          // flag.
           this.collector.addError(
-            messages.manifestFieldUnsupported('/background')
+            messages.manifestFieldUnsupported('/background/service_worker')
           );
           this.isValid = false;
         } else if (this.parsedJSON.manifest_version >= 3) {
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -901,10 +901,14 @@
       "type": "object",
       "properties": {
         "strict_min_version": {
-          "type": "string"
+          "type": "string",
+          "description": "Minimum version of Gecko to support.",
+          "pattern": "^[0-9]{1,3}(\\.[a-z0-9]+)+$"
         },
         "strict_max_version": {
-          "type": "string"
+          "type": "string",
+          "description": "Maximum version of Gecko to support.",
+          "pattern": "^[0-9]{1,3}(\\.[a-z0-9*]+)+$"
         }
       },
       "additionalProperties": {}
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -17,7 +17,7 @@
         "applications": {
           "allOf": [
             {
-              "$ref": "#/types/BrowserSpecificSettings"
+              "$ref": "#/types/DeprecatedApplications"
             },
             {
               "description": "The applications property is deprecated, please use 'browser_specific_settings'",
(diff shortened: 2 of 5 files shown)


src/parsers/manifestjson.js
```diff
@@ -514,11 +521,11 @@ export default class ManifestJSONParser extends JSONParser {
...
         if (!this.schemaValidatorOptions?.enableBackgroundServiceWorker) {
-          // Report an error and mark the manifest as invalid if
-          // background service worker support isn't enabled by
-          // the addons-linter feature flag.
+          // Report an error and mark the manifest as invalid if background
+          // service worker support isn't enabled by the addons-linter feature
+          // flag.
           this.collector.addError(
-            messages.manifestFieldUnsupported('/background')
+            messages.manifestFieldUnsupported('/background/service_worker')
```
willdurand, line 528: Minor tweak to reuse `manifestFieldUnsupported`

src/parsers/manifestjson.js
```diff
@@ -468,6 +468,13 @@ export default class ManifestJSONParser extends JSONParser {
...
+    if (this.parsedJSON.applications?.gecko_android) {
+      this.collector.addError(
+        messages.manifestFieldUnsupported('/applications/gecko_android')
+      );
+      this.isValid = false;
+    }
+
     if (this.parsedJSON.manifest_version < 3) {
       if (
         this.parsedJSON.browser_specific_settings &&
```
rpl, line 480: @willdurand I'm wondering if we should tweak also this check to make sure we will report the `IGNORE_APPLICATION_PROPERTY` message only when both `browser_specific_settings` and `applications` have a `gecko` property and `APPLICATIONS_DEPRECATED` otherwise, so that on a manifest.json like the following we would be using the latter instead of the former:

```
  "applications": {
    "gecko": { "id": "test@extension" }
  },
    "gecko_android": { ... }
  }
```

wdyt?

if we agree that it would be reasonable, then we would likely be tweaking this to something like:
```
     if (
        this.parsedJSON.browser_specific_settings?.gecko &&
        this.parsedJSON.applications?.gecko
      ) {
       ...  
```

and we would need to also tweak the test case `it('only emits a single warning when a MV2 extension uses both applications and browser_specific_settings', ...)` from `tests/unit/parsers/test.manifestjson.js`.

src/schema/imported/manifest.json
```diff
@@ -897,11 +897,45 @@
...
+      "properties": {
+        "strict_min_version": {
+          "type": "string"
+        },
+        "strict_max_version": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": {}
+    },
```
rpl, line 911: `GeckoAndroidSpecificProperties`'s `strict_min_version`/`strict_max_version` should get a format property merged into it as for the `FirefoxSpecificProperties` type's ones.

And so we should update `src/schema/updates/manifest.json` with something like:

```diff
diff --git a/src/schema/updates/manifest.json b/src/schema/updates/manifest.json
index 014b429e..fc8cdffd 100644
--- a/src/schema/updates/manifest.json
+++ b/src/schema/updates/manifest.json
@@ -18,6 +18,18 @@
           }
         }
       },
+      "GeckoAndroidSpecificProperties": {
+        "strict_min_version": {
+          "default": "42a1",
+          "description": "Minimum version of Gecko to support. Defaults to '42a1'. (Requires Gecko 45)",
+          "pattern": "^[0-9]{1,3}(\\.[a-z0-9]+)+$"
+        },
+        "strict_max_version": {
+          "default": "*",
+          "description": "Maximum version of Gecko to support. Defaults to '*'. (Requires Gecko 45)",
+          "pattern": "^[0-9]{1,3}(\\.[a-z0-9*]+)+$"
+        }
+      },
       "ExtensionID": {
         "maxLength": 80
       },
```

And the resulting part of this diff to look like the following as a side-effect of that addons-linter schema updates:

```diff
diff --git a/src/schema/imported/manifest.json b/src/schema/imported/manifest.json
index 096b39ae..3ca33352 100644
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -907,7 +907,17 @@
           "type": "string"
         }
       },
-      "additionalProperties": {}
+      "additionalProperties": {},
+      "strict_min_version": {
+        "default": "42a1",
+        "description": "Minimum version of Gecko to support. Defaults to '42a1'. (Requires Gecko 45)",
+        "pattern": "^[0-9]{1,3}(\\.[a-z0-9]+)+$"
+      },
+      "strict_max_version": {
+        "default": "*",
+        "description": "Maximum version of Gecko to support. Defaults to '*'. (Requires Gecko 45)",
+        "pattern": "^[0-9]{1,3}(\\.[a-z0-9*]+)+$"
+      }
     },
     "DeprecatedApplications": {
       "type": "object",
```

Let's also add test coverage for the gecko_android version strings formats.

rpl (changes requested):
@willdurand thanks, the tweaks applied on `messages.manifestFieldUnsupported` looks good to me, there seems to be some more tweaks needed to make sure we are going to apply on the `browser_specific_settings.gecko_android`'s `strict_min_version`/`strict_max_version` properties the same version format checks we apply on the ones part of the `browser_specific_settings.gecko` (and I'm wondering if we should tweak check we are using to determine if we need to report `IGNORE_APPLICATION_PROPERTY` or `APPLICATIONS_DEPRECATED` when both properties are included, but that is mainly a nit).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -4924,6 +5024,29 @@ describe('ManifestJSONParser', () => {
...
+        linter.collector
+      );
+
+      expect(manifestJSONParser.isValid).toEqual(true);
+      expect(linter.collector.errors).toEqual([]);
+
+      expect(linter.collector.warnings.length).toEqual(1);
+      assertHasMatchingError(linter.collector.warnings, {
+        code: messages.APPLICATIONS_DEPRECATED.code,
+      });
```
rpl, line 5048: Nit, we could consider similar tweaks to this test case as the ones from the other two comments in this review pass.

rpl (approved):
@willdurand lgtm, the comments that follows are just optional nits on the test side.

PR #4823, opened 2023-04-26
Do not use oneLine on localized strings

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4825

---

We shouldn't use `oneLine` in localized strings. That breaks extraction.
This PR fixes the existing localized strings (similar to how it is done
in mozilla/addons-frontend).

I'll submit another patch to configure the AMO ESLint plugin (which
should catch those issues).

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -69,12 +69,15 @@ export default class Linter {
     const { minManifestVersion, maxManifestVersion } = this.config;
     if (maxManifestVersion < minManifestVersion) {
       throw new AddonsLinterUserError(
-        i18n._(`
-        Invalid manifest version range requested:
-        --min-manifest-version (currently set to ${minManifestVersion})
-        should not be greater than
-        --max-manifest-version (currently set to ${maxManifestVersion}).
-      `)
+        i18n.sprintf(
+          i18n._(
+            `Invalid manifest version range requested:
+            --min-manifest-version (currently set to %(minManifestVersion)s)
+            should not be greater than
+            --max-manifest-version (currently set to %(maxManifestVersion)s).`
+          ),
+          { minManifestVersion, maxManifestVersion }
+        )
       );
     }
   }
--- a/src/linter.js
+++ b/src/linter.js
@@ -69,7 +69,7 @@ export default class Linter {
     const { minManifestVersion, maxManifestVersion } = this.config;
     if (maxManifestVersion < minManifestVersion) {
       throw new AddonsLinterUserError(
-        i18n._(oneLine`
+        i18n._(`
         Invalid manifest version range requested:
         --min-manifest-version (currently set to ${minManifestVersion})
         should not be greater than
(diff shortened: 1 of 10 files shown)


willdurand:
(I am trying to balance review requests)

src/messages/css.js
```diff
@@ -1,17 +1,15 @@
...
 import { i18n } from 'utils';
 
 export const CSS_SYNTAX_ERROR = {
   code: 'CSS_SYNTAX_ERROR',
   // This will be overriden by the reason passed from the error.
   message: i18n._('A CSS syntax error was encountered'),
-  description: i18n._(oneLine`An error was found in the CSS file being
+  description: i18n._(`An error was found in the CSS file being
     processed as a result further processing of that file is not possible`),
 };
```
rpl, line 9: @willdurand I also wonder if we should manually turn these strings to be actually one line as part of removing the `oneLine` tag. Not sure if that matters in practice (e.g. if not making those one line strings messes with how the table in the text format is formatted in the console).

src/linter.js
```diff
@@ -69,7 +69,7 @@ export default class Linter {
     const { minManifestVersion, maxManifestVersion } = this.config;
     if (maxManifestVersion < minManifestVersion) {
       throw new AddonsLinterUserError(
-        i18n._(oneLine`
+        i18n._(`
         Invalid manifest version range requested:
```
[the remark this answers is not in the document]
willdurand: fixed

src/messages/css.js
willdurand: These strings are wrapped at 80chars, that's why it uses backticks (and I kinda prefer that)

willdurand:
I found more occurrences of localized strings with interpolated values, all fixed now.

src/utils.js
```diff
@@ -151,6 +151,15 @@ export function getI18Data(locale) {
   return i18ndata;
 }
 
+// Functionality based on oneLine form declandewet/common-tags, copied from
```
Rob--W, line 154: form -> from

src/messages/javascript.js
```diff
@@ -150,8 +158,11 @@ export const DEPRECATED_CHROME_API = {
...
     code: apiToMessage(api),
-    message: i18n._(`"${api}" can cause issues when loaded temporarily`),
-    description: i18n._(oneLine`This API can cause issues when loaded
+    message: i18n.sprintf(
+      i18n._(`"$(api)s" can cause issues when loaded temporarily`),
+      { api }
+    ),
+    description: i18n._(`This API can cause issues when loaded
       temporarily using about:debugging in Firefox unless you specify
       applications|browser_specific_settings > gecko > id in the manifest.
```
Rob--W, line 167: Maybe not the point of this PR, but should we drop applications here?

Rob--W (approved):
r+wc

PR #4824, opened 2023-04-26
Use eslint-plugin-amo

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4826

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,13 +1,15 @@
 {
   "rules": {
     // This project uses `console.log()`.
-    "no-console": "off"
+    "no-console": "off",
+    "amo/i18n-no-tagged-templates": "error",
   },
   "env": {
     "node": true,
     "browser": true,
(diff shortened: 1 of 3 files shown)


willdurand:
I updated this PR to use the latest version of the plugin. This should be good to review now.

.eslintrc
```diff
@@ -1,14 +1,16 @@
...
+    "amo/i18n-no-tagged-templates": "error",
   },
   "env": {
     "node": true,
     "browser": true,
     "es6": true,
   },
-  "extends": ["amo/base"],
+  "plugins": ["amo"],
+  "extends": ["amo"],
```
diox, line 13: That includes a bunch of react rules addons-linter doesn't need though, right ? So why that change ? Is it just for consistency across our projects ?

willdurand: ah, right. That's a mistake...

PR #4831, opened 2023-04-27
Use dennis to lint our locale files

willdurand:


.circleci/config.yml
```diff
@@ -147,6 +147,18 @@ jobs:
       - run: npm run webext-test-functional
       - run: npm run smoke-test-eslint-version-conflicts
 
+  lint-locales:
+    working_directory: ~/addons-linter
+    docker:
+      - image: cimg/python:3.11.3
```
willdurand, line 153: We don't have python in the `cimg/node` images.

diox: Note: we didn't have coverage for Python 3.11 in dennis tests but it looks like it's working, I've submitted https://github.com/mozilla/dennis/pull/160

PR #4838, opened 2023-04-28
fix: Detect module in second child

regseb:
Fixes again https://github.com/mozilla/addons-linter/issues/4020#issuecomment-1527820431

In the detection of the type of a JavaScript file, the program walked only in the first child of a node (if the child was not an array).

### DIFF
--- a/tests/unit/scanners/test.javascript.js
+++ b/tests/unit/scanners/test.javascript.js
@@ -544,6 +544,24 @@ describe('JavaScript Scanner', () => {
       expect(jsScanner.sourceType).toEqual('module');
     });
 
+    it('should detect module in if', async () => {
+      const code = 'if (await Promise.resolve(true)) { }';
+
+      const jsScanner = new JavaScriptScanner(code, 'code.js');
+      await runJsScanner(jsScanner);
+
+      expect(jsScanner.sourceType).toEqual('module');
+    });
+
+    it('should detect module in ternary', async () => {
+      const code = 'const foo = true ? (1 && await Promise.resolve()) : 0;';
+
+      const jsScanner = new JavaScriptScanner(code, 'code.js');
+      await runJsScanner(jsScanner);
+
+      expect(jsScanner.sourceType).toEqual('module');
+    });
+
     it('should detect script', async () => {
       const code = oneLine`
         eval('foo');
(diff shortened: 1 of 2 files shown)


tests/unit/scanners/test.javascript.js
```diff
@@ -535,6 +535,15 @@ describe('JavaScript Scanner', () => {
       expect(jsScanner.sourceType).toEqual('script');
     });
 
+    it('should detect module in second child', async () => {
+      const code = 'const foo = await Promise.resolve();';
```
Rob--W, line 539: How about the following test cases:

```js
if (await Promise.resolve(true)) {
  // Do something
}
```

```js
const foo = true ? (1 && await Promise.resolve()) : 0;
```

I haven't looked at the AST/implementation too closely; these are interesting cases nevertheless.

regseb: I tested these two cases, the scanner detects well that it's module. I think that the scanner uses these `visitor-keys`:

- [`IfStatement.test`](https://github.com/eslint/eslint-visitor-keys/blob/7bd1fc1d483c2d0fdd5e0eddb2702f177372889c/lib/visitor-keys.js#L131)
- [`ConditionalExpression.consequent`](https://github.com/eslint/eslint-visitor-keys/blob/7bd1fc1d483c2d0fdd5e0eddb2702f177372889c/lib/visitor-keys.js#LL66C10-L66C20) and [`LogicalExpression.right`](https://github.com/eslint/eslint-visitor-keys/blob/7bd1fc1d483c2d0fdd5e0eddb2702f177372889c/lib/visitor-keys.js#L202)

---

I can add these two tests:

```JavaScript
    it('should detect module in if', async () => {
      const code = 'if (await Promise.resolve(true)) { }';

      const jsScanner = new JavaScriptScanner(code, 'code.js');
      await runJsScanner(jsScanner);

      expect(jsScanner.sourceType).toEqual('module');
    });

    it('should detect module in ternary', async () => {
      const code = 'const foo = true ? (1 && await Promise.resolve()) : 0;';

      const jsScanner = new JavaScriptScanner(code, 'code.js');
      await runJsScanner(jsScanner);

      expect(jsScanner.sourceType).toEqual('module');
    });
```

Rob--W: Please do, because that offers more coverage over the "trivial" cases from the original and the current patch.

Rob--W (approved):
Thanks!

Issue #4781, opened 2023-03-28
Enhance the compatibility detection for Android when `bss.gecko_android` is defined


willdurand:
See also: https://github.com/mozilla/addons-linter/issues/4780

---

Once we have support for the `bss.fenix_android` manifest key, we might be able to adjust some of our rules for Android. E.g., we currently warn about unsupported APIs on Android but that only takes the compatibility range from `bss.gecko`. If we have `bss.fenix_android`, we should probably use this compatibility range for the Android APIs.

PR #4856, opened 2023-05-04
Emit a notice when `bss.gecko_android.strict_max_version` is set

### LINKED ISSUE #4855: A notice should be displayed when strict_max_version is specified for gecko_android
From https://github.com/mozilla/addons-linter/issues/4780#issuecomment-1532855396

There's a difference in notices for this case: 

 `       "gecko": {
        "id": "testdummy6@test",
	"strict_min_version": "110.0",
	"strict_max_version":"100.0"
        } ` 

Here I get a notice:  "strict_max_version" not required.

For 	 `"browser_specific_settings": {
        "gecko": {
            "id": "testdummy6@test"
        } ,
		"gecko_android": {
            "strict_min_version": "113.0",
	   "strict_max_version":"100.0"
        }
    }`

There's no notice.

Expected result:
A notice should be displayed for gecko_android too when strict_max_version is used in manifest.json of an extension.

willdurand:
Fixes #4855

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -277,6 +277,36 @@ describe('ManifestJSONParser', () => {
     expect(manifestJSONParser.isValid).toEqual(true);
   });
 
+  it('should accept applications.gecko AND browser_specific_settings.gecko_android', () => {
+    const addonLinter = new Linter({ _: ['bar'] });
+    const json = validManifestJSON({
+      applications: {
+        gecko: {
+          strict_max_version: '100.0',
+        },
+      },
+      browser_specific_settings: {
+        gecko_android: {},
+      },
+    });
+
+    const manifestJSONParser = new ManifestJSONParser(
+      json,
+      addonLinter.collector
+    );
+
+    const { notices } = addonLinter.collector;
+    // This is only available if `applications.gecko` isn't lost when we deal
+    // with both `applications` and `browser_specific_settings`.
+    expect(notices).toEqual([
+      expect.objectContaining({
+        code: messages.STRICT_MAX_VERSION.code,
+        message: expect.stringMatching('strict_max_version'),
+      }),
+    ]);
+    expect(manifestJSONParser.isValid).toEqual(true);
+  });
+
   describe('id', () => {
     it('should return the correct id', () => {
       const addonLinter = new Linter({ _: ['bar'] });
@@ -1395,24 +1425,33 @@ describe('ManifestJSONParser', () => {
   });
 
   describe('strict_max_version', () => {
-    it('warns on strict_max_version', () => {
-      const addonLinter = new Linter({ _: ['bar'] });
-      const json = validManifestJSON({
-        browser_specific_settings: {
-          gecko: {
-            strict_max_version: '58.0',
+    it.each(['gecko', 'gecko_android'])(
+      'warns on %s.strict_max_version',
+      (key) => {
+        const addonLinter = new Linter({ _: ['bar'] });
+        const json = validManifestJSON({
+          browser_specific_settings: {
+            [key]: {
+              strict_max_version: '58.0',
+            },
           },
-        },
-      });
-      const manifestJSONParser = new ManifestJSONParser(
-        json,
-        addonLinter.collector
-      );
-      expect(manifestJSONParser.isValid).toEqual(true);
-      const { notices } = addonLinter.collector;
-      expect(notices[0].code).toEqual(messages.STRICT_MAX_VERSION.code);
-      expect(notices[0].message).toContain('strict_max_version');
-    });
+        });
+
+        const manifestJSONParser = new ManifestJSONParser(
+          json,
+          addonLinter.collector
+        );
+
+        expect(manifestJSONParser.isValid).toEqual(true);
+        const { notices } = addonLinter.collector;
+        expect(notices).toEqual([
+          expect.objectContaining({
+            code: messages.STRICT_MAX_VERSION.code,
+            message: expect.stringMatching('strict_max_version'),
+          }),
+        ]);
+      }
+    );
 
     it('does not warn on strict_max_version in language packs', () => {
       const addonLinter = new Linter({ _: ['bar'] });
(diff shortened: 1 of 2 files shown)


rpl (changes requested):
In addition to the inline review comment below, while trying this branch on a test extension, I noticed another pre-existing issue we may want to think about (not necessarily in this PR, we may as well file a separate issue and handle it in a separate PR):

with a manifest which include `bss.gecko_android` and `bss.gecko.strict_min_version` (or `applications.gecko.strict_min_version`) is < 113 (when `bsss.gecko_android` has been introduced based on the browser-compat-data) we show the following two warnings:

```
KEY_FIREFOX_UNSUPPORTED_BY_MIN…   Manifest key not supported   "strict_min_version" requires Firefox 111.0, which   manifest.json            
_VERSION                          by the specified minimum     was released before version 113 introduced support                            
                                  Firefox version              for "browser_specific_settings.gecko_android".                                
KEY_FIREFOX_ANDROID_UNSUPPORTE…   Manifest key not supported   "strict_min_version" requires Firefox for Android    manifest.json            
D_BY_MIN_VERSION                  by the specified minimum     111.0, which was released before version 113                                  
                                  Firefox for Android          introduced support for                                                        
                                  version                      "browser_specific_settings.gecko_android".
```

And I was thinking that we may handle that by introducing a new `firefoxAndroidStrictMinVersion` in the `utils.js` module, then in `checkKeySupport` skip the `if (support.firefox) { ... }` block if the key is `browser_specific_settings.gecko_android` and then in the `if (support.firefox_android) { ... }` use the `minVersion` returned by `firefoxAndroidStrictMinVersion`, something like the changes in the following diff:

```diff
diff --git a/src/parsers/manifestjson.js b/src/parsers/manifestjson.js
index 0d6fb81a..5a86b22a 100644
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -38,6 +38,7 @@ import JSONParser from 'parsers/json';
 import {
   basicCompatVersionComparison,
   firefoxStrictMinVersion,
+  firefoxAndroidStrictMinVersion,
   firstStableVersion,
   isToolkitVersionString,
   isValidVersionString,
@@ -144,7 +145,7 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   checkKeySupport(support, minVersion, key, isPermission = false) {
-    if (support.firefox) {
+    if (support.firefox && key !== "browser_specific_settings.gecko_android") {
       // We don't have to support gaps in the `@mdn/browser-compat-data`
       // information for Firefox Desktop so far.
       const versionAdded = support.firefox.version_added;
@@ -182,12 +183,13 @@ export default class ManifestJSONParser extends JSONParser {
       // not warn if the minVersion is in one of the few version gaps).
       const versionAddedAndroid = firstStableVersion(support.firefox_android);
 
-      if (basicCompatVersionComparison(versionAddedAndroid, minVersion)) {
+      const minVersionAndroid = firefoxAndroidStrictMinVersion(this.parsedJSON);
+      if (basicCompatVersionComparison(versionAddedAndroid, minVersionAndroid)) {
         if (!isPermission) {
           this.collector.addWarning(
             messages.keyFirefoxAndroidUnsupportedByMinVersion(
               key,
-              this.parsedJSON.applications.gecko.strict_min_version,
+              minVersionAndroid,
               versionAddedAndroid
             )
           );
@@ -195,7 +197,7 @@ export default class ManifestJSONParser extends JSONParser {
           this.collector.addNotice(
             messages.permissionFirefoxAndroidUnsupportedByMinVersion(
               key,
-              this.parsedJSON.applications.gecko.strict_min_version,
+              minVersionAndroid,
               versionAddedAndroid
             )
           );
diff --git a/src/utils.js b/src/utils.js
index 165cd83e..0b63d204 100644
--- a/src/utils.js
+++ b/src/utils.js
@@ -385,6 +385,22 @@ export function firefoxStrictMinVersion(manifestJson) {
   return null;
 }
 
+export function firefoxAndroidStrictMinVersion(manifestJson) {
+  if (
+    manifestJson.applications &&
+    manifestJson.applications.gecko_android &&
+    manifestJson.applications.gecko_android.strict_min_version &&
+    typeof manifestJson.applications.gecko_android.strict_min_version === 'string'
+  ) {
+    return parseInt(
+      manifestJson.applications.gecko_android.strict_min_version.split('.')[0],
+      10
+    );
+  }
+  // Fallback to gecko compatibility range.
+  return firefoxStrictMinVersion(manifestJson);
+}
+
 export function basicCompatVersionComparison(versionAdded, minVersion) {
   const asNumber = parseInt(versionAdded, 10);
   return !Number.isNaN(asNumber) && asNumber > minVersion;
```  

I'm pretty sure some of the existing test case is going to fail with these changes, and so that would be one more reason to consider this separately from what this PR is aiming for, and in this one just handle the notice message on strict_max_version when included in gecko_android.

willdurand:
@rpl about the other pre-existing issue, this is covered by https://github.com/mozilla/addons-linter/issues/4781, right? If so, can that be moved out of this PR? Thanks

rpl (approved):
lgtm (and thanks for adding the additional test case).

rpl:
> @rpl about the other pre-existing issue, this is covered by #4781, right? If so, can that be moved out of this PR? Thanks

yes, the pre-existing issue I mentioned in https://github.com/mozilla/addons-linter/pull/4856#pullrequestreview-1418435148 looks definitely part of what we have filed #4781, from my perspective it is absolutely ok (and actually better) to move that part out of this PR and consider that comment as part of a separate PR meant to fix #4781 specifically.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -277,6 +277,35 @@ describe('ManifestJSONParser', () => {
...
+      },
+      browser_specific_settings: {
+        gecko_android: {},
+      },
+    });
+
+    const manifestJSONParser = new ManifestJSONParser(
+      json,
+      addonLinter.collector
+    );
```
rpl, line 296: eslint is complaining that manifestJSONParser isn't used, we should fix this nit before landing this PR.

PR #4887, opened 2023-05-25
Warn about deprecated browser_style:true in MV3

### LINKED ISSUE #4873: Add deprecation warnings when browser_style is relied upon in MV3
`browser_style` is deprecated in MV3 (https://bugzilla.mozilla.org/show_bug.cgi?id=1827910).

The deprecation plan is described at https://bugzilla.mozilla.org/show_bug.cgi?id=1827910#c1.
Warn if MV3 AND:
- `options_ui.browser_style` is NOT false (i.e. void or true) AND `options_ui.open_in_tab` is NOT true (i.e. void or false).
- `sidebar_action.browser_style` is NOT false (i.e. void or true)
- `page_action.browser_style` is true
- `action.browser_style` is true.

The deprecation is currently not annotated in the schema, because of special cases around the conditions (https://phabricator.services.mozilla.com/D176811#5828582). When `browser_style` is eventually completely unsupported, we could rely on the schema and drop the special checks.

Rob--W:
Note that this does not warn about the use of options_ui or sidebar_action, despite their browser_style defaulting to true in Firefox 114 and earlier. This is because there are a very small number of affected extensions, and warning about these keys without the presence in manifest.json would be confusing to most extensions.

Fixes #4873.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -740,6 +740,7 @@ export default class ManifestJSONParser extends JSONParser {
     this.validateRestrictedPermissions();
     this.validateExtensionID();
     this.validateHiddenAddon();
+    this.validateDeprecatedBrowserStyle();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/parsers/manifestjson.js
```diff
@@ -791,6 +792,35 @@ export default class ManifestJSONParser extends JSONParser {
     }
   }
 
+  validateDeprecatedBrowserStyle() {
+    if (this.parsedJSON.manifest_version !== 3) {
```
willdurand, line 796: nit: this is a bit unusual. Shouldn't it be `< 3`?

src/parsers/manifestjson.js
```diff
@@ -791,6 +792,35 @@ export default class ManifestJSONParser extends JSONParser {
...
+        this.collector.addWarning({
+          instancePath,
+          ...messages.manifestFieldUnsupported(instancePath, errorParam),
+        });
+      }
+    };
+    checkBrowserStyleInManifestKey('action');
+    checkBrowserStyleInManifestKey('options_ui');
+    checkBrowserStyleInManifestKey('page_action');
+    checkBrowserStyleInManifestKey('sidebar_action');
```
willdurand, line 821: nit: I am guessing there is a reason for that but I would have replaced the function with a `for (const manifestKey of ["action", "options_ui", ...]) {}`

src/parsers/manifestjson.js
Rob--W: Then it would continue to warn in MV4 etc. As the comment on the next line emphasizes, this check is meant to cover MV2->MV3 migration.

PR #4903, opened 2023-06-06
Better validate add-on names

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/4904


rpl (approved):
As we discussed yesterday, we are going to start simple and gradually improve the validation in followups as much as we find valuable (e.g. in this version we are not going to account yet for a extension name using a i18n placeholder in the manifest field and one or more of the corresponding localized strings which would not pass this validation, we will consider adding that in a separate followup).

Issue #5152, opened 2024-01-02
Return a warning for the `service_worker` property


regseb:
### Describe the problem and steps to reproduce it:

```json
{
  "manifest_version": 3,
  "name": "Test case",
  "version": "1.0.0",
  "background": {
    "service_worker": "background.js",
    "scripts": ["background.js"]
  }
}
```

### What happened?

> Validation Summary:
>
> errors          1
> notices         0
> warnings        0
>
> ERRORS:
>
> | Code | Message | Description | File | Line | Column |
> | --- | --- | --- | --- | --- | --- |
> | MANIFEST_FIELD_UNSUPPORTED | "/background/service_worker" is not supported. | "/background/service_worker" is not supported. | manifest.json | | |

### What did you expect to happen?

- If the `background` property has a `scripts` and a `service_worker`: return no error (or only a warning) for the `service_worker` property.
- If the `background` property has only a `service_worker`: continue to return an error for the `service_worker` property.

### Anything else we should know?

- https://github.com/w3c/webextensions/issues/282
- [Background page fails to load if background.service_worker and background.scripts are both present](https://bugzil.la/1860304) (Bugzilla)
- [`background` / Browser support](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background#browser_support) (MDN)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-419)

PR #5156, opened 2024-01-08
Warn instead of error when service_worker is combined with background.scripts/page

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5152

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-423)

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -3179,6 +3179,47 @@ describe('ManifestJSONParser', () => {
       );
     });
 
+    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1860304
+    it.each([
+      { scripts: ['bg.js'] },
+      { page: 'bg.html' },
+      { scripts: [], page: 'bg.html' },
+    ])(
+      'emits a warning when background.service_worker is being used in combination with %o',
+      (backgroundProps) => {
+        const linter = new Linter({ _: ['bar'] });
+        const json = validManifestJSON({
+          background: {
+            service_worker: 'background_worker.js',
+            ...backgroundProps,
+          },
+        });
+        // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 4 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -3179,6 +3179,47 @@ describe('ManifestJSONParser', () => {
...
+          linter.collector,
+          {
+            io: {
+              files: { 'background_worker.js': '', 'bg.js': '', 'bg.html': '' },
+            },
+          }
+        );
+
+        expect(linter.collector.errors).toEqual([]);
+        // expect(manifestJSONParser.isValid).toEqual(true);
```
rpl, line 3209: @willdurand I believe you may have commented this while you were figuring out where the strict behavior related to the additional properties was coming from and forgot to uncomment it, it does currently pass as expected if uncommented.

would you mind to also move it after the assertion on the warnings along with uncommenting it? (in the past it was often annoying to get a failure from this assertion without having hit the more detailed assertion that follows it in this test and so I usually put this assertion at the end to make it easier to investigate regressions hit in ci jobs).

rpl (approved):
@willdurand lgtm, in the comments below just a nit for an assertion currently left as commented + some details about that schema update file history (at least the amount of history that I could gather without spending too much time on it).

rpl:
@willdurand forgot to mention, there is a typo in the commit message (pesent -> present), I took care of fixing the one in the PR subject message derived from the commit message

Rob--W (approved):
LGTM, thanks for the fix.

Note: the commit title states that its purpose is to emit a warning. That's not the objective: instead, the goal is to get rid of the hard error. Downgrading the error to a warning meets that goal, so this patch is indeed a right fix.

I suggest to change the commit message to: `Warn instead of error when service_worker is combined with background.scripts/page`

PR #5164, opened 2024-01-12
Do not warn about nesting css selectors when firefoxStrictMinVersion is >= 117 

### LINKED ISSUE #5102: Allow CSS nesting

Allow CSS nesting as it is supported by Firefox 117+

### What happened?

If i run the linter i get `INVALID_SELECTOR_NESTING` warning

that there is no warning about CSS nesting

https://caniuse.com/css-nesting

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-370)

schmop:
Fixes #5102

Since Firefox Version 117 CSS selector nesting is supported.
This is not reflected in the `rules/css/invalidNestingSelectors.js` of the addons-linter.
In this changeset I introduced a check against the `firefoxMinStrictVersion`, or `strict_min_version` which is specified in the `manifest.json`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-431)

### DIFF
--- a/src/messages/css.js
+++ b/src/messages/css.js
@@ -11,5 +11,7 @@ export const CSS_SYNTAX_ERROR = {
 export const INVALID_SELECTOR_NESTING = {
   code: 'INVALID_SELECTOR_NESTING',
   message: i18n._('Invalid nesting of selectors found'),
-  description: i18n._(`Selector nesting is supported from firefox version 117.0 and above`),
+  description: i18n._(
+    `Selector nesting is supported from firefox version 117.0 and above`
+  ),
 };
--- a/src/messages/css.js
+++ b/src/messages/css.js
@@ -11,5 +11,7 @@ export const CSS_SYNTAX_ERROR = {
 export const INVALID_SELECTOR_NESTING = {
   code: 'INVALID_SELECTOR_NESTING',
   message: i18n._('Invalid nesting of selectors found'),
-  description: i18n._(`Selectors should not be nested`),
+  description: i18n._(
+    `Selector nesting is supported from firefox version 117.0 and above`
+  ),
 };
--- a/src/messages/css.js
+++ b/src/messages/css.js
@@ -11,7 +11,5 @@ export const CSS_SYNTAX_ERROR = {
 export const INVALID_SELECTOR_NESTING = {
   code: 'INVALID_SELECTOR_NESTING',
   message: i18n._('Invalid nesting of selectors found'),
-  description: i18n._(
-    `Selector nesting is supported from firefox version 117.0 and above`
-  ),
+  description: i18n._(`Selector nesting is supported from firefox version 117.0 and above`),
 };
--- a/src/messages/css.js
+++ b/src/messages/css.js
@@ -11,7 +11,6 @@ export const CSS_SYNTAX_ERROR = {
 export const INVALID_SELECTOR_NESTING = {
   code: 'INVALID_SELECTOR_NESTING',
   message: i18n._('Invalid nesting of selectors found'),
-  description: i18n._(
-    `Selector nesting is supported from firefox version 117.0 and above`
-  ),
+  description: i18n._(`Selector nesting is supported from firefox version 117.0
+  and above`),
 };
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -47,17 +47,6 @@ import {
 } from 'utils';
 import BLOCKED_CONTENT_SCRIPT_HOSTS from 'blocked_content_script_hosts.txt';
 
-/**
- * @typedef {Object} Metadata
- * @property {string} id
- * @property {number} manifestVersion
- * @property {string} name
- * @property {number} type
- * @property {string} version
- * @property {string} firefoxMinVersion
- * @property {Set<string>} experimentApiPaths
- */
-
 async function getStreamImageSize(stream) {
   const chunks = [];
   for await (const chunk of stream) {
@@ -1297,6 +1286,15 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   /**
+   * @typedef {Object} Metadata
+   * @property {string} id
+   * @property {number} manifestVersion
+   * @property {string} name
+   * @property {number} type
+   * @property {string} version
+   * @property {string} firefoxMinVersion
+   * @property {Set<string>} experimentApiPaths
+   *
    * @returns {Metadata}
    */
   getMetadata() {
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -1293,6 +1293,7 @@ export default class ManifestJSONParser extends JSONParser {
    * @property {number} type
    * @property {string} version
    * @property {string} firefoxMinVersion
+   * @property {string} firefoxStrictMinVersion
    * @property {Set<string>} experimentApiPaths
    *
    * @returns {Metadata}
@@ -1304,10 +1305,14 @@ export default class ManifestJSONParser extends JSONParser {
       name: this.parsedJSON.name,
       type: PACKAGE_EXTENSION,
       version: this.parsedJSON.version,
+      // This is the `strict_min_version` value set in the `manifest.json` file
+      // for Firefox for desktop.
       firefoxMinVersion:
         this.parsedJSON.applications &&
         this.parsedJSON.applications.gecko &&
         this.parsedJSON.applications.gecko.strict_min_version,
+      // This is the strict min *major* version for Firefox for desktop.
+      firefoxStrictMinVersion: firefoxStrictMinVersion(this.parsedJSON),
       experimentApiPaths: this.getExperimentApiPaths(),
     };
   }
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -47,6 +47,17 @@ import {
 } from 'utils';
 import BLOCKED_CONTENT_SCRIPT_HOSTS from 'blocked_content_script_hosts.txt';
 
+/**
+ * @typedef {Object} Metadata
+ * @property {string} id
+ * @property {number} manifestVersion
+ * @property {string} name
+ * @property {number} type
+ * @property {string} version
+ * @property {string} firefoxMinVersion
+ * @property {Set<string>} experimentApiPaths
+ */
+
 async function getStreamImageSize(stream) {
   const chunks = [];
   for await (const chunk of stream) {
@@ -1285,6 +1296,9 @@ export default class ManifestJSONParser extends JSONParser {
     return apiPaths;
   }
 
+  /**
+   * @returns {Metadata}
+   */
   getMetadata() {
     return {
       id: this.getAddonId(),
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -1,8 +1,9 @@
 import * as messages from 'messages';
 
-import { basicCompatVersionComparisonGEQ } from '../../utils';
+import { basicCompatVersionComparison } from '../../utils';
 
-const CSS_NESTING_MIN_VERSION = 117;
+// Allowed version is 117.0 and above
+const CSS_NESTING_MIN_VERSION = 116;
 
 export function invalidNesting(
   cssNode,
@@ -10,7 +11,7 @@ export function invalidNesting(
   { startLine, startColumn, addonMetadata } = {}
 ) {
   if (
-    basicCompatVersionComparisonGEQ(
+    basicCompatVersionComparison(
       addonMetadata?.firefoxMinVersion,
       CSS_NESTING_MIN_VERSION
     )
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -2,7 +2,6 @@ import * as messages from 'messages';
 
 import { basicCompatVersionComparison } from '../../utils';
 
-// Allowed version is 117.0 and above
 const CSS_NESTING_MIN_VERSION = 117;
 
 export function invalidNesting(
@@ -11,10 +10,10 @@ export function invalidNesting(
   { startLine, startColumn, addonMetadata } = {}
 ) {
   if (
-    addonMetadata?.firefoxMinVersion &&
+    addonMetadata?.firefoxStrictMinVersion &&
     !basicCompatVersionComparison(
       CSS_NESTING_MIN_VERSION,
-      addonMetadata?.firefoxMinVersion
+      addonMetadata?.firefoxStrictMinVersion
     )
   ) {
     return [];
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -1,10 +1,22 @@
 import * as messages from 'messages';
 
+import { basicCompatVersionComparisonGEQ } from '../../utils';
+
+const CSS_NESTING_MIN_VERSION = 117;
+
 export function invalidNesting(
   cssNode,
   filename,
-  { startLine, startColumn } = {}
+  { startLine, startColumn, addonMetadata } = {}
 ) {
+  if (
+    basicCompatVersionComparisonGEQ(
+      addonMetadata?.firefoxMinVersion,
+      CSS_NESTING_MIN_VERSION
+    )
+  ) {
+    return [];
+  }
   const messageList = [];
   if (cssNode.type === 'rule') {
     for (let i = 0; i < cssNode.nodes.length; i++) {
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -18,6 +18,7 @@ export function invalidNesting(
   ) {
     return [];
   }
+
   const messageList = [];
   if (cssNode.type === 'rule') {
     for (let i = 0; i < cssNode.nodes.length; i++) {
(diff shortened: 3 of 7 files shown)


tests/unit/rules/css/test.invalidNesting.js
```diff
@@ -22,6 +26,23 @@ describe('CSS Rule InvalidNesting', () => {
...
+  it('should not report invalid nesting on sufficient version', async () => {
+    const code = oneLine`/* I'm a comment */
+      #something {
+        .bar {
+          height: 100px;
+        }
+      }`;
+    const cssScanner = new CSSScanner(code, 'fakeFile.css', {
+      addonMetadata: {
+        firefoxMinVersion: '117.0.1',
```
willdurand, line 38: It'd be good to test with `117` too since you introduced a different comparison function to allow equal values.

tests/unit/test.utils.js
```diff
@@ -602,11 +603,37 @@ describe('basicCompatVersionComparison', () => {
     expect(basicCompatVersionComparison('61', 60)).toBe(true);
   });
 
+  it('should return false when version added is equals than min version', () => {
```
willdurand, line 606:
```suggestion
  it('should return false when version added is equal to min version', () => {
```

tests/unit/test.utils.js
```diff
@@ -602,11 +603,37 @@ describe('basicCompatVersionComparison', () => {
...
+    expect(basicCompatVersionComparison('61.5.2', 61)).toBe(false);
```
willdurand, line 607: This makes me think that we have a bug in the current comparison function. I haven't looked in depth yet but I am wondering why we wouldn't allow this..

willdurand: I took another look and this `basicCompatVersionComparison` function seems to do the opposite of what I initially thought.

src/scanners/base.js
```diff
@@ -27,6 +27,18 @@ export default class BaseScanner {
...
+   * @property {import('../parsers/manifestjson').Metadata} addonMetadata
+   * @property {import('../collector').default} collector
+   * @property {string[]} disabledRules
+   * @property {string[]} existingFiles
+   * @property {boolean} privileged
+   *
+   * @param {string|Buffer|import('stream').Readable} contents
+   * @param {string} filename
+   * @param {ScannerOptions} options
+   */
```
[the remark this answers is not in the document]
schmop: These kind of JSDoc comments are the compromise of having types without using typescript. It helps documenting interfaces, allowing IDEs with code suggestions and helping new devs to faster move around foreign code. I typed parts where I struggled with the given interfaces and needed to dig deeper and documented it right away, so others don't have to. Is this kind of type-documenting style unwelcome in this repository? I can definitely just remove it.

src/rules/css/invalidNesting.js
```diff
@@ -1,10 +1,22 @@
...
+const CSS_NESTING_MIN_VERSION = 117;
+
 export function invalidNesting(
   cssNode,
   filename,
-  { startLine, startColumn } = {}
+  { startLine, startColumn, addonMetadata } = {}
 ) {
+  if (
+    basicCompatVersionComparisonGEQ(
```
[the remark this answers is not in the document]
schmop: That might be a bad abbreviation of "Greater or EQual", which is common in TeX, but might not sound familiar in other contexts. I will rename it the function to `isMajorVersionGreaterOrEqual(version, minimalVersion)`, but I am open to suggestions. Naming things is hard.

src/parsers/manifestjson.js
```diff
@@ -47,6 +47,17 @@ import {
...
+/**
+ * @typedef {Object} Metadata
+ * @property {string} id
+ * @property {number} manifestVersion
+ * @property {string} name
+ * @property {number} type
+ * @property {string} version
+ * @property {string} firefoxMinVersion
+ * @property {Set<string>} experimentApiPaths
+ */
```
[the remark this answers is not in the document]
schmop: I moved it down. I added JSDoc, because it helped me a great deal understand what structure I can rely on. I thought it might help future devs aswell, that do not know the internals yet.

tests/unit/rules/css/test.invalidNesting.js
schmop: I added a test

src/rules/css/invalidNesting.js
```diff
@@ -1,10 +1,22 @@
...
+      addonMetadata?.firefoxMinVersion,
+      CSS_NESTING_MIN_VERSION
```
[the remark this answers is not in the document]
schmop: I am pretty sure, the order was intact.
* The first parameter is the configured version as a string (or `undefined` if nothing is configured), that needs to be parsed before we can consume it. It has a form like `"133.2.6"` and the only thing we take a look at is the major-version part `133`.
* The second parameter is the required minimal version, that is hardcoded as a number, like `122`.

Example from the tests:
```js
  it('should return false when version added is undefined', () => {
    expect(basicCompatVersionComparison(undefined, 60)).toBe(false);
  });

  it('should return true when version added is bigger than min version', () => {
    expect(basicCompatVersionComparison('61', 60)).toBe(true);
  });
```
And indeed we can switch! That's great!

src/messages/css.js
```diff
@@ -11,5 +11,7 @@ export const CSS_SYNTAX_ERROR = {
 export const INVALID_SELECTOR_NESTING = {
   code: 'INVALID_SELECTOR_NESTING',
   message: i18n._('Invalid nesting of selectors found'),
-  description: i18n._(`Selectors should not be nested`),
+  description: i18n._(
+    `Selector nesting is supported from firefox version 117.0 and above`
+  ),
```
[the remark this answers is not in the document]
schmop: Prettier demands this in multiple lines, your formatting will result in a failing pipeline.

src/parsers/manifestjson.js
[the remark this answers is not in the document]
willdurand: I am not against but I think I'd rather see that added in a different PR.

src/messages/css.js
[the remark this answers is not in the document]
willdurand: I tried this suggestion locally and Prettier didn't say anything 🤔

src/scanners/base.js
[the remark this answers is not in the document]
willdurand: see other comment

src/rules/css/invalidNesting.js
```diff
@@ -1,10 +1,23 @@
...
+  { startLine, startColumn, addonMetadata } = {}
 ) {
+  if (
+    basicCompatVersionComparison(
+      addonMetadata?.firefoxMinVersion,
+      CSS_NESTING_MIN_VERSION
+    )
+  ) {
+    return [];
+  }
```
willdurand, line 20:
```suggestion
  }
  
```

tests/unit/rules/css/test.invalidNesting.js
```diff
@@ -22,6 +26,40 @@ describe('CSS Rule InvalidNesting', () => {
...
+      }`;
+    const cssScanner = new CSSScanner(code, 'fakeFile.css', {
+      addonMetadata: {
+        firefoxMinVersion: '117.0.0',
+      },
+    });
+
+    const { linterMessages } = await cssScanner.scan();
+    expect(linterMessages.length).toEqual(0);
+  });
```
willdurand, line 61: It looks like the two test cases are very similar, can we do something like this? cf. https://jestjs.io/docs/api#each

```suggestion
  it.each(["117.0.0", "117.0.1"])('should not report invalid nesting when firefoxMinVersion=%s', async (firefoxMinVersion) => {
    const code = oneLine`/* I'm a comment */
      #something {
        .bar {
          height: 100px;
        }
      }`;
    const cssScanner = new CSSScanner(code, 'fakeFile.css', {
      addonMetadata: {
        firefoxMinVersion,
      },
    });

    const { linterMessages } = await cssScanner.scan();
    expect(linterMessages.length).toEqual(0);
  });
```

willdurand (changes requested):
Thanks.  I requested some more changes but this is looking great.

src/rules/css/invalidNesting.js
```diff
@@ -1,10 +1,24 @@
 import * as messages from 'messages';
 
+import { basicCompatVersionComparison } from '../../utils';
+
+// Allowed version is 117.0 and above
+const CSS_NESTING_MIN_VERSION = 116;
```
willdurand, line 6: I don't think this is correct, though I understand why you set this constant to `117 - 1`. I think this would work a bit better, though:

```diff
diff --git a/src/rules/css/invalidNesting.js b/src/rules/css/invalidNesting.js
index b8cf9733..b51eab8f 100644
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -3,7 +3,7 @@ import * as messages from 'messages';
 import { basicCompatVersionComparison } from '../../utils';
 
 // Allowed version is 117.0 and above
-const CSS_NESTING_MIN_VERSION = 116;
+const CSS_NESTING_MIN_VERSION = 117;
 
 export function invalidNesting(
   cssNode,
@@ -11,9 +11,10 @@ export function invalidNesting(
   { startLine, startColumn, addonMetadata } = {}
 ) {
   if (
-    basicCompatVersionComparison(
-      addonMetadata?.firefoxMinVersion,
-      CSS_NESTING_MIN_VERSION
+    addonMetadata?.firefoxMinVersion &&
+    !basicCompatVersionComparison(
+      CSS_NESTING_MIN_VERSION,
+      addonMetadata?.firefoxMinVersion
     )
   ) {
     return [];
diff --git a/tests/unit/rules/css/test.invalidNesting.js b/tests/unit/rules/css/test.invalidNesting.js
index ecf05ea4..4ec67b11 100644
--- a/tests/unit/rules/css/test.invalidNesting.js
+++ b/tests/unit/rules/css/test.invalidNesting.js
@@ -5,28 +5,31 @@ import { VALIDATION_WARNING } from 'const';
 import CSSScanner from 'scanners/css';
 
 describe('CSS Rule InvalidNesting', () => {
-  it('should detect invalid nesting', async () => {
-    const code = oneLine`/* I'm a comment */
+  it.each(['60.5', '116.0', '116'])(
+    'should detect invalid nesting when firefoxMinVersion=%s',
+    async (firefoxMinVersion) => {
+      const code = oneLine`/* I'm a comment */
       #something {
         .bar {
           height: 100px;
         }
       }`;
-    const cssScanner = new CSSScanner(code, 'fakeFile.css', {
-      addonMetadata: {
-        firefoxStrictMinVersion: '60.5',
-      },
-    });
+      const cssScanner = new CSSScanner(code, 'fakeFile.css', {
+        addonMetadata: {
+          firefoxMinVersion,
+        },
+      });
 
-    const { linterMessages } = await cssScanner.scan();
-    expect(linterMessages.length).toEqual(1);
-    expect(linterMessages[0].code).toEqual(
-      messages.INVALID_SELECTOR_NESTING.code
-    );
-    expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
-  });
+      const { linterMessages } = await cssScanner.scan();
+      expect(linterMessages.length).toEqual(1);
+      expect(linterMessages[0].code).toEqual(
+        messages.INVALID_SELECTOR_NESTING.code
+      );
+      expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
+    }
+  );
 
-  it.each(['117.0.0', '117.0.1'])(
+  it.each(['117', '117.0', '117.1', '117.0.0', '117.0.1'])(
     'should not report invalid nesting when firefoxMinVersion=%s',
     async (firefoxMinVersion) => {
       const code = oneLine`/* I'm a comment */
```

willdurand (changes requested):

src/rules/css/invalidNesting.js
schmop: I applied the changes and added the test cases, where the order of `basicCompatVersionComparison` is relevant and breaks.
The second argument cannot be a string with complex version formats, it will get casted to a `number`. This worked in the special cases of Major versions (`116`) and Major and Minor versions (`116.1`), but not for Major, Minor and Patch Versions.

How do we proceed?

willdurand: Ha. Yeah, now I get it... We should pass `firefoxMinVersion` as second parameter like it is done here: https://github.com/mozilla/addons-linter/blob/0d3e83b8e576dc26b92b369c2debf83697a790d6/src/parsers/manifestjson.js#L158

That said, I overlooked an important detail: `minFirefoxVersion` here is processed here: https://github.com/mozilla/addons-linter/blob/0d3e83b8e576dc26b92b369c2debf83697a790d6/src/parsers/manifestjson.js#682

That's why it doesn't work like it should... What do you think of the following diff?

```diff
diff --git a/src/parsers/manifestjson.js b/src/parsers/manifestjson.js
index b160062b..7b4dae8f 100644
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -1293,6 +1293,7 @@ export default class ManifestJSONParser extends JSONParser {
    * @property {number} type
    * @property {string} version
    * @property {string} firefoxMinVersion
+   * @property {string} firefoxStrictMinVersion
    * @property {Set<string>} experimentApiPaths
    *
    * @returns {Metadata}
@@ -1304,10 +1305,14 @@ export default class ManifestJSONParser extends JSONParser {
       name: this.parsedJSON.name,
       type: PACKAGE_EXTENSION,
       version: this.parsedJSON.version,
+      // This is the `strict_min_version` value set in the `manifest.json` file
+      // for Firefox for desktop.
       firefoxMinVersion:
         this.parsedJSON.applications &&
         this.parsedJSON.applications.gecko &&
         this.parsedJSON.applications.gecko.strict_min_version,
+      // This is the strict min *major* version for Firefox for desktop.
+      firefoxStrictMinVersion: firefoxStrictMinVersion(this.parsedJSON),
       experimentApiPaths: this.getExperimentApiPaths(),
     };
   }
diff --git a/src/rules/css/invalidNesting.js b/src/rules/css/invalidNesting.js
index b8cf9733..cd6d22de 100644
--- a/src/rules/css/invalidNesting.js
+++ b/src/rules/css/invalidNesting.js
@@ -3,7 +3,7 @@ import * as messages from 'messages';
 import { basicCompatVersionComparison } from '../../utils';
 
 // Allowed version is 117.0 and above
-const CSS_NESTING_MIN_VERSION = 116;
+const CSS_NESTING_MIN_VERSION = 117;
 
 export function invalidNesting(
   cssNode,
@@ -11,9 +11,10 @@ export function invalidNesting(
   { startLine, startColumn, addonMetadata } = {}
 ) {
   if (
-    basicCompatVersionComparison(
-      addonMetadata?.firefoxMinVersion,
-      CSS_NESTING_MIN_VERSION
+    addonMetadata?.firefoxStrictMinVersion &&
+    !basicCompatVersionComparison(
+      CSS_NESTING_MIN_VERSION,
+      addonMetadata?.firefoxStrictMinVersion
     )
   ) {
     return [];
diff --git a/tests/unit/rules/css/test.invalidNesting.js b/tests/unit/rules/css/test.invalidNesting.js
index ecf05ea4..2ad372f4 100644
--- a/tests/unit/rules/css/test.invalidNesting.js
+++ b/tests/unit/rules/css/test.invalidNesting.js
@@ -5,30 +5,33 @@ import { VALIDATION_WARNING } from 'const';
 import CSSScanner from 'scanners/css';
 
 describe('CSS Rule InvalidNesting', () => {
-  it('should detect invalid nesting', async () => {
-    const code = oneLine`/* I'm a comment */
+  it.each([60, 116])(
+    'should detect invalid nesting when firefoxStrictMinVersion=%d',
+    async (firefoxStrictMinVersion) => {
+      const code = oneLine`/* I'm a comment */
       #something {
         .bar {
           height: 100px;
         }
       }`;
-    const cssScanner = new CSSScanner(code, 'fakeFile.css', {
-      addonMetadata: {
-        firefoxStrictMinVersion: '60.5',
-      },
-    });
+      const cssScanner = new CSSScanner(code, 'fakeFile.css', {
+        addonMetadata: {
+          firefoxStrictMinVersion,
+        },
+      });
 
-    const { linterMessages } = await cssScanner.scan();
-    expect(linterMessages.length).toEqual(1);
-    expect(linterMessages[0].code).toEqual(
-      messages.INVALID_SELECTOR_NESTING.code
-    );
-    expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
-  });
+      const { linterMessages } = await cssScanner.scan();
+      expect(linterMessages.length).toEqual(1);
+      expect(linterMessages[0].code).toEqual(
+        messages.INVALID_SELECTOR_NESTING.code
+      );
+      expect(linterMessages[0].type).toEqual(VALIDATION_WARNING);
+    }
+  );
 
-  it.each(['117.0.0', '117.0.1'])(
-    'should not report invalid nesting when firefoxMinVersion=%s',
-    async (firefoxMinVersion) => {
+  it.each([117, 118])(
+    'should not report invalid nesting when firefoxStrictMinVersion=%d',
+    async (firefoxStrictMinVersion) => {
       const code = oneLine`/* I'm a comment */
       #something {
         .bar {
@@ -37,7 +40,7 @@ describe('CSS Rule InvalidNesting', () => {
       }`;
       const cssScanner = new CSSScanner(code, 'fakeFile.css', {
         addonMetadata: {
-          firefoxMinVersion,
+          firefoxStrictMinVersion,
         },
       });
       const { linterMessages } = await cssScanner.scan();
```

schmop: Oh, I didn't know, there already was a function to extract the major version! Sweet!

willdurand (approved):
Excellent, thanks!

Issue #5194, opened 2024-02-09
TypeError: policy.toLowerCase is not a function


diox:
We had this [traceback](https://mozilla.sentry.io/issues/4351561706/events/55b4edb122cb4fa29b6deab7ab58008d/) in production a few times, I suppose because of some malformed add-on manifest:

```
policy.toLowerCase is not a function

/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/utils.js:282
  const directives = policy.toLowerCase().split(';');
                            ^
TypeError: policy.toLowerCase is not a function
    at parseCspPolicy (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/utils.js:282:29)
    at ManifestJSONParser.validateCspPolicyString (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/parsers/manifestjson.js:1152:38)
    at ManifestJSONParser.validateCspPolicy (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/parsers/manifestjson.js:1143:14)
    at ManifestJSONParser._validate (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/parsers/manifestjson.js:532:12)
    at new ManifestJSONParser (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/parsers/manifestjson.js:143:12)
    at Linter.getAddonMetadata (/deps/node_modules/addons-linter/dist/webpack:/addons-linter/src/lint...
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-461)

PR #5195, opened 2024-02-12
Prevent errors when non-string CSP values are defined in the manifest

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5194

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-462)

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -232,6 +232,7 @@ export default class ManifestJSONParser extends JSONParser {
           );
         } else if (
           typeof manifestKeyValue === 'object' &&
+          manifestKeyValue !== null &&
           Object.prototype.hasOwnProperty.call(manifestKeyValue, subkey)
         ) {
           this.checkCompatInfo(
@@ -1149,6 +1150,10 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   validateCspPolicyString(policy, manifestPropName) {
+    if (typeof policy !== 'string') {
+      return;
+    }
+
     const directives = parseCspPolicy(policy);
 
     // The order is important here, 'default-src' needs to be before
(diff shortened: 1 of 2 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -2248,6 +2248,24 @@ describe('ManifestJSONParser', () => {
         expect(warningsV3.length).toEqual(6);
       }
     );
+
+    // See: https://github.com/mozilla/addons-linter/issues/5194
+    it('should handle non-string values', () => {
+      const addonLinter = new Linter({ _: ['bar'] });
+      const json = validManifestJSON({ content_security_policy: [true] });
```
rpl, line 2255: @willdurand would you mind to also add a test manifest with a non string csp value when content_security_policy is in the object form? (e.g. even just making this same test executed for both this value and `{ content_security_policy: { extension_pages: true } }`).

src/parsers/manifestjson.js
```diff
@@ -1149,6 +1149,10 @@ export default class ManifestJSONParser extends JSONParser {
   }
 
   validateCspPolicyString(policy, manifestPropName) {
+    if (typeof policy !== 'string') {
```
rpl, line 1152: @willdurand I wonder if we should also tweak the caller (`validateCspPolicy') to only iterate over `Object.keys(policy)` if policy is not null and not an array (given that for the other primitive types but array, Object.keys would already be an empty array and not doing anything in practice), the tweak in this PR should already prevent the kind of unexpected issues we may be hitting but in the end if the csp value is an array then it is already invalid on its own and validating each element of the array not strictly necessary anymore.

rpl (approved):
@willdurand the fix looks reasonable as is, the inline comments are related to a small addition to the test case and considering not validating the csp strings at all if the value of the content_security_policy field is an array (mainly to avoid to validate values that would be in an invalid form anyway).

src/parsers/manifestjson.js
willdurand: I added a test case to cover `null`.

PR #5218, opened 2024-03-12
Extract and merge locales (2024-03-12)

willdurand:


locale/templates/LC_MESSAGES/messages.pot
```diff
@@ -52,7 +52,7 @@ msgid "Invalid nesting of selectors found"
 msgstr ""
 
 #: src/messages/css.js:14
-msgid "Selectors should not be nested"
+msgid "Selector nesting is supported from firefox version 117.0 and above"
```
flodolo, line 55: Firefox should have been uppercase here.

willdurand: ha, thanks. I'll fix that

PR #5263, opened 2024-04-22
Binary scanner should be able to check string-like buffers

### LINKED ISSUE #5258: Files with a flagged extension are not considered binary files
STR:

1. lint an add-on with a large `.exe` file (> 4MB)

Expected:

Maybe a message about the file being flagged due to its file extension.

Actual:

An error is emitted about the file being too big, and not a binary file. This is not correct.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-525)

willdurand:
I noticed this while trying to fix #5258. ~~It looks like - on my machine - the buffer contains characters, not numbers. Therefore, the check against the magic bytes fails.~~

```
NOTICES:

Code                Message                   Description                                                                                 File        Line   Column
FLAGGED_FILE_TYPE   Flagged file type found   Files whose names end with flagged extensions have been found in the add-on. The            win32/adb
                                              extension of these files are flagged because they usually identify binary components.
                                              Please see https://bit.ly/review-policy for more information on the binary content review
                                              process.
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-530)

### DIFF
--- a/src/scanners/binary.js
+++ b/src/scanners/binary.js
@@ -3,7 +3,7 @@ import * as messages from 'messages';
 import * as constants from 'const';
 
 export default class BinaryScanner extends BaseScanner {
-  static get fileStreamType() {
+  static get fileResultType() {
     return 'chunk';
   }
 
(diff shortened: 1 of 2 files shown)


willdurand:
This was reverted because I messed up the commit title...

PR #5264, opened 2024-04-22
Revert "Binary scanner should be able to check string-like buffers"

willdurand:
Reverts mozilla/addons-linter#5263


willdurand:
See: https://github.com/mozilla/addons-linter/pull/5263#issuecomment-2070027358

Issue #5275, opened 2024-05-05
Rename the artificial messages API / messages.json to i18nMessages.json


jobisoft:
### Describe the problem and steps to reproduce it:

The `addons-linter` introduces an artificial / additional schema file `messages.json`.

Thunderbird is using the same WebExtension mechanism as Firefox, since we share a large part of its code base. We do however have our own set of WebExtension APIs, which are of no use for browser vendors and are only shipped with Thunderbird. When we created these so called "MailExtension APIs", we made sure we do not collide with existing Firefox APIs.

One of those MailExtension APIs is our "messages API", which allows extensions to access the user's emails/messages.

We are currently evaluating, if we can use this `addons-linter` project directly, without maintaining our own fork. We do that by providing a zip-file, which has all the relevant schema files for Thunderbird:

[thunderbird_beta_schemas.zip](https://github.com/mozilla/addons-linter/files/15213105/thunderbird_beta_schemas.zip)

To verify the HtmlSourceEditor MailExtension [HtmlSourceEditor.zip](https://github.com/mozilla/addons-linter/files/15213116/HtmlSourceEditor.zip), we do the following:

- clone https://github.com/mozilla/addons-linter
- run `npm install`
- run `npm run build`
- run `node bin/addons-linter HtmlSourceEditor.zip` (this fails with UNSUPPORTED_API)
- run `node scripts/firefox-schema-import thunderbird_beta_schemas.zip`
- run `npm run build`
- run `node bin/addons-linter HtmlSourceEditor.xpi` (this passes)

The provided `thunderbird_beta_schemas.zip` file - however - does not include all our schema definitions. The schema of the mentioned `messages API` had to be excluded. The reason for this is: The artificial / additional schema file `messages.json` uses the same namespace as our messages API, which we cannot change.

### What happened?

We get the following error: **schema with key or id "messages" already exists**. Linting fails.

### What did you expect to happen?

Linting does not fail.

### Anything else we should know?

This can be fixed, by renaming the artificial messages API created by the `addons-linter` to `i18nMessages`. This should not have any negative consequences, as that schema file does not describe a real API, but is only used by the `addons-linter`.

I will file a PR shortly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-541)

PR #5279, opened 2024-05-06
Rename the artificial messages API / messages.json to i18nMessages.json

jobisoft:
Fixes https://github.com/mozilla/addons-linter/issues/5275

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-545)

### DIFF
--- a/src/schema/validator.js
+++ b/src/schema/validator.js
@@ -5,7 +5,7 @@ import { getDefaultConfigValue } from 'yargs-options';
 import { deepPatch } from 'schema/deepmerge';
 import schemaObject from 'schema/imported/manifest';
 import themeSchemaObject from 'schema/imported/theme';
-import messagesSchemaObject from 'schema/messages';
+import messagesSchemaObject from 'schema/i18nMessages';
 import {
   DEPRECATED_MANIFEST_PROPERTIES,
   MANIFEST_VERSION_DEFAULT,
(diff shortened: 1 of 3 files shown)


src/schema/validator.js
```diff
@@ -380,7 +380,7 @@ export class SchemaValidator {
     if (!this._localeValidator) {
       this._localeValidator = this._validator.compile({
         ...this.messagesSchemaObject,
-        $id: 'messages',
+        $id: 'i18nMessages',
         $ref: '#/types/WebExtensionMessages',
```
Rob--W, line 384:
```suggestion
        $ref: '#/types/WebExtensionI18nMessages',
```

Let's also rename the type, as well as the definition of this ID in the (newly renamed) i18nMessages.json file. Note: the `I18n` is the uppercase letter I (before J), not lowercase L (before M).

src/schema/validator.js
```diff
@@ -5,7 +5,7 @@ import { getDefaultConfigValue } from 'yargs-options';
 import { deepPatch } from 'schema/deepmerge';
 import schemaObject from 'schema/imported/manifest';
 import themeSchemaObject from 'schema/imported/theme';
-import messagesSchemaObject from 'schema/messages';
+import messagesSchemaObject from 'schema/i18nMessages';
```
Rob--W, line 8: The usual convention for the file name is to use underscore instead of camelCase, e.g. `i18n_messages.json`.

src/schema/validator.js
jobisoft: Done.

src/schema/validator.js
jobisoft: Done.

Issue #5261, opened 2024-04-19
Suggest dropping `incognito: "split"` when present in manifest.json


Rob--W:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1876924 for context.

`incognito: "split"` is not supported in Firefox, only in Chrome.

In Firefox, `incognito: "spanning"` is the only supported value that enables users to opt in to allowing extensions access in private browsing mode. The only other supported value is `incognito: "not_allowed"`, which prevents users from using an extension in private browsing mode.

In bug 1876924, I'm patching Firefox to allow Firefox to load extensions even when `incognito: "split"` is used in the manifest, with a fallback to "not_allowed"`. Extensions that do not rely on Chrome's "split" mode should remove the key from the manifest (or set the value to `incognito: "spanning"`). This enables users to use the extension in private browsing mode, if they opt in to it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-528)

ioanarusiczki:
I tried with "incognito":"split"

On AMO dev with linter 6.29.0

![warning on AMO dev](https://github.com/mozilla/addons-linter/assets/33448286/9068b97d-4fb6-4205-a503-3fe8f9bf70e6)

With linter 6.28.0

![error with amo stage](https://github.com/mozilla/addons-linter/assets/33448286/bd54a8b7-9df6-4290-9b20-4cea8a30fa59)



The [other 2 allowed values](https://developer.chrome.com/docs/extensions/reference/manifest#incognito) -> no problems , extension passed validation. 

With other values than allowed such as "incognito":"incognito" -> error "/incognito" must be equal to one of the allowed values

PR #5316, opened 2024-05-31
Suggest dropping incognito:"split" when present in manifest.json

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5261

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -829,6 +829,6 @@ export const INCOGNITO_SPLIT_UNSUPPORTED = {
   message: i18n._('incognito "split" is unsupported.'),
   description: i18n._(`The incognito "split" value is unsupported and will be
     treated as "not_allowed" in Firefox. Remove the key from the manifest
-    file.`),
+    file, if your extension is compatible with Firefox's private browsing mode.`),
   file: MANIFEST_JSON,
 };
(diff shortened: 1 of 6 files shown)


src/messages/manifestjson.js
```diff
@@ -823,3 +823,12 @@ export const VERSION_FORMAT_INVALID = {
     https://mzl.la/3h3mCRu (MDN Docs) for more information.`),
   file: MANIFEST_JSON,
 };
+
+export const INCOGNITO_SPLIT_UNSUPPORTED = {
+  code: 'INCOGNITO_SPLIT_UNSUPPORTED',
+  message: i18n._('incognito "split" is unsupported.'),
+  description: i18n._(`The incognito "split" value is unsupported and will be
+    treated as "not_allowed" in Firefox. Remove the key from the manifest
+    file.`),
```
Rob--W, line 832: Ee should not just recommend removal, but also point to documentation to explain how to develop the extension in a privacy-friendly way.

Since we don't have that yet (https://github.com/mdn/content/pull/33661/files#r1606120565), the next best alternative is to add a condition,


```suggestion
    file, if your extension is compatible with Firefox's private browsing mode.`),
```

(And maybe add a link to the incognito documentation? Any link is better than none?)

Rob--W (approved):

PR #5331, opened 2024-06-06
Prevent the use of the enterprise flag

### LINKED ISSUE #5330: Prevent the use of the enterprise flag
We want to prevent the use of the enterprise flag until we are ready to accept these enterprise add-ons on AMO.

willdurand:
Fixes #5330

### DIFF
--- a/src/messages/manifestjson.js
+++ b/src/messages/manifestjson.js
@@ -832,3 +832,11 @@ export const INCOGNITO_SPLIT_UNSUPPORTED = {
     file, if your extension is compatible with Firefox's private browsing mode.`),
   file: MANIFEST_JSON,
 };
+
+export const ADMIN_INSTALL_ONLY_PROP_RESERVED = {
+  code: 'ADMIN_INSTALL_ONLY_PROP_RESERVED',
+  message: i18n._('The "admin_install_only" property is reserved.'),
+  description: i18n._(`The "admin_install_only" property is reserved and can
+    only be used in enterprise add-ons.`),
+  file: MANIFEST_JSON,
+};
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -5478,4 +5478,90 @@ describe('ManifestJSONParser', () => {
       ]);
     });
   });
+
+  describe('admin_install_only', () => {
+    it.each([
+      {
+        manifest_version: 2,
+        applications: { gecko: { id: '@test-id', admin_install_only: true } },
+      },
+      {
+        manifest_version: 2,
+        applications: { gecko: { id: '@test-id', admin_install_only: false } },
+      },
+      {
+        manifest_version: 2,
+        browser_specific_settings: {
+          gecko: { id: '@test-id', admin_install_only: true },
+        },
+      },
+      {
+        manifest_version: 2,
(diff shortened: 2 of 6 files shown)


src/messages/manifestjson.js
```diff
@@ -832,3 +832,11 @@ export const INCOGNITO_SPLIT_UNSUPPORTED = {
     file, if your extension is compatible with Firefox's private browsing mode.`),
   file: MANIFEST_JSON,
 };
+
+export const ADMIN_INSTALL_ONLY_PROP_RESERVED = {
+  code: 'ADMIN_INSTALL_ONLY_PROP_RESERVED',
+  message: i18n._('The "admin_install_only" property is reserved.'),
+  description: i18n._(`The "admin_install_only" property is reserved and can
+    only be used in enterprise add-ons.`),
```
rpl, line 840: Nit: how about tweaking the description to state something like `property reserved for add-ons only distribuited by admins through enterprise policies`?

tests/unit/parsers/test.manifestjson.js
```diff
@@ -5478,4 +5478,40 @@ describe('ManifestJSONParser', () => {
       ]);
     });
   });
+
+  describe('admin_install_only', () => {
+    it.each([
+      { applications: { gecko: { admin_install_only: true } } },
+      { applications: { gecko: { admin_install_only: false } } },
+      { browser_specific_settings: { gecko: { admin_install_only: true } } },
+      { browser_specific_settings: { gecko: { admin_install_only: false } } },
```
rpl, line 5487: Thought: how about including an additional one where we do have both gecko and gecko_android properties, strict_min_version/string_max_version properties in the gecko_android one and admin_install_only in the gecko property?

just to make sure that if we would be tweaking the implementation we couldn't mistakenly look for the admin_install_only property from gecko_android and instead we still hit the expected linting error.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -5478,4 +5478,40 @@ describe('ManifestJSONParser', () => {
...
+    ])(
+      'emits an error when the admin_install_only flag is present - %o',
+      (manifestProps) => {
+        const linter = new Linter({ _: ['bar'] });
+
+        const manifestJSONParser = new ManifestJSONParser(
+          JSON.stringify({
+            manifest_version: 2,
```
rpl, line 5495: Thought: similarly I wonder if here we should test it out of both mv2 and mv3 extensions, just to make sure that if we tweak the implementation we can't mistakenly end up only checking this property for a particular manifest version and not for the other one.

rpl (approved):
@willdurand this PR looks pretty good and straightforward, I only collected one nit (related to some small proposed tweaks to the rule description, not really mandatory and we can always change the message later if we think it would be better to reword it) and a couple of thoughts on the test side (but mostly nice to have to ensure that if we change the implementation we can't mistakenly not report the error under certain corner cases).

Given that nothing collected in this review pass is that critical or blocking, I'm also approving this right away.

src/messages/manifestjson.js
willdurand: I mentioned "enterprise add-ons" because that's what AMO will use. I am going to keep the current message for now.

PR #5340, opened 2024-06-13
Introduce a --enterprise flag

willdurand:
Add a new CLI flag to control the checks for enterprise add-ons.


src/messages/manifestjson.js
```diff
@@ -840,3 +840,12 @@ export const ADMIN_INSTALL_ONLY_PROP_RESERVED = {
     only be used in enterprise add-ons.`),
   file: MANIFEST_JSON,
 };
+
+export const ADMIN_INSTALL_ONLY_REQUIRED = {
+  code: 'ADMIN_INSTALL_ONLY_REQUIRED',
+  message: i18n._('The "admin_install_only" property must be set to "true".'),
+  message: i18n._('The "admin_install_only" property is reserved.'),
```
Rob--W: Is this the right string?

src/parsers/manifestjson.js
```diff
@@ -524,8 +525,13 @@ export default class ManifestJSONParser extends JSONParser {
       typeof this.parsedJSON.applications?.gecko?.admin_install_only !==
       'undefined'
     ) {
-      this.collector.addError(messages.ADMIN_INSTALL_ONLY_PROP_RESERVED);
-      this.isValid = false;
+      if (!this.isEnterpriseAddon) {
+        this.collector.addError(messages.ADMIN_INSTALL_ONLY_PROP_RESERVED);
+        this.isValid = false;
+      } else if (!this.parsedJSON.applications?.gecko?.admin_install_only) {
```
Rob--W, line 531: This block must be outdented by one level and combined with the flag.

willdurand: what do you mean?

rpl (changes requested):
@willdurand looks pretty good to me, I have just one thought (well, maybe two packed into one :-P) about explicitly state in this new ADMIN_INSTALL_ONLY_REQUIRED linting error where the property is supposed to be set (mainly for when the error is reported for the property being required but not set at all, or set in the wrong part of the manifest).

Marking as request changes to signal this is currently pending on double-checking if that thought is something we should be covering in this PR before merging it.

src/messages/manifestjson.js
```diff
@@ -840,3 +840,11 @@ export const ADMIN_INSTALL_ONLY_PROP_RESERVED = {
...
   file: MANIFEST_JSON,
 };
+
+export const ADMIN_INSTALL_ONLY_REQUIRED = {
+  code: 'ADMIN_INSTALL_ONLY_REQUIRED',
+  message: i18n._('The "admin_install_only" property must be set to "true".'),
+  description: i18n._(`The "admin_install_only" property must be set to "true"
+    in an enterprise add-on.`),
+  file: MANIFEST_JSON,
+};
```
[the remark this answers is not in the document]
willdurand: I restricted the use of the flag to `bss.gecko.admin_install_only` and updated the message accordingly. We still verify the absence of the flag in `bss` and `applications` when `--enterprise` is _not_ passed.

We have a warning that indicates that `applications` is ignored when `bss` is defined already, so I didn't do anything here.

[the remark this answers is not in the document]
rpl: :+1: sgtm, that feels a reasonable balance.

As a side note: there is still an unfortunate corner case where the combined linting error messages reported may be slightly confusing, when the schema based validation may state that applications.gecko.admin_install_only has to be a boolean along with our new linting error added in this PR saying that browser_specific_settings.gecko.admin_install_only is required to be set to true.

In theory we could consider to tweak the error as reported from the schema validation from the ManifestJSONParser errorLookup method, but honestly I personally feel like it may not be worth to make this more complicate for that kind of corner case and so I'd **vote to leave the error collected by the schema based validation as is** at least for now.

To corner case can basically be hit with a manifest like the following:

```
{
  ...
    "gecko": {
      "id": "enterprise-addon@test"
    }
  },
  "applications": {
    "gecko": {
      "admin_install_only": "test"
    }
  }
}
```

The reported error will state:
- ```MANIFEST_FIELD_INVALID        "/applications/gecko/admin_install_only" must be boolean```
- ```ADMIN_INSTALL_ONLY_REQUIRED   "/browser_specific_settings/gecko/admin_install_only" property must be set to true```

rpl (approved):

src/messages/manifestjson.js
```diff
@@ -840,3 +840,12 @@ export const ADMIN_INSTALL_ONLY_PROP_RESERVED = {
     only be used in enterprise add-ons.`),
   file: MANIFEST_JSON,
 };
+
+export const ADMIN_INSTALL_ONLY_REQUIRED = {
+  code: 'ADMIN_INSTALL_ONLY_REQUIRED',
+  message: i18n._(`"/browser_specific_settings/gecko/admin_install_only"
+    property must be set to "true"."`),
```
flodolo, line 847: There's an extra " at the end

willdurand: https://github.com/mozilla/addons-linter/pull/5347

PR #5343, opened 2024-06-14
Remove --scan-file option

### LINKED ISSUE #5319: --scan-file Produces Different Linter Results for a File Compared to When the Entire Add-on is Linted

While working with the API, I noticed that `scanFile` caused addons-linter to produce different errors when `manifest.json `specifically was singled out. I tested the same scenario with the command line `--scan-file`, which returned the same unexpected results. 

The linter was run on [this](https://addons-dev.allizom.org/en-US/firefox/addon/adblock-plus/) add-on.

### What happened?

addons-linter produces `FILE_NOT_FOUND `errors for `manifest.json` when passed through `--scan-file` that are not produced when the entire add-on is scanned.

The same linting results should appear.

API results for `manifest.json` when entire extension folder is passed:
image

API results for `manifest.json` when linter option `scanFile` is set to ['manifest.json']:
image

[Command-line Result](https://github.com/user-attachments/files/15566493/Terminal.Saved.Output.txt)

willdurand:
Kinda fixes #5319

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,6 @@ const linter = linter.createInstance({
     output: 'none',
     boring: false,
     selfHosted: false,
-    // Lint only the selected files
-    //   scanFile: ['path/...', ...]
-    //
     // Exclude files:
     shouldScanFile: (fileName) => true,
   },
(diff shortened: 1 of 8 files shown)


rpl (approved):
@willdurand lgtm. Thanks a lot for taking the time to track down all the changes on implementation and test side that used to only be there for the "not really used" ("not that much officially supported") --scan-file option.

PR #5344, opened 2024-06-17
Enterprise add-ons should be treated as self-hosted add-ons

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5345

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -95,6 +95,7 @@ export default class ManifestJSONParser extends JSONParser {
       schemaValidatorOptions,
       io = null,
       isAlreadySigned = false,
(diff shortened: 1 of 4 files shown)


src/parsers/manifestjson.js
```diff
@@ -138,7 +142,7 @@ export default class ManifestJSONParser extends JSONParser {
 
       this.io = io;
       this.isAlreadySigned = isAlreadySigned;
-      this.isEnterpriseAddon = this.schemaValidatorOptions?.enterprise ?? false;
+      this.isEnterpriseAddon = isEnterprise;
```
willdurand, line 145: I noticed we didn't have to use the `schemaValidatorOptions` to pass this prop here.

Rob--W (approved):

PR #5356, opened 2024-06-18
Emit an error when an enterprise add-on defines an `update_url` in the manifest

### LINKED ISSUE #5355: Enterprise add-ons should not be able to define an `update_url` in the manifest
cf. title

willdurand:
Fixes #5355

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -2361,6 +2361,57 @@ describe('ManifestJSONParser', () => {
       expect(manifestJSONParser.isValid).toEqual(true);
       expect(addonLinter.collector.warnings.length).toBe(0);
     });
+
+    it('emits an error when the update_url is defined in an enterprise add-on', () => {
+      const linter = new Linter({ _: ['bar'] });
+
+      const manifestJSONParser = new ManifestJSONParser(
+        JSON.stringify({
+          manifest_version: 2,
+          name: 'some enterprise add-on',
+          version: '1',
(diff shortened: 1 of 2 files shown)


Rob--W (approved):

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2361,6 +2361,57 @@ describe('ManifestJSONParser', () => {
...
+          browser_specific_settings: {
+            gecko: {
+              id: '@test-id',
+              admin_install_only: true,
+              update_url: 'https://example.org',
+            },
+          },
+        }),
+        linter.collector,
+        { isEnterprise: true }
```
rpl, line 2382: Nit, the implementation also make sure that if the developer uses the linter locally and by mistake pass both --enterprise and --self-hosted, we still show the expected `messages.MANIFEST_UPDATE_URL`, it is not really mandatory (just a nice to have on the test coverage side) but we could consider to cover that explicitly by adapting this test case to run with this parameter set to `{ isEnterprise: true }` and then also with `{ isEnterprise: true, selfHosted: true }`.

rpl (approved):
lgtm, I added a comment below but it is just an optional nit (a "nice to have" addition on the test coverage side).

tests/unit/parsers/test.manifestjson.js
Rob--W: FYI: --enterprise implies --self-hosted due to https://github.com/mozilla/addons-linter/pull/5344

PR #5402, opened 2024-08-09
chore(ci): Import Firefox schema github workflow

### LINKED ISSUE #5388: Create Github action to automatically import schemas from Firefox
Our most recent schema imports (#5385, #5386) were delayed because these are manual tasks and it fell through the cracks while we were busy.

We should automatically create the issue and PR on a schedule, e.g. via Github actions' `schedule`: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

It is possible to automatically determine the upcoming branch/release dates via: https://whattrainisitnow.com/

We could run an import 1-2 week after a release reaches Beta.

rpl:
This pull request defines a new github actions workflow which run the Firefox schema import and optionally creates an issue and a pull request.

The workflow can be called manually or being called from another workflow (so that we can define a workflow that runs the schema import and creates issue and pull request on a schedule, but also allow us to run it manually if we need to).

Fixed #5388

TODO list:
- [x] use github API to detect changes applied to the target import branch
- [x] Look into the circle-ci failure (it is actually unrelated to the changes in this PR, it is due to recent changes on the web-ext side, which requires nodejs >= 18, separate PR #5418 is meant to deal with it)
- [ ] create a workflow to run beta schema import automatically on a weekly basis
- [ ] create a workflow to test for breakage central schema import automatically on a weekly basis (and file issue if import or tests are broken)

### DIFF
--- a/.github/workflows/firefox-schema-import.yml
+++ b/.github/workflows/firefox-schema-import.yml
@@ -1,0 +1,140 @@
+name: Import Firefox API Schema
+
+run-name: |
+  ref: ${{ github.ref_name }}
+  github-dev-branch: ${{ inputs.gecko-dev-branch }}
+
+concurrency:
+  group: firefox-schema-import-${{ inputs.gecko-dev-branch }}
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+    inputs:
+      gecko-dev-branch:
+        type: string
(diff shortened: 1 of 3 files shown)


rpl:
We may still want to automate part of the schema import process, but we agreed to look into some simpler approach than this one.

Issue #5407, opened 2024-08-17
Move `glob` dependency to devDependencies 


fregante:
It looks like `glob` is used only by build scripts but it’s set as a non-dev dependency:

https://github.com/search?q=repo%3Amozilla%2Faddons-linter+require%28%27glob%27%29&type=code

Should it be moved to devDependencies? I’ll send a PR this week, checking if any other dependencies are in this situation. 

Related:

- https://github.com/mozilla/web-ext/issues/3216

PR #5412, opened 2024-08-22
Drop unused dependencies

fregante:
- Closes https://github.com/mozilla/addons-linter/issues/5407

### Dropped

- [x] is-mergeable-object

### Moved to devDependencies

- [x] glob
- [x] tosource


fregante:
Weird, why would it fail with `You did not build addons-linter yet.` if the build ran without failing?

package.json
```diff
@@ -60,18 +60,14 @@
     "espree": "10.1.0",
     "esprima": "4.0.1",
     "fast-json-patch": "3.1.1",
-    "glob": "11.0.0",
     "image-size": "1.1.1",
-    "is-mergeable-object": "1.1.1",
```
fregante, line 65: This is the only dropped dependency

willdurand (approved):

PR #5414, opened 2024-08-22
Cleanup babel config

fregante:
- Drops unused dependencies
- Cleans babel config of unused transforms
- Renames config file to new name (old one is [sometimes](https://github.com/mozilla/addons-linter/pull/5415) ignored)


willdurand:
This looks reasonable to me, @rpl any thoughts?

willdurand (approved):

PR #5415, opened 2024-08-23
Bump os-locale (add support for ESM dependencies)

fregante:
- Picks up https://github.com/mozilla/addons-linter/pull/4989
- Partially includes https://github.com/mozilla/addons-linter/pull/5414
  - because Jest requires it to actually read `transformIgnorePatterns` correctly

`os-locale` v6 has huge savings

## v5

Screenshot 7

## v6

Screenshot 6


babel.config.json
```diff
@@ -0,0 +1,10 @@
+{
```
fregante, line 1: - Merge https://github.com/mozilla/addons-linter/pull/5414 first

jest.config.js
```diff
@@ -27,7 +28,7 @@ module.exports = {
     '^.+\\.js$': 'babel-jest',
     '^.+\\.txt$': '<rootDir>/tests/jest-raw-loader.js',
   },
-  transformIgnorePatterns: ['<rootDir>/node_modules/'],
+  transformIgnorePatterns: ['<rootDir>/node_modules/(?!os-locale)'],
```
fregante, line 31: Standard config for ESM packages

webpack.config.js
```diff
@@ -6,14 +6,19 @@ const webpack = require('webpack');
 // eslint-disable-next-line import/no-extraneous-dependencies
 const nodeExternals = require('webpack-node-externals');
 
+// ESM packages must be bundled because webpack will otherwise turn `require()` into `import()`
+// and make the import silently asynchronous. In most cases this breaks the import.
+const dependenciesToBundle = [
+  'os-locale', // Used exclusively in sync functions
+];
```
fregante, line 13: As expected, there's "no way" to `require` a ESM package synchronously. This is not a webpack/config issue. The only way for this to work is to turn `addons-linter` into an `type=module` package.

So this is the alternative: _do not_ keep `os-locale` as an external dependency, so that node does not try to `require` it at runtime.

webpack.config.js
```diff
@@ -6,14 +6,19 @@ const webpack = require('webpack');
...
+
 module.exports = {
   // Set the webpack4 mode 'none' for compatibility with the behavior of the
   // webpack3 bundling step.
   mode: 'none',
   entry: {
     'addons-linter': './src/main.js',
   },
-  target: 'node',
```
fregante, line 16: Webpack defaults to browserlist, if set. We have it already and it's set to `node 16`. This enables ESM support

fregante:
Ping

willdurand:
@fregante sorry to let you wait on this. The fact that we're essentially adding ESM support here is a bit scary to me (not necessarily a bad thing but not as straightforward as a simple version bump). We'll be looking into converting addons-linter to ESM before the end of the year (though that might not happen for various reasons...). In the meantime, we should probably look at this PR.

fregante:
"Adding ESM support" means bundling those ESM dependencies, with an explicit list of dependencies to bundle. No ESM dependency will actually be exposed to npm users. 
 
Once you do publish ESM, you can just drop/revert that part of the config and it will continue working.

willdurand:
Yeah, indeed.

Rob--W:
Thanks for looking into this @fregante . I'm closing this PR because rather than updating the dependency, we decided to drop the dependency altogether in https://github.com/mozilla/addons-linter/pull/5520, because the feature that depended on it was not used (#5521).

fregante:
What about the other ESM dependencies? The setup can work for other dependencies.

PR #5417, opened 2024-08-23
Drop sha.js dependency

fregante:


package.json
```diff
@@ -69,7 +69,6 @@
     "pino": "8.20.0",
     "relaxed-json": "1.0.3",
     "semver": "7.6.3",
-    "sha.js": "2.4.11",
```
fregante, line 72: From: https://github.com/mozilla/dispensary

Not sure why this was ever used, `crypto.createHash` has been here since Node 0.1.x

Rob--W (approved):

Issue #3910, opened 2021-09-13
Replace `relaxed-json` with `strip-json-comments`


fregante:
The [MDN documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json) mentions that:

> [manifest.json] is a JSON-formatted file, with one exception: it is allowed to contain "//"-style comments.

But RJSON is a bit more than that: https://github.com/phadej/relaxed-json#relaxed-json

Given this and that `relaxed-json` has [a few CLI-related dependencies](https://npmgraph.js.org/?q=relaxed-json@1.0.3), I'd suggest swapping it out for the more appropriate [strip-json-comments](https://github.com/sindresorhus/strip-json-comments), which has no dependencies.

Related:

- https://github.com/mozilla/addons-linter/issues/3601

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-19)

fregante:
In general, I'm a bit confused about this file:

https://github.com/mozilla/addons-linter/blob/872132e9661ea02423b1226a8926d1c9ce890810/src/parsers/json.js#L30

Why is esprima being used before RJSON at all? It seems that the whole `.parse` method should just be a plain call to `parse: json => JSON.parse(stripJsonComments(json))`

fregante:
Side note: both `espree` and `esprima` are used in this project, so `esprima` should probably be dropped anyway (because eslint still loads `espree` )

Rob--W:
RelaxedJSON is not used for parsing, but only as an attempt to get useful error messages. We can use `JSON.parse`, and if that fails, strip the comments and retry, and if that fails, try to extract a useful message from the original error (with line numbers for example).

For manifest.json and `_locales/<locale>/messages.json`, replacing the overly broad RJSON with stripJsonComments+JSON.parse should be safe, because Firefox itself already strips comments (without allowing a broader format): https://searchfox.org/mozilla-central/rev/4d90ff4537330d6b17cb956c0fadf759086d9bb7/toolkit/components/extensions/Extension.jsm#268-281

The linter also lints arbitrary JSON files:
- JSONParser uses RelaxedJSON: https://github.com/mozilla/addons-linter/blob/872132e9661ea02423b1226a8926d1c9ce890810/src/parsers/json.js#L6
[...]

PR #5420, opened 2024-08-25
Trim `relaxed-json` dependencies

fregante:
- Closes https://github.com/mozilla/addons-linter/issues/3910

As mentioned in https://github.com/mozilla/addons-linter/issues/3910#issuecomment-2308814995, `relaxed-json` has duplicate key detection so it cannot easily be replaced.

And we don't have to. The only problem with the package is that it has a lot of CLI dependencies.

Since the package has been stable/undeveloped for 5 years, I just forked it to publish an API-only version, documenting it: https://github.com/fregante/relaxed-json#readme. This was brought up to its developer [3 years ago](https://github.com/phadej/relaxed-json/issues/24) already but no response.

This drops _all_ of its deps:

Screenshot 1


fregante:
Failing CI is unrelated:

- https://github.com/mozilla/addons-linter/issues/5421

willdurand:
@fregante can you rebase please?

PR #5418, opened 2024-08-23
chore(circleci): bump nodejs version to from v16 to v18

rpl:
This PR bumps the nodejs versions used by the circleci jobs:

- test: to v18 (from v16)
- test-next: to v20 (from v18)
- test-alternate: to v22 (from v19)

Fixes test job issue while running recent web-ext tests on a recent unreleased web-ext version where nodejs >= 18 is required (e.g. see [example failure here](https://app.circleci.com/pipelines/github/rpl/addons-linter/39/workflows/463e9605-ba9a-46ee-bb48-0e73fc76f8e8/jobs/129)), introduced as a side effect of the changes from mozilla/web-ext#3217.

PR #5422, opened 2024-08-25
Actually use the calculated window width

### LINKED ISSUE #5419: The report table is always wider than terminal

Run the basic `addons-linter folder/` on an extension with warnings

### What happened?

The calculated window with is always ~15 tighter than it is. I'm using a mostly unmodified Terminal.app with Fish Shell. The same happens with `zsh`. Also same issue when used in VS Code's terminal windows.

`web-ext lint` is also affected.

### Demo

https://github.com/user-attachments/assets/3e765e7c-a098-4301-aaa8-0a05915ef3cd

fregante:
- Fixes #5419 

In my tests:

- real: 100 ch
- `maxColumns`: 98 ch
- printed: 122 ch

I don't know exactly why `columnify` is returning a line that is that long, but this fixes it.

### Demo

demo


fregante:
Failing CI is unrelated:

- https://github.com/mozilla/addons-linter/pull/5418

willdurand:
Thanks!

PR #5426, opened 2024-08-26
Output list instead of table

### LINKED ISSUE #5423: Output as list, not table

I'm not a fan of the table output, it's repetitive and it breaks file names on small terminals.

### Current

```
Code                              Message           Description                      File      
MANIFEST_FIELD_UNSUPPORTED        "/background/s…   "/background/service_worker"     ma
                                  ervice_worker"    is not supported.                          
                                  is not                                                       
                                  supported.                                                   
KEY_FIREFOX_UNSUPPORTED_BY_MIN…   Manifest key      "strict_min_version" requires    manife
_VERSION                          not supported     Firefox 126, which was                 
                                  by the            released before version 128                
                                  specified         introduced support for                     
                                  minimum Firefox   "optional_host_permissions".               
                                  version                                                      
KEY_FIREFOX_ANDROID_UNSUPPORTE…   Manifest key      "strict_min_version" requires    manife
D_BY_MIN_VERSION                  not supported     Firefox for Android 126, which         
                                  by the            was released before version                
                                  specified         128 introduced support for                 
                                  minimum Firefox   "optional_host_permissions".               
                                  for Android                                                  
                                  version                                                      
UNSAFE_VAR_ASSIGNMENT             Unsafe            Due to both security and         as
                                  assignment to     performance concerns, this may   dules/dom-
                                  innerHTML         not be set using dynamic         /index.js 
                                                    values which have not been                 
                                                    adequately sanitized. This can             
                                                    lead to security issues or                 
                                                    fairly serious performance                 
                                                    degradation.                               
UNSAFE_VAR_ASSIGNMENT             Unsafe            Due to both security and         as
                                  assignment to     performance concerns, this may   dules/doma
                                  innerHTML         not be set using dynamic         ex.js     
                                                    values which have not been                 
                                                    adequately sanitized. This can             
                                                    lead to security issues or                 
                                                    fairly serious performance                 
                                                    degradation.                               
UNSAFE_VAR_ASSIGNMENT             Unsafe call to    Due to both security and         as
                                  document.creat…   performance concerns, this may   dules/link
                                  eRange().creat…   not be set using dynamic         issues/ind
                                  eContextualFra…   values which have not been                 
                                  gment for         adequately sanitized. This can             
                                  argument 0        lead to security issues or                 
                                                    fairly serious performance                 
                                                    degradation.                               
UNSAFE_VAR_ASSIGNMENT             Unsafe call to    Due to both security and         as
                                  document.creat…   performance concerns, this may   dules/link
                                  eRange().creat…   not be set using dynamic         urls/index
                                  eContextualFra…   values which have not been                 
                                  gment for         adequately sanitized. This can             
                                  argument 0        lead to security issues or                 
                                                    fairly serious performance                 
                                                    degradation.                    
```

### Ideal

```
MANIFEST_FIELD_UNSUPPORTED
  "/background/service_worker" is not supported.
  "/background/service_worker" is not supported.
  manifest.json

KEY_FIREFOX_UNSUPPORTED_BY_MIN_VERSION
  Manifest key not supported by the specified minimum Firefox version
  "strict_min_version" requires Firefox 126, which was released before version 128 introduced support for "optional_host_permissions".
  manifest.json

KEY_FIREFOX_ANDROID_UNSUPPORTED_BY_MIN_VERSION
  Manifest key not supported by the specified minimum Firefox for Android version
  "strict_min_version" requires Firefox for Android 126, which was released before version 128 introduced support for "optional_host_permissions".
  manifest.json

UNSAFE_VAR_ASSIGNMENT
  Unsafe assignment to innerHTML
  Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
  assets/node_modules/dom-chef/index.js:96:13
  assets/node_modules/doma/index.js:6:5
  assets/node_modules/linkify-issues/index.js:20:24
  assets/node_modules/linkify-urls/index.js:15:24
```

fregante:
- Fixes #5423 

## Before

This isn't even the full table, it doesn't fit on my screen

Screenshot 3

## After

Screenshot 2

### DIFF
--- a/src/linter.js
+++ b/src/linter.js
@@ -180,11 +180,13 @@ export default class Linter {
           out.push(this.colorize(type)(`  ${code}`));
           out.push(`  ${list[0].message}`);
           list.forEach((message) => {
-            const location = message.file
-              ? `${message.file}${message.line ? `:${message.line}` : ''}${
-                  message.column ? `:${message.column}` : ''
-                }`
-              : 'N/A';
+            let location = message.file || 'N/A';
+            if (message.file && message.line) {
+              location += `:${message.line}`;
+              if (message.column) {
+                location += `:${message.column}`;
+              }
+            }
             out.push(`    ${location}`);
           });
           out.push('');
--- a/src/linter.js
+++ b/src/linter.js
@@ -178,7 +178,11 @@ export default class Linter {
 
         Object.entries(groupedMessages).forEach(([code, list]) => {
           out.push(this.colorize(type)(`  ${code}`));
-          out.push(`  ${list[0].message}`);
+          const parts = [list[0].message];
+          if (list[0].message !== list[0].description) {
+            parts.push(list[0].description);
+          }
+          out.push(`  ${parts.join(' ')}`);
           list.forEach((message) => {
             let location = message.file || 'N/A';
             if (message.file && message.line) {
--- a/src/linter.js
+++ b/src/linter.js
@@ -1,6 +1,5 @@
 import path from 'path';
 
-import columnify from 'columnify';
 import chalk from 'chalk';
 import { oneLine } from 'common-tags';
 import { lstat } from 'addons-scanner-utils/dist/io/utils';
@@ -10,7 +9,6 @@ import {
 } from 'addons-scanner-utils/dist/errors';
 import { Directory, Xpi, Crx } from 'addons-scanner-utils/dist/io';
 
-import { terminalWidth } from 'cli';
 import * as constants from 'const';
 import { BANNED_LIBRARIES, UNADVISED_LIBRARIES } from 'libraries';
 import * as messages from 'messages';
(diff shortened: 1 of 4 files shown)


src/linter.js
```diff
@@ -154,104 +152,43 @@ export default class Linter {
...
+          list.forEach((message) => {
+            const location = message.file
+              ? `${message.file}${message.line ? `:${message.line}` : ''}${
+                  message.column ? `:${message.column}` : ''
+                }`
+              : 'N/A';
+            out.push(`    ${location}`);
+          });
+          out.push('');
+        });
```
fregante, line 191: @willdurand @rpl This code is straight out of Claude. We can bikeshed the output but I think it's a vast improvement already.

Codebase cleanup and tests update will follow after approval

willdurand: It would have been good to discuss in https://github.com/mozilla/addons-linter/issues/5423 first. I don't really have a strong opinion on the current linter output but - at the same time - this works relatively well and that isn't necessarily an area where I'd change code, but that's my personal opinion on this.

fregante: I opened a PR right after the issue in order to demo the output and changes required. It's always easier to discuss things when there's an exact desired output, especially in a project that evolves very slowly.

fregante:
@rpl @willdurand I believe the improvements here are evident from my screenshots:

- the output is shorter
- there's no duplicate explanation
  Screenshot 3
- there's no bad wrapping of long text
  Screenshot 2
- paths are always clickable
  Screenshot 4
- row and col are part of the path, meaning IDEs can reach them in the same click
  Screenshot
- the table implementation isn't even right because the error/warning tables are not aligned
  Screenshot

The only reason not to move forward with this is because it already works, albeit in this suboptimal fashion. Since I'm currently available to work on this, allow me to complete this change and merge this PR. I don’t really like keeping PRs open for years

src/linter.js
```diff
@@ -154,104 +152,43 @@ export default class Linter {
...
+
+        Object.entries(groupedMessages).forEach(([code, list]) => {
+          out.push(this.colorize(type)(`  ${code}`));
+          out.push(`  ${list[0].message}`);
+          list.forEach((message) => {
+            const location = message.file
+              ? `${message.file}${message.line ? `:${message.line}` : ''}${
+                  message.column ? `:${message.column}` : ''
+                }`
+              : 'N/A';
```
Rob--W, line 187: Could you split this string construction in separate variables, so that there are no three ternary expressions embedded in one?

For example something like the following (untested, but just to show how it could be unrolled into something that is a bit more readable):

```suggestion
            let location = message.file || 'N/A';
            if (message.file && message.line) {
              location += `:${message.line}`;
              if (message.column) {
                location += `:${message.column}`; 
              }
            }
```

Rob--W (changes requested):
Thanks for improving the readability and dropping one dependency. We're willing to switch the output format.

Could you adjust the text and update the tests so that they pass?

Thanks for this, and apologies for the review delay!

fregante:
Most likely not intentional, it should appear just after the message as in the original order. In some cases it does seem to be identical to the message so it should likely be dropped there instead of repeating it.

PR #5520, opened 2024-12-04
Remove localization in this project

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5521


tests/unit/test.utils.js
```diff
@@ -79,25 +78,6 @@ describe('i18n._()', () => {
 describe('gettext()', () => {
   it('should return localizable message', () => {
     expect(i18n.gettext('This is a test')).toEqual('This is a test');
```
rpl, line 80: optional nit: it looks like i18n.gettext is only being used in the test.utils.js tests now, given that it is an exact duplicate of i18n._, we could file it as a followup or get rid of it in this PR with a couple more changes:

```diff
diff --git a/src/utils.js b/src/utils.js
index 55249e51..33df1b43 100644
--- a/src/utils.js
+++ b/src/utils.js
@@ -143,7 +143,6 @@ function oneLineTranslationString(translationKey) {
 // re-enable localization of the project in the future.
 export const i18n = Object.freeze({
   _: (str) => oneLineTranslationString(str),
-  gettext: (str) => oneLineTranslationString(str),
   sprintf: (fmt, args) => {
     let str = fmt;
     // Replace every `%(key)s` or `%(key)d` placeholder with the actual value.
diff --git a/tests/unit/test.utils.js b/tests/unit/test.utils.js
index 6e5893e2..cd946321 100644
--- a/tests/unit/test.utils.js
+++ b/tests/unit/test.utils.js
@@ -63,26 +63,13 @@ describe('getRootExpression()', () => {
 });
 
 describe('i18n._()', () => {
-  it('should return one-line strings', () => {
-    expect(
-      i18n._(
-        `This
-         is
-         a
-         test`
-      )
-    ).toEqual('This is a test');
-  });
-});
-
-describe('gettext()', () => {
   it('should return localizable message', () => {
-    expect(i18n.gettext('This is a test')).toEqual('This is a test');
+    expect(i18n._('This is a test')).toEqual('This is a test');
   });
 
   it('should return one-line strings', () => {
     expect(
-      i18n.gettext(
+      i18n._(
         `This
          is
          a
```

rpl (approved):
lgtm, only an optional nit (mentioning in case we feel it may be worth including that in this PR instead of filing it as a followup, but I'm fine either way).

tests/unit/test.utils.js
willdurand: I am going to add that to this PR.

PR #5586, opened 2025-02-24
feat: Imported Firefox 136 schema data

### LINKED ISSUE #5585: Import Firefox 136 API schema
This issue is about importing updated WebExtensions API JSONSchema files from Firefox 136.

rpl:
The updated schema data includes the following changes

Firefox 136:
- [Bug 1930334 - Implement background.preferred_environment](https://bugzilla.mozilla.org/show_bug.cgi?id=1930334)

Fixes #5585

### DIFF
--- a/src/schema/imported/manifest.json
+++ b/src/schema/imported/manifest.json
@@ -31,6 +31,7 @@
         },
         "name": {
           "type": "string",
+          "description": "Name must be at least 2, at should be at most 75 characters",
           "preprocess": "localize",
           "maxLength": 45,
           "minLength": 2
@@ -138,22 +139,13 @@
                     {
                       "type": "object",
                       "properties": {
+                        "service_worker": {
+                          "$ref": "manifest#/types/ExtensionURL",
+                          "min_manifest_version": 3
+                        },
                         "page": {
                           "$ref": "manifest#/types/ExtensionURL"
                         },
-                        "persistent": {
-                          "type": "boolean",
-                          "max_manifest_version": 2,
-                          "default": true
-                        }
-                      },
-                      "required": [
-                        "page"
-                      ]
-                    },
-                    {
-                      "type": "object",
-                      "properties": {
                         "scripts": {
                           "type": "array",
                           "items": {
@@ -162,6 +154,7 @@
                         },
                         "type": {
                           "type": "string",
+                          "description": "Only supported for page/scripts; not for service_worker yet, see bug 1775574",
                           "enum": [
                             "module",
                             "classic"
@@ -171,24 +164,20 @@
                           "type": "boolean",
                           "max_manifest_version": 2,
                           "default": true
+                        },
+                        "preferred_environment": {
+                          "type": "array",
+                          "items": {
+                            "type": "string",
+                            "enum": [
+                              "service_worker",
+                              "document"
+                            ],
+                            "onError": "warn"
+                          }
                         }
                       },
-                      "required": [
-                        "scripts"
-                      ]
-                    },
-                    {
-                      "type": "object",
-                      "properties": {
-                        "service_worker": {
-                          "$ref": "manifest#/types/ExtensionURL"
-                        }
-                      },
-                      "postprocess": "requireBackgroundServiceWorkerEnabled",
-                      "required": [
-                        "service_worker"
-                      ],
-                      "min_manifest_version": 3
+                      "postprocess": "checkRequiredManifestBackgroundKeys"
                     }
                   ]
                 },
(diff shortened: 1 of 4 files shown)


src/schema/validator.js
```diff
@@ -680,6 +684,50 @@ export class SchemaValidator {
...
+        ];
+        return false;
+      }
+      return true;
+    };
+
+    validator.addKeyword({
+      keyword: 'postprocess',
+      validate: validateRequiredManifestBackgroundKeys,
+    });
```
rpl, line 730: note: this addition was needed due to the changes to the structure of the new schema (which is not including a `required` property anymore and the various formats allowed are not an `anyOf` anymore), and so I opted to register a custom keyword for "postprocess" which only report errors for the postprocess named `checkRequiredManifestBackgroundKeys`, which enforces the expected required properties similarly to [how we are now doing that on the Gecko side here](https://searchfox.org/mozilla-central/rev/fcfb558f8946f3648d962576125af46bf6e2910a/toolkit/components/extensions/Schemas.sys.mjs#298-306).

The error reported has keyword set to "required" to then be turned into a addons-linter error with code "MANIFEST_FIELD_REQUIRED" (which then interpolate the invalid property name into the error message), given that it feels a better fit for this error than the generic `JSON_INVALID` code.

src/schema/updates/manifest.json
```diff
@@ -41,10 +41,12 @@
...
                     "anyOf": [
-                      {},
-                      {},
                       {
-                        "min_manifest_version": 3
+                        "properties": {
+                          "service_worker": {
+                            "min_manifest_version": 3
+                          }
+                        }
```
rpl, line 49: note: this change on the `src/schema/updates/manifest.json` was needed to ensure the `min_manifest_version`: 3 will still get applied to the `service_worker` manifest json property schema (otherwiser it was being applied to the entire `background` JSONSchema definition, which was part of the reasons for the test failures being hit by the initial version of this patch).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -2465,25 +2465,59 @@ describe('ManifestJSONParser', () => {
...
+        const expectedErrorMessage = `"/background" requires at least one of ${expectedRequiredProps}.`;
+        expect(errors).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              instancePath: '/background',
+              code: messages.MANIFEST_FIELD_REQUIRED.code,
+              message: expect.stringMatching(expectedErrorMessage),
+            }),
+          ])
+        );
```
rpl, line 2517: note: I replaced the previous call to `assertHasMatchingError` with one using toEqual/arrayContaining/objectContaining while I was here, because the latter approach provide a better (more detailed and more likely to be immediately actionable) error message than the [one that hitting an `assertHasMatchingError` logs](https://app.circleci.com/pipelines/github/mozilla/addons-linter/6675/workflows/1cf5015a-4b39-4134-94b5-fc7ef8bbafed/jobs/21014?invite=true#step-106-20192_57).

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3271,10 +3305,9 @@ describe('ManifestJSONParser', () => {
       expect(linter.collector.errors).toEqual(
         expect.arrayContaining([
           expect.objectContaining({
-            instancePath: '/background',
-            code: 'JSON_INVALID',
+            code: 'MANIFEST_FIELD_UNSUPPORTED',
             message: expect.stringMatching(
-              /"\/background" .* has invalid extra properties/
+              /"\/background\/service_worker" is not supported/
```
rpl, line 3310: This error is not [the one we raise from here in manifestjson.js](https://github.com/mozilla/addons-linter/blob/3d5d6ce72994745c3cf48921579873b1f076d52c/src/parsers/manifestjson.js#L576-L585), instead of the generic one that anyOf ajv validation used to raise.

src/schema/imported/manifest.json
```diff
@@ -31,6 +31,7 @@
         },
         "name": {
           "type": "string",
+          "description": "Name must be at least 2, at should be at most 75 characters",
```
willdurand, line 34: FWIW I fixed this typo in https://bugzilla.mozilla.org/show_bug.cgi?id=1950334

willdurand (approved):

src/schema/imported/manifest.json
rpl: ty, I'll leave that fixed typo to be covered from the next import for 137 data once it gets to beta in the next couple of weeks.

PR #5632, opened 2025-04-15
Emit an error when data_collection_permissions is specified in the manifest

willdurand:


wagnerand:
Please note there is an ongoing conversation on slack about the name of the key.

rpl (approved):

PR #5663, opened 2025-05-20
Add a CLI option to enable data collection permissions support

### LINKED ISSUE #5649: Add a CLI option to enable data collection permissions support
We need a new CLI option to allow data collection permissions, which are currently disabled via #5634.

willdurand:
Fixes #5649

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -5764,7 +5764,7 @@ describe('ManifestJSONParser', () => {
   });
 
   describe('data_collection_permissions', () => {
-    it('emits an error when data_collection_permissions is specified', () => {
+    it('emits an error when data_collection_permissions is specified and the support is disabled', () => {
       const linter = new Linter({ _: ['bar'] });
 
       const manifestJSONParser = new ManifestJSONParser(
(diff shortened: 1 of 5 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -5786,5 +5787,25 @@ describe('ManifestJSONParser', () => {
         ),
       ]);
     });
+
+    it('does not emit an error when data_collection_permissions is specified and the support is enabled', () => {
```
rpl, line 5791: followup: it would be good to add a few smoke tests that makes sure that the validation of the data_collection_permissions format is in place as expected, and one that confirms that setting data_collection_permissions to a truthy value that is not valid from a JSONSchema perspective doesn't get the missing data_collection_permissions error but it gets the schema validation error and the manifest to be considered invalid as expected.

note: as a side note, this PR is currently forked from the main branch before we imported the updated schema data that includes also the data_collection_permissions JSONSchema, and so to be able to add those tests we either need to do it on a rebased version of this PR or in a followup PR branched from a more recent tip.

rpl (approved):
looks good to me, only one nit on the test side and one followup mentioned in the inline review comments below.

Issue #5669, opened 2025-05-23
Include required data collection permissions


willdurand:
Due to some import issues, we don't have the schema for _required_ data collection permissions. Let's add that, we'll fix the import separately.

ioanarusiczki:
@willdurand I assume this can be verified on AMO dev once the waffle switch is enabled ?

ioanarusiczki:
I've checked all scenarios from the [document ](https://docs.google.com/document/d/1UFVCzAr-MDrBzHNcxdv4vq8MoSIqeTKfrIXyjsLCXdI/edit?tab=t.0) with ` --enable-data-collection-permissions` and I didn't have problems.

Also this passes validation 

`"data_collection_permissions": {
			"required": [],
			"optional": []
				}`

The one below will throw errors and does not pass validation 

`"data_collection_permissions": {
			"required": ["make it error"],
			"optional": ["another error"]
				}`

willdurand:
> [@willdurand](https://github.com/willdurand) I assume this can be verified on AMO dev once the waffle switch is enabled ?

yes, but that isn't implemented yet AFAIK

ioanarusiczki:
@willdurand yes, there's an issue filed , it's open. Atm on -dev/stage it throws the error with "property is reserved"

PR #5670, opened 2025-05-23
Include required data collection permissions

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/5669


rpl (approved):
lgtm, thanks @willdurand

**note:** #5671 is tracking replacing this workaround with a proper fix on the schema import side.

PR #5675, opened 2025-05-27
Bump node versions

willdurand:
I disabled a step for now, hoping to re-enable it when https://github.com/mozilla/web-ext/issues/3429 has been fixed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,10 @@ jobs:
       # run integration tests using an addons-linter binary in a
       # production-like environment
       - run: npm run test-integration:production
-      - run: npm run webext-test-functional
+      # TODO: re-enable this step once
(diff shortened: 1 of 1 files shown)


willdurand:
> I disabled a step for now, hoping to re-enable it when [mozilla/web-ext#3429](https://github.com/mozilla/web-ext/issues/3429) has been fixed.

I confirmed that https://github.com/mozilla/web-ext/pull/3519 would "fix" this failure.

rpl (approved):
lgtm, as we already agreed this is disabling just the CI task running the web-ext functional tests (which is mainly used to catch potential regressions on integrating new addons-linter version in the web-ext cli tool) as short term workaround to unblock bumping the nodejs version in this repository (and then soon enough we will re-enable it, once we have fixed the underlying issue on the web-ext side).

Issue #5671, opened 2025-05-23
Fix import script to handle "required" and "properties" schema properties


willdurand:
We recently discovered a bug in our import script that caused the `required` data collection permissions to be missing. Fixing that uncovered another issue with some props named `properties`. Let's fix that.

PR #5706, opened 2025-06-23
Add validation logic for required data collection permissions

### LINKED ISSUE #5705: Prevent `none` from being specified in `data_collection_permissions` with other required permissions
STR:
- Run the linter on an add-on with a manifest containing under `gecko` (pass `--enable-data-collection-permissions:`)
```json
"data_collection_permissions": {
  "required": [
    "none",
    "bookmarksInfo"
  ]
}
```

Expected results:
- `none` raises a validation error or warning, since it should only be used alone

Actual results:
- Validation passes

willdurand:
Fixes #5705

---

This PR adds validation logic to enforce at least one data collection
permission in the `required` list, and it makes sure that `none` cannot
be specified with other permissions.


docs/rules.md
```diff
@@ -139,6 +139,7 @@ Rules are sorted by severity.
 | `ADMIN_INSTALL_ONLY_REQUIRED`                           | error    | The `browser_specific_settings.gecko.admin_install_only` property must be set to "true" in an enterprise add-on                                                                    |
 | `DATA_COLLECTION_PERMISSIONS_PROP_RESERVED`             | error    | The `browser_specific_settings.gecko.data_collection_permissions` property is reserved for future usage and cannot be used in an extension at the moment                           |
 | `MISSING_DATA_COLLECTION_PERMISSIONS`                   | notice   | The `browser_specific_settings.gecko.data_collection_permissions` will be required in the future. Please add this key to the manifest                                              |
+| `NONE_DATA_COLLECTION_IS_EXCLUSIVE`                     | error    | "none" must be specified without any other data collection permissions                                                                                                             |
```
rpl, line 142: nit: not that big of a deal, but I wonder:

- if it may be worth to more explicitly mention that what this rule describes refers only to the required part of the data collection (in case some reader may be confused and think that that's also for optional data collection, where use of "none" as required along others set as optional is actually fine and expected).
- if the opposite wording may be slightly better, e.g. `"none" must NOT be specified with other required data collection permissions`.

src/schema/updates/manifest.json
```diff
@@ -21,6 +21,7 @@
             "properties": {
               "required": {
                 "type": "array",
+                "minItems": 1,
```
rpl, line 24: thought: maybe we should briefly mention in #5671 (the issue tracking fixing the underlying issue that forced us to copy this entire part of the schema here in the schema updates) that if we are removing this workaround then we should not forget that we need a `minItems` property to still be applied on the JSONSchema for the `required` property.

To be fair at least of the test added in this patch would fail, and so we can't really miss it, if we decide to add a mention it would be mainly to remind it to ourselves to not have to hit the failure and have to look into why).

src/parsers/manifestjson.js
```diff
@@ -1269,6 +1267,24 @@ export default class ManifestJSONParser extends JSONParser {
...
+    const { required } = permissions;
+    const requiredPermissions = Array.isArray(required) ? required : [];
+
+    if (
+      requiredPermissions.includes('none') &&
+      requiredPermissions.length > 1
+    ) {
+      this.collector.addError(messages.NONE_DATA_COLLECTION_IS_EXCLUSIVE);
+      this.isValid = false;
+    }
```
rpl, line 1285: thought: @willdurand should we also report a linting error if there are optional permissions but without any required data collection set in the manifest?

(if I'm not mistaken in that case we still want `required` to be set, either to `["none"]` or to a set of required data collection permissions.

rpl (commented):
@willdurand this pull request looks good to me, I have collected just a couple of nits and thoughts.

Before approving the PR I'd like to touch base about the thought related to the case where the addon has some optional data collection permissions but no required data collection at all.

note: as a side note, even if we agree that is a case that needs to be covered, I'd be ok also to sign off this PR with that part deferred to a separate followup.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -5909,5 +5909,85 @@ describe('ManifestJSONParser', () => {
...
+              data_collection_permissions: {},
+            },
+          },
+        }),
+        linter.collector,
+        { schemaValidatorOptions: { enableDataCollectionPermissions: true } }
+      );
+
+      assertHasMatchingError(linter.collector.errors, {
+        code: messages.MANIFEST_FIELD_REQUIRED.code,
```
rpl, line 5986: thought: for completeness of the test coverage it may be worth also adding a basic test case for a valid manifest with both required and optional data collection permissions, wdyt?

e.g. even just a basic one like the one that follows:
```js
  it('does not emit an error on valid required and optional data collection permissions', () => {
      const linter = new Linter({ _: ['bar'] });

      const manifestJSONParser = new ManifestJSONParser(
        JSON.stringify({
          manifest_version: 2,
          name: 'some name',
          version: '1',
          browser_specific_settings: {
            gecko: {
              data_collection_permissions: {
                required: ["none"],
                optional: ["technicalAndInteraction", "locationInfo"],
              },
            },
          },
        }),
        linter.collector,
        { schemaValidatorOptions: { enableDataCollectionPermissions: true } }
      );

      expect(linter.collector.errors).toEqual([]);
      expect(manifestJSONParser.isValid).toEqual(true);
    });
```

tests/unit/parsers/test.manifestjson.js
```diff
@@ -5909,5 +5909,85 @@ describe('ManifestJSONParser', () => {
       ]);
       expect(manifestJSONParser.isValid).toEqual(true);
     });
+
+    it('emits an error when "none" is specified with other data collection permissions', () => {
```
rpl, line 5913: thought: Given that we had to workaround the import of the schema definitions for the data collection permissions more than we would have liked to, it may be also worth to consider if adding another test case to verify that we emit a lintint error as expected when "technicalAndInteraction" is wrongly listed in the required permission, wdyt?

rpl (approved):
@willdurand lgtm, the following two comments are thoughts related to some additional test coverage that may be worth considering (one for a valid manifest with both required and optional data collection permissions, the other one is for an invalid manifest with "technicalAndInteraction" wrongly listed as a required data collection permission).

Issue #4942, opened 2023-07-01
Request to Increase Extension Upload File Size Limit


Nick-1979:
Dear Firefox Development Team,

I am writing to inform you of an issue I encountered while attempting to update my extension on Firefox. Unfortunately, despite successfully updating the extension on Chrome and Edge with new features, I am currently unable to do the same on Firefox due to the existing file size limitation.

The error message states that the file size is too large to parse.

I kindly request that you consider increasing the file size limit for extension updates on Firefox. This adjustment would allow developers like me to provide the same enhanced functionality to Firefox users that is already available on other browsers.

Thank you for your attention to this matter. I greatly appreciate the efforts of the Firefox development team and hope for a positive resolution.

<img width="1040" alt="Screenshot 2023-07-02 at 01 17 54" src="https://github.com/mozilla/addons-linter/assets/46442452/3f0c133b-c781-48e3-8224-1defea7ab64c">
<img width="1138" alt="Screenshot 2023-07-02 at 01 18 14" src="https://github.com/mozilla/addons-linter/assets/46442452/a0eea7ed-b2ef-436b-ad9d-0b3842a234b5">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDLINT-211)

diox:
Hi,

This has already been reported in https://github.com/mozilla/addons-linter/issues/4748 for a specific case - [`tesseract.js`](https://github.com/naptha/tesseract.js).

We have to have some kind of limit to avoid abuse as parsing and analyzing submitted code is time consuming and costly as the size increases.

What limit would work in your case ?

Nick-1979:
Hi @diox,

Thank you for responding. In my situation, a file size of 5 MB is acceptable.

diox:
I'm not opposed to a bump to 5 MB - there are several developers that it would benefit and it's been a while since we last increased that limit - but we need to make some internal architecture changes in our private scanners before considering this.

Nick-1979:
Hello,

I would like to know if there is any progress on resolving this issue or any estimate on when we can update our extensions without encountering this problem. 

Thank you for your hard work and attention.

diox:
Hi,

[...]

PR #5765, opened 2025-08-21
Bump max file size limit to 10MB

ssahillppatell:
- [X] This PR relates to an existing open issue and there are no existing PRs open for the same issue.

https://github.com/mozilla/addons-linter/issues/4942#issuecomment-3211112371

As per this comment, I am hoping to get the limit of max file size to be increased to a size that won't require someone, at least in the near future, to open an issue regarding the file size limit

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -101,15 +101,15 @@ export const MANIFEST_VERSION_MIN = 2;
 export const MANIFEST_VERSION_MAX = 3;
 
 // This is the limit in megabytes of a file we will parse (eg. CSS, JS, etc.)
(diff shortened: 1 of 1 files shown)


ssahillppatell:
hi team. Sorry to bother. But this is blocking us from publishing our addon(https://github.com/Prastavna/xodeforces) in the Firefox Add-on Store.

Rob--W:
We have recently bumped the limit and cannot keep bumping the limit forever.

As a work-around:

The Chrome Web Store version of your extension contains a 6 MB worker file. If you split the functionality in separate files and use `importScripts`, the behavior will be equivalent and you can upload without being blocked by the 5 MB file size limit.

PR #5838, opened 2025-10-30
Update background.service_worker message

### LINKED ISSUE #5252: Emit helpful message when background.service_worker is used without background.scripts/page
addons-linter was patched before to emit only a warning when `background.service_worker` is used without scripts/page (https://github.com/mozilla/addons-linter/issues/5152). The warnings are still not very helpful.

Instead of a generic "/background/service_worker" is not supported.", we should print a helpful message that encourages the use of `scripts`. For more details on why and what, see my write-up for `web-ext run` at https://github.com/mozilla/web-ext/issues/3101

The relevant implementation was last touched for #5152, at
https://github.com/mozilla/addons-linter/blob/40fa69a2f0274dc2c4bd4bb6c1a6d35ac979416e/src/parsers/manifestjson.js#L568-L576

dotproto:
Fixes #5252

This PR updates the error/warning message returned when an extension's `manifest.json` file contains a `background.service_worker` property without declaring `background.scripts` or `background.page`.


src/messages/manifestjson.js
```diff
@@ -888,3 +888,13 @@ export const MISSING_ADDON_ID = {
     future. See https://mzl.la/3PLZYdo for more information.`),
   file: MANIFEST_JSON,
 };
+
+export const BACKGROUND_SERVICE_WORKER = {
+  code: 'BACKGROUND_SERVICE_WORKER',
+  message: i18n._(`"/background/service_worker" is not supported.`),
+  description: i18n._(`The "background" property must contain either "scripts" 
+    or "page". See
```
Rob--W: Instead of offering two options, let's offer "scripts" as a fallback. The MDN docs provides more context:

```suggestion
  description: i18n._(`"background.scripts" must be specified as a fallback for
    "background.service_worker". See
```

PR #5841, opened 2025-11-03
Migrate to ESLint v9

willdurand:
This PR upgrades ESLint in this project, which requires change in the code since we're using ESLint for the JS scanner and on the rest of the project itself (as we are using the same ESLint version for the linting of the project).

### DIFF
--- a/tests/unit/helpers.js
+++ b/tests/unit/helpers.js
@@ -68,17 +68,9 @@ export function getRuleFiles(ruleType) {
   });
 }
 
-/**
- * Get variables in the current scope
- * @param {object} scope current scope
- * @param {string} name name of the variable to look for
- * @returns {ASTNode} The variable object
- *
- * Copied from ESLint tests; used to test {allowInlineConfig: false} settings.
- */
-export function getVariable(scope, name) {
+export function getVariable(variables, name) {
   let variable = null;
-  Array.from(scope.variables).some((v) => {
+  Array.from(variables).some((v) => {
     if (v.name === name) {
       variable = v;
       return true;
--- a/tests/unit/scanners/test.javascript.js
+++ b/tests/unit/scanners/test.javascript.js
@@ -333,7 +333,7 @@ describe('JavaScript Scanner', () => {
                 ? sourceCode.getScope(node)
                 : context.getScope();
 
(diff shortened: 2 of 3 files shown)


tests/unit/rules/javascript/test.no_implied_eval.js
```diff
@@ -80,8 +82,8 @@ describe('no_implied_eval', () => {
       description: [NO_IMPLIED_EVAL.description],
     },
     {
-      code: 'execScript("x = 1;");',
-      message: [expectedErrorMessage],
+      code: 'window.execScript("x = 1;");',
```
willdurand: I had to change this because of https://github.com/eslint/eslint/issues/19923, but I am not entirely sure that's what we want.

tests/unit/scanners/test.javascript.js
```diff
@@ -318,20 +318,28 @@ describe('JavaScript Scanner', () => {
...
-        };
+      test: {
+        create(context) {
+          const sourceCode = context.sourceCode ?? context.getSourceCode();
+
+          return {
+            Program(node) {
+              const scope = sourceCode.getScope
+                ? sourceCode.getScope(node)
+                : context.getScope();
```
rpl: Nit: similar comment here on the test side as the previous one related to the backward compatibility logic not being strictly necessary.

tests/unit/helpers.js
```diff
@@ -78,7 +78,7 @@ export function getRuleFiles(ruleType) {
  */
 export function getVariable(scope, name) {
```
rpl: question: @willdurand if I'm reading the diffs correctly, it looks like the getVariables function we use on the implementation side is now expecting to get the `variables` passed as the first parameter, while this similarly named one on the test side expected to get the `scope` object. I do wonder it we should consider keeping their type signature similar or rename one of the two to not be confused with each other (to be fair one is a test-only helper, I haven't dig into the historical reasons for having both of them around tbh).

rpl (approved):
@willdurand this PR looked good to me, I had some thoughts about some backward compatibility logic that we may not need, and about two utility functions (one of the test side one on the implementation side) having the same name but different type signatures. None of them seems a blocking issue and so I'm also signing this off as is.

PR #5899, opened 2026-01-19
fix: clarify message emitted by the linter on the background.service_worker manifest property

### LINKED ISSUE #5252: Emit helpful message when background.service_worker is used without background.scripts/page

rpl:
This PR splits the linting message emitted on the `background.service_worker` manifest key into an error (`BACKGROUND_SERVICE_WORKER_NOFALLBACK`, emitted when both `background.scripts` and `background.page` are missing and the unsupported property is missing a Firefox compatible fallback property) and a warning (`BACKGROUND_SERVICE_WORKER_IGNORED`, which is emitted when at least one between `background.scripts` and `background.page` is also present in the manifest, and so the message is just a reminder to the developer that the service worker one is ignored and they should make sure the Firefox-compatible fallback included provides the appropriate Firefox compatibility).

The goal of this change is to make these linting message more immediately clear and actionable for the extensions developers.

This PR also includes and extends the changes originally part of #5838, and so it also supersedes the previous PR.

Fixes #5252

### DIFF
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -3357,9 +3357,11 @@ describe('ManifestJSONParser', () => {
       { scripts: ['bg.js'] },
       { page: 'bg.html' },
       { scripts: [], page: 'bg.html' },
-      { scripts: [], page: 'bg.html' },
       // Expect the warning to be emitted also if preferred_environment is set but doesn't include
       // "document" as one of the preferred_environment.
+      // NOTE: These tests don't need to set a strict_min_version in this test because the
+      // assertion on the warnings currently check for the specific warnings expected and
+      // it doesn't fail when other unrelated warnings are emitted.
       { page: 'bg.html', preferred_environment: ['service_worker'] },
       { scripts: ['bg.js'], preferred_environment: ['service_worker'] },
     ])(
@@ -3412,13 +3414,8 @@ describe('ManifestJSONParser', () => {
         page: 'bg.html',
         preferred_environment: ['service_worker', 'document'],
       },
-      {
-        scripts: [],
-        page: 'bg.html',
-        preferred_environment: ['service_worker', 'document'],
-      },
     ])(
-      'omits warning on background.service_worker set along with %o and preferred_environment set accordingly',
+      'omits warning on background.service_worker if preferred_environment is set along with %o',
       (backgroundProps) => {
         const linter = new Linter({ _: ['bar'] });
         const json = validManifestJSON({


src/messages/manifestjson.js
```diff
@@ -888,3 +888,26 @@ export const MISSING_ADDON_ID = {
...
+  code: 'BACKGROUND_SERVICE_WORKER_IGNORED',
+  message: i18n._(
+    `Unsupported "/background/service_worker" manifest property is ignored by Firefox.`
+  ),
+  description:
+    i18n._(`The "/background/service_worker" manifest property is unsupported and ignored on Firefox.
+    Please make sure "/background/scripts" or "/background/page" properties are providing appropriate Firefox compatibility.
+    See https://mzl.la/4r6SF1L for more information.`),
+  file: MANIFEST_JSON,
+};
```
rpl: @Rob--W I also felt that using a single message for both error and warning linting message was going to be confusing (also given that in the rules.md file each rule code should have a single severity, and so it should be either `error` or `warning` and so it can't and shouldn't be both), and so I did split it into separate warning and an error rule code.

I've also reworded a bit and tried to account for the feedback you provided in the original PR, let me know what do you think.

rpl:
> LGTM, thanks!
> 
> If we want to, we could allow extension developers to opt out of the warning if they set `preferred_enviroment` in the background key, which is implemented in Firefox and Safari to allow extensions to signal the preference for a document-based background page over a service worker, when the browser supports both. For documentation, see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background#:~:text=preferred_environment

That sounds reasonable to me too and so I took a look to determine if I could roll the additional changes into this PR or if it felt like something to defer to a separate followup.

I've pushed an update to this PR with the tweaks applied through [dd74cb3](https://github.com/mozilla/addons-linter/pull/5899/commits/dd74cb3d4e33805a1ee22c588193352417cbd520), if you also feel that may be enough, then I'd be more than happy to include it in this PR, otherwise I'm happy to open a separate pull request to discuss about that part a bit more and get to an agreement, while merging this one without that part.

@Rob--W let me know what do you think.

tests/unit/parsers/test.manifestjson.js
```diff
@@ -3377,17 +3383,72 @@ describe('ManifestJSONParser', () => {
...
+        page: 'bg.html',
+        preferred_environment: ['service_worker', 'document'],
+      },
+      {
+        scripts: [],
+        page: 'bg.html',
+        preferred_environment: ['service_worker', 'document'],
+      },
+    ])(
+      'omits warning on background.service_worker set along with %o and preferred_environment set accordingly',
```
Rob--W: Nit: put the dynamic `%o` at the end so that all similar tests have the same prefix and it is easier to see which part of the test is unique in the test output.

Rob--W (approved):

Rob--W (approved):

Issue #5932, opened 2026-02-17
Replace `chalk` with Node.js' native `styleText`


erosman:
ESLint has already replaced  `chalk` with Node.js' native `styleText` in ESLint 10.

Details: [Change Request: Replace `chalk` with Node.js' native `styleText` in v10](https://github.com/eslint/eslint/issues/20012)

Rob--W:
Closing because https://github.com/mozilla/addons-linter/pull/5949 was merged.

PR #5949, opened 2026-02-26
chore(deps): replace chalk with native util.styleText

roli-lpci:
## Summary

- Replaces `chalk` dependency with Node.js native `util.styleText` API
- Bumps minimum Node.js version from `>=18.0.0` to `>=20.12.0` (required for `util.styleText`)
- All color methods use `String()` coercion for safe non-string handling

## Changes

- `src/linter.js` — `chalk.Instance({ enabled })` replaced with plain object of color functions that delegate to `styleText()` or pass through via `String` in boring mode
- `tests/integration/run-as-production-env.js` — `chalk.red()`/`chalk.green()` replaced with `styleText('red', ...)`/`styleText('green', ...)`
- `package.json` — Removed `chalk`, bumped `engines.node` and `browserslist` to `>=20.12.0`

## Testing

- 66/66 test suites pass, 1700 tests
- ESLint: 0 errors
- Webpack production build: clean
- No test changes needed (existing stubs work with new object shape)

Addresses #5932


Rob--W (changes requested):
Bumping the Node version would require a major version bump, which is too big of an ask for this refactor.

Please maintain support for older versions. I consider it to be an acceptable degradation of functionality by falling back to just printing the message as is without coloring (for simplicity), or including a simple, minimal set of escape sequences if you'd like to maintain the coloring on older Node.

Considering that Node 18 is already EOL for almost a year at this point, with Node 20 to follow soon (in less than 2 months), I think that we don't have to worry about old Node.

Therefore I'd be inclined to go with the simple approach: no coloring as a fallback.

Rob--W (commented):
How did you generate the package-lock.json file? I see that it includes several changed, and it is unclear why these changes were made.

I checked the package-lock file to see if the removed dependency would disappear. While the entry is deleted from the dependencies, it is still listed. I checked mentions of `chalk"` in the lockfile and it seems that there is one dependent (eslint) that itself is a peer dependency, and others but those are just dev dependencies.

As the code changes itself seem to be AI-generated, I'd like to know whether the lock file was also AI-generated, or created by standard tooling. And if so, what versions did you use?

roli-lpci:
Node v25.2.1, npm 11.6.2.

I am what many would call a nontechnical developer with a pipeline that is highly steered, monitored, submitting PRs while I learn proper orchestration, validation, testing, etc. If you would not like any more submissions from me, let me know and I can abstain going forward. Thanks!

tests/integration/run-as-production-env.js
```diff
@@ -1,8 +1,15 @@
...
 const path = require('path');
 const { spawn } = require('child_process');
+const util = require('node:util');
+
+// Graceful fallback: use styleText when available (Node >= 20.12),
+// otherwise return the plain text without coloring.
+const colorize =
+  typeof util.styleText === 'function'
+    ? (style, text) => util.styleText(style, String(text))
+    : (_style, text) => String(text);
```
Rob--W:
```suggestion
```

tests/integration/run-as-production-env.js
```diff
@@ -15,12 +22,14 @@ const npmScript = process.argv[2];
 const jestTestsPath = process.argv[3];
 
 if (!npmScript) {
-  console.error(chalk.red('Missing mandatory npm script to run'));
+  console.error(colorize('red', 'Missing mandatory npm script to run'));
```
Rob--W:
```suggestion
  console.error('Missing mandatory npm script to run');
```

tests/integration/run-as-production-env.js
```diff
@@ -15,12 +22,14 @@ const npmScript = process.argv[2];
...
   process.exit(1);
 }
 
 if (!jestTestsPath) {
-  console.error(chalk.red('Missing mandatory path to the tests dir to run'));
+  console.error(
+    colorize('red', 'Missing mandatory path to the tests dir to run')
+  );
```
Rob--W:
```suggestion
  console.error('Missing mandatory path to the tests dir to run');
```

tests/integration/run-as-production-env.js
```diff
@@ -45,7 +54,7 @@ function getPackedName() {
 }
 
 function createPackage(tmpDirPath) {
-  console.log(chalk.green('Create a pre-release npm package archive'));
+  console.log(colorize('green', 'Create a pre-release npm package archive'));
```
Rob--W:
```suggestion
  console.log('Create a pre-release npm package archive');
```

tests/integration/run-as-production-env.js
```diff
@@ -69,7 +78,10 @@ function createPackage(tmpDirPath) {
 
 function unpackTarPackage(packagePath, destDir) {
   console.log(
-    chalk.green(['Unpacking', packagePath, 'package into', destDir].join(' '))
+    colorize(
+      'green',
+      ['Unpacking', packagePath, 'package into', destDir].join(' ')
+    )
```
Rob--W:
```suggestion
    ['Unpacking', packagePath, 'package into', destDir].join(' ')
```

tests/integration/run-as-production-env.js
```diff
@@ -82,7 +94,7 @@ function unpackTarPackage(packagePath, destDir) {
 }
 
 function installPackageDeps(packageDir) {
-  console.log(chalk.green('Install production package dependencies'));
+  console.log(colorize('green', 'Install production package dependencies'));
```
Rob--W:
```suggestion
  console.log('Install production package dependencies');
```

tests/integration/run-as-production-env.js
```diff
@@ -150,6 +162,8 @@ tmp
     { unsafeCleanup: true }
   )
   .catch((err) => {
-    console.error(err.stack ? chalk.red(err.stack) : chalk.red(err));
+    console.error(
+      err.stack ? colorize('red', err.stack) : colorize('red', String(err))
+    );
```
Rob--W:
```suggestion
    console.error(err);
```

tests/integration/run-as-production-env.js
```diff
@@ -105,7 +117,7 @@ function installPackageDeps(packageDir) {
 
 function runIntegrationTests(packageDir) {
   console.log(
-    chalk.green('Running integration tests in production-like environent')
+    colorize('green', 'Running integration tests in production-like environent')
   );
```
Rob--W:
```suggestion
  console.log('Running integration tests in production-like environent');
```

Rob--W (commented):
Suggested edits to directly apply (dropping styleText from the `run-as-production-env.js` test-only script.

After applying these and recomputing package-lock.json this PR should be good to go. The styleText fallback in src/linter.js doesn't hurt, so I'll keep it there.

tests/integration/run-as-production-env.js
```diff
@@ -105,7 +117,7 @@ function installPackageDeps(packageDir) {
 
 function runIntegrationTests(packageDir) {
   console.log(
-    chalk.green('Running integration tests in production-like environent')
+    colorize('green', 'Running integration tests in production-like environent')
```
Rob--W:
```suggestion
    'Running integration tests in production-like environent'
```

Rob--W:
Tests are passing but code coverage dropped because the fallback for 20.0.0 is never taken when the tests are run. Since the package.json declares node support >=20.0.0 and styleText was introduced in Node 20.12.0, I'm fine with keeping it in for now at the "expense" of a tiny bit of lost code coverage.

I'm going to merge this. Thanks for getting this started!

Issue #5946, opened 2026-02-25
DeprecationWarning: `url.parse()`


erosman:
When running addons-linter, there is the following error:

> (node:30826) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.

It appears to ralate to https://github.com/mozilla/addons-linter/blob/04ba891d7c960fcfe6fbf2593a8cf292c09fd906/src/utils.js#L214

PR #5962, opened 2026-03-05
Remove the obsolete openDialog rules 

willdurand:
Closes https://github.com/mozilla/addons-linter/issues/5946


src/utils.js
```diff
@@ -211,19 +209,25 @@ export function ensureFilenameExists(filename) {
...
+  try {
+    const { protocol } = new URL(urlInput);
+    // Check protocol is chrome: or resource: if set.
+    // Details on the chrome protocol are here: https://goo.gl/W52T0Q
+    // Details on resource protocol are here: https://goo.gl/HHqeJA
+    return LOCAL_PROTOCOLS.includes(protocol);
+  } catch {
+    // If the URL has a scheme but failed to parse (e.g. http://example.com%),
+    // treat it as non-local.
+    if (/^\w+:\/\//.test(urlInput)) {
```
Rob--W: This kind of check is dangerous if this were to be a security check, because if someone were to find a way to get the URL validation to fail here (but somehow passing when used in the browser), then the `LOCAL_PROTOCOLS` allowlist could be bypassed.

The context and caller expects false unless the URL is certainly safe.

PR #6014, opened 2026-04-13
Remove the use of node-fetch, use native fetch instead

willdurand:


diox (approved):
We pretty much haven't used `dispensary` since it was merged though :(

PR #6106, opened 2026-07-10
Allow placeholder to be replaced more than once in an l10n string

### LINKED ISSUE #6107: Placeholders in l10n strings are not all replaced when the same placeholder appears more than once
The last strings added in #6078 are not rendered correctly, see the `description` in:

```
Code                              Message                                      Description                                                                             File            Line   Column
MANIFEST_THEME_CSS_GRADIENT_MI…   Theme CSS gradients require                  CSS gradients in theme images are only supported in Firefox 153 and above. Please set   manifest.json
N_VERSION                         "strict_min_version" to be set to "153" or   "browser_specific_settings.gecko.strict_min_version" to "%(minVersion)s" or higher in
                                  above                                        your manifest.json
```

That's because our no-op implementation of `sprintf()` doesn't account for the same placeholder appearing more than once.

willdurand:
Fixes https://github.com/mozilla/addons-linter/issues/6107

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -143,7 +143,7 @@ export const i18n = Object.freeze({
     let str = fmt;
     // Replace every `%(key)s` or `%(key)d` placeholder with the actual value.
     for (const key of Object.keys(args)) {
-      str = str.replace(new RegExp(`%\\(${key}\\)(d|s)`), args[key]);
+      str = str.replaceAll(new RegExp(`%\\(${key}\\)(d|s)`, 'g'), args[key]);
     }
     return str;
   },
--- a/tests/unit/parsers/test.manifestjson.js
+++ b/tests/unit/parsers/test.manifestjson.js
@@ -4343,9 +4343,16 @@ describe('ManifestJSONParser', () => {
           { io: { files: {} } }
         );
         expect(manifestJSONParser.isValid).toEqual(false);
-        assertHasMatchingError(linter.collector.errors, {
-          code: messages.MANIFEST_THEME_CSS_GRADIENT_MIN_VERSION,
-        });
+        expect(linter.collector.errors).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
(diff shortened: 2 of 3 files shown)


tests/unit/parsers/test.manifestjson.js
```diff
@@ -4343,9 +4343,16 @@ describe('ManifestJSONParser', () => {
           { io: { files: {} } }
         );
         expect(manifestJSONParser.isValid).toEqual(false);
-        assertHasMatchingError(linter.collector.errors, {
-          code: messages.MANIFEST_THEME_CSS_GRADIENT_MIN_VERSION,
-        });
+        expect(linter.collector.errors).toEqual(
```
rpl, line 4346: @willdurand could we add also test coverage on the tests/unit/test.utils.js side?

rpl (approved):
lgtm, thanks @willdurand (the inline comment before is related to consider adding test coverage also on the test.utils.js side, given that it already has test coverage for this utility).

tests/unit/parsers/test.manifestjson.js
willdurand: absolutely, sorry that I didn't think about it myself...

src/utils.js
```diff
@@ -143,7 +143,7 @@ export const i18n = Object.freeze({
     let str = fmt;
     // Replace every `%(key)s` or `%(key)d` placeholder with the actual value.
     for (const key of Object.keys(args)) {
-      str = str.replace(new RegExp(`%\\(${key}\\)(d|s)`), args[key]);
+      str = str.replaceAll(new RegExp(`%\\(${key}\\)(d|s)`, 'g'), args[key]);
```
Rob--W, line 146: Is it intentional that % in the replacement string can be reformatted again?

If not, then it might be safer to do a single replace and use the callback function to look up the replacement, like:

```javascript
// Drop the loop, only this:
str = str.replace(
  /%\((\w+)\)(d|s)/g,
  (full, key, suffix) => args[key] ?? full
);
```

PR #6108, opened 2026-07-13
Avoid re-scanning replacement values in l10n strings

willdurand:
See: https://github.com/mozilla/addons-linter/pull/6106#discussion_r3559005244

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -139,14 +139,12 @@ function oneLineTranslationString(translationKey) {
 // re-enable localization of the project in the future.
 export const i18n = Object.freeze({
   _: (str) => oneLineTranslationString(str),
-  sprintf: (fmt, args) => {
-    let str = fmt;
-    // Replace every `%(key)s` or `%(key)d` placeholder with the actual value.
-    for (const key of Object.keys(args)) {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -139,14 +139,10 @@ function oneLineTranslationString(translationKey) {
...
-    // Replace every `%(key)s` or `%(key)d` placeholder with the actual value.
-    for (const key of Object.keys(args)) {
-      str = str.replaceAll(new RegExp(`%\\(${key}\\)(d|s)`, 'g'), args[key]);
-    }
-    return str;
-  },
+  sprintf: (fmt, args) =>
+    // Replace every `%(key)s` or `%(key)d` placeholder with the actual value
+    // in a single pass so replacement values are never re-scanned.
+    fmt.replace(/%\((\w+)\)(d|s)/g, (full, key) => args[key] ?? full),
```
Rob--W: Presumably the string inputs are internal, right? Otherwise something like `%(__proto__)s` could result in `[object Object]` to appear. Not the end of the world, but if we want to do something against it, we could add an `Object.hasOwn` check. Previously we used `Object.keys` which didn't traverse prototypes either.

Rob--W (approved):
Looks good, with one piece of feedback. I haven't verified whether there are any keys that would not match `\w+`, as I assume that you already checked that.

PR #6156, opened 2026-08-21
Allow update_url for enterprise add-ons

### LINKED ISSUE #6155: update_url should be allowed for enterprise addons
There was discussion in the PRD about update_urls, but it predated some ExtensionSettings changes.

We allow update_url to be set via ExtensionSettings, so there should be no reason we don't allow update_url to be set in the extension as well.

The usecase here is that enterprises install the extension and get updates from the provider. They don't update the extension in policy every time.

chrstinalin:
Fixes https://github.com/mozilla/addons-linter/issues/6155.

Allows enterprise add-ons to use update_url.

### DIFF
--- a/src/parsers/manifestjson.js
+++ b/src/parsers/manifestjson.js
@@ -769,10 +769,7 @@ export default class ManifestJSONParser extends JSONParser {
       });
     }
 
-    if (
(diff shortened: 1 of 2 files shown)


diox (approved):

mkaply:
When does this rollout to https://addons.allizom.org?

Thanks!

PR #6183, opened 2026-09-11
Add a new release process

willdurand:
This PR adds a new workflow to release the linter, using PRs.

Here is an example: https://github.com/mozilla/addons-scanner-utils/pull/1152#issuecomment-5633183536

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [master]
+
+permissions: {}
+
+jobs:
+  draft-release:
+    if: |
+      github.event.pull_request.merged == true &&
+      startsWith(github.event.pull_request.head.ref, 'releases/')
+    runs-on: ubuntu-latest
+    permissions:
+      # Required to create the draft release.
+      contents: write
+      # Required to comment the draft release on the pull request.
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      startsWith(github.event.pull_request.head.ref, 'releases/')
+    runs-on: ubuntu-latest
+    permissions:
+      # Required to create the draft release.
+      contents: write
+      # Required to comment the draft release on the pull request.
+      pull-requests: write
+
+    steps:
+      - uses: mozilla/addons/.github/actions/release@442a70320c838d93bd0c34708471b8c5ca7ce86f # 1.1
```
rpl, line 25: thought: do we want to leave to ourselves an inline comment here to remind us to update the "out release process"  link on the README.md side?

(alternatively we could link in the README.md the last version of the action readme file, https://github.com/mozilla/addons/tree/main/.github/actions/release#release-process, but that may end up pointing a version of the README.md that may not correspond to the version of the action actually used and so it may be potentially misleading too).

rpl (approved):
lgtm, just one nit/thought described briefly below (not really blocking, more of a thought to determine how to avoid the link to the release docs in the README and the action version used in the github workflow to go out of sync).

